Issue #269, opened 2019-10-09
Tina runs lots of code even in my website's production build


dwalkr:
![Tina, you fat lard](https://user-images.githubusercontent.com/15221702/66491868-7ec48580-ea81-11e9-8f08-61e03f46144a.png)

## Description

Short version: https://github.com/tinacms/tinacms/issues/268 occurs even in the website's production build.

The Tina sidebar is set to be `hidden` when building the prod site. Since the editing sidebar is not accessible, we should avoid running **any** Tina code since it's just wasted electricity.

It's worth noting that, semantically, _hiding the sidebar_ doesn't really communicate that Tina code will be removed from the production build. Other solutions that involve conditionally running Tina but require a different configuration are welcome.

### Steps to reproduce

1. Create a Gatsby site and add Tina to it
2. In `gatsby-config.js`, set the sidebar to be hidden on production like this:

```javascript
{
  resolve: "@tinacms/gatsby-plugin-tinacms",
  options: {
    sidebar: {
      hidden: process.env.NODE_ENV === "production",
      position: "fixed",
    },
  },
},
```

### Expected result

Tina does not run on production

### Actual result

The Tina sidebar is inaccessible and content can't be edited, but Tina code still impacts page performance

PR #276, opened 2019-10-09
feat: forms are neve registered in production

ncphillips:
Addresses some of the concern of #269 

Not a complete fix


SimonKlausLudwig:
Is there a way to completly disable tinaCMS in production build ?

PR #310, opened 2019-10-16
Individual package builds

jamespohalloran:
**This is not to be merged. Left open as a reference for future work**

The goal of this PR was to ultimately allow packages to be built independently from one another, without using the --hoist option for lerna: This was in an effort to be able to npm link or yarn link to individual packages (with their complete set of dependencies in the local node_modules).
Typescript had to be downgraded to get around a styled-components import error. 

The packages will now build without error, but I still wasn't able to npm link to individual packages


jamespohalloran:
A few required steps to get going:
`lerna exec -- npm link`
Copy the eslintignore file from the tinacms repo into your gatsby site
from inside your site, run `npm link ../tinacms/node_modules/react` (to only load one version of react).
Currently leads to a new error
```
Error: React-Hot-Loader: AppContainer should be patched
▶ 17 stack frames were collapsed.
(anonymous function)
/Users/jamesohalloran/Documents/forestry/tinacms-demo/.cache/app.js:40
  37 |  *
  38 |  * Let's warn if we find service workers in development.
  39 |  */
> 40 | if (`serviceWorker` in navigator) {
  41 |   navigator.serviceWorker.getRegistrations().then(registrations => {
  42 |     if (registrations.length > 0)
  43 |       console.warn(
```
Will pick up from there eventually

ncphillips:
closing but leaving the branch

PR #315, opened 2019-10-17
Add missing step & fix path in Readme "To get started"

adamsilverstein:
* Fix an error when setting up for the first time and running `npm run bootstrap` without first running `npm install`. 

> lerna ERR! ENOLERNA `lerna.json` does not exist, have you run `lerna init`?

* Correct path into demo folder
* Fix similar missing package error when running demo without first running npm install:

> Error: Unable to find plugin "gatsby-tinacms-json". Perhaps you need to install its package?

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,11 +23,11 @@ To get started:
 ```bash
 git clone git@github.com:tinacms/tinacms.git
 cd tinacms
-npm run bootstrap
+npm install && npm run bootstrap
 
 # Start the Gatsby demo
-cd packages/demo/demo-gatsby
-npm run start
+cd packages/demo-gatsby
+npm install && npm run start
 ```
 
 ### Commands


spbyrne:
`npm run bootstrap` installs dependancies; `npm install` isn't a required step for either case. It seems the error you're encountering is coming from something else. I'm going to troubleshoot.

README.md
```diff
@@ -23,11 +23,11 @@ To get started:
...
 git clone git@github.com:tinacms/tinacms.git
 cd tinacms
-npm run bootstrap
+npm install && npm run bootstrap
 
 # Start the Gatsby demo
-cd packages/demo/demo-gatsby
-npm run start
+cd packages/demo-gatsby
+npm install && npm run start
```
ncphillips: Thanks a lot for the PR @adamsilverstein 😄  

This `npm install` will actually break the development environment. It will replace all the local links with the actual npm packages. The issue you hit was due to an issue with the demo site not referencing the local version numbers. I have #314 open to prevent that from happening again, and I will fix that blow up shortly.

```suggestion
npm run start
```
```

spbyrne: I'll get this updated after merge. Thanks @adamsilverstein!

Issue #307, opened 2019-10-14
Formatted Date not showing on initial load


kendallstrautman:
## Description

Upon passing in date field options to a form, on initial load the unformatted date string shows. If you open the date picker and select a new date, the new formatted date will be reflected unless you navigate to another page or refresh and you'll see the raw string again.

### Steps to reproduce

Pass in formatDate options to the date field in gatsby-demo, open the sidebar.

### Expected result

I would expect to see the formatted date filling the input.

### Actual result

The raw unformatted date string fills the input until a date is selected.

jamespohalloran:
PR open: https://github.com/tinacms/tinacms/pull/322

ncphillips:
Closed in #322

Issue #257, opened 2019-10-08
Date picker gets crazy


DirtyF:
## Description

Hours and minutes are spinning endlessly in the date picker

### Steps to reproduce

1. Click on a date field
2. Try to click on hours or minutes

### Expected result

Increment or decrement hours and minutes

### Actual result

Infinite loop

![tina-crazy-date-picker](https://user-images.githubusercontent.com/103008/66414894-9ad11400-e9fa-11e9-8dc0-c444d46da3e0.gif)

### Environment

latest macOS/Firefox

```
@tinacms/gatsby-plugin-tinacms": "next"
@tinacms/gatsby-tinacms-git": "next"
@tinacms/gatsby-tinacms-json": "next"
@tinacms/react-tinacms-remark": "next"
```

kendallstrautman:
Assuming this was fixed in #322, please reopen if behavior persists

PR #322, opened 2019-10-17
Date format fix

jamespohalloran:
Fix for: https://github.com/tinacms/tinacms/issues/307 
Also:
- Uses 'format' from plugin if not set on field
- Adds "field" as a param for 'parse' & 'format'.
- Hides TimeFormat by default (because it's currently buggy, and will require more attention)
- Changes default date display to: MMM DD, YYYY

### DIFF
--- a/packages/tinacms/src/plugins/fields/DateFieldPlugin.tsx
+++ b/packages/tinacms/src/plugins/fields/DateFieldPlugin.tsx
@@ -26,8 +26,7 @@ import { useEffect, useState, useRef } from 'react'
 import { useFrameContext } from '../../components/SyledFrame'
 import styled from 'styled-components'
 import { InputCss } from '@tinacms/fields'
-import * as _moment from 'moment'
-const moment = _moment //https://github.com/jvandemo/generator-angular2-library/issues/221#issuecomment-355945207
+import { format, parse } from './dateFormat'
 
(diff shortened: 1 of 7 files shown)


packages/tinacms/src/plugins/fields/DateFieldPlugin.tsx
```diff
@@ -57,6 +58,7 @@ export const DateField = wrapFieldsWithMeta<InputProps, DatetimepickerProps>(
             onFocus={input.onFocus}
             onChange={input.onChange}
             open={isOpen}
+            timeFormat={false}
```
ncphillips: Why default this to false?

jamespohalloran: It's a bit of a workaround. 
[react-datetime](https://github.com/YouCanBookMe/react-datetime) uses a separate dateformat and a timeformat, but when we [export the date value](https://github.com/tinacms/tinacms/blob/1d168ee01a774e1a58d4d7868c1161b13ea2d7a5/packages/tinacms/src/plugins/fields/dateFormat.ts#L36), we need a format with date & time combined.

The timepicker & non-utc-timezones also can cause issues (https://github.com/tinacms/tinacms/issues/257) with the picker, so I thought disabling it by default would simplify things for the majority of use-cases.

PR #323, opened 2019-10-18
demo: package deps added

mddanishyusuf:
There is one dependency is missing for "demo-gatsby" and that is "gatsby-tinacms-json".

image


jamespohalloran:
From the diff, it looks like the packages are the same before and after. Did you run `npm run bootstrap` on the project root beforehand?

mddanishyusuf:
> 
> 
> From the diff, it looks like the packages are the same before and after. Did you run `npm run bootstrap` on the project root beforehand?

Yes, I follow the way is giving on the README.

spbyrne:
I tested with a fresh clone & following the readme instructions and everything is working correctly. It may have been resolved with the addition of `npm install` or with the addition of `--force-local` to the bootstrap command.

I noticed `npm run build` is missing from the instructions; I'll add that.

Issue #256, opened 2019-10-08
Wrong date displayed


DirtyF:
## Description

Date displayed is incorrect on site on form update.

<img width="1150" alt="tina-cms-date-update" src="https://user-images.githubusercontent.com/103008/66414170-2a75c300-e9f9-11e9-9fb1-80ce48608e2c.png">

### Steps to reproduce

1. Open Sidebar on a blog post on tinacms-site
2. Change date to October 16th

### Expected result

October 16 is displaying on the site

### Actual result

October 15 is displayed instead

### Environment

Timezone CEST.

jamespohalloran:
PR: https://github.com/tinacms/tinacms/pull/326

kendallstrautman:
Mentioned in this issue - https://github.com/tinacms/tinacms/issues/545

spbyrne:
When I revert the change made in PR #326 (setting the `utc` prop on `ReactDatetime`) the date-picker works for me and selects the correct date instead of the previous day. I'll try to break down my understanding of what's going on.

The UTC prop does the following: "When true, input time values will be interpreted as UTC (Zulu time) by Moment.js. Otherwise they will default to the user's local timezone.". 

My local timezone is ADT, which is UTC - 3 hours. I tested with the prop both true and false, by clicking "April 1st". I expected my local time (UTC set to false) to be 3 hours behind, but it didn't work like that:

If UTC is set to true, "April 1st" will return `2020-04-01T00:00:00.000Z`
If UTC is set to false, "April 1st" will return `2020-04-01T03:00:00.000Z`

Regardless of what's returned, the parse function in `dateFormat.ts` seems to convert it as if going from UTC to local time. So in my case, that means going back 3 hours.

With UTC set to true, the returned date is actually midnight and so 3 hours behind is the previous day. With UTC set to false, the parsed date lands on midnight of April 1st.

I don't understand everything going on here so I just wanted to show my findings. I think removing this UTC prop could solve the issue in most cases, but I also think there's something strange going on with the parse function that could also solve the problem.

PR #326, opened 2019-10-18
improvement: Use utc as default for datepicker

jamespohalloran:
Displaying in the users timezone causes issues with the datepicker: https://github.com/tinacms/tinacms/issues/256

E.g When a user chooses Jan 26, it might then display Jan 25, or Jan 27 depending on the current time of day, and the browser's timezone setting. 

If a user wants to configure the display setting to use the user's timezone they can, but this will fix issues for the majority of users.


ncphillips:
🤙maybe add a comment above that line either explaining the rationale, or just linking to this issue

Issue #209, opened 2019-10-02
Allow commit+push to be configured


ncphillips:
Right now every commit is paired with a push. This is required for production use, but it is not always wanted when editing files locally. The developer should be able to configure the git-api, or the git-client, to decide when to push.

lyleunderwood:
For my use case it would be really nice if some kind of pull request-oriented workflow could be figured out. I think what would really make sense for me would be branch management. If we could:

- Change branch from the tina sidebar (this may be somewhat feasible currently using custom fields, I haven't looked into it yet), and
- Protect branches, `master` for example, forcing the user to select a new / different branch.

It would also be nice to be able to prompt the user for a commit message when saving.

If I had those tools I could probably cobble together a workflow to open pull requests automatically using github actions.

Overall it seems like there is lots of room for expansion with the git integration.

ncphillips:
Great suggestions @lyleunderwood!

I think each of these deserves they're own feature request:

* #332 - Change branch from the tina sidebar 
* #333 - Could make protected branches possible
* #334 - Commit messages

abtris:
I can point that I have pre-commit hooks automatically for every repo and that blocks any remark form and you get just ERROR: `fatal: Invalid object name 'HEAD'.` that isn't helpful.  I don't think so that  ERROR `fatal: No configured push destination.` is nice too and blocking edit function for this doesn't make sense.

lyleunderwood:
Yes, pushing should definitely be configurable. Would it make sense to make pushing a separate action?

ncphillips:
@abtris the error about `No configured push destination` occurs when the branch you're on is not in the remote. The `Invalid object name 'Head'` is less familiar, and unrelated to this issue in particular. Could you open a new Bug Report with as much detail as possible about the setup of your repo (i.e. pre-commit hooks) and the actions you're performing?

@lyleunderwood I think it depends on who the user is. If I was writing content locally I would want commit and push to be separate. But if I was setting up a site for a non-technical client using [TinaTeams](https://tinacms.org/teams/) then I would want them to be coupled. 

A potential API could look like:

```js
export default remarForm(BlogPostTemplate, {
  pushOnCommit: true
})
```

Then we could use some kind of environment variable for Tina Teams:

```js
export default remarForm(BlogPostTemplate, {
  pushOnCommit: process.env.NODE_ENV === "tinateams"
})
```

lyleunderwood:
[...]

PR #349, opened 2019-10-22
feat(git): implement pushOnCommit option

lyleunderwood:
/cc #209 

@ncphillips Does this approach make sense to you?

### DIFF
--- a/packages/@tinacms/api-git/src/commit.ts
+++ b/packages/@tinacms/api-git/src/commit.ts
@@ -24,6 +24,7 @@ export interface CommitOptions {
   message: string
   name?: string
   email?: string
+  push?: boolean
 }
 
 /**
@@ -35,6 +36,7 @@ export async function commit({
   name,
   email,
   pathRoot,
+  push,
 }: CommitOptions) {
   let options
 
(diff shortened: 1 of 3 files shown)


jamespohalloran:
Nice! I'm curious what others think the default for `PUSH_ON_COMMIT` should be. Maybe `false` until we validate it a bit more?

lyleunderwood:
Right now it's just emulating the current behavior, so making it `false` would actually be a breaking change.

jamespohalloran:
> Right now it's just emulating the current behavior, so making it false would actually be a breaking change.
Ah, my bad! you're totally right.

dwalkr:
+1 for the separate `push` route! This is looking pretty good to me. What else needs to be done to get this out of draft status?

ncphillips:
I'm feeling the same way. Nice job @lyleunderwood 👍

lyleunderwood:
I was thinking about trying to tag some UI onto this, but based on discussions elsewhere I'll probably hold off. So I'll probably just update documentation and whatnot, then take it out of draft.

lyleunderwood:
Hmm, I can't seem to find any documentation on the _current_ config options, so I guess I'm not gonna worry about that right now.

PR #353, opened 2019-10-23
refactor: move `createJestConfig` into the @tinacms/scripts package

### LINKED ISSUE #265: Move `createJestConfig` into the @tinacms/scripts package
## Summary

the createJestConfig helper function is declared at the root of the repo. It should be moved into `

### Basic example

Currently, using that script involves making a require outside of the package dir: 

```javascript
const createJestConfig = require('../../create.jest.config.js')
const pack = require('./package')

module.exports = createJestConfig(pack)
```

With this change it will be imported like so:

```javascript
const { createJebstConfig } = require('@tinacms/scripts')F
const pack = require('./package')

module.exports = createJestConfig(pack)
```

### Motivation

It's a little messy and fragile to import it the other way.

bolariin:
As requested in the issue create.jest.config.js was moved into @tinacms/scripts
and all files that referenced `createJestConfig` were updated.

closes #265


ncphillips:
Hey @bolariin thanks a lot for opening this PR!

Could you confirm that executing `npm run test` from the root of the repository does in fact run the tests? 

`@tinacms/scripts` doesn't have a `main` so  I don't see how it could be exporting the `createJestConfig` function. And since the and the build checks didn't run on your PR, I'm not sure that they still work.

bolariin:
I encountered an error when I run the tests. You mentioned I'm missing `main`. I'm new to this, would it possible to provide more detail on what you mean?

I'm not sure `npm run test` works as expected because when I run it on master, I see a lot of `No tests found, exiting with code 0`.

ncphillips:
Of course! 

From the [npm docs](https://docs.npmjs.com/files/package.json#main): 

```
The main field is a module ID that is the primary entry point to your program. That is, if your package is named foo, and a user installs it, and then does require("foo"), then your main module’s exports object will be returned.

This should be a module ID relative to the root of your package folder.
```



You might set `"main": "index.js"` and have it's contents be:

```js
exports.createJestConfig = require("./create.jest.config.js")
```

Don't take my word on it though, this might not be exactly right

> I'm not sure npm run test works as expected because when I run it on master, I see a lot of No tests found, exiting with code 0.

That's okay. Not all the packages have tests 😬

ncphillips:
Sidenote: `npm run test` is intentionally quiet. Running `lerna run test --stream` will print the full output

bolariin:
Thanks a lot for explaining, the quiet nature was definitely confusing at first.

PR #354, opened 2019-10-23
update release instructions

dwalkr:
`git push` does not push tags. Need to run `git push` to push changelog commits and `git push --tags` to push tags.


ncphillips:
Does `git push --tags` only push the tags, and not any commits?

DirtyF:
@ncphillips I _think_ so

dwalkr:
yes

PR #355, opened 2019-10-23
fix: add check if filepath isn't outside git repository

### LINKED ISSUE #250: @tincms/api-git should not allow files outside of the repo to be changed
## Summary

When creating, writing, or deleting files using the `@tinacms/api-git` endpoints, it should not be possible to create a file outside of the repo. 

Further control over where content can and cannot be written to may be required in the future, but this is a good first start.

## Basic Example

The following call should be rejected from the server:

```js
cms.api.git.writeToDisk({
  fileRelativePath: '../some-outside-file.json',
  contents: 'some-malicious-data',
])
```

### Motivation

This is a rather important security concern.

abtris:
- Try resolve #250 
- Can't find test or where I can put tests.


ncphillips:
Thanks for the PR @abtris! 

We unfortunately don’t have tests for this yet, but we put them beside the file under test.

I will give this a proper review tomorrow morning

abtris:
```
$ ./node_modules/.bin/jest
 PASS   @tinacms/api-git  src/file-writer.test.ts
  ✓ Check path if is in git repo - bad (3ms)
  ✓ Check path if is in git repo - good (1ms)

Test Suites: 1 passed, 1 total
Tests:       2 passed, 2 total
Snapshots:   0 total
Time:        3.103s
Ran all test suites.
```
Local jest works, but `npm run test` doesn't look like run this test.

abtris:
This can be related https://github.com/tinacms/tinacms/pull/353

packages/@tinacms/api-git/src/file-writer.ts
```diff
@@ -25,13 +25,26 @@ let count = 0
...
+}
 
+export function writeFile(filepath: string, content: string, repoAbsolutePath: string) {
+  count++
+  if (DEBUG) console.info(`checkFilePathIsInGitRepo? ${checkFilePathIsInGitRepo(filepath, repoAbsolutePath)}`)
   if (DEBUG) console.info(`request ${count} received`)
-
-  cacheCommand(filepath, content)
-  tryToWrite()
+  if (checkFilePathIsInGitRepo(filepath, repoAbsolutePath)) {
```
ncphillips: I would prefer to move `checkFilePathIsInGitRepo` out of `file-writer.ts` and into `router.ts`

Currently `writeFile` is a low level API in the domain of the filesystem. It just _eventually_ writes the given string to the given path. 

This PR introduces the concept of a git repository to `writeFile`

Instead, undo the changes here and move the check up so that `router.ts` looks something like:

```typescript
    try {
      const fileIsInRepo = checkFilePathIsInGitRepo(
        fileAbsolutePath,
        REPO_ABSOLUTE_PATH
      )
       
      if (fileIsInRepo) {
        writeFile(fileAbsolutePath, req.body.content)
      }
      res.json({ content: req.body.content })
    } catch (e) {

File: packages/@tinacms/api-git/src/file-writer.ts [hunk 2]
```diff
@@ -25,13 +25,26 @@ let count = 0
 
 const MAX_BUILD_TIME = 1000
 
-export function writeFile(filepath: string, content: string) {
-  count++
+export function checkFilePathIsInGitRepo(filepath: string, repoAbsolutePath: string) {
+  const fullpath = path.resolve(filepath)
+  const repopath = path.resolve(repoAbsolutePath)
+  if (fullpath.includes(repopath)) {
```
ncphillips: `fullPath.startsWith(repopath)` might be safer. Might have to strip any leading `/` though

packages/@tinacms/api-git/src/file-writer.test.ts
```diff
@@ -0,0 +1,13 @@
+import { checkFilePathIsInGitRepo } from './file-writer'
+
+test('Check path if is in git repo - bad', () => {
```
ncphillips: I would wrap these tests with a `describe` to make the output more readable:

```ts
describe("checkFilePathIsInGitRepo", () => {
  test('returns false if path not in repo', () => {
    // ...
  })

  test('returns true if path in repo', () => {
    // ...
  })
})
```

packages/@tinacms/api-git/src/file-writer.ts
abtris: Sure, not issue with startsWith. Not sure issue with leading `/`? Can you be more specific what you think that can be issue, we can cover as test case.

packages/@tinacms/api-git/src/file-writer.ts
abtris: ok, I wil move into router.

packages/@tinacms/api-git/src/file-writer.test.ts
abtris: ok, I don't have experience with jest this is looks like mocha ;-)

abtris:
I will squash change at the end after you approve final state.

packages/@tinacms/api-git/src/router.ts
```diff
@@ -80,7 +94,13 @@ export function router(config: GitRouterConfig = {}) {
       console.log(fileAbsolutePath)
     }
     try {
-      writeFile(fileAbsolutePath, req.body.content)
+      const fileIsInRepo = checkFilePathIsInGitRepo(
+        fileAbsolutePath,
+        REPO_ABSOLUTE_PATH
```
ncphillips: I just realized this should be:

```suggestion
        CONTENT_ABSOLUTE_PATH
```

If the site is in a subdirectory (i.e. if you have multiple sites in one repo) you shouldn't be able to edit content outside of that site.

ncphillips: It would be good to update the variable names (i.e. `fileIsInRepo`) accordingly.

packages/@tinacms/api-git/src/router.ts
```diff
@@ -80,7 +94,13 @@ export function router(config: GitRouterConfig = {}) {
...
+      )
+      if (fileIsInRepo) {
+        writeFile(fileAbsolutePath, req.body.content)
+      }
```
ncphillips: Right now this operation silently fails. It should instead throw an error. Something like:

```suggestion
      } else {
        throw new Error(`Failed to write to: ${fileRelativePath} \nCannot write outside of the content directory.`)
      }
```

abtris:
I hope that I fix everything.

packages/@tinacms/api-git/src/router.ts
abtris: what you suggest as name? `fileIsInSite` ?

ncphillips:
The tests failed when merged to master. Next time please make sure to run the tests after refactoring.

abtris:
Will be better add circleci to run tests?

PR #368, opened 2019-10-24
build(danger): check (js|tsx?) files for license

ncphillips:
Dangerfile checks js and ts files for the license header. Fails if they do not exist.

## TODO

- [ ] Make work on all lines of license

Currently fails on:

> Licensed under the Apache License, Version 2.0 (the "License");

Seems like the 2.0 makes the regex not work


ncphillips:
```
`Licensed under the Apache License, Version 2.0 (the "License");`.match(
    new RegExp(`Licensed under the Apache License, Version 2.0 (the "License");`
)  // false

`Licensed under the Apache License, Version 2.0 (the "License");`.match(
    new RegExp(`Licensed under the Apache License, Version`
) // true
```

I think it has something to do with `2.0`. I’m guess I need to escape it somehow.

ncphillips:
@swhalley helped me figure it out

Issue #370, opened 2019-10-24
Many lint warnings


lyleunderwood:
176 warnings, 30 potentially auto-fixable.

Is the current eslint config actually what's desired, or are decisions still being made about that? I was thinking about opening PRs to remove lint warning package-by-package.

ncphillips:
The eslint config was added late. We fixed the errors and left warnings due to time constraints. We definitely need to go through and fix them at some point. If you're up for doing that, it would be greatly appreciated! 

But we need to switch `"react-hooks/exhaustive-deps": "warn"` off entirely. It is a bad rule. The assumption that you want to re-run for all deps is just not correct. Doing so breaks tinacms, so until that's turned off please don't run the auto-fixer.

PR #372, opened 2019-10-25
chore(linting): Turn off react-hooks/exhaustive-deps

lyleunderwood:
https://github.com/tinacms/tinacms/issues/370#issuecomment-546142922

_According to @ncphillips:_

> we need to switch `"react-hooks/exhaustive-deps": "warn"` off entirely. It is a bad rule. The assumption that you want to re-run for all deps is just not correct. Doing so breaks tinacms, so until that's turned off please don't run the auto-fixer.


ncphillips:
@all-contributors please add @lyleunderwood for code

Issue #382, opened 2019-10-27
On Windows - unable to save edits


teyc:
## Description

Attempting to edit a page results in failure when tinacms is deployed on Windows.

### Steps to reproduce

1. `gatsby new my-tinacms-starter https://github.com/tinacms/gatsby-starter-tinacms`

2. `cd gatsby-starter-tinacms`

3. `gatsby develop`

4. Select a page

5. Edit HTML

### Expected result

What should happen?

HTML successfully edited, and see a new git commit.

### Actual result

```
error: pathspec 'C:\dev\toyapps\my-tina-starter\C:\dev\toyapps\my-tina-starter\content\blog\hello-world\index.md' did not match any file(s) known to git
```

### Environment

- Windows 10

- node v12.13.0

### Notes

1.  The directory name was repeated twice 

     'C:\dev\toyapps\my-tina-starter\C:\dev\toyapps\my-tina-starter\content\blog\hello-world\index.md'

2.  The `fileRelativePath` does not start with `/content` on Windows

```{
  "componentChunkName": "component---src-templates-blog-post-js",
  "path": "/hello-world/",
  "webpackCompilationHash": "",
  "result": {
    "data": {
      "site": {
        "siteMetadata": {
          "title": "Gatsby Starter Blog",
          "author": "Kyle Mathews"
        }
      },
      "markdownRemark": {
        "id": "8c658351-c0c8-5559-83e7-64fb0d699edb",
        "excerpt": "This is my first post on my new fake blog! How exciting! I’m sure I’ll write a lot more interesting things in the future. Oh, and here’s a great quote from this…",
        "html": "<p>This is my first post ...</p>",
        "fileRelativePath": "C:/dev/toyapps/my-tina-starter/content/blog/hello-world/index.md",
        "rawFrontmatter": "{\"title\":\"Hello World\",\"date\":\"2015-05-01T22:12:03.284Z\",\"description\":\"Hello World\"}",
        "rawMarkdownBody": "\r\nThis is my first post on...\r\n",
        "frontmatter": {
          "title": "Hello World",
          "date": "May 01, 2015",
          "description": "Hello World"
        }
      }
    },
    "pageContext": {
      "isCreatedByStatefulCreatePages": false,
      "slug": "/hello-world/",
      "previous": null,
      "next": {
        "fields": {
          "slug": "/my-second-post/"
        },
        "frontmatter": {
          "title": "My Second Post!"
        }
      }
    }
  }
}
```

[...]

PR #383, opened 2019-10-27
Support Gatsby on Windows #382

teyc:
On Windows the resolution for `fileRelativePath` was  incorrect because the 
`source.fileAbsolutePath` has been normalised to use forward slashes on both `win32` and `posix`.
So that it reads like
    
    C:/project/tinacms/content/hello-world/index.md

To correctly calculate the relativePath, on windows, we need to turn `process.cwd()`
from 

    C:\project\tinacms\

to

    C:/project/tinacms/

### DIFF
--- a/packages/gatsby-tinacms-remark/gatsby-node.js
+++ b/packages/gatsby-tinacms-remark/gatsby-node.js
@@ -15,10 +15,11 @@ See the License for the specific language governing permissions and
 limitations under the License.
 
 */
-const { GraphQLString } = require("gatsby/graphql")
+const { GraphQLString } = require("gatsby/graphql");
+const slash = require("slash");
 
 exports.setFieldsOnGraphQLNodeType = ({ type }) => {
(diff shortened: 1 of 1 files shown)


ncphillips:
Looks good, but we’ll test this out on Mac before approving.

Also I noticed your commit messages don’t follow the conventional changelog format. Could you rebase and add the `fix:` prefix to the commit message?

Go to this page and to the heading “Amending the message of older or multiple commit messages”

Thanks!

https://help.github.com/en/articles/changing-a-commit-message

https://github.com/tinacms/tinacms/blob/master/CONTRIBUTING.md#making-commits

ncphillips:
I pulled this down and tested things out locally. 

Everything looks good! We'll approve once the commit message has been ammended.

ncphillips:
Thanks @teyc! I'm merging your change into `fix-windows-git-support` to confirm that CI still passes.

PR #409, opened 2019-10-30
WIP - fix(git): fix comitting when global name & email isn't set

jamespohalloran:
DO NOT MERGE - 
We may want these set through environment variables instead of gatsby-config settings. Needs discussion

If you don't have a global git `user.name` & `user.email` set, the commit will fail.
This sets your repo's git committer username & email. It might not be ideal to set this on your repo, but I haven't had any luck setting it on an individual commit

### DIFF
--- a/packages/@tinacms/api-git/src/commit.ts
+++ b/packages/@tinacms/api-git/src/commit.ts
@@ -38,14 +38,16 @@ export async function commit({
   pathRoot,
   push,
 }: CommitOptions) {
+  const repo = openRepo(pathRoot)
+
   let options
   if (email) {
     options = {
       '--author': `"${name || email} <${email}>"`,
     }
+    repo.addConfig('user.name', name || email)
+    repo.addConfig('user.email', email)
   }
-
-  const repo = openRepo(pathRoot)
 
   await repo.add(files)
   const commitResult = await repo.commit(message, files, options)


ncphillips:
I’m pretty darn sure this _was_ working. Do you have any ideas @dwalkr?

We were doing this but it ends up changing in the remote repo. This means commits made by anyone, with or without Tina, will be “made” by that person.

dwalkr:
I just looked this up on a hunch, and it seems like when you commit with the `--author` option git will still complain if you don't have a `user.email` set.

https://stackoverflow.com/questions/22058041/commit-without-setting-user-email-and-user-name

dwalkr:
per the linked thread, the best solution might be to specify the config inline with `-c`. Would be worth looking into whether we can use simple-git's `raw` method to run commands this way.

Alternatively, we can set the config (or use env vars mentioned in thread) to set a global default, and individual identities should still work when using `--author`

dwalkr:
To clarify, using the `--author` option should override any values in git config for the specified commit. But you need to have _something_ set in the git config (whether inline, via env var, or via `git config`) or it won't let you commit even if you specify name/email via `--author`

dwalkr:
Something to chew on: author and committer are two different things in git lingo. That's why specifying author via the `--author` option doesn't work without also configuring an identity via `user.name`/`user.email`, as the latter is configuring the identity of the committer, not necessarily the author.

It may make sense to always set the committer to some default value (in the case of teams, possibly `teams@tinacms.org` or something,) since the software is what's doing the committing, and authors aren't always aware that they're making commits (by design)

ncphillips:
I'm gonna close this since it's more in the scope of Tina teams

PR #418, opened 2019-10-31
fix: early return for empty watch values

### LINKED ISSUE #410: useCMSForm hook not respecting initialValues
First of all, thanks for this project - it's a step in the right direction for the JAM stack.

I'm filling this as a bug report as that's the way I would expect the `useCMSForm` hook to work but perhaps I misunderstood something as it's only mentioned briefly in the documentation.

## Description

`initialValues` passed to the `useCMSForm` hook are not respected in this minimal example:

```jsx
import React from "react"
import { useCMSForm } from "react-tinacms"

export default () => {
  const [values] = useCMSForm({
    id: 'form-identifier',
    fields: [{ name: "hideText", label: "Hide text", component: "toggle" }],
    initialValues: {
      hideText: true
    },
    label: "Hide text",
    onSubmit: (v) => { console.log("Submitted with: ", v) }
  })
  console.log("Rendered with: ", values)

  return <div>
    {!values.hideText &&
      <div style={{ marginLeft: '50%' }}>
        You shouldn't see me on startup
      </div>}
  </div>
}
```

The same goes for more complex forms - all passed `initialValues` are zeroed out and only constructed when the user interacts with the form.

### Steps to reproduce

1. Create a project using the [official Gatsby Quickstart](https://tinacms.org/docs/gatsby/quickstart)
2. Add a `page` with code as above

### Expected result

- `values.hideText` should be `true` when no changes are made through the UI
- `Rendered with: {hideText: true}` should be printed to the console anytime the component renders
- Submitting the form without making any changes should output `Submitted with: {hideText: true}`

### Actual result

The component gets rendered three times and while the first two renders happen with the proper value, with the third render `values` is an empty object.

### Environment

Tested on Linux and Windows with both versions @0.5.0 and @0.6.0-alpha.0 on top of an unmodified Gatsby quickstart

dwalkr:
closes #410

### DIFF
--- a/packages/react-tinacms/src/use-form.ts
+++ b/packages/react-tinacms/src/use-form.ts
@@ -104,7 +104,7 @@ function updateFormLabel(label?: string, form?: Form) {
  */
 function updateFormValues(values: any = {}, form?: Form) {
   React.useEffect(() => {
-    if (!form || Object.keys(values).length == 0) return
+    if (!form || typeof values === 'undefined') return
     form.updateValues(values)
   }, [form, values])
 }
--- a/packages/react-tinacms/src/use-form.ts
+++ b/packages/react-tinacms/src/use-form.ts
@@ -102,7 +102,7 @@ function updateFormLabel(label?: string, form?: Form) {
  *
  * This hook is useful when the form must be kept in sync with the data source.
  */
-function updateFormValues(values: any = {}, form?: Form) {
+function updateFormValues(values: any, form?: Form) {
   React.useEffect(() => {
     if (!form || typeof values === 'undefined') return
     form.updateValues(values)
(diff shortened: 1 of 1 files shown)


packages/react-tinacms/src/use-form.ts
```diff
@@ -104,7 +104,7 @@ function updateFormLabel(label?: string, form?: Form) {
  */
 function updateFormValues(values: any = {}, form?: Form) {
   React.useEffect(() => {
-    if (!form || typeof values === 'undefined') return
+    if (!form || Object.keys(values).length == 0) return
```
ncphillips: This sort of changes the semantics.

Did you see if removing the `= {}` from the arguments worked?

i.e.

```ts
function updateFormValues(values: any, form?: Form) {
```

dwalkr: yeah I guess that makes much more sense, updated

PR #440, opened 2019-11-04
fix: resets active form to null for multiple list

### LINKED ISSUE #438: When navigating btw pages, multiple forms defaulting to show first child
## Description

If you have multiple forms on a page, the default landing behavior is wonky. It might land on the list of forms at first, but as you navigate the default seems to be the first child of the array of forms. Let's look at the logic of how these multiple forms get rendered in the form view. 

### Steps to reproduce

Start up the grande project - https://github.com/tinacms/tina-starter-grande
Open the sidebar and navigate to other pages. 
Check out the initial view of a form when you land on a page.

### Expected result

When there are multiple forms and you navigate to to a new page with the sidebar open, the list of the multiple forms should show as the default.

### Actual result

It seems like the first child in the list is rendering.

### Environment
Firefox + MacOS
"gatsby-tinacms-git": "^0.2.6-alpha.0",
 "gatsby-tinacms-json": "^0.5.1-alpha.0",
 "gatsby-tinacms-remark": "^0.3.6-alpha.0",

kendallstrautman:
Should fix #438 but opens up new issue

From the thread:
the original problem

The problem is that the number of forms registered with Tina doesn't happen all at once. SoFormsView mounts and at first it thinks it has no forms, then it thinks it has one, then it finally knows it has 2 etc. So essentially this useSubscribable will set the form id when it thinks it has one form and won't update it again even if the component registered more forms down the line. If we add this reset to null, it will fix the initial problem... but opens up a new issue.
the new problem

The problem we run into now is that a flash of the individual form happens when the component only registers one component, but it finally lands on the list. Thoughts on how we could mitigate this?
overarching question

Could we somehow wait to render the FormView until all the forms have been registered to prevent render flashes?

Boot up the gatsby-demo to see the flash I am talking about.

Don't merge until we figure out a way to prevent the flash...
@spbyrne thoughts?

### DIFF
--- a/packages/tinacms/src/components/FormView.tsx
+++ b/packages/tinacms/src/components/FormView.tsx
@@ -42,15 +42,18 @@ import { FORM_HEADER_HEIGHT, SIDEBAR_HEADER_HEIGHT } from '../Globals'
 import { GroupPanel } from '../plugins/fields'
 
 export const FormsView = () => {
-  const [activeFormId, setActiveFormId] = useState<string>()
+  const [activeFormId, setActiveFormId] = useState()
   const cms = useCMS()
 
   /**
    * If there's only one form, make it the active form.
+   * If there's multiple forms, set to null so the list shows.
    */
   useSubscribable(cms.forms, () => {
     if (cms.forms.all().length === 1) {
       setActiveFormId(cms.forms.all()[0].id)
+    } else {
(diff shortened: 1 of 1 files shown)


kendallstrautman:
This is stale. Closing until we find a solution.

PR #463, opened 2019-11-13
Add loading indicator on save proccess

maciekgrzybek:
This PR is connected to the [ticket](https://github.com/tinacms/tinacms/issues/4490). 

feat: Add loading indicator on saving

- create new LoadingDots components
- add submitting props to the component (coming from final-form library)
- add conditional content rendering in save button

---
There were also a conversation about adding some kind of 'Saved' indicator, but I think that might be a separate PR if the team agrees on that.


spbyrne:
I'll test this out. Thanks for adding this! Definitely a much needed feature.

spbyrne:
This looks great! I like the animation.

PR #464, opened 2019-11-13
Add saving indicator for save button

spbyrne:
@maciekgrzybek created a loading dots component that can be used to indicate something is happening after a button press for asynchronous actions and implemented it on the form save button.

I also added a `busy` state to the button component. Right now it only sets `cursor: wait;`, but we could add to that in the future.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5271,9 +5271,9 @@
 			}
 		},
 		"abab": {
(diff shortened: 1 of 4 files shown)


ncphillips:
@all-contributors please add @maciekgrzybek for code

Issue #460, opened 2019-11-12
gatsby-tinacms-remark fails inelegantly if gastby-transformer-remark is missing


ncphillips:
## Description

`gatsby-tinacms-remark` should have `gatsby-transformer-remark` as a peerDependency

### Steps to reproduce

* Create a site that does not use `gatsby-transformer-remark`
* Run `yarn add gatsby-tinacms-remark` 

### Expected result

It should warn about `gatstby-transformer-remark` being a peer dependency

### Actual result

No peer dependency warnings.

Trying yo start the server returns this error

```
 ERROR #85901  GRAPHQL

There was an error in your GraphQL query:

Unknown type "MarkdownRemark".

File: node_modules/gatsby-tinacms-remark/src/remark-fragment.ts:22:26

failed extract queries from components - 0.191s
```

maciekgrzybek:
I could take a look at this one if it's ok.

ncphillips:
That would be great! One other thing: the error should also say that it needs to be listed as a plugin

maciekgrzybek:
Ok I'll add it then, thanks :)

maciekgrzybek:
So @ncphillips just to clarify, do we want to add the message when installing the tina plugin? or on the build?

ncphillips:
At the very least during install and ideally when running as well.

Setting it as a `peerDependency` would automatically make it show up when installing the plugin.

I'm not sure how you could throw the error whens starting though.

maciekgrzybek:
I've added the PR - https://github.com/tinacms/tinacms/pull/483

defaultz12:
I am still seeing Markdown issue using gatsby.
- "gatsby-tinacms-remark": "^0.5.1",
- "gatsby-transformer-remark": "^2.6.0", and ^2.6.38

maciekgrzybek:
Is it the same error as mentioned [here](https://github.com/tinacms/tinacms/issues/460#issue-521793473) ?

arielwebriq:
I figured it out. I was doing the manual set-up. It needs to have a gatsby markdown ready before setting up tinaCMS.

witcradg:
> I figured it out. I was doing the manual set-up. It needs to have a gatsby markdown ready before setting up tinaCMS.

This needs to be documented. It bit me too.

ncphillips:
I'm not sure anything more can be done for this. Closing for now.

PR #483, opened 2019-11-19
Remark plugin peer dependency

maciekgrzybek:
This is PR connected to [issue](https://github.com/tinacms/tinacms/issues/460).


ncphillips:
We may want to loosen that. e.g. `>=2.6`

Kind of an aside, it would be good to know what the actual acceptable version range is. For example, does it work with 2.2? What about 1.0? This investigation could be a separate issue & PR

maciekgrzybek:
Hey, yeah that would make sense, just changed that. Do you want me to create a ticket to investigate the version's compatibility?

ncphillips:
Awesome! Yes, that would be greatly appreciated @maciekgrzybek 👍

PR #486, opened 2019-11-21
RFC 0001 – @tinacms/react-core

ncphillips:
Implements [RFC 0001](https://github.com/tinacms/rfcs/blob/master/0001-tinacms-react-core.md)

* Code defined in `react-tinacms` is now defined in `@tinacms/react-core`
* `tinacms` now re-exports everythign from `@tinacms/react-core`
* `tinacms` now re-exports important types from `@tinacms/core`
* `react-tinacms` has the same API as before, but just re-exports those things from `tinacms`


ncphillips:
Note: I don’t want this to go into the next release. I want to make sure to test it out on `@next` with all our starters etc.

Issue #476, opened 2019-11-18
Field description doesn't wrap


spbyrne:
## Description

When adding a description to a Tina field, it stays on one line and doesn't wrap as needed.

### Steps to reproduce

Add a long description to a field.

### Expected result

I'd expect the description to wrap if it's wider than the sidebar.

### Actual result

It stays on one line.

![image](https://user-images.githubusercontent.com/5075484/69080354-2c786c00-0a12-11ea-904e-9188deafd0b1.png)

### Environment

macOS 10.15.1, Chrome

maciekgrzybek:
I'll take a look at this one.

PR #487, opened 2019-11-21
Update final-form-builder.tsx styled component

Spraynard:
Removing `white-space: nowrap` from FieldsGroup div styling

Saw #476 and figured I could tackle that issue.

Seemed to just involve removing the `white-space` directive to go back to a `normal` default.

My guess is that this was added in order to have titles themselves not wrap? If so, that functionality is probably gone as well, but I don't know if it's a breaking change to go from no-wrap to wrap.

Here's a screenshot of the change on the gatsby demo:

Screen Shot 2019-11-21 at 3 30 26 PM


spbyrne:
Awesome work @Spraynard, thanks! I'll take a look at this tomorrow morning :smile:

Spraynard:
Thank you :)

spbyrne:
We had a discussion at one point around text wrapping. For labels it was intentional that they don't wrap; we have `text-overflow: ellipsis;` on the label element. I think if we want to support text wrapping specifically for the description, we could put `white-space: normal;` on the `FieldDescription` component in `packages/tinacms/src/plugins/fields/wrapFieldWithMeta.tsx` and leave the `final-form-builder.tsx` white-space definition as-is.

spbyrne:
Screen Shot 2019-11-22 at 11 12 07 AM

Spraynard:
Alright I think that we're all good now. I've updated `wrapFieldWithMeta.tsx` `FieldDescription` component styling to have `white-space: normal` defined.

From the looks of it, the layout that I'm getting is the same as what you were requesting. Example below.

tina-cms-resolution

Thank you very much for the guidance.

spbyrne:
Thanks @Spraynard !

spbyrne:
@all-contributors please add @Spraynard for code

PR #490, opened 2019-11-22
lock SiteWrapper Component from scrolling when sidebar is open

### LINKED ISSUE #219: Website starts scrolling if you're scrolling in the sidebar and reach the top/bottom
If you're scrolling in the sidebar and hit the top or bottom it will start scrolling the actual website. I don't think this is the expected behaviour and it's preventable with javascript.

In addition, it sometimes 'locks' the scrolling to the website so that if you begin scrolling the other direction it doesn't transfer scrolling back to the sidebar unless you interact with it in some way. This behaviour is quite jarring.

narendrasoni1989:
Toggle scrolling overflow:hidden/visible when side bar is open/closed.

Fix  #219

### DIFF
--- a/packages/tinacms/src/components/Tina.tsx
+++ b/packages/tinacms/src/components/Tina.tsx
@@ -84,7 +84,7 @@ const SiteWrapper = styled.div<{ open: boolean; position: SidebarPosition }>`
   opacity: 1 !important;
   background-color: transparent !important;
   background-image: none !important;
-  overflow: visible !important;
+  overflow: ${props => (props.open ? 'hidden' : 'visible')} !important;
   position: absolute !important;
   top: 0 !important;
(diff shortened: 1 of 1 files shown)


spbyrne:
Taking a look at this now. Thanks for the contribution! 👍

spbyrne:
I think this could be problematic as it forces you to close the sidebar in order to scroll down the page that you're editing. I think this is a good first take on a solution, but I think the problem it presents is potentially worse than the problem it solves.

I really appreciate you taking a look at this, though! The more the original issue comes up the less I think it's an 'issue' vs. simply a quirk of how nested scrolling containers work. Any solution may end up being more problematic vs. the original problem, so it's no fault of your solution.

What do you think, @ncphillips?

PR #498, opened 2019-11-27
New Package: react-tinacms-blocks

ncphillips:
`react-tinacms-blocks` provides a re-usable `Blocks` component that knows how to render data from the tinacms `BlocksFieldPlugin`. 

**Usage**

```js
import { Blocks } from "react-tinacms-blocks"

const PageBlogForm = {
  // ...
}

export default remarkForm(function Page(props) {
  const { frontmatter } = props.data.markdownRemark

  return (
    <Blocks 
      name="rawFrontmatter.blocks"
      data={props.data.markdownRemark} 
      templates={{ heading, banner, content }}
     />
  )
}, PageBlockForm)
```

**Test Output**
```
 PASS  src/blocks.spec.tsx
  Blocks
    without data
      ✓ renders nothing into the containing element (2ms)
    with some data
      when there are no templates
        ✓ renders nothing into the containing element (1ms)
        if (process.env.NODE_ENV === 'development')
          ✓ throws an error (3ms)
      when there is a matching template
        and it has Component defined
          ✓ renders that Component (2ms)
        and it does not have a Component defined
          ✓ renders nothing into the containing element
          if (process.env.NODE_ENV === 'development')
            ✓ throws an error (5ms)
```

### DIFF
--- a/packages/react-tinacms-blocks/package.json
+++ b/packages/react-tinacms-blocks/package.json
@@ -1,5 +1,4 @@
 {
-  "private": true,
   "name": "react-tinacms-blocks",
   "version": "0.0.0",
   "license": "Apache-2.0",
--- a/packages/react-tinacms-blocks/package.json
+++ b/packages/react-tinacms-blocks/package.json
@@ -1,4 +1,5 @@
 {
+  "private": true,
   "name": "react-tinacms-blocks",
   "version": "0.0.0",
   "license": "Apache-2.0",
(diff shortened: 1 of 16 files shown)


packages/react-tinacms-blocks/package.json
```diff
@@ -0,0 +1,27 @@
+{
+  "private": true,
```
ncphillips:
```suggestion
```

PR #500, opened 2019-11-28
refactor: centralize close button def to modal provider

spbyrne:
CloseButton was being defined in three places and passed to ModalHeader. ModalHeader now accepts a title string and a close function instead of taking children and will use consistent Title and CloseButton styles for all modals.

We should likely move these components out of the ModalProvider, but at least for now they're standardized and centralized.

I've also improved the contrast on the close button, making it more prominent.


spbyrne:
The `Build, Test, Lint` script is failing with this error again: 
```
semantic error TS2742: The inferred type of 'ModalHeader' cannot be named without a reference to '@tinacms/fields/node_modules/@tinacms/styles'.
```

I'm not getting that locally; are my types wrong for `ModalHeader`?

packages/tinacms/src/components/modals/ModalProvider.tsx
```diff
@@ -110,14 +111,49 @@ export const ModalOverlay = styled.div`
...
+  }
+`
+
+interface ModalHeaderProps {
+  title: string
+  close(): void
+}
+
+export const ModalHeader = styled(
+  ({ title, close, ...styleProps }: ModalHeaderProps) => {
```
ncphillips: What do you think of using `children` instead of `title`:

```jsx
<ModalHeader close={close}>Reset</ModalHeader>
```

spbyrne: Ah, that's smart & cleans it up a bit. I'll make that change.

spbyrne:
> Looks good. Left a comment for your consideration.

I've update `ModalHeader` to use `children` instead of a `title` prop. Seems like a good improvement.

spbyrne:
Thanks for you fixes 🙏 They look good to me.

PR #507, opened 2019-11-29
react-tinacms-blocks: Add support for inline editing

ncphillips:
Adds support for adding, removing, and moving blocks while editing content inline.

image


spbyrne:
Tested and working for me. This is going to be amazing. I can't wait to start styling this.

PR #525, opened 2019-12-10
Git ssh key

jamespohalloran:
Committing on Gatsby Cloud isn't working after leaving the environment idle for 1 hour. (Once the token is becoming stale). 
The user should be able to add a `SSH_KEY` env var to get around this. 

feat(git): allow user to specify SSH key for commits


packages/@tinacms/api-git/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@tinacms/api-git",
-  "version": "0.5.0-alpha.0",
+  "version": "0.5.0-alpha.1",
```
ncphillips:
```suggestion
  "version": "0.5.0-alpha.1",
```

PR #531, opened 2019-12-14
feat: allow user to user enter/return key to create new blog entry

chriswillsflannery:
A small fix that I think adds to a more seamless UX. In the form modal, upon completion of a new blog post title, user can simply press their 'return' key to create new post rather than having to click (or tab to) 'create' button


ncphillips:
Thanks for the PR Chris! I’ll take a look today 🤜🤛

ncphillips:
@all-contributors please add @chriswillsflannery for code

PR #533, opened 2019-12-15
Fix: colorpicker

### LINKED ISSUE #436: Feat: Color swatches
## Summary

Would be awesome to extend the color picker to have the option of passing in an array of colors that would comprise a swatch from which users could select.

### Basic example

We could import [Block Picker](http://casesandberg.github.io/react-color/) from react-color and pass in an array of color values given to us by the field options. Then conditionally render either the `SketchPicker` or the `BlockPicker` in `Popover`. Perhaps we add an additional flag in the field options so the dev could pass in whether they want `swatches` or `picker`.

### Motivation

 This will allow devs (& designers) to limit the color options and maintain a cohesive theme.

weibenfalk:
You can now set which picker you want and default colors as below: 

  {
      label: "Heading color",
      name: "frontmatter.heading_color",
      component: "color",
      colors: ["#ff0000", "#ffff00", "#00ff00", "#0000ff"],
      widget: "sketch"
    },

Also ... took the liberty to change to a functional component with hooks as state instead =) Have commented out the class component if you rather want to use that one. This one looks cleaner though.

Fix #436


packages/@tinacms/fields/src/ColorPicker/ColorPicker.tsx
```diff
@@ -131,6 +132,8 @@ export const Cover = styled.div`
 
 interface Props {
   colorFormat: ColorFormat
+  userColors: string[]
+  widget: string
```
ncphillips: The `widget` property is optional, and we can be more specific about which strings are acceptable

```suggestion
  widget?: "sketch" | "block"
```

packages/@tinacms/fields/src/ColorPicker/ColorPicker.tsx
```diff
@@ -140,82 +143,148 @@ interface State {
...
-    )
-  }
+  return (
+    <>
+      <Swatch
+        onClick={handleClick}
+        colorRGBA={getColorRGBA}
+        colorFormat={getColorFormat}
+      />
+      {displayColorPicker ? (
```
ncphillips: There's a couple changes I'd like to see here.

First, pull the `SketchPicker` and `BlocksPicker` into two components with a common props interface. 

Then pick the widget with a lookup and throw an error if there's no match

 ```js
    const WIDGETS = { sketch: SketchWidget, blocks: BlocksWidget }

    export function ColorPicker(props) {
      const Widget = WIDGETS[props.widget]
 
      if (!Widget) throw new Error("Some kind of helpful error message. Not configured properly")
 
      // ....
      return (
        <>
          <Swatch ... />
          { displayColorPicker && <Widget ... /> }
       </>
       )
    }
```

This will make it easier to add new widgets without changing the source of the `ColorPicker` component.

File: packages/@tinacms/fields/src/ColorPicker/ColorPicker.tsx [hunk 3]
```diff
@@ -140,82 +143,148 @@ interface State {
...
+            />}{' '}              
+          </Popover>
+          <Cover onClick={handleClose} />
+        </>
+      ) : null}{' '}
+    </>
+  )
 }
+
+// export class ColorPicker extends React.Component<Props, State> {
```
ncphillips: We don't need this comment

packages/tinacms/src/plugins/fields/ColorFieldPlugin.tsx
```diff
@@ -23,11 +23,18 @@ import { InputProps } from 'react-select/lib/components/Input'
 
 export interface ColorFieldProps {
   colorFormat: string
+  colors: string[]
+  widget: string
```
ncphillips:
```suggestion
  widget?: "sketch" | "block"
```

packages/@tinacms/fields/src/ColorPicker/ColorPicker.tsx
weibenfalk: Hey @ncphillips ... I think I got this one. Not really sure. But made a new commit so check it out. =) Also .. I haven't used TS for so long so bare with me if there's something that look strange there. Also .. you have to look into what error message you want.

Removed the old class component also that was commented out

weibenfalk: It's a little annoying that the BlockPicker and SketchPicker uses different names on some props that are essentially the same ... Otherwise these extra components wouldn't be needed.

packages/@tinacms/fields/src/ColorPicker/ColorPicker.tsx
```diff
@@ -131,91 +132,107 @@ export const Cover = styled.div`
...
+const WIDGETS = { sketch: SketchWidget, block: BlockWidget }
+
+export const ColorPicker: React.FC<Props> = ({
+  colorFormat,
+  userColors = presetColors,
+  widget = 'sketch',
+  input,
+}) => {
+  const Widget = WIDGETS[widget]
+  if (!Widget) throw new Error('You must specify a widget type.')
```
ncphillips:
```suggestion
  if (!Widget) throw new Error('Invalid Color Picker widget. The options are "sketch" and "block"')
```

ncphillips:
👍 great stuff @weibenfalk. I just made a suggestion for the error message. Once that's committed I'll approve

weibenfalk:
Wonderful to read @ncphillips ... feels awesome to help out and be a (little) part of the code base =)

weibenfalk:
Just one more note from me on this one. Since I added Dismissible to this it's going to have that functionality always from the component itself. The thing I did here is to send in the context to the newly created frame prop from tinacms package for it to work. If you don't want Dismissible integrated into the colorpicker itself (in the scoped folder @tinacms) and rather have it applied in the tinacms package. Some stuff have to be refactored. For example the Swatch has to be a component that's imported and used from tinacms instead of inside the colorpicker component itself. Because that component will also disappear if the colorpicker is wrapped with Dismissible from tinacms when Dismissible is hiding the component.  Meaning that ... if not visible there's no way to show the colorpicker again as the button for showing it will disappear. =) 
So ... it's a little bit more work that demands a little bit more thinking on the desired structure of the components. If that is desired instead.

ncphillips:
Sorry for the slow response @weibenfalk .

I don't quite grasp the issue with Dismissible that you're talking about. But `react-dismissible` is a totally separate package. Feel free to add it to the dependencies of `@tinacms/fields` and import it from there.

I'm just getting caught up after the holidays now. I might be able to help more later.

ncphillips:
The diff on this PR is actually totally screwed up. I'm going to merge it into that CI branch and open an new PR. You can make any further changes there

PR #536, opened 2019-12-16
Test: Add test for removing plugin

JeffersonBledsoe:
A small test which behaves the same as the `add plugin` test to check that a plugin is removed


ncphillips:
@all-contributors please add @JeffersonBledsoe for test

PR #543, opened 2019-12-18
enhancement(teams): site access should now be scoped to their namespace

jamespohalloran:
Site access can now be scoped to a namespace. This updates the api to check access based on the new teams api spec

Users will need to add `TINA_TEAMS_NAMESPACE` to their env.
If a user has access to the scoped namespace & the current url.. then they can access the site


packages/@tinacms/teams/src/authMiddleware.ts
```diff
@@ -113,26 +114,38 @@ export function redirectNonAuthenticated(req: any, res: any, next: any) {
...
+    )
+    res.status(422)
+    res.sendFile(missingNamespaceError)
+  } else {
+    const unauthorizedView = path.join(
+      __dirname + '/../public/unauthorized.html'
+    )
+    fetch(
+      `${TINA_TEAMS_API_URL}/sites/${encodeURIComponent(
+        TINA_TEAMS_NAMESPACE || ''
```
dwalkr: What's the purpose of the fallback here?

jamespohalloran: oops! no purpose. removed!

PR #547, opened 2019-12-21
Created a select Field

weibenfalk:
I've created a standard controlled select element. Haven't applied styles yet. You can use it like this: 

{
      label: "New Shiny Select",
      name: "frontmatter.cool",
      component: "select",
      values: ["Tina rules!", "Love this!", "How cool!"]
},

<TinaField
          name="rawFrontmatter.cool"
          Component={MySelect}
          type="select"
>
          <p>{post.frontmatter.cool}</p>
</TinaField>


ncphillips:
Thanks @weibenfalk  I love how simple this is! I think it's a great starting point to just accept a list of strings.

I'm sure @spbyrne will want to think about styling, but the only change on my mind, would be to rename `values` to `options`. 

In the future this will not just be a list of values, we'll need to support more types of `options` :

```ts
interface SelectFieldProps {
  label: string
  name: string
  component: string
  values: OptionOrValue[]
}

interface Option {
  label: string
  value: Value
}

type Value = string | number
```

For this PR I would only ask for the property rename. 

Extending the Select component to accept numbers or objects can be done later. Of course if you want to add that feel free, but just know that you don't have to for this PR to be merged.

Happy Holidays! 🎉

weibenfalk:
Great to hear! Thanks!

I’m very much into not over complicating things when no need for it. Also not bring in external libraries for something that’s possible to do fairly easy yourself :) 

This is a base to build from. Think it will be easy to extend functionality as needed in the future. 

If you want me to style it also just tell me and I will do that.

Will fix the renaming asap. 

Happy holidays to you too! Love working on this codebase! 

Thomas 



Skickat från min iPhone

weibenfalk:
Fixed the naming now @ncphillips and pushed it

spbyrne:
I wrote some basic styles and created a PR to your branch, @weibenfalk

weibenfalk:
@spbyrne Great! Do I just merge it in or do you do that? Guess your team want to check it out so I shouldn't merge it to my branch myself?

weibenfalk:
Realized that this doesn't work with inline editing now. Thought it was possible to give <TinaField> an options prop with the array also but didn't work. I'm not that familiar with the codebase yet so I easily can find where to fix this. Have looked around and haven't found it. Is it somewhere in the <TinaField> component itself I have to change something?

So ... my first question is. I guess this should work in inline editing also? And where do I fix so it's possible to send in the array with options to the <TinaField> component also so it is possible to access that array in the Select component itself? Because now it gets it values from the field where you set up the form structure. That's why it breaks on inline editing.

weibenfalk:
I got this to work in the inline editor also by doing below. Is it something that you like? 

The Select component: 

```
export interface SelectProps {
  name: string
  input: any
  field: SelectFieldProps
  disabled?: boolean
  options?: string[]
}

export const Select: React.FC<SelectProps> = ({ input, field, options }) => {
  const selectOptions = options || field.options
  const [selectValue, setSelectValue] = useState(input.value)
  const handleChange = (e: React.ChangeEvent<HTMLSelectElement>) =>
    setSelectValue(e.currentTarget.value)

  return (
    <SelectElement>
      <select
        id={input.name}
        value={selectValue}
        onChange={handleChange}
        {...input}
      >
        {selectOptions ? (
          selectOptions.map(option => (
            <option value={option} key={option}>
              {option}
            </option>
          ))
        ) : (
          <option>{input.value}</option>
        )}
      </select>
    </SelectElement>
  )
}
```

In the TinaField component I just spread the rest of the props to the component itself. That way it is possible to pass in custom props. 

The TinaField component: 

```
export function TinaField({
  Component,
  children,
  ...fieldProps
}: TinaFieldsProps) {
  const isEditing = useContext(EditingContext)
  if (!isEditing) return children || null

  return (
    <Field {...fieldProps}>
      {({ input, meta }) => {
        return <Component input={input} meta={meta} {...fieldProps} />
      }}
    </Field>
  )
}
```

packages/demo-gatsby/src/templates/blog-post.js
```diff
@@ -248,6 +273,18 @@ const BlogPostForm = {
...
+      name: "frontmatter.cool",
+      component: "select",
+      options: [100, "Love this!", "How cool!"]
+    },
+    {
+      label: "Testing Number Component",
+      name: "frontmatter.testNumber",
+      component: "number",
+      steps: 3
+    },
```
spbyrne: This outputs as `Unrecognized field type` in the sidebar; maybe this shouldn't be a part of the select PR? Otherwise this is good to merge in, tested and looks good. This is meant to be a simple first step, after all.

packages/demo-gatsby/src/templates/blog-post.js
```diff
@@ -113,15 +120,33 @@ function BlogPostTemplate(props) {
...
+          <p>{post.frontmatter.cool}</p>        
         </TinaField>
+        {/* <TinaField
+          name="rawFrontmatter.testNumber"
+          Component={Number}
+          type="number"
+        >
+          <p>Number: {post.frontmatter.testNumber}</p>
+            {console.log(typeof post.frontmatter.testNumber)}
+        </TinaField> */}
```
spbyrne: If this is tied to the number input maybe we should pull this out, too?

weibenfalk:
Hey. No the number input is something I tried and I forgot to remove that code. Sorry! Shouldn’t be there.

Skickat från min iPhone

weibenfalk:
I got the number field to work but it writes the number as a string in the markdown With quotes around. It works anyway though. Don’t know how to change that. Every input field in HTML is just a string type. No number type in HTML. Does it matter? 

Skickat från min iPhone

spbyrne:
No worries. We can deal with the number input separately.

packages/demo-gatsby/src/templates/blog-post.js
```diff
@@ -113,15 +120,33 @@ function BlogPostTemplate(props) {
         <button onClick={() => setIsEditing(p => !p)}>
           {isEditing ? "Stop Editing" : "Start Editing"}
         </button>
+
         <TinaField
           name="rawFrontmatter.draft"
           Component={MyToggle}
-          type="checkbox"
+          type="toggle"
```
ncphillips: This isn't the name of the component, it's a prop that gets passed to `Field` in `react-final-form`. It's necessary for the toggle field to work properly as a checkbox. 

Scroll to the bottom to read the [`value`](https://final-form.org/docs/react-final-form/types/FieldProps) section.

packages/demo-gatsby/src/templates/blog-post.js
```diff
@@ -113,15 +120,33 @@ function BlogPostTemplate(props) {
...
         >
           {post.frontmatter.draft && (
             <small style={{ color: "fuchsia" }}>Draft</small>
           )}
+        </TinaField><br />
+
+        <TinaField
+          name="rawFrontmatter.cool"
+          Component={MySelect}
+          type="select"
```
ncphillips: This is not necessary
```suggestion
```

weibenfalk:
@ncphillips ... I've cleaned up the code now and remove the Number stuff ... Pushed it but forgot that this one was from the original fork before I had contributor access to the original TinaCMS repo. So it's in a new branch there now.

spbyrne:
@weibenfalk Is it possible for you to update this PR to merge from that branch, or should I create a new PR from your new branch?

weibenfalk:
Nolan said that I can create a new PR but you can do that instead if you want?

spbyrne:
Sounds good, go ahead!

PR #548, opened 2019-12-23
Fixed fallback for broken images on image field issue #203

### LINKED ISSUE #203: Image field needs a placeholder for broken images
When an image field's src is set to "", it doesn't render a placeholder. This makes it difficult to upload an image, as the clickable dropzone area is very small.
Screen Shot 2019-10-02 at 9 32 02 AM

weibenfalk:
fixes #203

I've fixed a fallback solution when image is broken. Just added a svg icon with a broken image and set the min-height to 100px ... that will show the icon on broken images. Used the color from the sidebar on the toggle button.

Skärmavbild 2019-12-23 kl  16 57 22


kendallstrautman:
Possibly fixes #203

spbyrne:
I'm going to approve and merge this, but when it comes time to use that SVG icon in another place we'll have to pull it out into a separate file (works great for now, though). Looks good! 😄

weibenfalk:
Great to hear! Thanks! 

Skickat från min iPhone

PR #552, opened 2019-12-29
refactor: dismissible

weibenfalk:
This is no open issue. I played around, learning the codebase, and refactored the react-dismissible package to use hooks instead. I rarely use classes anymore and love hooks. In this case it may be a little more elegant to use the useRef hook to reference the area. So ... use it if you want otherwise just ignore =)


weibenfalk:
Ahhh did I miss this. Will look into it ASAP.

Skickat från min iPhone

weibenfalk:
Hey .. I've looked into this now. Have you tested it and get infinity loops? It shouldn't do that. The effect doesn't do anything that triggers a re-render .. for example setting state or something. So this effect don't need a dependency array. I've console logged out the effect and it isn't causing any infinity loop for me here. Renders as it should ... 

Also. I'm returning a cleanup function that deletes the previous event listeners on unmount.

But ... it probably won't hurt to have it there if you still want it? =)

weibenfalk:
I've pushed a new commit now. Refactored the useEffect and added dependencies anyways. Also moved the functions inside of the useEffect to avoid using useCallback on them. 

But ... there's something fishy with the escape key. Don't know if it's my touchbar that is not working. It worked before and now I can't use it to dismiss. Do you have the same issue @ncphillips ? 
And if so ... is the problem also in the class version of Dismissible.

spbyrne:
The escape key is working fine for me. This looks good. Also I don't see any event listener issue anymore. This is working much more reliably for me.

ncphillips:
You're right, it doesn't trigger an infinite loop, but it will re-run on every render. Things outside of Dismissible might cause a rerender. Maybe not necessary but just a better-safe-then-sorry situation.

Moving the functions inside the useEffect is a great idea 👍 

As to the click behaviour, you introduced a common `stopAndPrevent` function for both `handleEscape` and `handleClick`. In the old code they did two different things:

`handleEscape` just did

```
event.stopPropagation()
```

whereas `handleClick` does
```
event.stopPropagation() 
event.stopImmediatePropagation()
event.preventDefault()
```

Now both do the later.  I assume this is the problem.

weibenfalk:
@ncphillips Ok ... I totally agree =) 

Regarding the escape key. I tried it with both using the stopAndPrevent and also by just calling stopPropagation ... It worked with both first. That's why I used it on both. Have changed it back here. May be some thing with my escape key locally then. Or .. don't know if something is different cause my escape key is a touch area on the touchbar. =) Really have no idea .. it just stopped working. But worked a day ago.

spbyrne:
Yeah that's odd! I tried both with my external keyboard & my touchbar esc key, no issue with either for me.

packages/react-dismissible/src/Dismissible.tsx
```diff
@@ -47,64 +48,63 @@ export interface Props {
...
+      )
+    }
+    // Clean up event listeners on unmount
+    return () => {
+      documents.forEach(document => {
+        document.body.removeEventListener('click', handleDocumentClick)
+        document.removeEventListener('keydown', handleEscape)
+      })
+    }
+  }, [click, documents, escape, disabled, onDismiss])
```
ncphillips: A new `documents` array is created on every render, which means the effect would rerun on every render. It needs to be the two values listed individually to prevent rerunning.
```suggestion
  }, [click, document, customDocument, escape, disabled, onDismiss])
```

PR #557, opened 2020-01-02
Custom git remote

jamespohalloran:
Allow git remote to be configurable. Required to host development environment with certain providers (e.g Heroku) when git remote is not already set


packages/@tinacms/api-git/src/router.ts
```diff
@@ -84,9 +84,21 @@ async function createSSHKey(pathRoot: string) {
...
       mode: 0o600,
     })
+  }
+}
 
-    updateRemoteToSSH(pathRoot)
+async function configureGitRemote(pathRoot: string) {
+  await createSSHKey(pathRoot)
+  if (process.env.GIT_REMOTE) {
+    await updateOrigin(pathRoot, process.env.GIT_REMOTE)
```
ncphillips: We should log a warning message if there's already an existing `origin` that we're overwriting

Issue #558, opened 2020-01-04
Customizable width of the sidebar


weibenfalk:
## Summary

Think it would make sense to add an option to set a custom width to the sidebar. If you think it's a good idea I can try to implement this for the gatsby plugin to see how it works and create a PR. My idea is to set a max-width so it will be responsive and scale down on mobile devices. 

Haven't looked at the components/plugins for the form fields in detail so don't know if it will be a bigger project refactoring them. But guessing they are set to scale to the width of the sidebar?

### Basic example

```
{
    resolve: 'gatsby-plugin-tinacms',
    options: {
      sidebar: {
        position: 'displace',
        maxWidth: '400px'  // could also just be width
      },
    },
}
```

### Motivation

The sidebar can feel small on bigger screens.  Would be great to have the possibility to decide how wide the sidebar should be.

DirtyF:
Would be even better if you could resize it without any optional setting.

weibenfalk:
Yes, that would be great ... but guess that requires a lot more work as it will affect the whole site when the user resizes dynamically. Or it is much simpler than I think right now without having thought it through =) 

Guess it would be great to still be able to set the initial width though. There's probably also use cases when you don't want the user to be able to resize but only set a max-width.

weibenfalk:
Check out the below video. Is this something you want? Still a lot of refactoring to do as I just got it working =) And it doesn't seem to work with inline editing yet. Haven't looked at why yet ... It changes size when I press the inline editing button and I have no clue why now as I haven't tampered with that at all. I created a new component, a resizerBar, and added it to Tina with some small tweaks and local states. 

I also didn't want to re-render the whole sidebar on each pixel move with the mouse. That's why I created this kind of "ghosting" process where the user drags the line to the desired width first. Then I update the width of the sidebar and everything in it. This way I just have to re-render the resize-bar itself when it moves and update its local state. 

As you can see there are also some tweaks that have to be made to the form fields so they will have full width. And I didn't test it with the "fixed" version of the sidebar yet. Just the "displace" one. 

Anyways ... below is the video: 

https://www.dropbox.com/s/1zv811ixk7pwwfa/tina-resizable-test.mov?dl=0

DirtyF:
Wow, I didn't think you'll go the hard way. 😊 
Would it help with the rendering if the sidebar was kind of blurry on drag?
It mostly raises the question: how large should the sidebar be in different contexts? 
When do we need a larger sidebar?
[...]

Issue #479, opened 2019-11-18
Groups/Group Lists/Blocks are broken inside Global Forms


spbyrne:
## Description

The way groups/group lists/blocks are displayed only works when they're displayed in a regular form. When you open one inside a global form, it gets rendered in the sidebar instead of inside the global form.

### Steps to reproduce

Add a global form that contains a group, or use `useGlobalJsonForm` on an existing json form that contains a group.

### Expected result

I'd expect the panel to open inside the modal.

### Actual result

What happened.
![image](https://user-images.githubusercontent.com/5075484/69080991-6c8c1e80-0a13-11ea-808c-596848b84c31.png)

timonweber:
I’ve just experienced this myself. It would be great if someone can fix it!

ncphillips:
A fix is coming soon @timonweber. @spbyrne is doing a bit of work that will make a fix possible.

spbyrne:
My previous attempt at fixing this used refs to measure the appropriate spot to render the panel. It ended up being overly complicated and prone to failure.

Nolan has set up a form portal that can be used to move any sub component up to the root form level, which will make this a lot easier. I'm currently working on this and should have it ready today.

PR #565, opened 2020-01-08
feat: resize sidebar

weibenfalk:
Hey. Here's a working version of the resize function that we discussed. Haven't gone through the code in detail so probably can use some optimization. Added a min width now also. Think it would also be a good idea to have an option to turn resizing on and off when setting up the sidebar.


packages/tinacms/src/components/ResizerBar.tsx
```diff
@@ -0,0 +1,129 @@
...
+export const ResizerBar = ({
+  resizerBarWidth,
+  open,
+  sideBarWidth,
+  setSideBarWidth,
+}: ResizerBarProps) => {
+  const [barPos, setBarPos] = React.useState(sideBarWidth - resizerBarWidth / 2)
+  const isResizing: any = React.useRef()
+
+  const resize = (e: any) => {
```
ncphillips: It might be best to have `useCallback` wrap these:
* `resize`
* `handleMouseUp`
* `handleMouseDown`

weibenfalk: Is that necessary? I'm not using any useEffect in this one with the functions in a dependency array that will create an infinite loop. The recommendation is to use useCallback on expensive operations ... Some times the useCallback can actually decrease performance.

ncphillips: Better safe then sorry. A new function instance is being created and passed to the `Resizer` component on every re-render. With the memo around it that won't happen, and we can be certain that nothing funky is happening. Every render may cause `Resizer` to re-render unnecessarily. I'm especially wary since those callbacks aren't just being used in React, they're being added and removed from the `document`.

weibenfalk: Ok. I can add them. This is an interesting discussion as there are many opinions on this. Some say that you don't need them and the useCallback itself create more overhead cost than leaving it out when it is small functions like this. =) The useCallback itself is a function that you have to call with a dependency array also ... 

Do you want to wrap the whole component in React.memo also or do you mean the useCallback on the three inline functions only?

weibenfalk: Pushed a new commit now. Added useCallback and also changed to mouse cursor to a resize cursor instead .. seems to be working fine. =)

weibenfalk: Took a more deep dive in thinking through this. I see why you want to have the useCallback on the callback functions to be safe. It's a good tool to use when certain child components shouldn't re-render when the parent component that holds the callback function re-renders. 

In this case the Styled Component named Resizer should actually re-render itself on each re-render. So the useCallback won't have any real effect on performance. I haven't measured anything but think it actually can slow stuff down as it is an extra function that does dependency checks via the dep. array (and also more code to be written). 

I'm not arguing against you. I'm just really curious cause we all have different approaches and opinions on stuff like this. =) It's a widely discussed topic in the React world and my learnings is that you don't need useCallback if it's not a real expensive function. Or if you don't want to re-render child components that gets the callback function as props. In this case the whole Resizer component is an "isolated" component that don't render any child components more than itself. 

And yet again ... I'm just curious to hear out your opinion cause you seem really experienced and I like discussions like this. It is you who has the final saying on the code and I will absolutely change it to however you want the code. Truly respect your knowledge!  =)

ncphillips: I'm glad you did the dive @weibenfalk! I didn't dig too deeply on this particular case. 

I made the suggestion based on a heuristic: *if you're adding/removing the callbacks to something outside of the React Component it's safest to wrap them with `useCallback`*. In this situation, as well as with `Dismissible`, we're using the callbacks as event listeners on the document, so the heuristic applies.

In simpler examples I would opt for inline functions. In this counter example there's no need to memo-ize the callback:

```js
function Counter() {
  const [count, setCount] = useState(0)

  return <button onClick={() => setCount(count => count + 1)}> + </button>
} 
```

In that example there is little room for confusion or error. It might be mildly less performant but it's not a big deal. This component has callbacks, but those callbacks do not have any external side-effects.

*The main reason for this heuristic is the potential for memory leaks.* In the above example, a new `onClick` will be created on every render, but we know for certain that old one will be discarded. In the situations where were adding those callbacks to objects that live outside of our React component (i.e. the `document`) we need to be more careful. If we misunderstand how scope+hooks interact, then each render will create another function that never get garbage collected. This is what results in the memory leak. 

Using `useCallback` in those situations gives me some peace of mind. The potential risk of not using `useCallback` it is much higher than it's mental overhead cost.

weibenfalk: Thanks for that long answer. It's good to know how you're thinking. I also see your point as it is callbacks with references outside of the React Component. So .. in this case it was not for performance issues as I thought .. it is for have peace in mind. =) 

Regarding the inline function example. As it is a functional component it shouldn't matter if the function is inline. Both inlines and the other functions inside the functional component will get recreated on each re-render. So it should be acceptable to have it as a regular function also and not inline. =)

I've read a lot of opinions and also the React documents a few times since the release of hooks and they all say that useCallbacks isn't necessary for performance reasons. If you don't have a lot of child components that receive the callback as a prop. That will trigger re-renders of the child components also ... So the recommendation is to not overuse them cause you are "afraid" of performance issues with the function being recreated on each render. That's why I asked. 

Don't think I've come across any article about using them with callbacks that uses eventlisteners on document ... So your opinions is great reading for me! Love to learn more stuff!

ncphillips: By "inline" I meant "not-memoized". 

The above example is equivalent to:

```js
function Counter() {
  const [count, setCount] = useState(0)

  const increment = () => setCount(count => count + 1)

  return <button onClick={increment}> + </button>
}
```

Again, I haven't dug into this. It's quite possibly fine. I would personally just use `useCallback` because I'm certain it will be fine. Convincing myself that not using it is fine would take more work then I'm willing to put in. It's a decision based on expediency rather than correctness. 

If after digging in you find that this does not endlessly add callbacks to the document and does not trigger expensive re-renders, then not wrapping them is cool with me.

ncphillips:
I'm good with this you but the `package-lock.json` is conflicting. 

You'll have to merge `master` into this branch and run `npm run bs`. I believe NPM should correct the merge conflict for you automatically.

packages/tinacms/src/components/ResizerBar.tsx
weibenfalk: No no .. it's totally fine! I was just curious and also fun to discuss React topics as I'm passionate about React.

weibenfalk:
Merged the master now and pushed it again. Hope that resolved it. 

You'll have to discuss if you want to have this functionality. As this was not an open issue or demand from your side. I played around for fun trying it out.

ncphillips:
Hey @weibenfalk I've noticed a few bugs with this:

1. I'm only able to drag sidebar to the right
1. If I mouse-up to the left of it it doesn't cancel out, I need to click to make it exit that behaviour
1. @spbyrne when wider, the form's scrollbar is no longer at the edge of the sidebar

weibenfalk:
Hmm @ncphillips I had that behavior first but thought I got rid of that. Must be because of the iFrame somehow don't register the mouse events on the document. Do you have any good solution? =) 

The sidebar has a minimum width also ... So it won't allow you to make it smaller than that. But You can't drag it to the left even when it is big and you want to drag it back?

spbyrne:
If it's an iFrame issue that could be reason to expedite the iFrame removal PR. I know pretty much exactly what I need to get done to merge that PR so I'll bring it up on Monday to see where it fits in to our priorities.

weibenfalk:
@spbyrne @ncphillips So you're going to remove the iFrame? In that case maybe I should not look into this right now cause I'm pretty sure that's the iFrame that makes things more difficult. Had a lot of issues first that the event listeners didn't register when hovering over the iFrame ... thought I got rid of it but guess not =)

I'm curious .. what is the iFrame for? =)

ncphillips:
👍 sounds good @weibenfalk.

The iframe is there to isolate he styling of the sidebar from the styling of the website. It's getting in the way of a couple things though, so we're looking to remove it.

weibenfalk:
Aha… thanks! I see the point then. =) IFrames are bound to be problematic though =)

weibenfalk:
Hey @ncphillips  ... Noticex that you have removed the iFrame from master now. Sweet! Have made some minor changes to the resize functionality and it works locally for me now. Try it out and see if it works on your end. 😄

spbyrne:
@weibenfalk this is working great for me now, nicely done. 👍  I'm going to make a couple small tweaks before approving this.

weibenfalk:
Sounds great! =)

spbyrne:
The only remaining issue that I see is the UI for groups, group lists & blocks (issue #479). I started tackling this last week as it also affects modals, but the solution I was working on didn't pan out. We're going to try fixing this with portals instead, which will fairly significantly change the underlying layout CSS for the panels. We should have that ready this week as it's a pressing issue for a number of reasons now.

weibenfalk:
Sweet! Looks great @spbyrne ... probably better to have it show up on hover as you did now. Great work!

kendallstrautman:
Should close #558

weibenfalk:
Is this going to be implemented? =)

ncphillips:
This issue was being blocked by #479 which was fixed yesterday. I think it might be possible to move forward with this now 👍

weibenfalk:
Aha ... I see. Would be great to see this implemented. I actually think it
enhances the Tina experience to be able to resize it =)

Den lör 22 feb. 2020 kl 17:59 skrev NCPhillips <notifications@github.com>:

ncphillips:
Hey @weibenfalk! It's been a while and there's been a ton of work done to the sidebar since it was opened. We're going to close this PR for now but I'll mention #558 here so it's linked up, and we'll keep the branch so it can be used as a reference. As always, thanks for the effort here it's really appreciated!

weibenfalk:
No worries! I'm just glad if I helped with something. Great to see that
your'e progressing forward also. =)

Den tors 16 apr. 2020 kl 15:46 skrev NCPhillips <notifications@github.com>:

Issue #567, opened 2020-01-08
feat: display focus on images in the editor


DirtyF:
## Summary

Give visual feedback to the editors when an image is selected, to prevent accidental deletion.

### Basic example

Given I'm editing text after an image
When I hit backspace 
Then I can not delete an image before having visual feedback (border) that the image is currently selected and that if I hit backspace again, it will be removed.

### Motivation

Better editing experience.
Being able to select image to perform actions

PR #570, opened 2020-01-08
Adding node view for image node.

jpuri:
The draft PR at this stage shows approach I am planning to take to fix this https://github.com/tinacms/tinacms/issues/567

I am creating a node view for images, this node view will highlight image when it is falling at selection. Highlight can be added to a wrapper around image or the image itself. 

Selected image is not necessarily focused, thus its is required to highlight image at selection in editor.


jpuri:
Screenshot 2020-01-09 at 10 00 53 PM

Images now have selected state highlighted with blue outline, also image has margin around it. Designs changes are discussed with @spbyrne .

Image will be selected on back arrow, behaviour on backspace is same as before. I think that should not be changed, its standard key mapping  - to delete content on backspace.

packages/@tinacms/fields/src/Wysiwyg/node-views/ImageView/image.css
```diff
@@ -0,0 +1,22 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+
+*/
+
+.tinacms-image {
+  display: inline-block;
+  padding: 1em;
+}
```
jpuri: Padding will be applied to image and not image wrapper.

jpuri:
Behaviour of gap cursor can be improved further to make editor more intuitive.

packages/@tinacms/fields/src/Wysiwyg/node-views/ImageView/image.css
spbyrne: What if we had `margin: 1em 0;` and 0 padding? (outline right on the image, no horizontal spacing between image and margins)

Not sure I like that the image isn't left aligned to the text; is that only when the focus state is applied? We don't necessarily have to address all the design concerns in this PR, I can tweak the styles later as well.

jpuri: Hey @spbyrne , I have done this change to styles. You can do any more required style changes to images here: https://github.com/tinacms/tinacms/pull/570/files#diff-85a5bec6b762da44b21a435e5fc7c8bdR21

spbyrne: Sweet, thanks! 👍

spbyrne:
This image UI is really nice! I like the indicator for where it gets dropped when you’re dragging it around. I think it's using a green color right now which we can swap for the blue used on the image outline.

I noticed a strange bug and I’m wondering if it’s related; when I’m editing in the WYSIWYG my cursor is visible as expected, but after dragging an image into the editor and going back to edit text my cursor is no longer visible.

ncphillips:
Great stuff @jpuri! 

A note for future PRs: we use [conventional commits](https://www.conventionalcommits.org/en/v1.0.0-beta.4/) in this project. It helps us automatically determine what the version bump and and generate CHANGELOG files.

ncphillips:
@all-contributors please add @jpuri for code

PR #574, opened 2020-01-08
chore!: introduce @tinacms/forms package

ncphillips:
BREAKING CHANGE: Forms related concepts are no longer declared
in the @tinacms/core package.

This change removes the concepts of Forms and Fields from `@tinacms/core` into it's own package. This means `@tinacms/core` is now solely responsible for managing plugins and API integrations. 

`tinacms` now introduces `cms.forms`


ncphillips:
`lerna` handles version bumping. It will automatically bump it to `0.1.0` when it's released

ncphillips:
Thanks for mentioning that @DirtyF I just noticed the commits wouldn't actually cause the version bump. I had to switch the first commit from "chore" to "feat"

PR #576, opened 2020-01-09
chore: update dependencies with security issues

ncphillips:
There's a security issue with `handlebars`. I've tried updating the packages that depend on handlebars:

* `lerna`
* `jest`
* `rollup`

I think ta bad version is still getting included. 

No tests or docs updates will be required.


ncphillips:
Closing. There's a bug with the new version of `jest` that's causing the tests to hang.

PR #591, opened 2020-01-13
Generate docs for tinacms core

ncphillips:
How to check them out?

```
lerna run docs
```

Then open `packages/@tinacms/core/docs/index.html` in your browser.


ncphillips:
There is a [monorepo plugin](https://github.com/amiller-gh/typedoc-plugin-monorepo) but I haven't looked into it yet.

Alternatively, each package could generates it’s own API doc independently and we could write some scripts to stitch them together.

PR #597, opened 2020-01-14
Sticky Wyisiwyg Menu

ncphillips:
CI Build


spbyrne:
I was getting an error that the lock files were out of sync; I ran `npm run install && npm run bs` and commited the changes.

PR #607, opened 2020-01-15
Fix gatsby plugin dependency

ncphillips:
This PR is trying to fix some dependency issues that come up. 

The various gatsby plugins are currently depending directly on `tinacms` which can break your build. 

This PR makes `tinacms` a peerDep of all each the plugins. 

Similarly, the sub-plugins now have `gatsby-plugin-tinacms` as a peerDep rather then a dep.

### !Breaking!

This would require a change to docs describe how to get started with TinaCMS. 

```yarn add gatsby-plugin-tinacms styled-components```

will now need to be

```yarn add tinacms gatsby-plugin-tinacms styled-components```

We should make sure to update any effected blog posts and notify anyone else who wrote a blog post.

### DIFF
--- a/packages/@tinacms/api-git/package-lock.json
+++ b/packages/@tinacms/api-git/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "@tinacms/api-git",
-  "version": "0.5.1",
+  "version": "0.5.2",
   "lockfileVersion": 1,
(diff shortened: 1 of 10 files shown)


ncphillips:
Everything got out of sync and also I think I can do this more piecemeal

PR #612, opened 2020-01-16
feat: adding highlighting to table selection

jpuri:
Improved selection on tables:

table


DirtyF:
Can you explain what problem does it solve? What actions can you perform on select?

jpuri:
Hey @DirtyF : this will be useful to do any action on whole column or row. In following task plan is to add option to delete a whole cell or row after selecting it.

PR #617, opened 2020-01-16
Group/Blocks Global Forms UI Fix

spbyrne:
- Groups/Group Lists/Blocks are now passed the dimensions of the form they're displayed in so they can render appropriately via a prop called formBoundingBox.
- The FormView component now uses a ResizeObserver to update the formBoundingBox as the form is resized or moved.


spbyrne:
Nevermind, this is awful.

PR #633, opened 2020-01-20
Adding delete column/row option to table.

jpuri:
Delete option for selected columns / rows in table.
Screenshot 2020-01-20 at 7 21 14 PM
Screenshot 2020-01-20 at 7 21 22 PM

### DIFF
--- a/packages/@tinacms/fields/src/Wysiwyg/ProseMirrorCss.tsx
+++ b/packages/@tinacms/fields/src/Wysiwyg/ProseMirrorCss.tsx
@@ -17,7 +17,6 @@ limitations under the License.
 */
 
 import { css } from 'styled-components'
-import { color } from '@tinacms/styles'
 
 const proseMirrorTableStyles = `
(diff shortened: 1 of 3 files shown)


jpuri:
@spbyrne : plz share ur suggestion about better highlight colour for selected row/column or for better styling for trash icon.

spbyrne:
@jpuri I've adjusted some of the styles.

- Adjusted colors
- Increased row/col handle size & added pointer cursor
- Turned delete button into icon button and moved the click event from the wrapper
- Adjusted the way the handles are sized and placed; they were using px values to compensate for the table cell padding. Since we can't always control the cell padding I've set them to `top: 0; & left: 0;` and positioned via the transform property. In testing on chrome/firefox/safari locally this seems to fix an issue I was experiencing where there was a 1px gap between these handles.
- I've removed the focus outline from the WYSIWYG. This outline was getting in the way of things and will essentially be replaced by the blocks editor focus ring (when inside a block) and isn't needed otherwise.
- I've removed the left/right margin from the table so it remains full width in editing mode.

PR #637, opened 2020-01-20
Fix @tinacms/api-git `/show/:path` endpoint

ncphillips:
Possible fix for several reports

There are three pieces to this bug.

First, on windows path's can contain forward slashes, or
double back slashes. For example

    data/author.json

or

    data\\author.json

Second, Node's `path.join` returns the second on Windows.

Third, the `git show` command behaves oddly with the former:

    git show HEAD:data/author.json

Returns the contents of the file

    git show HEAD:data/dne.json

Throws an error, and finally

    git show HEAD:data\\author.json

returns an empty string.

This very well may be a bug in Git. I'm not sure.

### DIFF
--- a/packages/@tinacms/api-git/src/router.ts
+++ b/packages/@tinacms/api-git/src/router.ts
@@ -302,7 +302,7 @@ export function router(config: GitRouterConfig = {}) {
 
   router.get('/show/:fileRelativePath', async (req, res) => {
     try {
-      const fileRelativePath = path
+      const fileRelativePath = path.posix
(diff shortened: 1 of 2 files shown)


dwalkr:
Node provides `path.posix` which will always run the POSIX implementation of path methods. Since show requires a POSIX path, I think it would be appropriate (and simpler) to just call `path.posix.join` here. What do you think?

https://nodejs.org/api/path.html#path_path_posix

ncphillips:
That very well may be the better option.

ncphillips:
I wonder if there's an ESLint option for that?

PR #638, opened 2020-01-20
fix: FieldPlugins for text have a default to empty string

### LINKED ISSUE #632: FieldPlugins for text should default to empty string, instead of undefined
## Summary

By default, a cleared text field's value is `undefined`. Instead it should be an empty string: `''`.

This applies to several plugins:

- [ ] TextFieldPlugin
- [ ] TextareaFieldPlugin
- [ ] SelectFieldPlugin
- [ ] MarkdownFieldPlugin
- [ ] ColorFieldPlugin
- [ ] ImageFieldPlugin

### Basic example

Currently users must add a `parse` function on their text field definitions to get empty strings:

```js
{
  name: "frontmatter.title",
  component: "text",
  parse: value => value || '',
}
```

This should not be required.

### Motivation

The current behavior is unexpected for most use cases. This assertion is based off of #630 as well as our experience with Forestry.io

### Solution

Add the `parse` function above to the `TextFieldPlugin` directly. 

Look at the [DateFieldPlugin](https://github.com/tinacms/tinacms/blob/master/packages/tinacms/src/plugins/fields/DateFieldPlugin.tsx#L76-L81) for reference

NickCis:
This PR fixes #632 .

I label it as a fix, but, I'm not sure if it's the correct label.

### DIFF
--- a/packages/tinacms/src/plugins/fields/ColorFieldPlugin.tsx
+++ b/packages/tinacms/src/plugins/fields/ColorFieldPlugin.tsx
@@ -20,6 +20,7 @@ import * as React from 'react'
 import { ColorPicker } from '@tinacms/fields'
 import { wrapFieldsWithMeta } from './wrapFieldWithMeta'
 import { InputProps } from 'react-select/lib/components/Input'
+import { parse } from './textFormat'
 
 export interface ColorFieldProps {
   colorFormat: string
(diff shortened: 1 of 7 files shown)


ncphillips:
@all-contributors please add @NickCis for code, tests

PR #646, opened 2020-01-21
fix(wysiwyg): prevent selection crash from arrows

ncphillips:
Wysiwyg was crashing when using arrows to move around in the document. Turns out we weren't accessing Selection. 

Don't cast to `any`.

### DIFF
--- a/packages/@tinacms/fields/src/Wysiwyg/plugins/keymap.ts
+++ b/packages/@tinacms/fields/src/Wysiwyg/plugins/keymap.ts
@@ -17,7 +17,7 @@ limitations under the License.
 */
 
 import { NodeType, Schema } from 'prosemirror-model'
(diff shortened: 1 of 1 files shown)


packages/@tinacms/fields/src/Wysiwyg/plugins/keymap.ts
```diff
@@ -301,11 +301,12 @@ export function isNotAList(state: EditorState) {
   const { $from, $to } = state.selection
   const range = $from.blockRange(
     $to,
-    node => !!(
-      node.firstChild &&
-      (node.firstChild.type == ordered_list ||
-        node.firstChild.type == bullet_list)
-    )
+    node =>
```
ncphillips: This is just prettier formatting changes

PR #649, opened 2020-01-21
refactor(sidebar state): moved onto TinaCMS

ncphillips:
Looking for everyone's thoughts on this. 

I started mulling over the Sidebar state after checking out @weibenfalk's thoughts on configuring the text on the Save/Reset buttons.

This change reifies the sidebar state and places it on the TinaCMS class as a property. I think this approach has a couple benefits:

1. The API feels more righter
1. It removes the need for the `SidebarContext` in the React app, instead relying on the pre-existinig CMSContext. 

Here is how it would be used.

### Checking sidebar state:

    cms.sidebar.isOpen

### Changing sidebar state:

    cms.sidebar.isOpen = true

### Subscribing Option 1

    import { useCMS, useSubscribable } from "tinacms"
    const cms = useCMS()

    useSubscribable(cms.sidebar)

### Subscribing Option 2:

    const sidebar = useSidebar()

### DIFF
--- a/packages/tinacms/src/components/Tina.tsx
+++ b/packages/tinacms/src/components/Tina.tsx
@@ -18,13 +18,12 @@ limitations under the License.
 
 import * as React from 'react'
 import { ModalProvider } from './modals/ModalProvider'
-import { SidebarContext } from './sidebar/SidebarProvider'
 import styled, { ThemeProvider } from 'styled-components'
 import { TinaReset, Theme, DefaultTheme, ThemeProps } from '@tinacms/styles'
 import { Sidebar } from './sidebar/Sidebar'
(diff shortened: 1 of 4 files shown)


weibenfalk:
I think it's great to remove the SidebarContext when it's not needed. One question .. how can you use "private" in the JS Class? Thought private wasn't supported. =) Isn't a private field still experimental and it uses # instead?

ncphillips:
It's a [Typescript feature](http://www.typescriptlang.org/docs/handbook/classes.html#public-private-and-protected-modifiers). 

The ECMAscript private fields syntax is coming in 3.8, but we're not on that yet.

PR #652, opened 2020-01-21
feat: adds next-tinacms-markdown package

kendallstrautman:
This PR adds the `next-tinacms-markdown` package with helpers for editing Markdown in Next.js 
- `useMarkdownForm`
- `useLocalMarkdownForm`
- `useGlobalMarkdownForm`

It also adds the `demo-next` package for testing out next+tina development features.


kendallstrautman:
Still Todo: 
- Need to add a few helpers for global, inline and render props etc. 
- Figure out the bootstrap error

kendallstrautman:
I don't understand where those generated style component files came from when I merged master. Perhaps something left over when I was messing around on the `tinacms-ui` branch?

packages/next-tinacms-markdown/package.json
```diff
@@ -0,0 +1,40 @@
+{
+  "name": "next-tinacms-markdown",
+  "version": "0.1.0-alpha.0",
+  "license": "Apache-2.0",
+  "author": "Nolan Phillips",
```
ncphillips: No I'm not 😉

packages/next-tinacms-markdown/package.json
```diff
@@ -0,0 +1,40 @@
...
+    "test": "tsdx test --env=jsdom",
+    "lint": "tsdx lint"
+  },
+  "peerDependencies": {
+    "next": ">=9",
+    "react": ">=16",
+    "tinacms": ">=0.11"
+  },
+  "dependencies": {
+    "next": ">=9",
```
ncphillips: `next`, `react`, and `tinacms` should just be in `peerDependencies` and `devDependencies`.

I assume this came from `next-tinacms-json`?

packages/next-tinacms-markdown/README.md
```diff
@@ -0,0 +1,6 @@
+# next-tinacms-markdown
+
+This package provides helpers for editing content stored in Markdown files in the local filesystem.
+
+- `useMarkdownForm`
+- `useLocalMarkdownForm`
```
ncphillips:
```suggestion
- `useLocalMarkdownForm`
- `useGlobalMarkdownForm`
```

packages/next-tinacms-markdown/src/use-markdown-form.ts
```diff
@@ -0,0 +1,128 @@
...
+ */
+export function useMarkdownForm (
+  markdownFile: MarkdownFile,
+  options: Options = {}
+) {
+  const cms = useCMS()
+
+  const [valuesInGit, setValuesInGit] = useState()
+
+  const valuesOnDisk = useMemo(
```
ncphillips: I vaguely think the `useMemo` is unnecessary here. That callback isn't doing any processing, just creating a new object. I think `const valuesOnDisk = markdownFile` would be fine. 

I like making the new variable just for the sake of making it easier to understand.

I could be wrong on this. We could do a quick check to see if it causes a bunch of unnecessary rendering without the `useMemo`

packages/next-tinacms-markdown/src/use-markdown-form.ts
```diff
@@ -0,0 +1,128 @@
...
+  const writeToDisk = useCallback(formState => {
+    cms.api.git.writeToDisk({
+      fileRelativePath: markdownFile.fileRelativePath,
+      content: toMarkdownString(formState.values),
+    })
+  }, [])
+
+  useWatchFormValues(form, writeToDisk)
+
+  return [values || {markdownBody: markdownFile.markdownBody, frontmatter: markdownFile.frontmatter}, form]
```
ncphillips:
```suggestion
  return [values || markdownFile, form]
```

ncphillips:
Woops! I realized this was a draft PR _after_ I added my comments. 😬

packages/next-tinacms-markdown/package.json
kendallstrautman: oh yeah I just copied it over

packages/next-tinacms-markdown/src/use-markdown-form.ts
kendallstrautman: I was having a hard time getting it to work without `useMemo` but I will do some more testing

kendallstrautman:
haha all good!

packages/next-tinacms-markdown/package.json
kendallstrautman: is `gray-matter` okay being in `dependencies`? I saw that's how it was configured for `gatsby-tinacms-remark`

ncphillips: Yup!

Here's a pseudocode rule of thumb:
```
if package_X is a plugin for package_Y
then package_Y is a peerDependency of package_X
```

packages/next-tinacms-markdown/src/use-markdown-form.ts
ncphillips: Okay cool! If that's the case then I would just add `markdownFile.fileRelativePath` to the dependency array

packages/next-tinacms-markdown/package.json
kendallstrautman: The build fails now without `tinacms` in `dependencies`...which makes sense. Thoughts on how to fix?

ncphillips: If it's in `peerDependencies` ti should also be in `devDependencies`

kendallstrautman:
Going to add inline helpers in another PR, wanted to get all these changes in 👌🏼

PR #658, opened 2020-01-22
fix(deps): stripped out unnecessary deps

ncphillips:
tinacms should be a peer
the others aren't actually used


DirtyF:
Could https://www.npmjs.com/package/depcheck help?

ncphillips:
Possibly! I'll look at making a danger script that uses depcheck

PR #661, opened 2020-01-22
fix: improve wysiwyg sticky toolbar

### LINKED ISSUE #758: WYSIWYG toolbar in sidebar editor jumps to top of page when scrolling
## Description
https://www.loom.com/share/0376dae8696849f8ba1a62ad5def7a8c

### Expected result
When scrolling inside the body of the page, the sidebar shouldn't be affected

#### Tools

| Tool    | Version |
| ------- | ------- |
| Chrome |         |
| iOS      |         |

spbyrne:
This fixes a couple little issues with the sticky wysiwyg toolbar:

- A placeholder is used to occupy the toolbar space when it's fixed to the top of the screen. This prevents content from jumping around when the toolbar sticks/unsticks.
- There's a resize event that will adjust this placeholder size as needed
- There's a sticky prop on the menu that defaults to true. We'll set this to false when it's in the sidebar.
- There was a `null` error coming up a lot that's fixed
- There were times when the toolbar went full width; this is fixed.
- Sets 'sticky' to false for WYSIWYG menu when in sidebar

Closes #758


spbyrne:
# :frowning_face:

PR #677, opened 2020-01-24
feat(next-tinacms-json): adds jsonForm general HOC

kendallstrautman:
This is a general purpose HOC to use when you don't need inline editing. I wanted it to work seamlessly with `inlineJsonForm`, so you could just switch between them. I could use feedback on the best way to pass the `data` or `jsonFile` props. ~~Right now, the content from `jsonFile` will be passed as `data`.~~ The content from `jsonFile` is passed as `jsonFile`

~~Use looks like:~~ **See comments below for updated use example**
```js
function Index({data}) {
  return (
     <h1>
       {data.title}
     </h1>
  )
}

const formOptions = {
  //...
}

const EditableIndex = jsonForm(Index, formOptions)
```


ncphillips:
In the gatsby plugins they pass it as `data: { markdowonRemark }` to match the queries. 

I would either say pass it as `jsonFile` or `data: { jsonFile }`.

What do you think @dwalkr ?

packages/next-tinacms-json/src/json-form.hoc.tsx
```diff
@@ -0,0 +1,39 @@
...
+
+import React from 'react'
+import { FC } from 'react'
+import { NextPage } from 'next'
+import { JsonFile, Options } from './use-json-form'
+import { useLocalJsonForm } from './use-local-json-form'
+
+interface JsonFormProps extends Options {
+  jsonFile: JsonFile
+  options?: Options
```
ncphillips: `JsonFormProps` both extends `Options` and has an `options` property. I would go with just one or the other.

kendallstrautman: good call!

kendallstrautman:
I was just making the `markdownForm` HOC...and I think it makes sense to restructure the props to pass `jsonFile` instead of `data`. Seems more clear imo.

PR #685, opened 2020-01-27
hide menu toggle button if no screen registered

### LINKED ISSUE #680: ability hide global settings button
## Summary

global settings form always is visible even if there's no global form registered and after clicking on it an empty drawer shows up. it should be possible to hide global settings button to prevent confusions.

2020-01-27-14:35:29-screenshot

### Motivation

when there's not any global form or settings, its button still showing up and it might confuse some users.

sijad:
fixes #680


packages/tinacms/src/components/sidebar/Sidebar.tsx
```diff
@@ -42,42 +42,47 @@ export const Sidebar = () => {
   useSubscribable(cms.screens)
   const [menuIsVisible, setMenuVisibility] = useState(false)
   const [ActiveView, setActiveView] = useState<ScreenPlugin | null>(null)
+  const allScreens = cms.screens.all();
+  const showMenu = allScreens.length < 0;
```
ncphillips: Slightly off 😉 

```suggestion
  const showMenu = allScreens.length > 0;
```

PR #698, opened 2020-01-29
Auto generated api documentation

jpatters:
This builds and publishes auto generated api documentation.


typedoc.js
```diff
@@ -18,6 +18,7 @@ limitations under the License.
 
 module.exports = packageJson => {
   const gitRevision = `${packageJson.name}@${packageJson.version}`
+  const outDir = `${__dirname}/docs/${gitRevision}`
```
ncphillips: This is nice 👍

.github/workflows/docs.yml
```diff
@@ -0,0 +1,27 @@
...
+on: pull_request
+
+jobs:
+  build:
+    runs-on: ubuntu-18.04
+    steps:
+      - uses: actions/checkout@v1
+      - uses: actions/setup-node@v1
+        with:
+          node-version: '10.x'
```
ncphillips: Double check if this is needed. IIRC It's in the build script for running on macos

.github/workflows/docs.yml
```diff
@@ -0,0 +1,27 @@
...
+
+      - name: Install
+        run: |
+          npm ci
+          npm run bs
```
ncphillips: I could see this not being necessary

.github/workflows/docs.yml
jpatters: It might not be necessary, but I would rather be explicit about the version of node that it is using.

ncphillips:
@all-contributors please add @jpatters for infra and docs

PR #699, opened 2020-01-29
Inline editing revamp

ncphillips:


packages/react-tinacms-inline/src/inline-form.tsx
```diff
@@ -0,0 +1,88 @@
...
+  children: React.ReactElement | React.ReactElement[] | InlineFormRenderChild
+}
+
+export interface InlineFormRenderChild {
+  (props: InlineFormRendeChildOptions):
+    | React.ReactElement
+    | React.ReactElement[]
+}
+
+export type InlineFormRendeChildOptions = InlineFormState &
```
jamespohalloran: typo on InlineFormRendeChildOptions

Issue #692, opened 2020-01-28
Allow gatsby-plugin-tinacms to be configured to not add Tina component automatically


DaniTulp:
## Summary

I think disabling the sidebar on certain pages would be helpful. I'm creating a store right now and for my use case it would be useful when navigating to the checkout page. 

It would be great to have a little more control over when it's shown.

### Basic example
I don't have a concrete example but maybe there could be an option added like: `manualInit: true`. When set to `true` you would have to call a function to let Tina use the page, this way you could decide on which pages it should be enabled. I do believe this should be an opt-in feature.

If anything I'm saying sounds dumb or impossible, I would love to learn why. I haven't actually dug into the code yet.

### Motivation
For me personally it makes a lot of sense to add this feature, also looking at when cloud (Tina Teams, Gatsby cloud) is more supported and non technical users have to use it. It would solve some UX problems when accessing the sidebar when nothing is actually editable.

If I could be of any assistance I would also love to contribute to this awesome project.

Thanks for all the great work ✌️

ncphillips:
Here's one option for solving this:

* remove`gatsby-plugin-tinacms` from your project.
* copy that plugins [onClientEntry](https://github.com/tinacms/tinacms/blob/master/packages/gatsby-plugin-tinacms/gatsby-browser.tsx#L32-L34) to your `gastby-browser.js`
* Add the `<Tina>` component manually in your site, but only on the pages you want 

I could make an example of this later if you're interested @DaniTulp  but I'm a bit busy for next couple days.

DaniTulp:
Thanks a lot @ncphillips! I managed to get it working.
I do feel like it could be a little easier by setting an option in the `gatsby-config.js` like I described above. That way you only have to do the third step in this solution + you get to keep all your other options inside the config (theming, position). I would be willing to make this work.

ncphillips:
I like that direction! 

I like that it makes use of the existing API to do this–we definitely lean towards smaller simpler APIs so that's a big win for me. If I understand you correctly, it wouldn't need to change to the `tinacms` at all, only the `gatsby-plugin-tinacms`.

I suspect we might need too make a slight change to the [TinaCMS](https://github.com/tinacms/tinacms/blob/master/packages/tinacms/src/tina-cms.ts) class though. We may need to  switch to having some of those sidebar options given to it [here](https://github.com/tinacms/tinacms/blob/master/packages/gatsby-plugin-tinacms/gatsby-browser.tsx#L33).

[...]

PR #704, opened 2020-01-30
#692 Initial functionality for manual initializing TinaCMS for Gatsby.

DaniTulp:
#692  Currently it is very rough and untested. Please let me know what you think and what needs to be changed. @ncphillips


packages/tinacms/src/tina-cms.ts
```diff
@@ -74,4 +83,8 @@ export class SidebarState extends Subscribable {
     this._isOpen = nextValue
     this.notifiySubscribers()
   }
+
+  get position(): "fixed" | "float" | "displace" | "overlay" {
```
ncphillips: I would like to name this type:

```ts
export type SidebarPosition = "fixed" | "float" | "displace" | "overlay" 
```

And then 

```ts
get position(): SIdebarPosition
```

DaniTulp:
@ncphillips I've updated the pull request based on your feedback! Let me know what you think!

PR #706, opened 2020-02-01
Refactoring Experiment: cms plugins as a proxy

ncphillips:
The main benefit of the `PluginTypeManager` was that it let the user find-or-create `PluginTypes` instead of manually initializing them. 

```ts
cms.plugins.findOrCreate('field')
cms.plugins.findOrCreate('content-creator')
```

In reality it was a glorified lookup table with one extra feature.

I somehow discovered the [Proxy](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Proxy) class in Javascript. 

This PR does two things:

* removes the `PluginTypeManager` class and replaces it with a Proxy object. 
* renames `PluginType` to `PluginCollection` which feels way more natural

This means plugins collections can be accessed by name. The below snippet is functionally equivalent to the above snippet.

```ts
cms.plugins.field
cms.plugins['content-creator']
```


ncphillips:
Cool experiment. Too busy right now. I don't feel like exploring this.

PR #707, opened 2020-02-02
feat: add gatsby-tinacms-mdx package

jsjoeio:
This is an active WIP. @ncphillips helped me a little bit but we figured it might be better to start with a draft PR to get some help asynchronously. (I've been too busy with a new job to coordinate outside work. I think this might be easiest.)

## Summary

The goal of this PR is to add a new package called `gatsby-tinacms-mdx` to the tinaCMS/Gatsby package family so that users can use tinaCMS with their Gatsby/MDX sites similar to how they can use it with Markdown. 

I struggled to get my local dev env working well to develop this plugin so I ended up developing it in isolation [here](https://github.com/jsjoeio/gatsby-tinacms-mdx).

### Approach 

I essentially copied the exact same approach in `gatsby-tinacms-remark` but for `mdx` specifically. That seemed to work for the _most_ part but this isn't 100% complete and is not a fully-functioning state. 

### What I Need Help With

- Guidance on testing this locally in the monorepo and in a Gatsby demo site (I haven't had much luck following the instructions in the README/CONTRIBUTING hence why I broke it into a separate repo)
- Feedback on what might be missing to get this fully-functioning
- Any other support that can help us get this ready for the community

Many thanks in advance ❤️

PR #713, opened 2020-02-04
Add gatsby mdx plugin

kendallstrautman:
Copied from [original PR](https://github.com/tinacms/tinacms/pull/707) by @jsjoeio 

This is an active WIP. @ncphillips helped me a little bit but we figured it might be better to start with a draft PR to get some help asynchronously. (I've been too busy with a new job to coordinate outside work. I think this might be easiest.)

## Summary

The goal of this PR is to add a new package called `gatsby-tinacms-mdx` to the tinaCMS/Gatsby package family so that users can use tinaCMS with their Gatsby/MDX sites similar to how they can use it with Markdown. 

I struggled to get my local dev env working well to develop this plugin so I ended up developing it in isolation [here](https://github.com/jsjoeio/gatsby-tinacms-mdx).

### Approach 

I essentially copied the exact same approach in `gatsby-tinacms-remark` but for `mdx` specifically. That seemed to work for the _most_ part but this isn't 100% complete and is not a fully-functioning state. 

### What I Need Help With

- Guidance on testing this locally in the monorepo and in a Gatsby demo site (I haven't had much luck following the instructions in the README/CONTRIBUTING hence why I broke it into a separate repo)
- Feedback on what might be missing to get this fully-functioning
- Any other support that can help us get this ready for the community

Many thanks in advance ❤️

### Unresolved reviews from original PR 
- [x] Lets remove the the `.d.ts` files and add them to the .gitignore
- [x] Nolan's notes on naming `mdx`instead of `_mdx`

Nolan: The whole component would look like this:

```ts
interface MdxFormProps extends Partial<FormOptions<any>> {
  mdx: MdxNode
  render(renderProps: { form: Form; mdx: any }): JSX.Element
  timeout?: number
}

export function MdxForm({ mdx: _mdx, render, ...options }: MdxFormProps) {
  const [mdx, form] = useLocalMdx(_mdx, options)

  return render({ form: form as Form, mdx })
}
```

The `mdx: _mdx` lets you pass it in as `mdx` and then use it as `_mdx`

- [x] Note for @ncphillips  - maybe DeleteAction should be part of the gatsby-tinacms-git plugin. This action could work for Json, Markdown, Mdx, anything stored in git.
- [x] `tinacms` needs to be a peerDep
- [x] @paulbunker noted: I was having trouble getting this branch to build untill I added `"@tinacms/form-builder": "0.2.13"` as a dependency


packages/gatsby-tinacms-mdx/src/mdxFormHoc.tsx
```diff
@@ -0,0 +1,80 @@
...
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+
+*/
+
+import * as React from 'react'
+import { FormOptions, Form } from 'tinacms'
+import { TinaForm } from '@tinacms/form-builder'
```
kendallstrautman: `TinaForm` can be imported directly from `tinacms`. This will solve @paulbunker's issue with needing `form-builder` as a dep

jsjoeio:
Okay, I think I've made all the changes listed by @kendallstrautman, minus the DeleteAction note. 

Thanks for moving this over and cleaning things up a bit! 🙌🏼

packages/gatsby-tinacms-mdx/src/useMdxForm.tsx
```diff
@@ -0,0 +1,226 @@
...
+    })
+  }, [])
+
+  /* eslint-disable-next-line react-hooks/rules-of-hooks */
+  useWatchFormValues(form, writeToDisk)
+
+  return [mdx, form]
+}
+
+export function useLocalMdx(
```
kendallstrautman: This is small, but I think naming this `useLocalMdxForm` will be more cohesive with the other packages. Along with `useGlobalMdxForm`

kendallstrautman:
How are we planning to render the mdx while editing? Are we making a special field?

packages/gatsby-tinacms-mdx/src/useMdxForm.tsx
jsjoeio: Totally agree with you. Consistency is key. Great catch! Just fixed.

jsjoeio:
> How are we planning to render the mdx while editing? Are we making a special field?
> 
> I totally could have set it up wrong but seeing some funky stuff testing the `useLocalMdx` hook...

When I had things in a semi-working state, this is exactly what I ran into. I wasn't exactly sure how to approach this 😅 

Maybe this is what @PaulBunker was working on [here](https://github.com/tinacms/tinacms/pull/714/files#r374678364).

Is that the missing piece?

PaulBunker:
Yes I have editing working :)

kendallstrautman:
@PaulBunker Sweet!! Yeah that 'sticky' menu is another bug being addressed in this [PR](https://github.com/tinacms/tinacms/pull/661).

ncphillips:
Just merged the PR from @PaulBunker should be working better now

PaulBunker:
I'm going to do a little more testing now.

jsjoeio:
Hmm... attempting to test locally on a new machine but running into issues during the build process.

### Steps 

1. Check out `tinacms` repo
2. Switch to branch - `add-gatsby-mdx-plugin`
3. Run `npm install`
4. Run `npm run bootstrap`
5. Run `npm run build`
6. See error:

```shell
Error: /Users/jprevite/dev/oss/tinacms/packages/@tinacms/api-git/src/upload.ts(22,17): 
semantic error TS4058: 
Return type of exported function has or is using name 'Multer' from 
external module "/Users/jprevite/dev/oss/tinacms/node_modules/@types/multer/index" 
but cannot be named.
```

Simplest fix I found was to add `any` as the return type for `createUploader`. Wondering how we might want to solve that. Also found this [S/O post](https://stackoverflow.com/a/43901135).

jsjoeio:
Tested and the form shows up okay on the `mdx` page but when I click save, it's "forever saving"

image

Though when I refresh, the save persists so makes me think something might be up with the saving state not knowing that it's completed and can stop showing the loading icon. 

Demo:

 (ignore the freaky cursor 😅 I don't know why it did that)

PaulBunker:
I'm working a problem where if you only have one file with a certain frontmatter key and you are editing the mdx body. If the body doesn't compile then the frontmatter key is removed from the inferred type in the graphql layer and the whole page craps out. I have a local fix where I am adding the createSchemaCustomization lifecycle method. You could also fix by adding a dummy file with the key that goes missing. Looking for a more generic fix. Watch this space.

PaulBunker:
Upon further inspection I think messing about with the types is the wrong fix, bad jsx/html will mash up the page the file gets saved the server tries to reload and cand process that and then Sh*t hits the fan. 

Perhaps some way of validating before anything is saved to disk. I am concerned about package size and performance though :/

[This](https://www.npmjs.com/package/@mdx-js/runtime) is one option but I'm open to thoughts suggestions.

kendallstrautman:
Played around a bit and it seems to work great so far! @jsjoeio I saw that same saving behavior, but I'm curious if you were being prompted for your password in the terminal. Once I entered my password, the commit & push went through and the saving 'stopped'.

Apologies for the 'push' in my tests, I forgot I didn't make a new test branch off this one and Tina is pretty forward with the push 🖖

jsjoeio:
Ah, to be honest, I didn't even look at my terminal while I was doing this. I'll give it another try when I get some time.

Note, if that part about entering your password is not in the docs already, we should definitely add it!

PaulBunker:
You don't have your ssh keys set up? 😳

jsjoeio:
> You don't have your ssh keys set up? 😳

My ssh keys are set up. Maybe I'm misunderstanding something?

PaulBunker:
I can only imagine if you need to put your password in it's because you either don't have an ssh key attached to your repo so it might be asking for a password when you push or you haven't got ssh-agent running. 

I was only joking anyway Apologies. 🤡

jsjoeio:
Remaining TODOs:
- [ ] Test plugin via class implementation (similar to [RemarkForm](https://tinacms.org/docs/gatsby/markdown/#2-the-render-props-component-remarkform))
- [ ] Fix conflicts

jsjoeio:
Hmm...Scratching my head trying to figure out why the build step is failing

> Error: /home/runner/work/tinacms/tinacms/packages/gatsby-tinacms-mdx/src/useMdxForm.tsx(27,8): semantic error TS2307: Cannot find module 'tinacms'.

- `tinacms` is listed a `devDependency` and a `peerDependency`

jsjoeio:
Strange that we're getting this build error:

```shell
Error: /home/runner/work/tinacms/tinacms/packages/gatsby-tinacms-mdx/src/useMdxForm.tsx(27,8): semantic error TS2307: Cannot find module 'tinacms'.
```

I was able to get around it locally by adding the `gatsby-tinacms-mdx` plugin to the ignore field in the `lerna.json`, running `npm run build` then removing it from the ignore field and building again.

Does this mean it

ncphillips:
IMO this is a good `v0.1.0.` 

There is one major issue with it: if you write invalid MDX then gatsby blows up and you have to manually fix the file from the filesystem. This is really not good, but I think it's okay for v0.1.0. We can merge it into `master` and open a new issue.

ncphillips:
@all-contributors please add @jsjoeio for code

PR #714, opened 2020-02-04
Add gatsby mdx plugin paul

PaulBunker:
The purpose of this pull request is for visibility with others contributing to `add-gatsby-mdx-plugin` branch


packages/gatsby-tinacms-mdx/gatsby-node.js
```diff
@@ -36,7 +36,7 @@ exports.setFieldsOnGraphQLNodeType = ({ type }) => {
       rawMarkdownBody: {
         type: GraphQLString,
         resolve: source => {
-          return JSON.stringify(source.rawBody)
+          return source.rawBody.replace(new RegExp(/^---[\s\S]*---[\n\r]*/), '')
```
PaulBunker: I tried to find a way to do this without a regex.

I also toyed with the idea of having a helper called `removeFrontmatter` and a comment in there explaining the regex.

Open to your thoughts

ncphillips: This is a good start.

I think you're right. This seems like a good usecase to write some comprehensive tests for. i.e. Extracting it to a function inside a new `remove-frontmatter.ts` file and write some tests to make sure it works. I almost guarantee that this RegExp doesn't capture all the situations we need to handle so a comprehensive test suite will be useful. 

Here's a weird example that might break:

```markdown
---
title: hello---
---
World
```

DirtyF: If it helps here's the regexp used by Jekyll to parse YAML front matter: https://github.com/jekyll/jekyll/blob/master/lib/jekyll/document.rb#L13

PaulBunker: Looks like `gray-matter` is being used elsewhere in the plugin. Perhaps this is the wrong thing to be doing anyway. It solved my use-case but may have an effect on others. 

I could use  `matter(mdx.rawMarkdownBody).content`[here](https://github.com/tinacms/tinacms/blob/9c15dd618ae0934a1afac545180c71570a340be0/packages/gatsby-tinacms-mdx/src/to-markdown.ts#L27)

PR #722, opened 2020-02-06
REFERENCE: Media manager

ncphillips:
### This PR is a reference for future work.

The work will be split up into multiple PRs rather then done all at once. 

- #733 Media Store with uploading only

### Shows Off

- Media Manager GUI
  - [ ] Looks Pretty
  - Selecting
    - [x] select individual images
    - [ ] select multiple images
  - Deleting images
    - [x] confirmation dialog
    - [x] deletes all selected images
  - Insert Images
    - [x] Allow selected image to be inserted 
  - Upload Images
    - [ ] File picker for images
    - [x] Drag image(s) onto media manager to upload them
- [x] Global Media Manager, implemented as a screen plugin
- [x] Media Dialog `cms.media.open({ ... })`
- Media Store
  - [ ] Dummy media store educates users
- Image Field Plugin
  - [x] File upload happens through `cms.media.store`
  - [x] Clicking opens the Media Dialog
- [ ] Git Media Store
- [ ] `gatsby-tinacms-git` should set the git media store on the cms
 
Things intentionally left out:

* Search
* Pagination
* Metadata management for individual files
* Media Input plugins (i.e. unsplash)

### DIFF
--- a/packages/tinacms/src/media.ts
+++ b/packages/tinacms/src/media.ts
@@ -1,0 +1,96 @@
+import { Subscribable } from '@tinacms/core'
+
+/**
+
+Copyright 2019 Forestry.io Inc
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+
+*/
+
+/**
+ * Media represents a single media object (file + metadata) in a media store
+ */
+export interface Media {
+  reference: string // internal reference used by the store
+  src: string // filepath or URI to the media object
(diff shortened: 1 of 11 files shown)


packages/tinacms/src/media.ts
```diff
@@ -0,0 +1,111 @@
...
+*/
+
+/**
+ * Media represents a single media object (file + metadata) in a media store
+ */
+export interface Media {
+  reference: string // internal reference used by the store
+  src: string // filepath or URI to the media object
+  previewSrc: string // filepath or URI to a preview of the media (will this always be a URI?)
+}
```
DirtyF: What about other metadata associated to a media? Images should have an alt and title for instance.

ncphillips: Out of scope for this PR. I'll update the description.

Metadata is dependent upon a lot of factors. The file format, storage provider, and usage in the site all effect things. It's far too early to make some kind of opinionated decision about this. Right now, we're leaning towards making it depending on the format. For example, using exif for images, metadata tags for svgs, and god knows what for PDFs. All this is just way more than we want to consider for this initial PR.

DirtyF: (I was simply suggesting a media should be accessible by default, `alt` and `title`, are standards in CommonMark and in HTMLsince 1993.)

PR #725, opened 2020-02-07
Mdx testing

PaulBunker:
The purpose of this branch is because we have multiple people contributing to this feature


PaulBunker:
It seems to me that we should switch the names of the functions in the module. This works for me but I feel like useMdxForm should be imported into the page, I could be wrong though. I've attempted switching some names but I'm running into typescript issues so I'll just leave this here while I figure it out.

ncphillips:
The name changes pushed up now look good to me 👍

PR #728, opened 2020-02-10
Make mdx forms more resilient. 

PaulBunker:
### Problem
Users can add bad data (or even just be in the process of typing good data) and this causes the Gatsby dev server to throw errors and omit pages from the build.

 - Validate mdx before saving to disk to avoid build problems.


ncphillips:
Something really wonky happened here. There's a ton of CHANGELOG and package-lock.json changes

PaulBunker:
I think I need to merge master into the other branch.

Just found `buble-jsx-only` which could solve step one of the problem. I didn't want to stop to try and fix it the merge issue because I'm sick of computers 😂, instead I'm having a headache trying to install the dependency without running npm install in the packages directory

PaulBunker:
That didn't fix it. 😳 I'll copy my changes into a fresh branch and push a new PR

PR #731, opened 2020-02-10
wip: add test for valid markup before save

PaulBunker:
### Problem
Users can add bad data (or even just be in the process of typing good data) and this causes the Gatsby dev server to throw errors and omit pages from the build.

 - Validate mdx before saving to disk to avoid build problems.

UPDATE: I've found a way to check if the form field will cause gatsby to throw errors. Not sure how to go about adding an error state to the form and disabling the save button


PaulBunker:
Next figure out how to set the error state on the form field if we are in an error state

packages/gatsby-tinacms-mdx/src/useMdxForm.tsx
```diff
@@ -161,8 +163,33 @@ export function useMdxForm(
...
 
+  const isValidMdxState = (mdxContent: any) => {
+    const wrapped = `<div>${mdxContent}</div>`
+    try {
+      transform(wrapped, {
+        objectAssign: 'Object.assign',
+      })
+      return true
+    } catch (err) {
+      console.log('🚨 Not saving until your mdx is vallid')
```
DirtyF:
```suggestion
      console.log('🚨 Not saving until your MDX is valld')
```

PaulBunker:
For now I will just use

```
 {
      label: "Body",
      name: "rawMarkdownBody",
      component: "textarea",
      validate: function(value) {
        const wrapped = `<div>${value}</div>`
        try {
          transform(wrapped, {
            objectAssign: "Object.assign",
          })
        } catch (err) {
          return "🚨 Malformed MDX please fix before save"
        }
      },
    },
```

In the page component where I define my tina form fields. It's not ideal because I'm not setting the state of the form on't the field but it seems to work better than doing the check in the plugin at the moment. I'll have another crack asap but I think I need to look a little more into Final Form.

PaulBunker:
image

PaulBunker:
kindafix

ncphillips:
Why did you decide to close this @PaulBunker ?

PaulBunker:
I just thought it looked messy. I've not given up just struggling a little.

PR #734, opened 2020-02-11
Git remote configuration refactor

### LINKED ISSUE #178: Uploaded images should be committed to the repo

jpatters:
fixes #178


packages/@tinacms/api-git/src/router.ts
```diff
@@ -24,114 +24,55 @@ import * as express from 'express'
...
-  const CONTENT_ABSOLUTE_PATH = path.join(REPO_ABSOLUTE_PATH, CONTENT_REL_PATH)
+export function router(config: Partial<GitRouterConfig> = {}) {
+  const {
+    pathToRepo,
+    pathToContent,
+    defaultCommitMessage,
+    defaultCommitName,
+    defaultCommitEmail,
+    pushOnCommit
+  } = setDefaults(DEFAULT_OPTIONS, config)
```
ncphillips: Object spread allows you to do this without directly calling Object.assign

```suggestion
  } = { ...DEFAULT_OPTIONS, ...config }
```

jpatters: Since config is a partial, will the types be correct?

ncphillips: I believe so

packages/gatsby-tinacms-git/gatsby-node.ts
```diff
@@ -19,6 +19,8 @@ limitations under the License.
 import { router as gitRouter, GitRouterConfig, configureGitRemote } from '@tinacms/api-git'
 
 exports.onCreateDevServer = ({ app }: any, options: GitRouterConfig) => {
-  configureGitRemote(options)
+  if (process.env.TINA_CEE !== undefined) {
```
ncphillips: Add comment about why it's a undefined check.
Double check the conventions for that too.

packages/@tinacms/api-git/src/repo.ts
```diff
@@ -0,0 +1,201 @@
...
+  fileIsInRepo(filepath: string) {
+    const fileIsInRepo = checkFilePathIsInParent(
+      filepath,
+      this.contentAbsolutePath
+    )
+    if (fileIsInRepo) {
+      return true
+    } else {
+      throw new Error(
+        `Failed to write to: ${filepath} \nCannot write outside of the content directory.`
```
ncphillips: This method shouldn't throw the error. That should happen in `writeFile` and `deleteFiles`. Right now `deleteFiles` would throw an error saying "Failed to write"

e.g.

```ts
if (this.fileIsInRepo(fileAbsolutePath)) {
  writeFile(fileAbsolutePath, contents)
} else {
  throw new Error(
    `Failed to write to: ${filepath} \nCannot write outside of the content directory.`
  )
}

File: packages/@tinacms/api-git/src/utils/paths.ts [hunk 4]
```diff
@@ -16,21 +16,17 @@ limitations under the License.
...
-  try {
-    return await repo.show([`HEAD:${fileRelativePath}`])
-  } catch (e) {
-    return fs.readFileSync(path.join(pathRoot, fileRelativePath), 'utf8')
-  }
+import * as os from 'os'
+
+export function checkFilePathIsInParent(filepath: string, parent: string) {
+  const fullpath = path.resolve(filepath)
+  const repopath = (() => {
```
ncphillips: Instead of putting this in an immediately invoked function I think you could just do:

```ts
    let repopat: string
    if (os.type() === 'Windows_NT') {
      repopath = path.resolve(parent).replace(/\\+$/, '') + '\\'
    } else {
      repopath = path.resolve(parent).replace(/\/+$/, '') + '/'
    }
```

packages/@tinacms/api-git/src/utils/paths.test.ts
```diff
@@ -0,0 +1,35 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+
+*/
+
+import * as path from 'path'
+
+import { checkFilePathIsInParent } from './paths'
+
+describe('checkFilePathIsInParent', () => {
```
ncphillips: We run the tests on windows too. We could probably have a couple tests to check the behaviour on windows.

jpatters: I'm not sure what you mean. These tests do run on windows and verify that the behaviour is the same as on linux.

ncphillips:
@all-contributors please add @jpatters for code

PR #748, opened 2020-02-13
Tinacms constructor corrections

ncphillips:
Some recent "refactoring" actually broke APIs for hiding and theming the sidebar. This PR corrects that.

### DIFF
--- a/packages/gatsby-plugin-tinacms/options.ts
+++ b/packages/gatsby-plugin-tinacms/options.ts
@@ -1,3 +1,21 @@
+/**
+
+Copyright 2019 Forestry.io Inc
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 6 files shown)


packages/gatsby-plugin-tinacms/options.ts
```diff
@@ -0,0 +1,24 @@
...
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+
+*/
+
+import { TinaCMSConfig } from 'tinacms/build/tina-cms'
```
dwalkr: wrong import path?

ncphillips: Sure is. Apparently that's not exported.

packages/tinacms/src/components/Tina.tsx
```diff
@@ -31,25 +31,34 @@ export interface TinaProps {
...
   useSubscribable(cms.sidebar)
   const theme: ThemeProps['theme'] = React.useMemo(
     () => ({
-      tinacms: merge(DefaultTheme, themeOverrides) as Theme,
+      tinacms: merge(DefaultTheme, cms.sidebar.theme, themeOverrides) as Theme,
     }),
     [DefaultTheme, themeOverrides]
   )
 
+  if (typeof !hidden === 'undefined') {
```
dwalkr: typo? `!` will always cast as boolean

PR #765, opened 2020-02-16
Use form default initial values

ncphillips:
Last piece that makes `react-tinacms-inline` work with tinacms.org

I made sure to check that both the development and built version of the site still work. I think it's clear sailing from here. 🤞

### DIFF
--- a/packages/@tinacms/react-core/src/use-form.ts
+++ b/packages/@tinacms/react-core/src/use-form.ts
@@ -108,7 +108,7 @@ function createForm(options: FormOptions<any>, handleChange: any): Form {
  * This hook is useful when dynamically creating fields, or updating
  * them via hot module replacement.
  */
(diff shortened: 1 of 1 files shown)


packages/@tinacms/react-core/src/use-form.ts
```diff
@@ -48,7 +48,20 @@ export const useCMSForm = useLocalForm
...
+   *
+   * If the form helper is using `watch.values`, which would contain
+   * the current state of the form, then we set that to the `initialValues`
+   * so the form is initialized with some state.
+   *
+   * This is beneficial for SSR and will hopefully not be noticeable
+   * when editing the site as the actual `initialValues` will be set
+   * behind the scenes.
+   */
+  options.initialValues = options.initialValues || watch.values
```
ncphillips: This is the important line in this PR

Issue #754, opened 2020-02-14
Enable/Disable Tina from inside <Tina> context 


jamespohalloran:
## Summary

I would like to enable/disable the Tina "toggle" button from within a page. 

### Basic example

I have a NextJS site with a main layout using: 
```

export default withTina(Site, {
  cms: {
// ...
  },
  sidebar: {
    hidden: true,
  },
})
```

From within a view, I have a "preview" prop (based from getStaticProps). This preview prop isn't available in the main layout. I would like to be able to toggle Tina's edit-mode from within this view based on this `preview` property.

### Motivation

Showing/Hiding the sidebar toggle is a bit tricky when it's not based on build environment variables

ncphillips:
The [isOpen](https://github.com/tinacms/tinacms/blob/master/packages/tinacms/src/tina-cms.ts#L101-L108) API could be a good model:

```ts
get hidden() {
  return this._hidden
}
set hidden(val: boolean) {
  this._hidden = nextValue
  this.notifySubscribers()
}
```

kendallstrautman:
Something like this - https://github.com/tinacms/tinacms/pull/769/commits/bf2956c3be73f9aff9843dcaeb0ac500d9cd285e - via #769?

jamespohalloran:
Looks good to me!

ncphillips:
Implemented by #769

PR #769, opened 2020-02-19
Adds `hidden` Setter to Sidebar

kendallstrautman:
So that one can manually hide or show the sidebar like this: 

```js
  const cms = useCMS()
  cms.sidebar.hidden = true
```

Attemps to address #754 

- [x] Documentation PR Accepted


packages/tinacms/src/tina-cms.ts
```diff
@@ -106,4 +106,16 @@ export class SidebarState extends Subscribable {
     this._isOpen = nextValue
     this.notifiySubscribers()
   }
+
+  private _isHidden: boolean = this.hidden
+
+  get isHidden() {
```
ncphillips: Looks good! `SidebarState` already has the `hidden` flag. If this was switched to `hidden` then it would be a non-breaking change.

kendallstrautman: Good call, [this commit](https://github.com/tinacms/tinacms/pull/769/commits/12f71620bd878d814069f2e9a6bea20bf4a087b0) should do that.

kendallstrautman:
Docs PR → https://github.com/tinacms/tinacms.org/pull/108

PR #784, opened 2020-02-20
@tincams/api-git warns about missing Repo

ncphillips:
There was an API change in `@tinacms/api-git#router`. It was accounted for in the gatsby plugin but not for Next. This adds a warning about that and updates the `demo-next`.

### DIFF
--- a/packages/@tinacms/api-git/src/router.ts
+++ b/packages/@tinacms/api-git/src/router.ts
@@ -49,6 +49,22 @@ export function router(repo: Repo, config: Partial<GitRouterConfig> = {}) {
     pushOnCommit,
   }: GitRouterConfig = { ...DEFAULT_OPTIONS, ...config }
 
+  if (!repo) {
+    /**
+     * Incase js users forget to pass in `repo`.
+     */
+    throw new Error(
+      '@tinacms/api-git#router(repo, config): Parameter `repo` is is missing.'
+    )
+  } else if (!(repo instanceof Repo)) {
+    /**
+     * Maintains backwards compatibility. Types are intentionally
+     * left out to discourage people from using it.
+     */
+    const repoConfig: any = repo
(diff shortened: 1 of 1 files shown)


packages/@tinacms/api-git/src/router.ts
```diff
@@ -49,6 +49,15 @@ export function router(repo: Repo, config: Partial<GitRouterConfig> = {}) {
     pushOnCommit,
   }: GitRouterConfig = { ...DEFAULT_OPTIONS, ...config }
 
+  /**
+   * JS users might not pass in `repo`.
+   */
+  if (!repo || repo instanceof Repo) {
+    throw new Error(
+      '@tinacms/api-git#rotuer(repo, config): Parameter `repo` is an instance of `Repo` '
```
dwalkr:
```suggestion
      '@tinacms/api-git#router(repo, config): Parameter `repo` is not an instance of `Repo` '
```

packages/@tinacms/api-git/src/router.ts
```diff
@@ -49,6 +49,22 @@ export function router(repo: Repo, config: Partial<GitRouterConfig> = {}) {
     pushOnCommit,
   }: GitRouterConfig = { ...DEFAULT_OPTIONS, ...config }
 
+  if (!repo) {
+    /**
+     * Incase js users forget to pass in `repo`.
+     */
+    throw new Error(
+      '@tinacms/api-git#router(repo, config): Parameter `repo` is is missing.'
```
dwalkr:
```suggestion
      '@tinacms/api-git#router(repo, config): Parameter `repo` is missing.'
```

error messages are hard

ncphillips:
@jpatters what's `atob`? It's being used in `configure.ts` but it isn't being imported. That seems like it might be a bug 🤔

PR #790, opened 2020-02-21
Fix next tinacms markdown

ncphillips:
Navigating between blog posts caused problems.

The `writeToDisk` would always write to the _first_ file you visited.

This was because the callback was grabbing `fileRelativePath` via closure, instead of through form values


packages/demo-next/data/blog/apple.md
```diff
@@ -0,0 +1,3 @@
+---
+title: Banana
```
jamespohalloran:
```suggestion
title: Apple
```

packages/next-tinacms-markdown/src/use-markdown-form.ts
```diff
@@ -68,6 +68,7 @@ export function useMarkdownForm(
   const label = options.label || markdownFile.fileRelativePath
   const fields = options.fields || []
   const actions = options.actions || []
+  console.log(id)
```
jamespohalloran:
```suggestion

```

PR #798, opened 2020-02-24
Fix editor scrolling when opening modals.

jpuri:
It turned out fix for scroll in storybook was simpler. Here is PR to remove autoFocus from inputs in modals.


DirtyF:
We still want to focus on the URL field by default when adding a link.

jpuri:
Hey @DirtyF , let me try to do that using JS code.

DirtyF:
Given I run gatsby-demo
When I click on an image
Then modal is not centered

image_modal-798_

jpuri:
Hey @DirtyF : seems to work at my end in both chrome and safari, 
Screenshot 2020-02-24 at 6 49 20 PM

Can you share a recording may be.

jpuri:
Hey @DirtyF , I just added to PR code to focus inputs using javascript.

DirtyF:
Scrolling issues are gone, modal is now centered on image select.

When I select an image, then title field is focused in the modal (shoult be alt IMO)
When I toggle on the link, URL field is not focused.

Not a blocker to merge this PR, we can focus on focus issue in another PR 😉

packages/@tinacms/fields/src/Wysiwyg/state/plugins/Menu/ImageMenu/index.tsx
```diff
@@ -82,6 +83,12 @@ export default (props: FloatingImageMenu) => {
     setAlt(node.attrs.alt)
   }, [selectedImage.node])
 
+  useEffect(() => {
+    setTimeout(() => {
+      if (inputRef.current) inputRef.current.focus()
+    })
+  }, [inputRef])
```
ncphillips: The `inputRef` object never actually changes. If you want this to re-run every time it's value changes then it would look like this:

```ts
  }, [inputRef.current])
```

If you only want it to run the once then it would be:

```ts
  }, [])
```

jpuri: I think this conditions can be improved, I will check.

jpuri: I tested this and found that current condition is working well.

PR is good to be merged.

Issue #783, opened 2020-02-20
gatsby-demo: error when toggling draft on and off again


DirtyF:
## Description

```gherkin
Given I am editing a blog post on gatsby demo
When I toggle the draft off and on again
Then the following error is displayed 👇 
```

```sh
Error: useRemarkForm(markdownRemark) 'markdownRemark' was not found on the top node of the graphql query

1. Check if the 'markdownRemark' attribute is included in the GraphQL query. For example:

export const pageQuery = graphql`
  query BlogPostBySlug($slug: String!) {
    markdownRemark(fields: { slug: { eq: $slug } }) {
      ...TinaRemark
      // etc...
    }
 }

2. If you are using an alias, For example:

   export const pageQuery = graphql`
     query BlogPostBySlug($slug: String!) {
       markdownRemark: markdownRemark(fields: { slug: { eq: $slug } }) {
          ...TinaRemark
       // etc...
       }
    }

  then you can optionally use the 'queryName' option to specify a new property in place of 'markdownRemark':

    export const remarkForm(BlogPostTemplate, { queryName: markdownRemark })
`
```

### Expected result

No error when toggling draft status.

#### Tinacms Package Versions

current `master`:

```
demo-gatsby@1.6.3 /Users/frank/tina/tinacms/packages/demo-gatsby
├── UNMET DEPENDENCY @tinacms/core@^0.7.1
├── UNMET DEPENDENCY @tinacms/fields@^0.5.0
├── UNMET DEPENDENCY @tinacms/form-builder@^0.2.15
├── UNMET DEPENDENCY @tinacms/icons@^0.5.2
├── UNMET DEPENDENCY @tinacms/styles@^0.1.2
├── UNMET DEPENDENCY gatsby-plugin-tinacms@^0.1.24
├── UNMET DEPENDENCY gatsby-tinacms-git@^0.2.19
├── UNMET DEPENDENCY gatsby-tinacms-json@^0.8.8
├── UNMET DEPENDENCY gatsby-tinacms-mdx@^0.0.1
├── UNMET DEPENDENCY gatsby-tinacms-remark@^0.7.3
├── UNMET DEPENDENCY gatsby-tinacms-teams@^0.1.4
├── UNMET DEPENDENCY react-tinacms-blocks@^0.3.1
├── UNMET DEPENDENCY tinacms@^0.13.1
```

#### Tools

| Tool    | Version |
| ------- | ------- |
| browser |  Safari       |
| os      |     10.15.3    |
[...]

PR #802, opened 2020-02-24
chore: avoid double-sourcing content in gats demo

dwalkr:
This is likely causing some weird bugs in the gatsby demo, I suspect it is the cause of #783


DirtyF:
I still get the screen of death on this branch...

dwalkr:
@DirtyF try running `gatsby clean` in the demo before starting the dev server again, and let me know if it's still happening

DirtyF:
@dwalkr sadly, it does.

PR #806, opened 2020-02-25
build: add husky to make code prettier

ncphillips:
**What?** Executes `prettier` in a pre-commit hook for all staged files.

**Why?** Contributors don't always have a Prettier extension installed in their editor. There's a number of files that are formatted incorrectly because of this. This will fix existing files and prevent files from being committed without being formatted in the future.

* [Prettier Docs](https://prettier.io/docs/en/precommit.html#option-2-pretty-quickhttpsgithubcomazzpretty-quick)


ncphillips:
I don't know what's happening. This commit is in master and I can't merge this PR. Closing.

PR #807, opened 2020-02-25
fix: blockquote option to toggle and not nest.

jpuri:
Blockquote menu option to toggle.


ncphillips:
I'll look at fixing `demo-gatsby`

@DirtyF the `demo-next` is still working

PR #813, opened 2020-02-25
Fix demo gatsby

ncphillips:


packages/demo-gatsby/gatsby-config.js
```diff
@@ -67,9 +67,8 @@ module.exports = {
     {
       resolve: `gatsby-source-filesystem`,
       options: {
-        path: `${__dirname}/content`,
+        path: `${__dirname}/content/mdx`,
```
DirtyF: What about Markdown content?

PR #817, opened 2020-02-26
Add option to prevent server overload

PaulBunker:
Sometimes the build takes longer than a second and causes issued in gatsby.js development server

Add option to increase the time between file writes.

(This took longer than the number of lines of code would suggest 😂)

Future work Listen to gatsby __emitter for the result of a file write being completed and take the guesswork / magic number out

- [x] [Documentation PR Approved](https://github.com/tinacms/tinacms.org/pull/151)

### DIFF
--- a/packages/@tinacms/api-git/src/server.ts
+++ b/packages/@tinacms/api-git/src/server.ts
@@ -27,7 +27,6 @@ export interface GitServerConfig extends GitRouterConfig {
   pathToRepo: string
   pathToContent: string
   gitRemote?: string
-  maxBuildTime?: string
   sshKey?: string
 }
 
--- a/packages/@tinacms/api-git/src/server.ts
+++ b/packages/@tinacms/api-git/src/server.ts
@@ -27,6 +27,7 @@ export interface GitServerConfig extends GitRouterConfig {
   pathToRepo: string
   pathToContent: string
   gitRemote?: string
+  maxBuildTime?: string
   sshKey?: string
 }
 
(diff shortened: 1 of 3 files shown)


ncphillips:
I like the `file-writer` changes 👍 

However, I'd rather not change the `server` or `gatsby-node`. I don't think the server should be setting environment variables. It should be up to the user to set the environment variable on their machine or to the config of their container config.

packages/@tinacms/api-git/src/server.ts
```diff
@@ -27,6 +27,7 @@ export interface GitServerConfig extends GitRouterConfig {
   pathToRepo: string
   pathToContent: string
   gitRemote?: string
+  maxBuildTime?: string
```
ncphillips: The server doesn't actually use this variable, so it shouldn't expect to receive it.

```suggestion
```

packages/gatsby-tinacms-git/gatsby-node.ts
```diff
@@ -32,9 +32,11 @@ exports.onCreateDevServer = (
     pathToContent,
     gitRemote,
     sshKey,
+    maxBuildTime,
     ...routerOptions
   } = options
 
+  process.env.MAX_BUILD_TIME = maxBuildTime || '1000'
```
ncphillips: I don't think we should be here. The user should be responsible for setting their environment variables
```suggestion
```

PaulBunker:
This was the only way I could figure out how to set the variable from inside the gatsby-config.js file do you think it would still work without these lines?

ncphillips:
You would run the command like:

```
MAX_BUILD_TIME=3000 gatsby develop
```

[Twilio has a good doc on node environment variables](https://www.twilio.com/blog/2017/08/working-with-environment-variables-in-node-js.html)

ncphillips:
Just tested it out, works great. Just gotta update the docs before merging.

ncphillips:
Just renamed the variable.

Went with `TINA_GIT_DEBOUNCE_MS` to make it it's purpose more clear.

PR #822, opened 2020-02-28
feat: added API for triggering alerts

### LINKED ISSUE #821: TinaCMS Alerts API
## Summary

Visual alerts on the success or failure of actions is important for a variety of cases. TinaCMS would benefit greatly from a consistent API for creating alerts.

- [x] Finalize API

**The details of this API are not final. Feel free to add any comments or concerns.**

**Any changes will be added to to this document. You can view the edit history to gain more context.**
### Examples

**Four Types of Alerts**
```ts
cms.alerts.info("Quebec is a province in Canada.")    // Gray
cms.alerts.error("You can't use cheddar in poutine!") // Red
cms.alerts.warn("We'd prefer you speak French")       // Yellow
cms.alerts.success("La séparation est finie !")         // Green
```

**Default**

By default, alerts will be visible for 2000 ms.

```ts
cms.alerts.success("La séparation est finie !")
```

**Timeout**

A `number` can be passed as the second argument to indicate how long the
alert should be visible. 

```ts
cms.alerts.success("La séparation est finie !", 5000)
```

**Manual Close**

Alternatively, a falsy value can be passed as the second argument to indicate that 
the alert should be visible until the user dismisses.

```ts
cms.alerts.warn("We'd prefer you speak French", 0)
cms.alerts.warn("We'd prefer you speak French", false)
cms.alerts.warn("We'd prefer you speak French", null)
```

**Programmatic Close**

The alert functions return a callback that will dismiss the alert.

```ts
const closeAlert = cms.alerts.info("Quebec is a province in Canada.", false)   

waitForQuebecToSeparate()
  .then(() => closeAlert())
  .then(() => cms.alerts.success("La séparation est finie i!"))
```

**Interfaces**
```ts
interface Alerts {
  info(message: string, until?: number): () => void
  error(message: string, until?: number): () => void
  warn(message: string, until?: number): () => void
  success(message: string, until?: number): () => void
}
```

### Motivation

Use cases this supports:

- Success/Failure when saving forms
- Success/Failure when creating PR

### Related Issues

- #600

ncphillips:
Closes #821

Example:

```
cms.alerts.success("Form Saved", 1000)
```

- [x] Alert Styling

## Details

- Alerts are kept in a map
- The unique key is the alert's message
- Default timeout is 3000ms

UI

- Clicking on Alert dismisses it
- Alerts stack up


packages/tinacms/src/tina-cms/alerts.ts
```diff
@@ -0,0 +1,80 @@
...
+    message: string,
+    timeout: number = 3000
+  ): () => void {
+    const alert = {
+      level,
+      message,
+      timeout,
+    }
+
+    this.alerts.set(alert.message, alert)
```
dwalkr: This will cause two alerts with the same message to overwrite each other. I'm guessing this is a feature and not a bug, but is there a scenario where we might not want that to happen?

ncphillips: Honestly not sure 🤷‍♂

dwalkr: because pushing subsequent alerts with the same message doesn't even change their appearance in alert order, in the event 2 alerts with the same message (e.g. "saved successfully") are pushed, there is no visual feedback

therefore I have elected to generate a unique ID per alert

PR #847, opened 2020-03-05
chore: remove outdated template dir

ncphillips:
This directory hasn't been used to create
new packages in a while. IMO it's no longer
relevant, as we'll probably want to use tsdx


ncphillips:
Just looking for a place to do this.

@all-contributors please add @mitchmac for blog

PR #867, opened 2020-03-10
react-tinacms-editor

### LINKED ISSUE #865: WYSIWYG Package
In order to be able to load the editor only when it's needed (when you click on edit in an inline field for instance) we need to out the WYSIWYM in its own package.

This could save a bunch on JS on the initial load and makes Tina significantly faster.

ncphillips:
Pull the `Wysiwyg` into it's own package

closes #865


packages/tinacms/package.json
```diff
@@ -47,10 +47,12 @@
     "lodash.merge": "^4.6.2",
     "react-beautiful-dnd": "^11.0.5",
     "react-datetime": "^2.16.3",
-    "react-dismissible": "^1.1.5"
+    "react-dismissible": "^1.1.5",
+    "react-tinacms-editor": "^0.0.0"
```
DirtyF: `0.0.0`?

ncphillips: the version number will get bumped automatically before publishing

PR #871, opened 2020-03-10
Move inline blocks and fields from tinacms.org

kendallstrautman:
- [x] Update demo-gatsby to new inline blocks api

Notes: 
- Going to add inline image in another PR
- The Blocks Controls are a little wonky and not showing up in `demo-gatsby`, assuming this is due to styling stuff cause the controls seem mounted.


kendallstrautman:
Trying to figure out how best to handle the `react-tinacms-inline` dependency within `Wysiwyg`... We need to import `InlineField` from it. 

When I make it a direct dependency, I get this error `lerna ERR! ECYCLE Dependency cycles detected, you should fix these!`. And when I make it a peerDep, I get this error ` Cannot find module 'react-tinacms-inline'`

ncphillips:
Hmmm...that makes sense. Maybe just put inlineWysiwyg in react-tinacms-inline for now.

kendallstrautman:
For some reason I'm getting a weird error where the block template is undefined when toggling edit mode. Looking at the template and data it seems okay...

packages/demo-gatsby/src/components/blog-blocks.js
```diff
@@ -16,95 +16,77 @@ limitations under the License.
...
-                index={props.index}
-                templates={props.templates}
-              />
-            </div>
-          )
-      }}
-    />
+    <InlineForm form={form}>
+      <InlineBlocks name="rawFrontmatter.blocks" blocks={BLOCKS}></InlineBlocks>
+    </InlineForm>
```
ncphillips: So much simpler lol

packages/demo-gatsby/src/components/blog-blocks.js
```diff
@@ -16,95 +16,77 @@ limitations under the License.
...
 // Image Block Component
 function EditableImage(props) {
   return (
-    <BlockWrapper {...props}>
+    <BlocksControls index={props.index}>
       <img {...props.data} />
-    </BlockWrapper>
+    </BlocksControls>
   )
 }
```
ncphillips: omg that's so much better

packages/react-tinacms-inline/src/inline-field-textarea.tsx
```diff
@@ -0,0 +1,100 @@
...
+    width: calc(100% + 2rem);
+    height: calc(100% + 2rem);
+    border: 3px solid ${color.primary()};
+    border-radius: ${radius()};
+    opacity: 0;
+    pointer-events: none;
+    z-index: 1000;
+    transition: all 150ms ease-out;
+  }
+`
```
ncphillips: Maybe this should be in it's own file?

```
src/style/input-focus-wrapper.tsx
```

Thoughts @spbyrne?

packages/react-tinacms-inline/src/inline-wysiwyg.tsx
```diff
@@ -0,0 +1,44 @@
...
+
+import * as React from 'react'
+import { InlineField } from './inline-field'
+import { Wysiwyg } from 'react-tinacms-editor'
+
+interface InlineWysiwygFieldProps {
+  name: string
+  sticky?: string
+  children: any
+}
```
ncphillips: We should add [`format`](https://github.com/tinacms/tinacms/blob/fix-tina-provider-name/packages/react-tinacms-editor/src/Wysiwyg.tsx#L29-L34) to the props and pass it through.

```suggestion
interface InlineWysiwygFieldProps {
  name: string
  sticky?: string
  children: any
  format?: Format
}
```

packages/react-tinacms-inline/src/inline-field-textarea.tsx
kendallstrautman: yeah im down with that

packages/react-tinacms-inline/src/inline-field-textarea.tsx
```diff
@@ -0,0 +1,68 @@
...
+        }
+        return <>{input.value}</>
+      }}
+    </InlineField>
+  )
+}
+
+export const InlineTextarea = styled(({ ...styleProps }) => {
+  return <TextareaAutosize {...styleProps} />
+})`
```
ncphillips:
```suggestion
export const InlineTextarea = styled(TextareaAutosize)`
```

kendallstrautman: yeah thats much easier to read

PR #873, opened 2020-03-11
fix: renamed Tina to TinaProvider

ncphillips:
The new name the purpose of the component.

The old name is still accessible, but it has been deprecated

```tsx
import { TinaCMS, TinaProvider } from "tinacms"

const cms = new TinaCMS()

function App () {
  return <TinaProvider cms={cms}> ... </TinaProvider>
}
```

### DIFF
--- a/packages/tinacms/src/index.ts
+++ b/packages/tinacms/src/index.ts
@@ -30,7 +30,13 @@ export { TinaCMS as CMS } from './tina-cms'
 /**
  * Tina Sidebar
  */
-export { Tina, TinaProviderProps as TinaProps } from './components/Tina'
+export {
+  TinaProvider,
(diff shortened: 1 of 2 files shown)


packages/tinacms/src/index.ts
```diff
@@ -30,7 +30,7 @@ export { TinaCMS as CMS } from './tina-cms'
 /**
  * Tina Sidebar
  */
-export { Tina, TinaProps } from './components/Tina'
+export { Tina, TinaProviderProps as TinaProps } from './components/Tina'
```
dwalkr: do we also need to export `TinaProvider` here?

ncphillips: Oh shit that was a bad rename

PR #874, opened 2020-03-11
Move Theme System to CSS Custom Properties

spbyrne:
## Current Issues With Theme
1. It's a constant source of TS errors
2. The theme object is constructed by merging the user's sidebar config theme with the default theme. This is cumbersome and creates problems with components used in isolation.
3. Customizing the theme can't be done via global CSS
4. The theme helpers are an unnecessary abstraction / nuisance 
5. It's difficult to locally theme components as needed

## Proposed Solution:
Strip out this entire theme system and replace it with root level css custom properties. 

## Benefits of Solution:
1. No CSS helper functions needed; all the properties are available everywhere.
2. The global theme can be easily overwritten by the user in their site styles at a global or local level.
3. No more `@tinacms/styles` related TS errors! 

## Breaking Changes
1. Anything using theme helpers will break. They need to be replaced with references to the custom properties: `${color.grey(2)}` changes to `var(--tina-color-grey-2)`. I've done this for things inside the monorepo.
2. Theme configuration via passing a theme to the sidebar component will no longer work. This is replaced by overriding these custom properties in your own stylesheet, giving you more control.

## Documentation

- [x] Why we’re making the change (Blog Post)
- [x] How to use the new theming system (Docs updates and additions)
- [ ] How to migrate from the old theming system to the new one (Release Notes)


ncphillips:
The code all looks good to me 👍 

We need some eyes actually testing it out

DirtyF:
Did some local tests (sorry for the unrelated commits). This does not seem to change anything related to this PR scope.

ncphillips:
From a technical standpoint I'm happy, but we need to prep some documentation.

We need to explain 3 things:

1. Why we’re making the change (Blog Post)
2. How to use the new theming system (Docs updates and additions)
3. How to migrate from the old theming system to the new one (Release Notes)

PR #882, opened 2020-03-14
Add webpack helper package

jeffsee55:
This should allow us to easily work across repos locally, see README


packages/@tinacms/webpack-helpers/readme.md
```diff
@@ -0,0 +1,43 @@
...
+
+...
+
+module.exports = {
+  webpack: (config, { buildId, dev, isServer, defaultLoaders, webpack }) => {
+    if (dev) {
+      tinaWebpackHelpers.aliasTinaDev(config, '../tinacms')
+    }
+    return config
+  },
```
ncphillips: Is this specifically Next.js?

packages/@tinacms/webpack-helpers/package.json
```diff
@@ -0,0 +1,20 @@
...
+  "main": "index.js",
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/tinacms/tinacms.git",
+    "directory": "packages/@tinacms/webpack-alias"
+  },
+  "scripts": {
+    "watch": "tinacms-scripts watch",
+    "build": "tinacms-scripts build",
+    "dev": "tinacms-scripts dev"
```
ncphillips: These commands are unnecessary:

```suggestion
    "watch": "echo 'watch'",
    "build": "echo 'build'",
    "dev": "echo 'dev'"
```

packages/@tinacms/webpack-helpers/readme.md
```diff
@@ -0,0 +1,43 @@
+### Tina Webpack Helpers
+
+---
+
+Linking apps to a monorepo can be tricky. Tools like `npm link` are buggy and introduce inconsistencies with module resolution. If multiple modules rely on the same package you can easily end up with multiple instances of that package, this is problematic for packages like `react` which expect only one instance. If your app uses `webpack` you can get around these issues by ensuring that your dependencies come from a specific path on your system. This is especially helpful for working on the TinaCMS monorepo while using it's packages in your app. `@tinacms/webpack-helpers` makes it easy to set up:
```
DirtyF:
```suggestion
Linking apps to a monorepo can be tricky. Tools like `npm link` are buggy and introduce inconsistencies with module resolution. If multiple modules rely on the same package you can easily end up with multiple instances of that package, this is problematic for packages like `react` which expect only one instance. If your app uses `webpack` you can get around these issues by ensuring that your dependencies come from a specific path on your system. This is especially helpful for working on the TinaCMS monorepo while using its packages in your app. `@tinacms/webpack-helpers` makes it easy to set up:
```

ncphillips: Github diff failed. What was changed?

ncphillips:
Note: This doesn't work Gatsby plugins.

ncphillips:
I added documentation to the README for gatsby and it's plugins

packages/@tinacms/webpack-helpers/readme.md
DirtyF: using it's packages -> using its packages

kendallstrautman:
We should add this to the contributing docs 

- [ ] - PR for docs accepted

ncphillips:
@all-contributors please add @jeffsee55 for infra

PR #903, opened 2020-03-17
fix: use TinaProvider instead of Tina

DirtyF:
Following #877

### DIFF
--- a/packages/demo-gatsby/gatsby-browser.js
+++ b/packages/demo-gatsby/gatsby-browser.js
@@ -19,13 +19,11 @@ distributed under the License is distributed on an "AS IS" BASIS,
 // custom typefaces
 import "typeface-montserrat"
 import "typeface-merriweather"
-import {Tina} from "tinacms";
-import * as React from "react";
+import * as React from "react"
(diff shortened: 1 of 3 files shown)


packages/demo-gatsby/gatsby-browser.js
```diff
@@ -19,13 +19,11 @@ distributed under the License is distributed on an "AS IS" BASIS,
...
-export function wrapRootElement({element}) {
-    return (
-        <Tina cms={window.tinacms}>
-            {element}
-        </Tina>
-    )
+import * as React from "react"
+import { TinaProvider, TinaCMS } from "tinacms"
+
+const cms = new TinaCMS()
```
ncphillips: No need to create that instance
```suggestion
import { TinaProvider } from "tinacms"
```

ncphillips:
@all-contributors please add @DirtyF for code

packages/demo-gatsby/gatsby-browser.js
DirtyF: I wasn't sure 😊, thanks for the review.

dwalkr:
should this be `chore` and not `fix`?

PR #905, opened 2020-03-17
Feat: Adds inline image field

kendallstrautman:
To test, start up `demo-next` and head to the info page. Toggle edit mode and click on the image, select a new one from your local computer (or drag it over) and you should see it commit / update.

- Got rid of `previewSrc` since there's no sidebar where you would view the preview.
- ~~Difficult to test in `demo-next` because husky doesn't allow the media commits.~~ configured the git router
- [x] Still need to add a focus ring


packages/react-tinacms-inline/src/inline-field-image.tsx
```diff
@@ -0,0 +1,106 @@
...
+
+import * as React from 'react'
+import { InlineField } from './inline-field'
+import { useCMS } from 'tinacms'
+import { useDropzone } from 'react-dropzone'
+import { InputFocusWrapper } from './styles'
+
+interface InlineImageProps {
+  name: string
+  path?: string
```
ncphillips: Unused prop
```suggestion 
```

packages/react-tinacms-inline/src/inline-field-image.tsx
```diff
@@ -0,0 +1,106 @@
...
+  parse(filename: string): string
+  uploadDir(form: any): string
+  children?: any
+}
```
ncphillips: removes `path` and set the type for `form`

```suggestion
import { useCMS, Form } from 'tinacms'
import { useDropzone } from 'react-dropzone'
import { InputFocusWrapper } from './styles'

interface InlineImageProps {
  name: string
  parse(filename: string): string
  uploadDir(form: Form): string
  children?: any
}
```

PR #908, opened 2020-03-18
Inline Blocks Fixes

spbyrne:
Work in progress.

Done:
- Change settings button to an icon button
- Replace delete block icon with trash icon
- Style inline text input
- Handle block focus with `InlineBlocksContext`
- Lose block focus when clicking outside a block
- Add block ddl functionality
- Fix add block UI
- Improve block control grouping/arrangement


kendallstrautman:
Just gonna paste some photos here of inline blocks controls things I see while testing. I don't think we necessarily have to fix them all before merging that PR, but the ones we don't we'll make issues for.

Screen Shot 2020-03-19 at 5 15 08 PM
For some reason this extra height for the block menu items is happening and I couldn't suss out where this was coming from

Screen Shot 2020-03-19 at 5 13 50 PM
the right: -1rem; value for the blocks controls makes it bleed off the page when a block is 100% width

kendallstrautman:
Also copied from Slack: 

We need to play around with coordinating click handlers. Right now there's some odd behavior when trying to add blocks. sometimes it like glitches out and when you click `AddBlock` it thinks your trying to 'unfocus' the block or reactivate another one. Or if you click a block to add it will just refocus on another block instead of adding the new one. 

Again I don't necessarily think we need to solve this right now, let's see if it really is a problem when we use it cause the blocks in the next demo don't have any padding etc, which I think will help mitigate this.

spbyrne:
Screen Shot 2020-03-20 at 10 41 34 AM

I'm not opposed to setting the block controls to `right: 0;`. In fact the rounded focus ring makes this work quite nicely.

spbyrne:
Yeah I've been playing around trying to iron out all the click issues. Basically clicking inside the focus ring & blocks controls are the 'top level' handlers that should prevent further events from firing, otherwise the 'click outside' handler is fired.

spbyrne:
It's the 'click outside' handler that's causing all the issues, essentially. That event is bound on a per-block basis. If we could handle that at the parent level it might fix things, but the parent doesn't have access to child refs.

spbyrne:
Screen Shot 2020-03-20 at 10 56 35 AM

Here's the grouping I have set up right now. I moved the settings cog so that the delete button is always on the far right. Moving the add block button balances it out nicely.

spbyrne:
Screen Shot 2020-03-20 at 11 36 50 AM

Improved add block menu sizing & spacing. Merged master so we're using CSS custom properties now. Next I'll tackle the click event issues.

kendallstrautman:
Don't merge yet, there's still an issue were looking into

spbyrne:
I'm stripping out focus loss stuff right now, we can merge without it.

spbyrne:
Screen Shot 2020-03-20 at 5 12 01 PM

I pulled the universal clear code and added a clear button. Fixed for now.

kendallstrautman:
I saw a new issue related to `z-index` where the add block menu was going underneath the blocks controls of lower block items. Visibly you didn't see this, but it was preventing you from being able to click the button and add the new block. I saw there was a lot of z-index stuff going on so I commented out a few `z-index` values and focused on fixing the blocks controls. 

I set it up so that the `z-index` is set for each block depending on the index value. So the first block is 1000, the second block is 999 and so on. This will make it so the blocks controls and the children (add menu etc.) will always be on 'top' of the subsequent blocks. Not sure if this is the best fix but I think its a fine approach for now. 

I left the `z-index` values just commented out instead of removing them in case we see some issues later and can easily reinstate. But I think setting `z-index` on any children of the blocks controls won't do anything, since the value of blocks controls will take precedent.

PR #911, opened 2020-03-18
@tinacms/react-modals

ncphillips:
Extracts the modal components to a new package


packages/@tinacms/react-modals/src/ModalProvider.tsx
```diff
@@ -21,17 +21,18 @@ import { useCallback, useState, ReactNode } from 'react'
...
-import { Z_INDEX } from '../../Globals'
 import {
   TinaReset,
   Button,
   padding,
   font,
   color,
   // eslint-disable-next-line @typescript-eslint/no-unused-vars
-  ThemeProps,
+  // ThemeProps,
```
dwalkr: should this be removed?

ncphillips: yeah 👍

dwalkr:
```suggestion
```

PR #913, opened 2020-03-18
Add missing Field type attributes for Image fields

marklawlor:
Values taken from packages/tinacms/src/plugins/fields/ImageFieldPlugin.tsx

### DIFF
--- a/packages/@tinacms/forms/src/form.ts
+++ b/packages/@tinacms/forms/src/form.ts
@@ -47,6 +47,9 @@ export interface Field {
     field: Field
   ): string | object | undefined
   defaultValue?: any
+  previewSrc?: (form: any, field: Field) => string
+  uploadDir?: (form: any) => string
+  clearable? : boolean  
   fields?: Field[]
 }
 


ncphillips:
Hey Mark! Thanks for the PR 🙌 

We actually intentionally left those properties off of `Field`. It's purpose is to list only the common properties across all field plugins. Subclasses of `Field` are free to accept additional properties, however those properties won't be listed on `Field` itself. 

That said, I'm definitely interested in improving the API. Is there a particular problem you were hoping to solve with this PR?

marklawlor:
This code throws and error in typescript

```
Type '{ name: string; label: string; component: string; previewSrc: (formValues: any, { input }: any) => any; uploadDir: () => string; clearable: boolean; }' is not assignable to type 'Field'.
  Object literal may only specify known properties, and 'previewSrc' does not exist in type 'Field'.
```

```
  const [data] = useLocalJsonForm(jsonFile, {
    fields: [
      {
        name: "myImage",
        label: "My Image",
        component: "image",

        previewSrc: (formValues, { input }) => {
          return input.value;
        },
        uploadDir: () => {
          return "public";
        },
        clearable: false
      }
    ]
  });
```

ncphillips:
Riiight, that's an awkward typing issue. 

For example, one weird thing is that if you don't declare the image field directly in line then it works:

```tsx
const imageField = {
  name: 'image',
  component: 'image',
  previewSrc: (formValuesany: any, { input }: any) => {
    return input.value
  },
  uploadDir: () => {
    return 'public'
  },
  clearable: false,
}

new Form({
  id: '',
  label: '',
  onSubmit() {},
  fields: [{ name: 'title', component: 'text' }, imageField],
})
```

Anyway, this might work better:
```tsx
export interface Field {
  [key: string]: any
  name: string
  // ...
```

It allows extra properties to be used in `Field`

marklawlor:
I think the best solution is to be specific with the typings eg
```
export type Field = TextField | ImageField | ComponentField

export interface TextField {
  component: "text" 

``

**Comment by marklawlor** [comment]
I would always advocate for stricter typings if possible

For example:

```
export type Field = TextField | ImageField | ComponentField

export interface AbstractField {
    name: string
    label?: string
    description?: string
    component: React.FC<any> | string | null
    parse?: (value: any, name: string, field: Field) => any
    format?: (value: any, name: string, field: Field) => any
    validate?(
      value: any,
      allValues: any,
      meta: any,
      field: Field
    ): string | object | undefined
    defaultValue?: any
    fields?: Field[]
}

export interface TextField extends AbstractField {
  component: "text" 
}

export interface ImageField extends AbstractField {
  component: "image"
  previewSrc: (form: any, field: Field) => string
  uploadDir?: (form: any) => string
  clearable? : boolean  
}

export interface ComponentField extends AbstractField {
  component: React.Element
  [key: string]: any
}
```

This way you can get per component type errors

I'm happy to update the PR if you can point me in the direction of all the field types and their required fields

**Comment by ncphillips** [comment]
I generally would too, it's a bit of an ergonomics issue. The problem is that new field plugins can be added. If someone makes a `MapFieldPlugin` then they shouldn't have to modify `@tinacms/forms` in order to make the types work. Something like this could maybe work:

```ts
type AnyField = Field & { [key:string]: any }

export class Form<V = any, F extends Field = AnyField> implements Plugin {
  id: any
  label: string
  fields: F[]
  initialValues: V
```

The defaulting to `AnyField` means it could work as I described above:

```ts
export const imageField = new Form({
  fields: [
    { name: 'title', component: 'text' },
    {
      name: 'image',
      component: 'image',
      previewSrc: (formValuesany: any, { input }: any) => {
        return input.value
      },
      uploadDir: () => {
        return 'public'
      },
      clearable: false,
    },
  ],
  // ....
})
```

But if you wanted to add even more type-safety you could pass a second generic in:

```ts
interface ImageField extends Field {
  previewSrc?: (values: any, props: any) => string
  uploadDir?: () => string
  clearable?: boolean
}

interface CakeField extends Field {
  cake?: boolean
}

type AvailableFields = Field | ImageField | CakeField

export const imageField = new Form<any, AvailableFields>({
  fields: [
    { name: 'title', component: 'text' },
    {
      name: 'image',
      component: 'image',
      previewSrc: (formValuesany: any, { input }: any) => {
        return input.value
      },
      uploadDir: () => {
        return 'public'
      },
      clearable: false,
    },
  ],
  // ....
})
```

ncphillips:
Closing in favour of #1004 

Thanks a lot for bringing this up @marklawlor 

@all-contributors please add @marklawlor for ideas

PR #914, opened 2020-03-18
Inline Blocks Fixes: AddBlockMenu renders block options dropdown

kendallstrautman:
## What this does: 

- Merges master
- Basically re-implemented what was setup in the original `react-tinacms-blocks` 
`AddBlockMenu`. 
- The inline text inputs weren't inheriting the `font-family` or `font-weight`. We may need to add a few more font properties to cover our bases. Without this fix, when I toggled edit mode the type went from serif to sans-serif and changed weights.
- Set a `max-width` and centered the blocks in the demo cause it was difficult to see the controls.

## Notes:

There's still some focus issues, where the block options go 'underneath' the input. But I figured we could get this in and then tweak UI styles as needed! 

Feel free to test in `demo-next` on the blocks page.


spbyrne:
Awesome!

PR #915, opened 2020-03-19
feat: image drag drop support in wysiwym

jpuri:
Implementing drag-drop for images.

For feature to work i demo pages we need to pass image upload fn as prop.


packages/react-tinacms-editor/src/state/plugins/Menu/Menu.tsx
```diff
@@ -184,7 +176,7 @@ export const Menu = (props: Props) => {
               <ItalicControl view={view} />
               <UnderlineControl view={view} />
               <LinkControl view={view} />
-              <ImageMenu editorView={editorView} imageUpload={imageUpload} />
+              {/* <ImageMenu editorView={editorView} imageUpload={imageUpload} /> */}
```
jpuri: As discussed over slack we can create a modal in editor for image upload and then enable this menu option.

packages/react-tinacms-editor/stories/wysiwyg.stories.tsx
```diff
@@ -67,16 +67,19 @@ const WithImage = () => {
...
+          console.log(files)
+          return [
+            new Promise(resolve => {
+              setTimeout(function() {
+                resolve('https://i.imgur.com/2FCfbgg.jpg')
+              }, 250)
+            }),
+          ]
+        },
+        previewUrl: (str: string) => str,
```
jpuri: Storybook here demos way to use `ImageProps`.

packages/react-tinacms-editor/src/state/plugins/Image/index.ts
```diff
@@ -18,55 +18,86 @@ limitations under the License.
...
+          const imagePromises = uploadImages(files)
+          imagePromises.forEach(promise => {
+            promise.then(url => {
+              const { state, dispatch } = editorView
+              insertImage(state, dispatch, url)
+            })
+          })
+        }
+        return false
+      },
```
jpuri: `handleDrop` implemented to upload images.

packages/react-tinacms-editor/src/types.ts
```diff
@@ -0,0 +1,22 @@
...
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+
+*/
+
+export interface ImageProps {
+  upload?: () => Promise<string>[]
```
ncphillips: I think this will end up being:

```
  upload?: () => Promise<string[]>
```

jpuri: Can you plz confirm, I can refactor code accordingly.

ncphillips: [Yup](https://github.com/tinacms/tinacms/blob/master/packages/@tinacms/media/src/media.ts#L59).

MediaStore returns array of Media objects. These will be passed as strings:

```js
upload: async (files) => {
  const media = await cms.media.store.persist(files)
  return media.map(toSrc)
}
```

jpuri: Hey @ncphillips , PR is updated.

packages/react-tinacms-editor/src/state/plugins/Image/index.ts
jpuri: I extended it to handle paste also.

DirtyF:
I had to manually add `static/images` but now it works.

PR #919, opened 2020-03-20
Roadmap: Propose design collaboration feature

blaggacao:
Most probably this might be out of scope, for now. In such case, please just bluntly reject.

Else:

- I'm the developer and product owner
- I'm working with a copy writer for texting (tinaCMS yeay!)
- I'm working with a designer for the graphics, layout and overall design

While tinaCMS vastly enables collaboration with the copy writer, for collaboration with the designer, I'd need to look out for other means.

The proposal consists therefore of:
- Leveraging tina's existing overlay UX componentes to provide component selection and attach a "comment box" to each page component
- Commiting those conversations with all necesary and useful metadata within a tinacms-collaboration.json at the project's root.

### DIFF
--- a/ROADMAP.md
+++ b/ROADMAP.md
@@ -21,6 +21,7 @@ Two high-level sets of issues have been identified:
 - Collaboration
   - Identifying commit authors
   - Locking forms
+  - On Design: overlay discussion on page components
 - Media management
 - Better media support in the markdown wysiwyg
 - Improving the inline editing experience


blaggacao:
Actually, such "on component commenting" also enables better colab with the copy writer, so the "On Design:" prefix might be unwarranted, it was the first use case popping to my mind, though.

ncphillips:
Hey, that's a really cool idea! But yeah, it's a complex issue and it's pretty out of scope for Tina right now. There's definitely a possibility that you could prototype something like this yourself though! If you decide to try that I would be glad to help out however possible, but it's not something that we're going to consider as part of our roadmap in the near future. 

@all-contributors please add @blaggacao for ideas

PR #934, opened 2020-03-24
Refactor tina fields

jeffsee55:
WIP - move stuff


packages/@tinacms/fields/src/components/TextField.ts
```diff
@@ -29,6 +29,6 @@ export interface TextFieldProps extends a {
   ref?: any
 }
 
-export const TextField = styled.input<TextFieldProps>`
+export const BaseTextField = styled.input<TextFieldProps>`
```
ncphillips: I don't think this should have been changed

jeffsee55: The `TextFieldPlugin` module exports `TextField` as well. This is just a WIP but there are a few of these types of collisions that I need to think about

ncphillips: oof 🤔

packages/@tinacms/fields/src/plugins/BlocksFieldPlugin.tsx
```diff
@@ -570,7 +570,7 @@ const Panel = function Panel({
   )
 }
 
-export default {
+export const BlocksFieldPlugin = {
```
jeffsee55: 👍

PR #943, opened 2020-03-27
Add description to block, group & group list

### LINKED ISSUE #910: Group Fields don't exhibit description
## Description

Group Fields doen't exhibit their description as defined by form config.

#### Tinaform config:
```jsx
const TinaCmsFormOptions = {
  label: "No Group Description",
  fields: [
    {
      label: "Foo",
      name: "rawJson.foo",
      description: "Foo description not shown",
      component: "group",
      fields: [
        {
          label: "Bar",
          name: "bar",
          description: "Bar description shown",
          component: "text",
        },
      ],
    },
  ],
}
```
#### data.json
```json
{
  "foo": {
    "bar": "hey there"
  }
}
```

Render a form with the above form config (afaik, he form config can be placed on any form. Thus no need to match the underlying datastructure to reproduce).

### Expected result

When selecting the group, just below the title, the description should be shown

### Actual result

Group description is shown nowhere, not even as tooltip (I checked)

#### Tinacms Package Versions

```text
├─┬ gatsby-plugin-tinacms@0.1.28
│ └─┬ tinacms@0.17.0
│   ├── @tinacms/core@0.7.3
│   ├─┬ @tinacms/fields@0.8.0
│   ├─┬ @tinacms/form-builder@0.3.0
│   ├─┬ @tinacms/forms@0.3.0
│   ├── @tinacms/icons@0.7.0
│   ├─┬ @tinacms/react-core@0.2.9
│   ├── @tinacms/styles@0.3.0
│   └─┬ react-tinacms-editor@0.1.0
├─┬ gatsby-tinacms-git@0.4.1
│ ├─┬ @tinacms/api-git@0.7.0
│ └── @tinacms/git-client@0.6.1
├─┬ gatsby-tinacms-json@0.8.12

```

#### Tools

n/a

spbyrne:
Blocks, groups, and group lists weren't being wrapped with `wrapFieldsWithMeta`. For the group field I simply added this wrapper, but blocks and group lists handle their own labelling so I've added the ability for them to accept a description for now. 

Now all three of these fields will accept a description, and some styling inconsistencies have been fixed.

Closes #910

### DIFF
--- a/packages/tinacms/src/plugins/fields/BlocksFieldPlugin.tsx
+++ b/packages/tinacms/src/plugins/fields/BlocksFieldPlugin.tsx
@@ -32,6 +32,12 @@ import { GroupPanel, PanelHeader, PanelBody } from './GroupFieldPlugin'
 import { Dismissible } from 'react-dismissible'
 import { IconButton } from '@tinacms/styles'
 import { useFormPortal } from '@tinacms/react-forms'
+import { FieldDescription } from './wrapFieldWithMeta'
+import {
+  GroupListHeader,
+  GroupListMeta,
+  GroupLabel,
+} from './GroupListFieldPlugin'
 
 export interface BlocksFieldDefinititon extends Field {
   component: 'blocks'
@@ -78,7 +84,7 @@ interface BlockFieldProps {
   tinaForm: Form
 }
 
-const Blocks = function({ tinaForm, form, field, input }: BlockFieldProps) {
+const Blocks = ({ tinaForm, form, field, input }: BlockFieldProps) => {
   const addItem = React.useCallback(
     (name: string, template: BlockTemplate) => {
       let obj: any = {}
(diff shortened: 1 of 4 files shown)


packages/tinacms/src/plugins/fields/BlocksFieldPlugin.tsx
```diff
@@ -99,7 +105,15 @@ const Blocks = function({ tinaForm, form, field, input }: BlockFieldProps) {
...
-        <GroupLabel>{field.label || field.name}</GroupLabel>
+        <GroupListMeta>
+          <GroupLabel>{field.label || field.name}</GroupLabel>
+          {field.description && (
+            <FieldDescription>{field.description}</FieldDescription>
+          )}
+        </GroupListMeta>
+        {field.description && (
+          <FieldDescription>{field.description}</FieldDescription>
+        )}
```
ncphillips: I'm not really sure what this would look like with the description being rendered twice.

Could you post a screenshot?

ncphillips: Actually, since `GroupList` is different I'm guessing that second one shouldn't be there?

spbyrne: Fixed.

PR #945, opened 2020-03-30
style: set transition property to width

### LINKED ISSUE #944: CSS Transition: all effects every element on user's site/app
###  Description

This should be an easy one to improve upon.

Because `transition: all` is applied to the `SiteWrapper` every element inside also inherits this property causing every property on every element, to transition. These include things like font-size. In my scenario, in some cases, it is interfering with my site and causing font sizes to 'jitter' as CSS loads and font-size properties are animated.

### Steps to reproduce

Change the value of any property on any element inside `SiteWrapper`. For example either be using the inspector tool or by toggling a HTML class.

### Expected result

Tina should not interfere with user's styles.

### Actual result

All properties of user's app (inside `SiteWrapper` ) are animated.

This causes some issues/interference even when Tina is not in use as the styles are still applied to the wrapper. Please consider limiting the properties which need to be transitioned like `width` and only applying them when in use for the Tina sidebar.

spbyrne:
This sets the `SiteWrapper` component to only animate `width` instead of `all`. 

Closes #944

### DIFF
--- a/packages/tinacms/src/components/TinaProvider.tsx
+++ b/packages/tinacms/src/components/TinaProvider.tsx
@@ -92,7 +92,7 @@ const SiteWrapper = styled.div<{ open: boolean; position: SidebarPosition }>`
     isFixed(props.position) && props.open
       ? 'calc(100% - ' + SIDEBAR_WIDTH + 'px)'
       : '100%'} !important;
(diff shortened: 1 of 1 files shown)


kendallstrautman:
Thoughts on using `translateX` here? For perf reasons.

PR #948, opened 2020-03-30
fix: duplicate exports

ncphillips:
This was causing an error:

```
There was an error compiling the html.js component for the development server.
See our docs page on debugging HTML builds for help https://gatsby.dev/debug-html TypeError: Cannot redefine property: Input
> 1 | !function(n,t){"object"==typeof exports&&"undefined"!=typeof module?t(exports,require("@ti
```


kendallstrautman:
NVM

Issue #927, opened 2020-03-23
Feat (react-tinacms-inline): add dismissable to 'clear' block focus


kendallstrautman:
## Summary

Currently if you want to 'clear' focus on a block, you can hit the far right 'X' icon. It would be ideal if clicking anywhere outside of the block cleared focus as well.

<img width="822" alt="Screen Shot 2020-03-23 at 3 17 25 PM" src="https://user-images.githubusercontent.com/36613477/77364081-77382680-6d19-11ea-9973-536194d8c8cc.png">

We could use dismissable for this, as we do with the color picker or date fields. Or come up with our own solution.

### Motivation

Clicking outside of an area to reset focus is intuitive UI behavior. This behavior will also align better with our sidebar UI, where I believe the regular sidebar fields already do this. Also, the 'X' icon could be confused with a delete action.

DirtyF:
I got to admit I had no idea what this close icon did before you explained it. 😲

kendallstrautman:
Closed via #952

PR #952, opened 2020-03-31
chore: improve block focus handling

spbyrne:
This removes the 'clear focus' button from the block menu and instead handles clearing focus by clicking outside the block.


kendallstrautman:
Closes #927

PR #958, opened 2020-04-01
Adding the Github Open Auth package

jhuggett:


packages/react-tinacms-github/package.json
```diff
@@ -0,0 +1,17 @@
+{
+  "name": "react-tinacms-github",
+  "version": "1.0.0",
```
ncphillips, line 3: This will get bumped automatically when the package is released
```suggestion
  "version": "0.0.0",
```

PR #960, opened 2020-04-01
Adding the Github Open Auth package

jhuggett:
This adds Helpers for managing "Open Authoring" on a site using Github. 
This functionality is being pulled over from [tinacms.org](tinacms.org). 
https://github.com/tinacms/tinacms.org/pull/318

The GithubClient uses a proxy for the backend, which can be used to[ attach the GitHub access token. ](https://github.com/tinacms/tinacms.org/blob/master/github/api/req/proxy.ts)

For implementation details, see: https://github.com/tinacms/rfcs/pull/6

- [x] April 6th Changes Released


packages/react-tinacms-github/package.json
```diff
@@ -0,0 +1,17 @@
+{
+  "name": "react-tinacms-github",
+  "version": "1.0.0",
```
jamespohalloran, line 3:
```suggestion
  "version": "0.0.0",
```
as per:
https://github.com/tinacms/tinacms/pull/958#discussion_r401661039

packages/react-tinacms-github/src/toolbar-plugins/ForkNamePlugin.tsx
```diff
@@ -0,0 +1,49 @@
...
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+
+*/
+
+import { FieldMeta } from 'tinacms'
+import styled from 'styled-components'
+import { color } from '@tinacms/styles'
```
kendallstrautman, line 21:
```suggestion
```

packages/react-tinacms-github/src/toolbar-plugins/ForkNamePlugin.tsx
```diff
@@ -0,0 +1,49 @@
...
+  component: ForkInfo,
+})
+
+const MetaLink = styled.a`
+  display: block;
+  max-width: 250px;
+  overflow: hidden;
+  text-overflow: ellipsis;
+  font-size: 16px;
+  color: ${color.primary('dark')};
```
kendallstrautman, line 38:
```suggestion
  color: var(--tina-color-primary-dark);
```

kendallstrautman: That _should_ fix it

jhuggett: That solved it (at least it's no longer complaining, I can't fully test it yet), thanks

packages/react-tinacms-github/.gitignore
```diff
@@ -0,0 +1,9 @@
+*.log
+.DS_Store
+node_modules
+.cache
+.rts2_cache_cjs
+.rts2_cache_esm
+.rts2_cache_umd
+.rts2_cache_system
+dist
```
ncphillips, line 9: the root `.gitignore` has all these, so this `.gitignore` can be removed

packages/react-tinacms-github/src/open-authoring/OpenAuthoringProvider.tsx
```diff
@@ -0,0 +1,118 @@
...
+  exitEditMode: () => void
+  error?: any
+}
+
+interface AuthState {
+  authenticated: true
+  forkValid: true
+}
+
+export const OpenAuthoringProvider = ({
```
ncphillips, line 40: The name `OpenAuthoringProvider` seems both too broad and too specific to me. 

Too broad because it suggests it's a general solution to Open Authoring,  which it is not too.

Too specific because if this supports private repos (i.e. no forking, only authorized if you have access to the repo) in the future then it wouldn't be just for open authoring. 

What do you think of renaming to something like `GithubAuthenticationProvider`?

packages/react-tinacms-github/src/toolbar-plugins/useOpenAuthoringToolbarPlugins.tsx
```diff
@@ -0,0 +1,51 @@
...
+*/
+
+import { Form, useCMS } from 'tinacms'
+import { useEffect } from 'react'
+import { getForkName } from '../open-authoring/repository'
+import { PRPlugin } from './pull-request'
+import { ForkNamePlugin } from './ForkNamePlugin'
+import { Plugin } from 'tinacms'
+
+export const useOpenAuthoringToolbarPlugins = (
```
ncphillips, line 26: Maybe rename to something like `useGithubToolbarPlugins` or `useTinacmsGithubPlugins`

packages/react-tinacms-github/src/open-authoring/OpenAuthoringProvider.tsx
ncphillips: Since it also handles entering/exiting editing mode maybe `TinacmsGithubProvider`?

packages/react-tinacms-github/src/form/useGithubJsonForm.ts
```diff
@@ -0,0 +1,48 @@
...
+  formOptions: Options,
+  githubOptions: GithubOptions
+) {
+  const [values, form] = useGithubFileForm(
+    jsonFile,
+    formOptions,
+    githubOptions,
+    serialize
+  )
+  usePlugins(form as any)
```
dwalkr, line 46: I think this call to `usePlugins` is redundant, since `useGithubFileForm` calls `useLocalForm` which already adds the plugin.

so this could be simplified to `return useGithubFileForm(...)`

packages/react-tinacms-github/src/form/useGithubMarkdownForm.ts
```diff
@@ -0,0 +1,44 @@
...
+  formOptions: Options,
+  githubOptions: GithubOptions
+) {
+  const [values, form] = useGithubFileForm(
+    markdownFile,
+    formOptions,
+    githubOptions,
+    toMarkdownString
+  )
+  usePlugins(form as any)
```
dwalkr, line 42: same as `useGithubJsonForm`, this line can be removed

packages/react-tinacms-github/package.json
```diff
@@ -0,0 +1,42 @@
...
+    "dev": "tsdx build",
+    "test": "tsdx test --env=jsdom --passWithNoTests",
+    "lint": "tsdx lint"
+  },
+  "author": "",
+  "license": "Apache-2.0",
+  "peerDependencies": {
+    "react": ">=16.8",
+    "@tinacms/styles": "^0.4.0",
+    "@tinacms/fields": "^0.9.1",
```
ncphillips, line 18:
```suggestion
    "@tinacms/styles": ">=0.4",
    "@tinacms/fields": ">=0.9",
```

packages/react-tinacms-github/package.json
```diff
@@ -0,0 +1,42 @@
...
+    "styled-components": ">=4.1",
+    "tinacms": ">=0.13",
+    "js-cookie": "^2.2.1",
+    "final-form-submit-listener": "^1.0.0",
+    "@tinacms/icons": ">=0.7.0",
+    "next-tinacms-markdown": ">=0.2.9",
+    "btoa": ">=1.2.0",
```
ncphillips, line 25: I think a few of these could be `dependencies` instead of `peerDependencies`:

* `js-cookie`
* `final-form-submit-listener`
* `btoa` 

Maybe even `next-tinacms-markdown`

packages/react-tinacms-github/README.md
```diff
@@ -0,0 +1,73 @@
...
+      {...children}
+    </TinacmsGithubProvider>)
+}
+```
+
+Add error handling to our forms which prompt Github-specific action when errors occur (e.g a fork no longer exists).
+```ts
+// YourSiteForm.ts
+const YourSiteForm = ({ form, children }) => {
+  useGithubErrorListener(form)
```
ncphillips, line 33: I don't think this needs to be addressed now, but I'm pretty sure we could solve this without the user having to add this to their site. 

We might be able to add something like this to the `TinacmsGithubProvider`:

```
useSubscribable(cms.forms)
let activeForm = cms.forms.all()[0] // Right now assume there's only one form active on the page.
useGithubErrorListener(activeForm)
```

PR #961, opened 2020-04-01
Tinacms react screens

ncphillips:
- [x] Move `ScreenPlugin` interface and creator
- [x] Move `useScreenPlugin` react hook
- [x] Move React components for rendering screens (listing will still be integrated with the Sidebar


packages/@tinacms/react-screens/src/screen-plugin.tsx
```diff
@@ -0,0 +1,78 @@
...
+ * @param options
+ */
+export function createScreen<ExtraProps>({
+  Component,
+  props,
+  ...options
+}: ScreenOptions<ExtraProps>): ScreenPlugin<ExtraProps> {
+  return {
+    __type: 'screen',
+    ...options,
```
jamespohalloran, line 71:
```suggestion

```
options is spread twice

PR #962, opened 2020-04-01
fix: not possible to come out of code mark which is at end of line

### LINKED ISSUE #922: Can't move right of inline code at the end of the line
## Description

It's possible to get stuck inside code-ticks at the end of the line.

### Steps to reproduce

Go to the end of a document and write the following:
```markdown
`test`
```
It will get turned into inline code.

Press the left-arrow and then the right-arrow twice

Type any character and it will be inside the inline code.

### Expected result

Pressing right a second time should push you out of the inline code. An analogous behavior exists in Slack

jpuri:
If user is in code mark at end of line a arrow right will bring him out of it.

Fix #922

### DIFF
--- a/packages/react-tinacms-editor/src/state/plugins/Common/index.tsx
+++ b/packages/react-tinacms-editor/src/state/plugins/Common/index.tsx
@@ -17,9 +17,29 @@ limitations under the License.
 */
 
 import { Plugin, PluginKey, Transaction } from 'prosemirror-state'
+import { EditorView } from 'prosemirror-view'
 
 interface CommonPluginState {
   editorFocused: boolean
+}
+
+const exitCodeMarkOnArrowRight = (view: EditorView, event: KeyboardEvent) => {
+  // If user is at end of current line and there is a code mark an arrow-right will
+  // take user out of the code mark.
+  if (event.key !== 'ArrowRight') return false
+  const { selection, schema } = view.state
+  const { code } = schema.marks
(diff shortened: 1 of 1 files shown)


packages/react-tinacms-editor/src/state/plugins/Common/index.tsx
```diff
@@ -62,5 +63,26 @@ export const commonPlugin = new Plugin({
...
+      const selectionIsAtEnd =
+        selection.$to.node().nodeSize - 2 === selection.$to.parentOffset
+      if (!selectionIsAtEnd) return false
+      const { state, dispatch } = view
+      dispatch(
+        state.tr
+          .insertText(' ', selection.$to.pos)
+          .removeMark(selection.$to.pos, selection.$to.pos + 1, code)
+      )
+      return true
```
ncphillips, line 85: Awesome! Could you just extract this to it's own function? Maybe something like `exitCodeMarkOnArrowRight`

```ts
handleKeyDown(view, event) {
  return exitCodeMarkOnArrowRight(view, event)
}
```

DirtyF:
When you delete a line starting with `inline code` and type again on the same line
Then text is still formatted as `inline code`
When you hit return and backspace
Then text is no more formated as inline code

https://www.loom.com/share/0b47424528e844b2ae6aee06e3f7634a

jpuri:
Hey @ncphillips : PR is updated.

PR #964, opened 2020-04-01
style: add z-index custom properties

spbyrne:
This removes a lot of the aggressive z-index values and instead uses a set of CSS custom properties to set consistent z-index values for components throughout Tina. 

The sidebar will no longer have an incredibly high z-index. Instead, if you find your site has elements that render on top of the sidebar, you can adjust the z-index values used by Tina to compensate. 

Here are the values used: 

```
    --tina-z-index-0: 500;
    --tina-z-index-1: 1000;
    --tina-z-index-2: 1500;
    --tina-z-index-3: 2000;
    --tina-z-index-4: 2500;
```

Some elements will continue to use values like `-1` for local layering.


ncphillips:
Want to write up some docs for this?

spbyrne:
> Want to write up some docs for this?

https://github.com/tinacms/tinacms.org/pull/314

I've added the properties, but no explanation yet. Right now there's no rigid system for how they're used.

PR #974, opened 2020-04-02
Feat: Adds ToolbarState & ToolbarProvider

kendallstrautman:
## What this does: 

- Adds `ToolbarState`
- Creates `ToolbarProvider`, adds it to `TinaProvider`
- Instantiates `ToolbarState` class on `tinacms` 

The toolbar's default state is hidden. The toolbar will only mount if the `hidden` property is specified in the config: 

```js
    this.cms = new TinaCMS({
      sidebar: {
        position: 'overlay',
        hidden: true,
      },
      toolbar: {
        hidden: false,
      },
    })
```
Let me know if this is in the ballpark of what you had in mind @ncphillips — I know this isn't a long-term implementation but felt like these are good first steps to let folks opt-in to using the toolbar w/o breaking changes.

### DIFF
--- a/packages/@tinacms/react-toolbar/package.json
+++ b/packages/@tinacms/react-toolbar/package.json
@@ -40,7 +40,6 @@
     "@tinacms/react-core": "^0.2.9",
     "@tinacms/react-forms": "^0.1.0",
     "@tinacms/react-modals": "^0.1.0",
-    "@tinacms/react-screens": "^0.0.0",
     "@tinacms/styles": "^0.2.9",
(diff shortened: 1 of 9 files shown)


kendallstrautman:
For some reason in `TinaProvider`, I'm getting this on the `ToolbarProvider`:

```
 Property 'toolbar' does not exist on type 'IntrinsicAttributes & ToolbarProviderProps'.
```
But `toolbar` is def on the `ToolbarProviderProps` 🤷‍♀️. It's odd cause everything seems to work fine. Would love a second eye on that.

PR #975, opened 2020-04-06
Next tinacms GitHub

jamespohalloran:
This package provides helpers for managing the github auth token on the server.
When used with Next.js in the `pages/api` directory, these functions are mapped to `/api/*` endpoints.

Moved from [tinacms.org](https://github.com/tinacms/tinacms.org/compare/next-tinacms-github?expand=1)

PR #977, opened 2020-04-06
create github-tinacms-content package for fetching github content

jamespohalloran:
This package provides helpers for fetching content from github.

Moved from [tinacms.org](https://github.com/tinacms/tinacms.org/pull/323)


packages/github-tinacms-content/package.json
```diff
@@ -0,0 +1,25 @@
...
+  "peerDependencies": {
+    "atob": ">=2.1.0",
+    "gray-matter": ">=4.0.0",
+    "axios": ">=0.19.0"
+  },
+  "devDependencies": {
+    "atob": "2.1.2",
+    "gray-matter": "4.0.2",
+    "axios": "0.19.2"
+  }
```
ncphillips, line 24: I would say those can just be dependencies:
```suggestion
  "dependencies": {
    "atob": "2.1.2",
    "gray-matter": "4.0.2",
    "axios": "0.19.2"
  }
```

packages/github-tinacms-content/src/sourceProviderConnection.ts
```diff
@@ -0,0 +1,34 @@
...
+  }
+
+  return {
+    accessToken: previewData.github_access_token,
+    sourceProviderConnection: {
+      forkFullName: previewData.fork_full_name,
+      headBranch: previewData.head_branch || 'master',
+    },
+  }
+}
```
ncphillips, line 52: This feels really Next.js specific. Maybe belongs in `next-tinacms-github`, or maybe it needs to be renamed?

jamespohalloran: true, or maybe even just within the site for now.

jamespohalloran: Moved it back into tinacms.org

ncphillips:
I'm not sure I like the package name. 

This seems like an analogous package to `@tinacms/api-git`. A more appropriate name might be `@tinacms/api-github`.

ncphillips:
@dwalkr said 

> it's a little weird to me that to code for loading github data and the code for the rest of the github api operations are in different packages

Maybe this code should just be part of `next-tinacms-github` for now? Both of these packages are about Github on the backend and they're always going to be used together for the time being.

jamespohalloran:
> Maybe this code should just be part of `next-tinacms-github` for now? Both of these packages are about Github on the backend and they're always going to be used together for the time being.

The main reasoning was that this package doesn't have any ties to Next, but the `next-tinacms-github` relies on using Next's api functions /previews

ncphillips:
Yeah that's good reasoning. 

There's a balance between two conflicting principles here:

### Common Reuse Principle (CRP)

> Don't force users of a module to depend on other modules they don't need.

Don't force people who want the content fetching code to need to use the Next code. This pushes us to split them up.

### Common Closure Principle (CCP)

> Group modules into packages that change for the same reasons at the same time.

This principle pushes us to group them. Next is is the only supported framework at the moment. . we're talking about changing the way to build a Github-backed CMS, both of these modules are likely to require changes. 

---

I'm personally okay with either approach. 

Given that `@tinacms/api-git` already exists, it seems to me that splitting them would be consistent with how things are. It's also what makes me thinking `@tinacms/api-github` would be the more appropriate name.

jamespohalloran:
Moving this over to the next[ tinacms github package](https://github.com/tinacms/tinacms/pull/975)

PR #978, opened 2020-04-06
fix: image should be deleted only after selection

jpuri:
When using delete / backspace button image should be deleted only after selection.

@DirtyF : is it possible to get this tested on windows box also, is there a browserstack account for forestry ?


DirtyF:
Image selection with arrows or backspace before deleting works fine on macOS, couldn’t test it on Windows. https://www.loom.com/share/83cc38c9bcfe49fcad03ac6dd927b050

PR #996, opened 2020-04-08
Don't allow users to set cookie schema in api handlers

dwalkr:
Cookie schema was user-specified, perhaps to ensure interoperability with `react-tinacms-github` which contains the code that sets these cookies. After discussing with @jamespohalloran,  we determined that this degree of configuration was unnecessary and may even be a detriment down the line. This schema is now a hard-coded concern within the package, which has the benefit of simplifying userland implementation for these API routes

### DIFF
--- a/packages/next-tinacms-github/src/constants.ts
+++ b/packages/next-tinacms-github/src/constants.ts
@@ -1,3 +1,21 @@
+/**
+
+Copyright 2019 Forestry.io Inc
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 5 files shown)


packages/next-tinacms-github/src/constants.ts
```diff
@@ -0,0 +1,3 @@
+export const ACCESS_TOKEN_KEY = 'github_access_token'
+export const FORK_KEY = 'fork_full_name'
+export const HEAD_BRANCH_KEY = 'head_branch'
```
ncphillips, line 3: What if these were defined in `react-tinacms-github` and exported from there?

dwalkr: makes sense to me. it would mean introducing a hard dependency to `react-tinacms-github` but at this point I'm not sure that's really a concern

ncphillips: True. I'm good with magic variables temporarily being the same by "coincidence". 

That sounds much worse after I write it but I stand by that statement

PR #1001, opened 2020-04-10
fix: deleting code block

jpuri:
Fix deleting code block.


jpuri:
> I can delete a code block on backspace if that's what this issue is about.

Yes @DirtyF that is the issue, it does not works on master branch.

jpuri:
PR is good to be merged.

PR #1002, opened 2020-04-13
fix: core editor related code cleanup

jpuri:
Editor code refactoring and cleanup.

Creating context to get latest updates in EditorView.


packages/react-tinacms-editor/src/core/context/useProsemirror.tsx
```diff
@@ -0,0 +1,28 @@
...
+import * as React from 'react'
+import { EditorView } from 'prosemirror-view'
+
+export default (): {
+  editorView: { view: EditorView } | undefined
+  setEditorView: (view: EditorView) => void
+} => {
+  const [view, setEditorView] = React.useState<EditorView>()
+  return { editorView: view ? { view } : undefined, setEditorView }
+}
```
ncphillips, line 28: I like that you're disconnecting building the editor from keeping it in React.

I would suggest `useMemo` over `useState` though. It's essentially a wrapper for `useState` where you just create the object at startup time. This would makes it so `buildEditor` can just return the `view` instead of having to pass it `setEditorView`:

```suggestion
function useProsemirror(
  input: Input,
  el: HTMLDivElement | undefined | null,
  plugins: Plugin[] = [],
  imageProps: ImageProps = {},
  format?: Format
): {
  editorView: { view: EditorView } | undefined
  setEditorView: (view: EditorView) => void
} => {
  return React.useMemo(() => {
    return buildEditor(input, el, plugins, imageProps, format)
  })
}
```

ncphillips: Then again, it was `useState` before. I can't remember if there was a reason for that or not. So just take this comment as a suggestion for something to investigate.

packages/react-tinacms-editor/src/core/context/useProsemirror.tsx
```diff
@@ -0,0 +1,28 @@
+/**
+
+Copyright 2019 Forestry.io Inc
```
DirtyF, line 3: We're in 2020 now :)

DirtyF: Unrelated, but I'm just noticing this copyright isn't very flexible.

ncphillips: Legit, rope @jpatters in on that

packages/react-tinacms-editor/src/core/context/useProsemirror.tsx
jpuri: Hey @ncphillips : `setEditorView` is needed [here](https://github.com/tinacms/tinacms/pull/1002/files#diff-9f71814e207ca3a638f57f2a4c8615e0R64).

I plan to do some more cleanup on this. I plan to move editor state creation part to context provider. So context consumers will always get latest version of `editorView`.

ncphillips: Right, that’s awkward. I hope you find a nice alt

jpuri:
Code changes are complete and look good, I plan to do more testing tomorrow and move it for review.

jpuri:
PR is ready for review.

packages/react-tinacms-editor/src/core/context/editorState.tsx
```diff
@@ -0,0 +1,45 @@
...
+      {children}
+    </EditorStateContext.Provider>
+  )
+}
+
+export const EditorStateConsumer = EditorStateContext.Consumer
+
+export const useEditorStateContext = () => ({
+  ...useContext(EditorStateContext),
+})
```
ncphillips, line 63:
```suggestion
export const useEditorStateContext = () => useContext(EditorStateContext)

```

packages/react-tinacms-editor/src/core/index.tsx
```diff
@@ -44,34 +46,37 @@ export const Wysiwyg = styled(
...
     return (
       <WysiwygWrapper className="wysiwyg-wrapper">
         <link
           rel="stylesheet"
           href="https://codemirror.net/lib/codemirror.css"
         />
-        {editorView && (
+        <EditorStateProvider
+          input={input}
+          el={el}
```
ncphillips, line 64: I think you could remove `el` and just use `editorRef`:

```suggestion
          el={editorRef.current}
```

You would probably need

jpuri: I need to trigger a re-render here so that provider gets value of `el`.

ncphillips: Ahh yes I see. Could you add a comment explaining that above the `useState` for `el`?

jpuri: Sure I will do that

jpuri:
Hey @ncphillips : Let me fix default exports also.

PR #1008, opened 2020-04-15
docs: a few tweaks to the react-tinacms-githubREADME code examples

ncphillips:


packages/react-tinacms-github/README.md
```diff
@@ -39,35 +39,41 @@ const cms = new TinaCMS({
...
   })
+}
+
 const exitEditMode = () => {
-  fetch(`/api/reset-preview`).then(() => {
+  return fetch(`/api/reset-preview`).then(() => {
     window.location.reload()
   })
 }
-const YourLayout = ({ Component, pageProps, children }) => {
```
jamespohalloran, line 55: Isn't it the `pageProps` we want to deconstruct here? `error` is just a prop get on getStaticProps

ncphillips: This is a generic React example and `pageProps` is Next specific so it doesn't need to be there. Maybe you  could do that with Next, but in[ the demo site I setup](https://github.com/ncphillips/tg-demo/blob/master/pages/_app.tsx#L60-L62) I just passed `error` specifically.

jamespohalloran: I'm not sure what you mean, it looks like your demo is also using pageProps.error

ncphillips: That's in the `App`. The [Layout](https://github.com/ncphillips/tg-demo/blob/master/pages/_app.tsx#L20-L36) just accepts `error`

jamespohalloran: Okay, the component I'm describing in this example would be ["Site"](https://github.com/ncphillips/tg-demo/blob/master/pages/_app.tsx#L38), (which in your case isn't really a layout).
IMO if we go down a layer and remove the notion of pageProps, we should add a blurb somewhere (maybe that's in the docs and not the readme).

ncphillips: What would you want add in that blurb?

jamespohalloran: If you're using Next, I feel we are moving a layer of information (describing how to retrieve the error property passed from getStaticProps for an individual page). If you don't think it belongs in this readme, I'll just make a note to add it to our site docs

ncphillips: Ohhh I see 👍 Yeah let's put that in the `next-tinacms-github` docs

PR #1010, opened 2020-04-15
chore: fix client.createFork request

ncphillips:
Was using an env, now uses this.baseRepoFullName.


DirtyF:
How can we test this?

ncphillips:
I don't think this is specifically worth testing. Just a missed find-and-replace from an earlier change. All the requests in that file use `this.baseRepoFullName`, which is set from that environment variable elsewhere

PR #1012, opened 2020-04-15
Deprecate local global keywords

ncphillips:
Implemeting [RFC 0006](https://github.com/tinacms/rfcs/pull/7)

### DIFF
--- a/packages/demo-gatsby/src/components/bio.js
+++ b/packages/demo-gatsby/src/components/bio.js
@@ -26,7 +26,7 @@ limitations under the License.
 import React from "react"
 import { useStaticQuery, graphql } from "gatsby"
 import Image from "gatsby-image"
-import { useLocalJsonForm } from "gatsby-tinacms-json"
+import { useJsonForm } from "gatsby-tinacms-json"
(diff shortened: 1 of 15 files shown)


packages/demo-gatsby/src/components/bio.js
```diff
@@ -78,11 +78,13 @@ const Bio = () => {
     }
   `)
 
-  const [author] = useLocalJsonForm(data.dataJson, {
+  const [author, authorForm] = useJsonForm(data.dataJson, {
     label: "Author",
     fields,
   })
 
+  usePlugin(authorForm)
```
BPScott, line 86: This doesn't seem to be imported from anywhere? Should this be `useFormScreenPlugin`?

ncphillips: You're right on both counts! Thanks 😄

ncphillips: Thanks again, caught a couple other issues in the demos.

@all-contributors please add @BPScott for review

ncphillips:
@all-contributors please add @BPScott for review

PR #1013, opened 2020-04-15
Toolbar bug fixes

jhuggett:
- [x]  On page load, the state of the form isn't reflected properly in the toolbar (e.g. the save button is greyed out even if the form is dirty)

- [x] The discard button is working inconsistently if at all

- [x] The discard and save button **do not** grey out after the discard button is pressed just after page load


packages/@tinacms/react-toolbar/src/components/Toolbar.tsx
```diff
@@ -67,13 +69,26 @@ export const Toolbar = () => {
...
+
+  const reset = () => {
+    if (form) {
+      if (form.reset) {
+        form.reset()
+      } else if (form.finalForm.reset) {
+        form.finalForm.reset()
+      }
+    }
+  }
```
ncphillips, line 81: > // TODO: Form#reset should always exist

[This was actually fixed.](https://github.com/tinacms/tinacms/blob/master/packages/@tinacms/forms/src/form.ts#L126-L131) This can be simplified:
```suggestion

  const reset = () => {
    if (form) {
      form.reset()
    }
  }
```

packages/@tinacms/react-toolbar/src/components/Toolbar.tsx
```diff
@@ -51,12 +51,16 @@ export const Toolbar = () => {
...
-  // TODO: This doesn't return the correct value initially
+
+  const currentState = form?.finalForm.getState()
+
   const formState = useFormState(form, {
     pristine: true,
     submitting: true,
   })
 
+  const [state, setState] = React.useState<any>()
```
ncphillips:
```suggestion
  const [, setState] = React.useState<any>()
```

PR #1015, opened 2020-04-16
fix: add moment as dep to tinacms

ncphillips:
Adding `tinacms` to a fresh site yields an annoying `moment` is not defined issue. 

The version number is super wide to reduce the chance of multiple versions of moment.


ncphillips:
Because`react-datetime`, the component that we use for the date field, depends on `moment`.

I've also just never heard of luxon

PR #1024, opened 2020-04-17
Update error message on Github 404

jamespohalloran:
Tweak error message on Github 404


jamespohalloran:
Could probably still be improved, but the wording just felt weird to me

PR #1026, opened 2020-04-20
remove toolbar plugin args

ncphillips:
* `GithubEditingContext` provides `editMode`
* `useGithubToolbarPlugins` no longer has arguments


packages/react-tinacms-github/src/toolbar-plugins/useGithubToolbarPlugins.tsx
```diff
@@ -16,22 +16,27 @@ limitations under the License.
...
+export const useGithubToolbarPlugins = () => {
   const cms = useCMS()
+  const { editMode } = useGithubEditing()
+  const github: GithubClient = cms.api.github
+  const baseRepo = github?.baseRepoFullName
+  const baseBranch = github?.baseBranch
 
   useEffect(() => {
+    if (!baseRepo || !baseBranch) {
+      // Warn dev appropriately
```
jamespohalloran, line 37: Shall we throw an error here? (We'll also have to check if it's in editMode)

ncphillips: This was removed in the other PR

ncphillips:
Dif is all messed up and the name is misleading

PR #1030, opened 2020-04-20
GitHub client repo branch

ncphillips:
Add repo and branch names to client.

Satisfies DIP by allowing components and hooks to indirectly call the `getForkRepoName` and `getHeadBranch` functions which grab them from cookies. 

The GithubClient know knows the repo name and branch name for both the fork and the base


packages/react-tinacms-github/src/form/useGithubFileForm.ts
```diff
@@ -39,6 +39,7 @@ export const useGithubFileForm = <T = any>(
     fields: formOptions.fields || [],
     // save & commit the file when the "save" button is pressed
     onSubmit(formData) {
+      const github: GithubClient = cms.api.github
       return cms.api.github
```
dwalkr, line 43:
```suggestion
      return github
```

packages/react-tinacms-github/src/github-client/index.ts
```diff
@@ -73,7 +72,18 @@ export class GithubClient {
     })
   }
 
-  async fetchExistingPR(forkRepoFullName: string, headBranch: string) {
+  get repoFullName(): string {
+    return getForkName() || this.baseRepoFullName
```
jamespohalloran, line 76: I could see this maybe affecting behaviour in the short-term (If your fork cookie gets deleted, we might start committing to the base repo)

packages/react-tinacms-github/src/github-client/index.ts
```diff
@@ -73,7 +72,18 @@ export class GithubClient {
...
+  }
+
+  get branchName() {
+    return getHeadBranch()
+  }
+
+  async fetchExistingPR() {
+    const forkRepoFullName = this.repoFullName
```
jamespohalloran, line 84: misleading `forkRepoFullName` variable if it falls back to  `getForkName() || this.baseRepoFullName`

jamespohalloran:
If we're moving the cookie business inside The GIthubClient, I think [these helpers](https://github.com/tinacms/tinacms/blob/github-client-repo-branch/packages/react-tinacms-github/src/github-editing-context/repository.ts) should be private functions on the GIthubClient.

We're currently calling `setForkName` inside the `GithbubAuthModal`. This would cause the GithubClient to have an out of sync forkName

packages/react-tinacms-github/src/github-client/index.ts
ncphillips: Very true, I'll make it throw instead

ncphillips:
> If we're moving the cookie business inside The GIthubClient, I think these helpers should be private functions on the GIthubClient.

Great! Thanks, I'm going to change that

> We're currently calling setForkName inside the GithbubAuthModal. This would cause the GithubClient to have an out of sync forkName

I don't think this would be a problem. `repoFullName` is a getter, it fetches the value from the cookies on every request.

packages/react-tinacms-github/src/github-client/index.ts
```diff
@@ -71,9 +74,31 @@ export class GithubClient {
         base: this.baseBranch,
       },
     })
+
+    this.setCookie(GithubClient.FORK_COOKIE_KEY, pullRequest.full_name)
```
jamespohalloran, line 78: Did you mean to put this in the `createFork` function?

ncphillips: Yes that's how it was being used before.

ncphillips: But it was being done in [the modal](https://github.com/tinacms/tinacms/blob/master/packages/react-tinacms-github/src/github-editing-context/GithubAuthModal.tsx#L85-L86)

ncphillips: Nope I'm tripping out

PR #1043, opened 2020-04-22
GitHub client refactor

ncphillips:
This PR moves the basic authentication responsibilities from the `TinacmsGithubProvider` to the `GithubClient`

- [x] Merge #1036 with Auth Modal Changes


packages/react-tinacms-github/src/github-client/index.ts
```diff
@@ -26,15 +36,25 @@ export class GithubClient {
...
+    clientId,
+    authCallbackRoute,
+    baseRepoFullName,
+    baseBranch = 'master',
+  }: GithubClientOptions) {
     this.proxy = proxy
     this.baseRepoFullName = baseRepoFullName
     this.baseBranch = baseBranch
+    this.clientId = clientId
+    this.authCallbackRoute = authCallbackRoute
```
ncphillips, line 53: What if we check that these exist and throw an error if they don't. Then we could get rid of that new modal step. That would make it fail much earlier.

Of course the error would have to be verbose and give the same link to the docs

packages/react-tinacms-github/src/github-editing-context/TinacmsGithubProvider.tsx
```diff
@@ -16,64 +16,58 @@ limitations under the License.
...
   const [error, setError] = useState<any>(null)
   const cms = useCMS()
   const github: GithubClient = cms.api.github
-  const [authorizingStatus, setAuthorizingStatus] = useState<AuthState | null>(
-    null
-  )
+  const [authStep, setAuthStep] = useState<AuthStep>(null)
 
   const tryEnterEditMode = async () => {
+    const validSetup =
```
jamespohalloran, line 50: This line confuses me. Earlier we show the error modal if 'completeSetup' is true, but we set validSetup if 'completeSetup' is true

ncphillips: I believe that's a mistake

jamespohalloran: ah ok. Also I think this might be easier to read if that step is something like "invalidConfig" instead of "completeSetup"?

ncphillips: This steps totally gone now

packages/react-tinacms-github/src/github-editing-context/TinacmsGithubProvider.tsx
```diff
@@ -16,64 +16,51 @@ limitations under the License.
...
-    const forkValid = authorizingStatus?.forkValid || (await github.getBranch())
-
-    if (authenticated && forkValid) {
-      enterEditMode()
-    } else {
-      setAuthorizingStatus({
-        authenticated,
-        forkValid,
-      })
+      authStep === 'authenticate' || (await github.getUser())
```
jamespohalloran, line 51: Don't really understand this check. If we are on the 'authenticate' step, why should `authenticated = true` ?

packages/react-tinacms-github/src/github-editing-context/TinacmsGithubProvider.tsx
```diff
@@ -16,64 +16,51 @@ limitations under the License.
...
+    const forkValid = authStep === 'createFork' || (await github.getBranch())
```
jamespohalloran, line 52: Same with this check

packages/react-tinacms-github/src/github-editing-context/TinacmsGithubProvider.tsx
jamespohalloran: Ahh, actually I guess this behaviour hasn't really changed.

packages/react-tinacms-github/README.md
```diff
@@ -28,7 +28,12 @@ const REPO_FULL_NAME = process.env.REPO_FULL_NAME // e.g: tinacms/tinacms.org
 
 const cms = new TinaCMS({
   apis: {
-      github: new GithubClient('/api/proxy-github', REPO_FULL_NAME),
+    github: new GithubClient({
+      proxy: '/api/proxy-github',
+      clientId: process.env.GITHUB_CLIENT_ID,
+      authCallbackRoute: '/api/create-github-access-token'
+      baseRepoFullName: REPO_FULL_NAME,
```
mittonface, line 35: Seems inconsistent for no real reason to have GITHUB_CLIENT_ID be pulled directly from the env while assigning REPO_FULL_NAME to a local var.

ncphillips:
Nice catch, adjusted that 👍

Issue #862, opened 2020-03-09
Re-arrange wysiwym code


jpuri:
Re-organise folder structure for wysiwym to make it more meaningful.

DirtyF:
What do you suggest?

jpuri:
To start with I want to re-organise files and folders, to a structure like this:

```
- schema
   -- nodes
   -- marks
- components
   -- menu and related components
- plugins
   -- image
       -- prosemirror-plugin
       -- nodeView
       -- editingModal
   -- table
       -- prosemirror-plugin
       -- nodeView
       -- editingModal
   -- link
       -- prosemirror-plugin
       -- nodeView
       -- editingModal
   -- code-block
       -- prosemirror-plugin
       -- nodeView
       -- editingModal
....
```

jpuri:
New folder structure I would propose:

```
- context
    - editorState.tsx
- components
    - Editor
       - index..tsx
       - styles
           - CodeMirror.tsx
           - ProseMirror.tsx
    - Menu
    - InlineMenu
- plugins
    - block
    - inline
    - image
    - link
    - common
    - table
- schema
    - nodes
    - marks
    - index.ts
- Translator
```

jpuri:
Hey @ncphillips : Above is details of new folder structure I would propose.

On high level intent is we have folders for `prosemirror schema`, `react components`, `context`, etc.

Each prosemirror feature like `table`, `image`, `inline styles`, `lists`... will be a folder in plugin. It will club its related prosemirror plugin, menu components, commands, etc.

This will make architecture pluggable, adding removing any of these plugins will add remove the feature. We can go ahead to dynamically add plugins to editor like schema.

ncphillips:
Looks good to me!

PR #1045, opened 2020-04-23
chore: create separate components folder for all components

jpuri:
Create separate components folder.
Ref: https://github.com/tinacms/tinacms/issues/862#issuecomment-616667982


jpuri:
Checking this issue @DirtyF .

jpuri:
@DirtyF : it should be fixed after last commit.
Screenshot 2020-04-24 at 5 57 21 PM

jpuri:
Hey @spbyrne : image upload modal position is breaking in this PR:

Screenshot 2020-04-24 at 6 41 09 PM

It appears to be related to code of `MenuButtonDropdown`, can you plz check.

spbyrne:
I'm confused by a couple parts of this and I'm just looking for some clarity.

1. The `FormattingDropdown` component has become the `BlockControl` component, which used to be a file that (confusingly) contained `blockTool` (which was used by `FormattingDropdown`). Instead of `BlockControl` and `blockTool` maybe we should rename it 
 to `TitleControl` and `TitleOption` or something. Having them in separate files when they're so directly related doesn't seem necessary. 

2. The `ImageMenu` refs have been modified. Instead of creating a single `menuButtonRef` and passing it to both the `MenuButton` and `MenuButtonDropdown`, there's a large amount of additional logic added on lines 47-55 that doesn't seem necessary. 

The way the files were moved around combined with all the changes being in a single commit makes it hard to see what exactly changed. The diffs aren't very helpful.

I'll continue to troubleshoot the image modal issue.

jpuri:
Hey @spbyrne : the issue appears related to this piece of [code](https://github.com/tinacms/tinacms/blob/master/packages/react-tinacms-editor/src/state/plugins/Menu/MenuComponents.tsx#L134).

I can try to debug and find exact cause.

Refactoring is actually WIP, it will me much clearer as I refactor code for plugins also.

spbyrne:
I pushed a fix for the dropdown positioning. It was a combination of three things causing the issue:
1. The added ref logic in the `ImageMenu` component
2. The fact the offset measurement was taken from the offset child inside `MenuButtonDropdown`; this means only the first measurement would be accurate. This only worked because we were using `useMemo`, which would memorize the initial value. 
3. The use of `useMemo` caused issues when you closed a dropdown menu. The ref would be `null`, so `useMemo` would be triggered, but without a ref it would return 0. Now, it uses `useState` and only updates if there's a current ref available.

spbyrne:
Unrelated but I think the link button in the toolbar should be disabled if no text is highlighted. Currently it pops up the modal in the default position, but with no text selected it's useless.

jpuri:
Hey @spbyrne : I just pushed a commit to the PR which appears to be fixing this issue.

cc @DirtyF

jpuri:
Hey @spbyrne : the fix for modal positioning works great 👍

jpuri:
> Unrelated but I think the link button in the toolbar should be disabled if no text is highlighted. Currently it pops up the modal in the default position, but with no text selected it's useless.

Make sense, can you plz add this to issues of wysiwym.

DirtyF:
Link button issue added in #1059

Issue #1044, opened 2020-04-23
Select field label and value


DaniTulp:
## Summary

Say I'm using a SelectField for creating a form and I want to be able to have a different label from the value of the option, right now I can't do that with the default component.

### Basic example
The api I would propose would look like this
```     
{
    name: 'gridCols',
    label: 'Grid columns',
    component: 'select',
    options: [
       {value:  'grid-cols-3', label: '3' }
    ]
},
```
We could probably also keep the old API if you just pass in an array of Strings

### Motivation

I think the use case described above explains it well enough. I also think it would keep it inline with the other elements within forms, most of them have a way to modify the way something is displayed for the end-user.

ncphillips:
Absolutely! The type for option should be :

```
type Option = string | {
  value: string
  label: string
}
```

PR #1047, opened 2020-04-23
feat: added Option type for Select/SelectField

DaniTulp:
Closes issue #1044

### DIFF
--- a/packages/@tinacms/fields/src/components/Select.tsx
+++ b/packages/@tinacms/fields/src/components/Select.tsx
@@ -19,11 +19,18 @@ limitations under the License.
 import * as React from 'react'
 import styled from 'styled-components'
 
+type Option =
+  | string
+  | {
+      value: string
+      label: string
+    }
+
 interface SelectFieldProps {
   label?: string
   name: string
   component: string
-  options: string[]
+  options: Option[]
 }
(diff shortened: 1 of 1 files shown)


packages/@tinacms/fields/src/components/Select.tsx
```diff
@@ -46,11 +53,20 @@ export const Select: React.FC<SelectProps> = ({ input, field, options }) => {
         {...input}
       >
         {selectOptions ? (
-          selectOptions.map(option => (
-            <option value={option} key={option}>
-              {option}
-            </option>
-          ))
+          selectOptions.map((option: Option) => {
```
ncphillips, line 56: Great! 

My only ask is to make this a two step mapping:

```ts
selectOptions
  .map(toProps)
  .map(toComponent)
```

where `toMap` looks something like:

```ts
function toProps(option: Option) {
  if (typeof option === 'object') return option
  return { value: option, label: option }
}
```

DaniTulp: Great suggestion! Should be resolved now. Should I update the docs too?

ncphillips:
@all-contributors please add @DaniTulp for code

PR #1065, opened 2020-04-27
GitHub branching

ncphillips:
- [x] List Available Branches
- [x] Switch Branches
- [x] Create Branches
- [x] Add Confirmation dialogue before switching

image


packages/react-tinacms-github/src/toolbar-plugins/BranchSwitcherPlugin.tsx
```diff
@@ -0,0 +1,513 @@
...
+  const [branches, setBranches] = React.useState<Branch[]>([])
+  React.useEffect(() => {
+    github
+      .getBranchList()
+      .then(branches => {
+        setBranches(branches)
+        setBranchStatus('loaded')
+      })
+      .catch(() => {
+        setBranchStatus('error')
```
jamespohalloran, line 65: It doesn't look like we're handling an error branchStatus currently

ncphillips: Indeed we are not.

@spbyrne want to add a little "We had trouble loading branches. Please refresh to try again."

jamespohalloran:
Wondering if we need to scope the localStorage unsaved changes to workingRepo / branch
https://github.com/tinacms/tinacms/pull/1066#issuecomment-620620168

ncphillips:
Possibly, but I think that's outside the scope of this PR. The localstorage caching is not done by these packages.

packages/react-tinacms-github/src/toolbar-plugins/BranchSwitcherPlugin.tsx
spbyrne: Yeah good call.

spbyrne: Added 👍

PR #1066, opened 2020-04-28
react-tinacms-github: support working from origin

ncphillips:
If the user is authorized to write to the repository then there's no need to fork created

### DIFF
--- a/packages/react-tinacms-github/src/github-client/github-client.ts
+++ b/packages/react-tinacms-github/src/github-client/github-client.ts
@@ -129,6 +129,8 @@ export class GithubClient {
       return forkName
     }
 
+    this.setCookie(GithubClient.WORKING_REPO_COOKIE_KEY, this.baseRepoFullName)
+
     return this.baseRepoFullName
   }
 
--- a/packages/react-tinacms-github/src/github-client/github-client.ts
+++ b/packages/react-tinacms-github/src/github-client/github-client.ts
@@ -67,11 +67,10 @@ export class GithubClient {
     return this.getUser()
   }
 
-  isAuthorized() {
-    // If this 404's they don't have access.
-    // This works for now but a different
-    // implementation will be needed.
-    return this.getBranch()
+  async isAuthorized() {
+    const repo = await this.getRepository()
+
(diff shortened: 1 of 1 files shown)


packages/react-tinacms-github/src/github-client/github-client.ts
```diff
@@ -120,12 +125,11 @@ export class GithubClient {
...
 
-    if (!forkName) {
-      // TODO: Right now the client only works with forks. This should go away once it works with origin.
-      throw new Error('GithubClient cannot find name of fork')
+    if (forkName) {
+      return forkName
     }
 
-    return forkName
+    return this.baseRepoFullName
```
jamespohalloran, line 132: Looks good, my one worry is that baseRepo is used as a fallback instead of needing to be explicitly set as the working Repo within the cookie. 

For instance, if your fork cookie was to expire, or get deleted, we would switch you to the baseRepo which might not be ideal if you're about to save.

ncphillips: Just added the part that sets the cookie.

I don't think the cookie expiring will be a problem though. This change prevents you from ever working from a fork if you have permission to do so.

jamespohalloran:
Are there any localStorage implications we need to deal with here? I assume where you switch repos, you don't want your uncommited changes carrying over

packages/react-tinacms-github/src/github-client/github-client.ts
jamespohalloran: > I don't think the cookie expiring will be a problem though. This change prevents you from ever working from a fork if you have permission to do so.

Even if I had permission to edit on the base repo, I think I'd still want to have to explicitly switch my working repo to the base repo, rather than have it switch automatically if my WORKING_REPO_COOKIE expired.

ncphillips:
There's no way to switch repos so that wouldn't apply. It may apply for switching branches though

jamespohalloran:
Oh gotchya, so if you have access to the repo you just edit on the base by default instead of being able to create a fork. Makes sense.

Issue #1067, opened 2020-04-28
Create separate folder for schema


jpuri:
Ref: https://github.com/tinacms/tinacms/issues/862

PR #1068, opened 2020-04-28
chore: more code re-organizing

jpuri:
PR moves around some code and creates separate folder for schema.

Now next step will be cleaning up content inside plugin folder to group related. functionality like this:

```
- plugins
    - block
        - commands.ts
        - commands.test.ts
        - components
            - InlineMenu
    - inline
        - commands.ts
        - commands.test.ts
        - components
            - BlockMenu
    - image ...
    - link ...
    - common ...
    - table ...
    - default-plugins.ts
    - index.ts
```


jpuri:
Fixes https://github.com/tinacms/tinacms/issues/1067

PR #1071, opened 2020-04-28
chore: prevent user from creating PR on base branch

### LINKED ISSUE #1069: react-tinacms-github: Pull Request button is still visible when on the base branch
Screen Shot 2020-04-28 at 2 54 54 PM
It appears as if I can create a pull request from the base_repo master > master.

When you click "Create Pull Request", it fails

### Expected result

Should the pull request button be hidden when you are on the base_repo base_branch?

ncphillips:
image

Note: `baseBranch` and `workingBranch` contain the owner name if they're different?

closes #1069

### DIFF
--- a/packages/react-tinacms-github/src/toolbar-plugins/pull-request/PRModal.tsx
+++ b/packages/react-tinacms-github/src/toolbar-plugins/pull-request/PRModal.tsx
@@ -93,6 +93,19 @@ export const PRModal = () => {
     baseBranch = `${baseOwner}:${baseBranch}`
   }
 
+  if (workingBranch === baseBranch) {
+    return (
+      <PrModalBody>
+        <ModalDescription>
+          <p>
+            You are currently on the base branch: <b>{baseBranch}</b>.
+          </p>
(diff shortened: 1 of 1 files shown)


packages/react-tinacms-github/src/toolbar-plugins/pull-request/PRModal.tsx
```diff
@@ -93,6 +93,19 @@ export const PRModal = () => {
     baseBranch = `${baseOwner}:${baseBranch}`
   }
 
+  if (workingBranch === baseBranch) {
```
jamespohalloran, line 96: Should we also be checking if the Base Repo is the same as the working repo? 
We will want this to still work with forks

ncphillips: Nope, the baseBranch and workingBranch variables contain the owner name if they're different.

ncphillips: Confusing name, I might change them 🤔

PR #1075, opened 2020-04-29
fix: github-client branch name should default to base branch

ncphillips:
It was defaulting to 'master' even if that wasn't the base

Discovered this while trying to turn the tinacms/rfcs repo into a Next+Github site.


ncphillips:
@all-contributors please add @jpuri for review

Issue #1081, opened 2020-04-30
Link plugin


jpuri:
Ref: https://github.com/tinacms/tinacms/issues/862

PR #1082, opened 2020-04-30
chore: moving link related functionality to link plugin

jpuri:
Link plugin: https://github.com/tinacms/tinacms/issues/1081


DirtyF:
We don't validate URL links for now, it might be something this plugin could do

PR #1086, opened 2020-05-01
Refactor plugins

jpuri:
Moving all remaining functionalities to respective plugins.


jpuri:
PR is ready for review, it moves all code to respective plugins. It covers almost all of plugins and cleanup code for each.

Code now looks well organised [check here](https://github.com/tinacms/tinacms/tree/refactor_plugins/packages/react-tinacms-editor).

[Common plugin](https://github.com/tinacms/tinacms/tree/refactor_plugins/packages/react-tinacms-editor/src/plugins/Common) still needs some cleanup.

PR #1087, opened 2020-05-01
CMS Events

ncphillips:
### Core Changes
- [x] Put an EventBus on `CMS`

### React

- [ ] ~Introduce `useEvents` hook~ Probably gonna leave this out for now

### Refactor Old Subscription APIs
- [x] Plugins
- [x] Alerts

## RFC
tinacms/rfcs#0009

## Test Output

image


ncphillips:
As soon as I implemented `useEvent` I realized the API was weird.

```ts
useEvent(() => {
  console.log("Forms updated")
}, ['form:*:plugins']
```

This looks like `useEffect` but the second array behaves totally differently

Leaving it for now

ncphillips:
Right now I'm thinking of just not introducing hooks with this PR. 

That said here's a couple ideas. I'm leaning towards option 2.

### Option 1: Rerender on the specified Events
In this API the listener function is optional and it always re-renders
```tsx
  // Rerenders for any Sidebar Event
  useEventListener(['sidebar'])

  // Rerender with Callback
  useEventListener(['sidebar'], () => {
    // Do something
  })

interface Subscribable {
  subscribe(callback: any): () => void
}

export function useEventListener(events: string[], listener?: Function) {
  const [, s] = React.useState(0)
  React.useEffect(() => {
    return subscribable.subscribe(() => {
      s(x => x + 1)
      if (cb) cb()
    })
  })
}
```

### Option 2: Subscribe and callback
Lower-level. The listener is required and it doesn't necessarily trigger a re-render. 
```
  useEventListener({
    events: ['sidebar'],
    listener() {
      // Do stuff
    },
  })

interface EventListener {
  events?: string[]
  listener(): void
}

function useEventListener({ listener, events = [] }: EventListener) {
  const cms = useCMS()

  React.useEffect(() => {
    return cms.events.subscribe(listener, events)
  }, events)
}
```

dwalkr:
How about option 2, but it looks like option 1 - you just always have to pass a callback

```js
useEventListener(['sidebar'], () => {
    // Do something
})
```

On top of that: while I see the benefit of events being an array, with the wildcard syntax (`plugins:*:add`), most of the time I think developers will just be subscribing to a single event signature. I think DX might be improved if the array was optional:

```js
useEventListener('sidebar', () => {
    // Do something
})
```

This mirrors the standard [addEventListener syntax](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener), which IMO is a Very Nice Thing™

ncphillips:
Agreed. I'm going to change that now

dwalkr:
on second thought, while I do like that syntax, maybe it's too familiar (and generic-named). What about `useCMSEvent` instead...?

ncphillips:
I agree on that too.

I'm still not sure about adding the hook at all. But I'm siwtching the `EventBus#subscribe` interface around

packages/@tinacms/core/src/event.ts
```diff
@@ -0,0 +1,84 @@
...
+  handleEvent(event: CMSEvent) {
+    if (this.watchesEvent(event)) {
+      this.callback(event)
+      return true
+    }
+    return false
+  }
+
+  watchesEvent(currentEvent: CMSEvent) {
+    if (this.eventPattern === '*') return true
```
dwalkr, line 61: I feel like this line should be in `matchEventPattern`

packages/@tinacms/core/src/event.ts
```diff
@@ -47,7 +47,7 @@ export class EventBus {
 }
 
 export class Listener {
-  constructor(private eventPattern: string, private callback: Callback) {}
+  constructor(private matchEventPattern: string, private callback: Callback) {}
```
dwalkr, line 50: I think I was misunderstood, I didn't mean rename the var. FWIW the old name was better 😅

PR #1088, opened 2020-05-01
(gatsby-tinacms-remark) & (gatsby-tinacms-json): return forms in prod

### LINKED ISSUE #1054: Gatsby build fails with cannot read property 'finalForm' of null
## Description

When using InlineForm with Gatsby, the build process fails. @kendallstrautman explained it's likely because [useRemarkForm returns [markdown,null] in production](https://github.com/tinacms/tinacms/blob/5e11cb97730d84b1507ec8758b45fe2a9a4bf6e8/packages/gatsby-tinacms-remark/src/useRemarkForm.tsx#L52-L54) and the inline form doesn't expect the null value.

The problem occurs only once `InlineForm` is added to the page.

### Steps to reproduce

Here's a test case: https://github.com/spbyrne/final-form-null

### Expected result

When you run `gatsby build`, it should build successfully. 

### Actual result

```
WebpackError: TypeError: Cannot read property 'finalForm' of null
  
  - index.js:1 a
    node_modules/@tinacms/form-builder/build/index.js:1:733
```

kendallstrautman:
Fixes #1054 — curious if we should also make some changes with the call to `useWatchFormValues`...

_sidenote_ what is UP with these old merges and commits showing up on all my PRs??? 🤷‍♀️


ncphillips:
> curious if we should also make some changes with the call to useWatchFormValues...

I was thinking the same, but this is good on it's own as well.

Issue #1063, opened 2020-04-27
Several clicks required for Table floating menu items to trigger


ncphillips:
## Description

Must click multiple times to add/remove columns

### Steps to reproduce
1. Create a table
1. Click on the Add Column button; nothing happens
1. Click the Add Column button multiple times in a row; column is added

jpuri:
Hey @ncphillips : are you facing this in latest code from master ?

I created this few days ago: https://github.com/tinacms/tinacms/issues/1017

Issue #1017, opened 2020-04-17
Click event is not getting fired properly in table menus


jpuri:
## Description

I found an issue with table menus, click event is not getting fired properly. It requires 2-3 clicks for event to get captured.

It looks like event is getting captured somewhere else.

I have fixed it temporarily by using `onMouseDown` instead of `onClick` in this [PR](https://github.com/tinacms/tinacms/pull/1002). But it requires more debugging and fixing.

PR #1102, opened 2020-05-06
fix: issue with table menu click event not trigerred

jpuri:
Fixes issue with click event on table floating menus not working correctly.


jpuri:
Fixes https://github.com/tinacms/tinacms/issues/1063

Fixes https://github.com/tinacms/tinacms/issues/1017

Issue #1083, opened 2020-04-30
Image deletion issue


jpuri:
Steps:

- click image to select it
- close image overlay modal
- hit delete

Image should get deleted.

PR #1111, opened 2020-05-08
fix: close image popup and hitting delete button should delete image

jpuri:
Fix for image not getting deleted when popup over it is closed and delete button is hit.

Ref: https://github.com/tinacms/tinacms/issues/1083


jpuri:
Fixes https://github.com/tinacms/tinacms/issues/1083

Issue #1059, opened 2020-04-25
Error on deleting empty link


DirtyF:
```gherkin
Given I have no text selected
When I click on the link button
And I click delete
Then an error is blowing up
```

```js
RangeError: Index -1 out of range for <>
child
node_modules/prosemirror-model/dist/index.es.js:211
  208 | // index is out of range.
  209 | Fragment.prototype.child = function child (index) {
  210 |   var found = this.content[index];
> 211 |   if (!found) { throw new RangeError("Index " + index + " out of range for " + this) }
  212 |   return found
  213 | };
  214 | 
View compiled
markExtend
src/commands/link-commands.ts:33
  30 | let endIndex = $cursor.indexAfter()
  31 | 
  32 | // Clicked outside edge of tag.
> 33 | if (startIndex === $cursor.parent.childCount) {
     | ^  34 |   startIndex--
  35 |   endIndex--
  36 | }
View compiled
removeLinkBeingEdited
src/commands/link-commands.ts:93
  90 |  * @param {(tr: Transaction) => void} dispatch
  91 |  * @param {Object} attrs
  92 |  */
> 93 | export function updateLinkBeingEdited(
  94 |   state: EditorState,
  95 |   dispatch: Dispatch | null,
  96 |   attrs: object
View compiled
▶ 20 stack frames were collapsed.
```

**Expected**

- Link button is only shown when text is selected
- No error on delete

PR #1112, opened 2020-05-08
fix: link button should be disabled if selection is empty

jpuri:
Link button will be disabled if no text is selected.


jpuri:
Fixes https://github.com/tinacms/tinacms/issues/1059

PR #1117, opened 2020-05-08
Add example for when we might need to ignore an invalid markdown link

jeffsee55:
This is for discussing our Prosemirror issues in Forestry only


jpuri:
Hey @jeffsee55 : I updated PR with prosemirror extension to support `relref` links. I plan to add more test coverage for it, that is very important in this scenario.

@DirtyF : it might be useful in-fact to merge this in tina also, if we need to support hugo stuff.

DirtyF:
@jpuri I'll let that technical decision up to @ncphillips, for now, Tina doesn't really need this.

packages/react-tinacms-editor/src/translator/MarkdownTranslator/hugo-utils.ts
```diff
@@ -0,0 +1,37 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+
+*/
+
+import { Token } from './types'
+import { Schema } from 'prosemirror-model'
+
+const RELREF_LINK_REGEX = /\[(.+)\]\((\{\{\<\srelref ".*"\s>\}\}\))/g
```
DirtyF, line 22: While we're at it, can we also add support for `ref` here?
https://gohugo.io/content-management/cross-references/#use-ref-and-relref

DirtyF: FYI Hugo has also pipes and filters, could that regexp also support those as in https://gohugo.io/functions/rellangurl/

```
{{ "blog/" | absLangURL }}
{{ "blog/" | relLangURL }}
```

jpuri: @DirtyF : I can add more regex rules for this. It will be useful to have some samples of valid url for Hugo to test the REGEX. It is required to ensure that it works correctly in all possible combinations.

ncphillips:
Tina doesn't really support Hugo right now, so I would hold off on accepting this. 

Eitherway, I wouldn't want to SSG specific stuff in the `react-tinacms-editor` package, but maybe this could be a good candidate for a plugin?

jpuri:
Yep @ncphillips : plugin is indeed good idea. We can build this kind of extension points in parser and build plugins.

I agree it not idea to add customisations like this to core lib.

jeffsee55:
Hi @jpuri - I've added the example for shortcodes discussed in Basecamp, let me know if it's not clear what we're after here.

jpuri:
@jeffsee55 , @DirtyF : I have updated the PR with handling for links. The fix uses a REGEX for Hugo links.

I had discussion with Frank about these tags `{{% grid %}}` but there is apparently no way to fix whitespace issue while keeping markdown also valid.

ncphillips:
@jeffsee55 does this PR still need to be open?

jeffsee55:
No we can close it. Thanks

PR #1118, opened 2020-05-08
chore: update `gatsby-tinacms-*` TypeScript dependencies/files

bshackelford:
Recently started using TinaCMS with a Gatsby site and I'm really loving it so far. Great work and thank you to everyone involved. The Gatsby + TinaCMS site is built with TypeScript and after adding `gatsby-tinacms-remark` I started seeing TypeScript errors within the codebase.

```shell
node_modules/gatsby-tinacms-remark/src/to-markdown.ts:20:23 - error TS7016: Could not find a declaration file for module 'js-yaml'. '/Users/bshackelford/vsco/js/node_modules/js-yaml/index.js' implicitly has an 'any' type.
  Try `npm install @types/js-yaml` if it exists or add a new declaration (.d.ts) file containing `declare module 'js-yaml';`

20 import * as yaml from 'js-yaml'
                         ~~~~~~~~~

node_modules/gatsby-tinacms-remark/src/useRemarkForm.tsx:184:19 - error TS7030: Not all code paths return a value.

184     React.useMemo(() => {
                      ~~~~~~~

Found 2 errors.
```

The **first commit** addresses the first error by moving `@types/js-yaml` from `devDependencies` to `dependencies`. See [TypeScript's "Publishing" docs](https://www.typescriptlang.org/docs/handbook/declaration-files/publishing.html#dependencies):

> For that reason, we used `"dependencies"` and not `"devDependencies"`, because otherwise our consumers would have needed to manually install those packages. If we had just written a command line application and not expected our package to be used as a library, we might have used `devDependencies`.

The **second commit** addresses the second error by only publishing declaration files in `src` instead of all `*.ts` files. Since the `.ts` source files are included and TypeScript config files may differ (in my case `"noImplicitReturns": true`) this will potentially throw errors for consumers of the package. There may be a better way to prevent the TypeScript compiler from typechecking `.ts` files within `node_modules` so I'm open to feedback on this approach.


ncphillips:
Fantastic PR! Thanks a lot @bshackelford. I'm off now but I'll make sure this gets in Monday's release 👍

bshackelford:
@ncphillips Awesome! That would be super helpful, thanks so much. Have a great weekend.

Issue #771, opened 2020-02-19
Exclude TinaCMS + dependencies from production build


nibtime:
## Summary

I've been busy with getting a Gatsby site into production and I observed that the app bundle size became unusually large (1.7MB uncompressed), so I installed webpack bundle analyzer and generated a report. And it turned out, that TinaCMS is the main culprit, accounting for around 800-900KB of the size, including dependencies, which are irrelevant to the production build (but are needed for the development server with Tina of course).

I take guidance from the excellent articles [The Cost of JavaScript 2018](https://medium.com/@addyosmani/the-cost-of-javascript-in-2018-7d8950fbb5d4) and [The Cost of JavaScript 2019](https://v8.dev/blog/cost-of-javascript-2019). The overhead that stems from TinaCMS is almost as large as the median uncompressed bundle size. This is a huge obstacle for using Tina with production sites.

### Basic example
TinaCMS + its dependencies should not be present in production build. I have no idea though how this can be achieved, since Tina is integrated with code of the site. First thing that comes to mind is Webpack hacking. Maybe you already know that this is a problem and already have an idea how to fix this? If so, this should be part of the docs, since out-of-the-box this blow-up happens.

Here is the [report of bundle analyzer of my production build](https://github.com/tinacms/tinacms/files/4225790/report.zip)

And this is a regex to filter for dependencies, that are exclusively needed by Tina
`(@tinacms|prosemirror|codemirror|react-datetime|react-beautiful-dnd|markdown-it|final-form|styled-components|react-dropzone|react-color|tinycolor2|stylis).*.js`

### Motivation
Obviously faster page loads and less production traffic. Tina should not be that expensive.

ncphillips:
We're a little ways off from totally taking care of this problem. But I've opened PR #773 to hopefully mitigate future bloat.

I think a major win would be to take better advantage of dynamic imports. Codemirror and Prosemirror are pretty big, if we could split the wysiwyg into it's own chunk that would be quite helpful. I've done this for apps before, but I'm a little nervous about doing it for lbraries.

nibtime:
Good news, I found a way to hotfix this! Bundle size is now just around 490KB (that's like 1.2MB smaller 😃 ) and Tina is still up and running in development.

[Here's the new report](https://github.com/tinacms/tinacms/files/4233335/report.zip)

And the hotfix of course. It involves 2 parts:
[...]

PR #1134, opened 2020-05-12
Remove Markdown/Html Editors from Default Plugins

ncphillips:
## Problem

Tina adds a lot of Javascript to your site. 

As discussed in #771 a large portion of that comes from `react-tinacms-editor`. This package includes both `prosemirror` and `codemirror`, two fairly large projects for creating text editors. 

## Solution

In this branch I moved the `HtmlFieldPlugin` and `MarkdownFieldPlugin` definitions from `tinacms` to `react-tinacms-editor`. Doing this means that these fields will not be available by default. Instead, developers will have to add `react-tinacms-editor` themselves if they want to be able to use them. This gives developers the option to load these fields dynamically to reduce the amount of javascript being transferred.

## The Numbers

| gatsby-starter-blog | Total JS | Diff
| --- | --- | --- |
| without tinacms |  0.3 MB | --- |
| tinacms with  wysiwyg by default |  1.6 MB| +1.3 MB
| tinacms with wysiywg removed |  0.9MB | -0.7 MB

**Total JS:** Can be found by opening the network tab and only looking at JS requests. At the bottom of the panel is "Total Transferred"

_these numbers are only approximations_

### Loading Immediately. 

This is the simplest solution but it will increase your initial bundle size by 700kb

```tsx
import { HtmlFieldPlugin, MarkdownFieldPlugin } from "react-tinacms-editor"

cms.plugins.add(HtmlFieldPlugin)
cms.plugins.add(MarkdownFieldPlugin)
```

### Loading Dynamically

```js
import { useCMS } from "tinacms"

const Layout = ({ children }) => {
  const cms = useCMS()

  React.useEffect(() => {
    import("react-tinacms-editor").then(({ MarkdownFieldPlugin }) => {
      cms.plugins.add(MarkdownFieldPlugin)
    })
  }, [])

  return <div>{children}</div>
}
```

### Loading InlineWysiwyg Dynamically

This component will load the wysiwyg but only if the inline form is active.

```tsx
function InlineWysiwyg(props) {
  const { status } = useInlineForm()
  const [{ InlineWysiwyg }, setEditor] = React.useState({})

  React.useEffect(() => {
    if (!InlineWysiwyg && status === "active") {
      import("react-tinacms-editor").then(setEditor)
    }
  }, [status])

  if (InlineWysiwyg) {
    return <InlineWysiwyg {...props} />
  }

  return props.children
}
```

## Placeholder Plugin

Rather then just remove the plugin entirely, this PR also adds some placeholder plugins that tell the user about the deprecation.

- [x] Create the plugin
- [x] Fix the styles
- [x] Link to the propert docs

image


spbyrne:
Updated placeholder styles.

Screen Shot 2020-05-20 at 1 03 16 PM

packages/tinacms/src/plugins/fields/markdown.tsx
```diff
@@ -0,0 +1,78 @@
...
+      <FieldMeta name={props.input.name} label={`Deprecated: ${name} Field`}>
+        <PlaceholderParagraph>
+          In order to help improve bundle sizes the {name} Field has been
+          removed from the set of default fields.
+        </PlaceholderParagraph>
+        <PlaceholderParagraph>
+          See the docs to learn how to{' '}
+          <a
+            // TODO: Add actual link
+            href="https://tinacms.org/docs"
```
kendallstrautman, line 56:
```suggestion
            href="https://tinacms.org/docs/cms/plugins#adding-plugins"
```

packages/tinacms/src/plugins/fields/markdown.tsx
```diff
@@ -0,0 +1,78 @@
...
+    return (
+      <FieldMeta name={props.input.name} label={`Deprecated: ${name} Field`}>
+        <PlaceholderParagraph>
+          In order to help improve bundle sizes the {name} Field has been
+          removed from the set of default fields.
+        </PlaceholderParagraph>
+        <PlaceholderParagraph>
+          See the docs to learn how to{' '}
+          <a
+            // TODO: Add actual link
```
kendallstrautman, line 55:
```suggestion
```

PR #1135, opened 2020-05-12
Feat: Add StyleProvider and FontLoader

### LINKED ISSUE #926: Warning (@tinacms/styles): Inter font should not be imported via CreateGlobalStyle
## Description

This warning when developing:

> Please do not use @import CSS syntax in createGlobalStyle at this time, as the CSSOM APIs we use in production do not handle it well. Instead, we recommend using a library such as react-helmet to inject a typical <link> meta tag to the stylesheet, or simply embedding it manually in your index.html <head> section for a simpler app.

### Steps to reproduce

Go to `demo-next`, kick up development mode and you'll see this warning in the console.

### Expected result

No warning, and the font still imports properly.

### Actual result

The warning displayed, but it didn't seem to affect the font.

#### Tinacms Package Versions

```text
├─┬ @tinacms/api-git@0.7.0
├─┬ @tinacms/form-builder@0.3.0
│ ├─┬ @tinacms/forms@0.3.0
│ ├─┬ @tinacms/react-core@0.2.9
│ │ └─┬ @tinacms/forms@0.3.0
├── @tinacms/git-client@0.6.2
├── @tinacms/styles@0.4.0
├── next-tinacms-json@0.3.7
├─┬ react-tinacms@0.9.15
├─┬ react-tinacms-editor@0.2.0
│ ├── @tinacms/core@0.7.3
│ ├── @tinacms/icons@0.7.0
├── react-tinacms-inline@0.4.0
├─┬ tinacms@0.18.0
│ ├── @tinacms/alerts@0.1.0
│ ├─┬ @tinacms/fields@0.7.0
│ │ ├── @tinacms/styles@0.3.0
│ ├─┬ @tinacms/form-builder@0.2.16
│ ├─┬ @tinacms/forms@0.2.1
│ ├── @tinacms/media@0.1.0
│ ├── @tinacms/react-alerts@0.1.0
│ ├── @tinacms/react-modals@0.1.0
```

### LINKED ISSUE #1108: External reference in tinacms 0.19.0
Monitoring network traffic in firefox I noticed that en external css resource is loaded.

I would blame this part but I am not too familiar with styled-components. https://github.com/tinacms/tinacms/blob/da37710076665f6dfc06ca4352c5a287fa0de793/packages/%40tinacms/styles/src/Styles.tsx#L90

The overall system seems to work even though there not network available but I consider it odd that it tries to load this resource in runtime.

spbyrne:
### Adds FontLoader component

This component uses the **typekit/webfontloader** library to import our interface typeface only when Tina is active and styled.

### Renames some things in `@tinacms/styles`
- `TinaResetStyles` 👉 `tina_reset_styles` (since this is just plain css, the name shouldn't look like a component)
- `TinaReset` 👉 `StyleReset`

---

### Introduces `Theme` component into `TinaProvider`. 

`TinaProvider` now accepts a `styled : boolean` prop that conditionally renders `Theme`. 

```js
export function Theme() {
  return (
    <>
      <FontLoader />
      <GlobalStyles />
    </>
  )
}
```

This allows people to prevent styles (and the inter font) from being loaded when they don't need or want it, but it will default to `true` if no `styled` prop is passed.
```js
// Example: Tina styles and font don't load in production
<TinaProvider cms={this.cms} styled={process.env.NODE_ENV === 'production'}>
  <Component {...pageProps} />
</TinaProvider>
```

This also allows people to pick and choose components from `@tinacms/styles` to load. For example, they may not want to load the 'Inter' font, so they could do something like this: 
```js
import { GlobalStyles as TinaGlobalStyles } from '@tinacms/styles'

// Example: this config doesn't load external 'Inter' Font
<TinaProvider cms={this.cms} styled={false}>
  <TinaGlobalStyles />
  <Component {...pageProps} />
</TinaProvider>
```

Closes #926 
Also closes #1108 by giving a way to disable that external request to google fonts


packages/tinacms/src/components/TinaProvider.tsx
```diff
@@ -40,6 +40,7 @@ export const TinaProvider: React.FC<TinaProviderProps> = ({
   return (
     <CMSContext.Provider value={cms}>
       <ModalProvider>
+        <FontLoader active={true} />
```
Muthaias, line 43: Would it be completely against the intended use of the plugin system if you used it to configure or optionally replace the `FontLoader`?

kendallstrautman:
Closes #926

spbyrne:
Thanks @kendallstrautman 👍

packages/tinacms/src/components/TinaProvider.tsx
```diff
@@ -29,18 +29,20 @@ export interface TinaProviderProps {
...
   children,
   hidden,
   position,
+  styled = true,
 }) => {
   return (
     <CMSContext.Provider value={cms}>
       <ModalProvider>
-        <GlobalStyles />
+        <StyleProvider active={styled} />
```
ncphillips, line 45: I'm just realizing the `Provider` suffix is a bit misleading.

The `Provider` convention is a React Context thing. i.e.

```
const Style = React.createContext(...)

<Style.Provider><App /></Style.Provider>
```

The `GlobalStyles` name better fits the naming conventions for this use case.

Maybe the old `GlobalStyles` could be renamed to `GlobalCSS` to distinguish it from the `FontLoader`

ncphillips: Note that this also goes for the Toolbar, but no need to address that now.

kendallstrautman: we could call `TinaGlobalStyles` --> `TinaTheme`

ncphillips: That works 👌

kendallstrautman: or `StyleProvider` --> `TinaTheme`....might be nice

ncphillips:
```
Theme
    Fonts
    Styles
```

kendallstrautman:
Hows this: 

`StyleProvider` 👉 `GlobalStyles` (now lives in `Styles.tsx`)
`TinaGlobalStyles` 👉 `TinaTheme` (using the 'Tina' name for clarity since folks may use this component separately)

ncphillips:
I'm note sure the `Tina` part is necessary. They will importing it from the tinacms package, of course it's for Tina. If they need to namespace it they can do that themselves:
```ts
import { Theme as TinaTheme } from "@tinacms/styles"
```

kendallstrautman:
One more time lol

```
 Theme
    FontLoader
    GlobalStyles
```

kendallstrautman:
Should we also to the `cms.enabled` stuff?

ncphillips:
Yeah sure 👍

Issue #745, opened 2020-02-13
Strikethrough ~~Like this.~~ crashes wysiwyg


PaulBunker:
## Description
wysiwyg crashes if you add the following line in your markdown:

`Strikethrough uses two tildes. ~~Scratch this.~~`

### Steps to reproduce

create a markdown file with the following contents
```
Strikethrough uses two tildes. ~~Scratch this.~~
```
open the page and click edit (using inline editing and the wysiwyg)

### Expected result

you should be able to edit the page

### Actual result

What actually happened? The page crashes giving many errors. including the following.

```ErrorEvent {isTrusted: true, message: "Uncaught Error: Token type `s_open` not supported by Markdown parser", filename: "http://localhost:8000/commons.js", lineno: 6853, colno: 71000, …} 
    in AppContainer (created by HotExportedComponent)
    in HotExportedComponent (created by PageRenderer)
    in PageRenderer (at query-result-store.js:86)
    in PageQueryStore (at root.js:51)
    in RouteHandler (at root.js:73)
    in EnsureResources (at root.js:61)
    in LocationHandler (at root.js:119)
    in LocationProvider (created by Context.Consumer)
    in Context.Consumer (created by Location)
    in Location (at root.js:118)
    in Root (at root.js:126)```

#### Tinacms Package Versions
```text
├─┬ gatsby-plugin-tinacms@0.1.21
│ └─┬ tinacms@0.12.0
│   ├── @tinacms/core@0.7.0
│   ├─┬ @tinacms/fields@0.3.1
│   ├─┬ @tinacms/form-builder@0.2.13
│   ├─┬ @tinacms/forms@0.1.2
│   ├── @tinacms/icons@0.4.1
│   ├─┬ @tinacms/react-core@0.2.5
│   ├── @tinacms/styles@0.1.1
├─┬ gatsby-tinacms-git@0.2.18
│ ├─┬ @tinacms/api-git@0.5.4
│ └── @tinacms/git-client@0.4.1
```

PaulBunker:
Adding a gif to prove I'm not lying :)

![Feb-23-2020 11-11-58](https://user-images.githubusercontent.com/1537408/75110487-1096ee80-562f-11ea-8994-6757b2d00bd8.gif)

PaulBunker:
Gif for clarity

![Feb-23-2020 11-28-34](https://user-images.githubusercontent.com/1537408/75110549-b64a5d80-562f-11ea-9796-daac488e6b9d.gif)

[...]

PR #1137, opened 2020-05-14
feat: support for strike in wysiwym

jpuri:
Strikethough support on wysiwym.

@spbyrne : can you plz add suitable icon, here I have used `underline`  icon for strike.

Screenshot 2020-05-14 at 11 45 00 AM

@ncphillips : should we make this optional feature as some users may not want it.


DirtyF:
@jpuri we'll see later for toolbar options, let's add this by default for now.

kendallstrautman:
Closes #745

spbyrne:
Added strikethrough icon. 

Screen Shot 2020-05-15 at 12 06 10 PM

PR #1145, opened 2020-05-18
feat: raw mode implementation

### LINKED ISSUE #626: Add "raw mode" toggle for WYSIWYG
## Summary

The WYSIWYG editor should have a button in its toolbar to toggle between editing the raw content and editing in WYSIWYM mode

### Motivation

There are still some kinks with the WYSIWYG, and we should track them down and fix them, but in the meantime having a raw mode will allow editors to fix formatting gone awry from the CMS instead of having to open a text editor.

jpuri:
Basic raw mode implementation, it still needs some brushup.

mode_toggle

@spbyrne, @DirtyF  : some idea about the design on this feature will help me.

---
Closes #626


DirtyF:
We could use a Markdown editor Icon in the toolbar for a start, with an active/inactive state.

Similar to **Bold** when you click on the Markdown icon you switch from WYSIWYG to Raw Markdown view.

toolbar-markdown-icon

jpuri:
@DirtyF : If we plan to keep menu bar i raw mode, we need to implement each button also for raw mode, for instance pressing `B` in raw mode should bold text, etc.

DirtyF:
@jpuri I can see that coming later but for now raw mode would be just "let me edit Markdown source", so if all buttons are disabled in Markdown mode for now, it's fine for a start.

jpuri:
From code perspective not keeping toolbar altogether is convenient, keeping it disabled will still have more complexity.

spbyrne:
I think putting it in the toolbar is likely ideal (disable other controls when raw mode is enabled) but I'm going to mock up some options to see if an alternative might work. I don't think the markdown icon is very obvious vs. a word like 'raw' since not everyone is familiar with markdown.

jpuri:
@spbyrne : I can work to create a disabled toolbar option for now, we can enable them as we add implementation for the buttons.

It will take more development time, but that is ok if this is the planned design.

spbyrne:
Here's what both a text and an icon toggle in the toolbar would look like:

Screen Shot 2020-05-19 at 10 18 24 AM

Here's two options outside the toolbar:

Screen Shot 2020-05-19 at 10 27 13 AM

spbyrne:
I think we should definitely opt for the in-toolbar option.

jpuri:
Hey @spbyrne : that looks good 👍 
I will start working on option-1 in-toolbar, keeping buttons disabled for now.

DirtyF:
I agree an icon alone is never easy to understand but we have tooltips coming now, so we should be good.

spbyrne:
@DirtyF Good point! I've added the markdown icon for this purpose. This will be consistent with other toolbar options.

packages/react-tinacms-editor/src/components/RawModeEditor/index.tsx
```diff
@@ -0,0 +1,91 @@
...
+        <MarkdownEditor value={value} onChange={handleChange} />
+      ) : (
+        <WysiwygEditor
+          input={{
+            value,
+            onChange: handleChange,
+          }}
+          plugins={plugins}
+          sticky={sticky}
+          format="markdown"
```
ncphillips, line 84: The format should be passed in. This could be `html` as well.

jpuri: @ncphillips : is it planned to support raw HTML editing also ? I will prefer to create a different component for it and keeping this for just markdown.

Toolbar implementation for these modes: wysiwyg, markdown and html will be very different.

ncphillips: Right, nevermind then. 

When I was originally thinking of this feature I was comparing it to the Forestry raw text editor. It is a codemirror editor that support any text format at all and doesn't have a formatting menu.

DirtyF:
Getting a webpack error when trying to run the Gatsby demo:

```
 ERROR #98123  WEBPACK

Generating SSR bundle failed

Can't resolve './react-tinacms-editor.cjs.development.js' in '/Users/frank/tina/tinacms/packages/react-tinacms-editor/dist'

File: ../react-tinacms-editor/dist/index.js
```

DirtyF:
I couldn't test it in the demo, just in a storybook and yeah it still needs some love from our beloved designer.

jpuri:
Hey @spbyrne : I did some improvements to it, can you also take a look as you get time.

spbyrne:
raw mode toggle

Here's a preview of the updates to the toggle. The gif compression removes the grey background on the active markdown button, and the Tina css properties aren't available in storybook. In a regular site setting it should have a grey background & blue fill when active, which the MenuButton already provides.

I find it funny we have two identical components, one that enables markdown mode, and one that enables wysiwyg mode. When you click one button it's replaced with the other. A single toggle button makes a lot more sense to me but I don't have a good enough understanding of how this is set up so I assume it has to be done this way.

jpuri:
@DirtyF , @ncphillips, @spbyrne  : code changes are good to be merged as initial version of raw mode toggle. We should plan these 2 tasks more:

1. Unit test coverage of new components added.
2. Enabling menu options by implementing the functionality.

Issue #347, opened 2019-10-22
'Tags' field plugin


gabelluardo:
## Summary

Hi, first thank you for this pretty tool, I love it.
I had a trouble with Gatsby and group-list and I think would be great if there was a Field that could parse markdown list, in particular I have an MDX file with this header: 

```
title: Title
description: Description
tags:
  - tag1
  - tag2
  - tag3
```

I've tried to parse tags with group-list in this way:

```
fields: [
    {
      label: 'Tags',
      name: 'rawFrontmatter.tags',
      component: 'group-list',
      itemProps: (item) => ({
        label: item,
      }),
      defaultItem: 'new-tag',
      fields: [
        { component: 'text' }
      ],
    },
  ...
```

But seems this is not the proper way to use group-list, I've also tried to add a `name` option but nothing.

I used to do that with NetlifyCMS with this configuration: 

```
fields:
      - { name: tags, label: Tags, widget: list }
```

### Basic example

I think that something like this could be much more easy to use:

```
fields: [
    {
      label: 'Tags',
      name: 'rawFrontmatter.tags',
      component: 'list',
      fields: [
        { name: 'tag', component: 'text' }
      ],
    },
  ...
```

### Motivation

This tool is great I want use it and I would see more people use it.

ncphillips:
Hey @gabelluardo! Thanks for the request. I changed the title because what you're asking for is not specific to any data format (i.e. mdx, json, yaml)

PR #1146, opened 2020-05-19
Tags support

DaniTulp:
Adresses #347. I have taken some liberties with the styling but I've tried to use the @tinacms/styles. The way it works right now:

- No duplicates allowed (maybe make that configurable)
- When a Space is pressed, the tag gets created and the input gets cleared (maybe make this configurable as well or a different default)

Let me know what you think!

Screenshot 2020-05-19 at 09 08 41


spbyrne:
Hey @DaniTulp this looks great! I'm going to approve and merge it to a branch called `tags-field` so that we can continue to work on this before merging to master.

DaniTulp:
Sounds great! Let me know if I can help out with anything

PR #1150, opened 2020-05-19
fix: we weren't handling time formatting in any way

### LINKED ISSUE #1144: Date component with timeFormat doesn't allow the time to change
## Description

Enabling time in the date component (timeFormat: true or timeFormat: "HH:mm ZZ"), displays the time and shows the time editor, but clicking on the arrows changes the time and immediately reverts it back.

### Steps to reproduce

In the Tina-starter-grande project, in src/templates/post.js, add dateFormat: true, timeFormat: true to the date field. Navigate to http://localhost:8000/blog/introducing-tina-grande, click the edit icon, click the date and select the time. Attempting to change the time with the arrows will cause the number to briefly change and then immediately go back to its previous value.

### Expected result

Time should change when the arrows are clicked

### Actual result

User is unable to change the time.

#### Tinacms Package Versions

```text
├─ @tinacms/alerts@0.1.2
├─ @tinacms/api-git@0.7.0
├─ @tinacms/core@0.7.4
├─ @tinacms/fields@0.10.1
│  ├─ @tinacms/icons@^0.10.0
│  ├─ @tinacms/styles@^0.4.2
├─ @tinacms/form-builder@0.3.2
│  ├─ @tinacms/forms@^0.4.1
│  ├─ @tinacms/react-core@^0.2.11
├─ @tinacms/forms@0.4.1
├─ @tinacms/git-client@0.6.7
├─ @tinacms/icons@0.10.0
├─ @tinacms/media@0.1.0
├─ @tinacms/react-alerts@0.1.3
├─ @tinacms/react-core@0.2.11
│  ├─ @tinacms/core@^0.7.4
│  └─ @tinacms/forms@^0.4.1
├─ @tinacms/react-forms@0.1.3
│  ├─ @tinacms/form-builder@^0.3.2
├─ @tinacms/react-modals@0.1.3
├─ @tinacms/react-screens@0.1.2
├─ @tinacms/react-sidebar@0.1.3
│  ├─ @tinacms/core@^0.7.4
│  ├─ @tinacms/forms@^0.4.1
├─ @tinacms/react-toolbar@0.1.1
│  ├─ @tinacms/core@^0.7.3
│  ├─ @tinacms/fields@^0.10.1
│  ├─ @tinacms/forms@^0.4.1
├─ @tinacms/styles@0.4.2
├─ gatsby-plugin-tinacms@0.1.35
│  └─ tinacms@^0.19.1
├─ gatsby-tinacms-git@0.4.9
│  ├─ @tinacms/api-git@^0.7.0
│  └─ @tinacms/git-client@^0.6.7
├─ gatsby-tinacms-json@0.8.20
├─ gatsby-tinacms-remark@0.7.15
├─ react-tinacms-editor@0.5.1
│  ├─ @tinacms/core@^0.7.4
│  ├─ @tinacms/fields@^0.10.1
│  ├─ @tinacms/icons@^0.10.0
│  ├─ @tinacms/react-forms@0.1.2-alpha.0
│  ├─ @tinacms/react-forms@0.1.2-alpha.0
│  │  ├─ @tinacms/form-builder@^0.3.2-alpha.0
│  ├─ @tinacms/styles@^0.4.2
├─ react-tinacms-inline@0.4.6
├─ tinacms@0.19.1
│  ├─ @tinacms/alerts@^0.1.2
│  ├─ @tinacms/core@^0.7.4
│  ├─ @tinacms/fields@^0.10.1
│  ├─ @tinacms/form-builder@^0.3.2
│  ├─ @tinacms/forms@^0.4.1
│  ├─ @tinacms/icons@^0.10.0
│  ├─ @tinacms/media@^0.1.0
│  ├─ @tinacms/react-alerts@^0.1.3
│  ├─ @tinacms/react-core@^0.2.11
│  ├─ @tinacms/react-forms@^0.1.3
│  ├─ @tinacms/react-modals@^0.1.3
│  ├─ @tinacms/react-screens@^0.1.2
│  ├─ @tinacms/react-sidebar@^0.1.3
│  ├─ @tinacms/react-toolbar@^0.1.1
│  ├─ @tinacms/styles@^0.4.2
│  └─ react-tinacms-editor@^0.5.1

```

#### Tools

| Tool    | Version |
| ------- | ------- |
| browser |   Brave 1.8.96, Chrome 81.0.4044.138      |
| os      |   macOS Catalina      |
| node    |   v12.16.1      |
| yarn     |     1.22.4    |

mittonface:
Fixes #1144


packages/tinacms/src/plugins/fields/dateFormat.ts
```diff
@@ -45,13 +45,21 @@ export const parse = (
...
   field: DatetimepickerProps
 ) => {
-  const dateFormat =
-    typeof field.dateFormat === 'string'
-      ? field.dateFormat
-      : DEFAULT_DATE_DISPLAY_FORMAT
-  if (typeof val === 'string') {
-    const date = moment(val, dateFormat)
-    return date.isValid() ? date.toDate() : val
+  return format(val, _name, field)
```
mittonface, line 48: This method is _actually_ supposed to return a Date, not just the formatted string.

ncphillips:
IIRC we intentionally weren't handling time when `timeFormat` was undefined. I think that's so it could work as both a `date` field and a `datetime` field. 

Maybe @DirtyF or @jamespohalloran remembers?

packages/tinacms/src/plugins/fields/dateFormat.ts
ncphillips: `react-datetime` must be okay with using a `string`

mittonface:
I just thought of something I need to check. So I will look into it and reopen.

mittonface:
This has been a journey. 

* With my initial PR I hadn't considered how dates were actually being saved in the front-matter. They were being saved in the user-provided format, which might not be the worst but could end up painting the user into a corner if they ever decided they wanted to change formats.
* I had to make some adjustments for timezones. I went ahead and made the assumption that for now we always want to work in UTC. The best assumption we can make right now is that when a user inputs 8:00AM they want everyone else to also see 8:00AM. 
* There was seemingly _another_ bug where after saving the date would get inserted into the frontmatter as `2020-08-02T12:12:12Z` instead of `'2020-08-02T12:12:12Z'`, this was causing a parsing error. To fix this I made the `parse` function return a string in the default datetime format.

It seems to be all working now.


May-20-2020 11-25-54

kendallstrautman:
I'm still seeing some [weird behavior](https://www.loom.com/share/623edbd1f8af48238cbba3a08f24362b).

mittonface:
I'll dig in a bit more, maybe give it a try in another project other than tinacms.org. Thanks!

mittonface:
- I was able to reproduce the issue by not passing anything to `dateFormat` while also specifying a `timeFormat`. `dateFormat.ts` specifies a format to use in this case, but that format was not actually being passed into the `ReactDateTime` component. 
- Passing the default `dateFormat` to `ReactDateTime` seems to fix the issue and it will just get overwritten the same way `timeFormat={false}` is being overwritten. 
- It _might_ be worth it to make this field type more flexible, because at the moment it can't be configured to act exclusively as a time-picker (though that doesn't seem like a very desirable field type, and this limitation _is_ reflected in the docs).

mittonface:
@kendallstrautman @DirtyF  let me know if you can take another look at this! 

I think there is still some funkiness that people can get themselves into if they provide weird formats (i.e., specifying timeFormats w/ incomplete dateFormats, specifying dateFormat=false) but I don't think it's worth it to fix those or even trying to document all the ways people could get themselves into trouble.

PR #1152, opened 2020-05-19
Feat: Tags Field

### LINKED ISSUE #347: 'Tags' field plugin

spbyrne:
Thanks to @DaniTulp we have a simple tags field, [see original PR here](https://github.com/tinacms/tinacms/pull/1146).

I've tweaked the styling, swapped the trash icon for an 'x' icon (just feels more appropriate) and I've added 'enter' to the list of keys that trigger a tag gets added when you're in the text field.

Screen Shot 2020-05-19 at 4 38 57 PM

Closes #347 

---
### Todo
- [x] Docs PR Ready


packages/@tinacms/fields/src/plugins/TagsFieldPlugin.tsx
```diff
@@ -0,0 +1,139 @@
...
+    text-overflow: ellipsis;
+    overflow: hidden;
+  }
+`
+export const TagsFieldPlugin = {
+  name: 'tags',
+  Component: TagsField,
+  validate(value: any, values: any, meta: any, field: any) {
+    if (field.required && !value) return 'Required'
+  },
```
ncphillips, line 137:
```suggestion
```

packages/@tinacms/fields/src/plugins/TagsFieldPlugin.tsx
```diff
@@ -0,0 +1,139 @@
...
+    <>
+      <BaseTextField
+        value={value}
+        onChange={event => setValue(event.target.value)}
+        placeholder={field.placeholder}
+        onKeyPress={event => {
+          if (event.key === ' ' || event.key === 'Enter') {
+            event.preventDefault()
+            addTag(value)
+            return
```
ncphillips, line 52:
```suggestion
```

packages/@tinacms/fields/src/plugins/TagsFieldPlugin.tsx
```diff
@@ -0,0 +1,139 @@
...
+  const [value, setValue] = React.useState<string>('')
+  const addTag = React.useCallback(
+    (tag: string) => {
+      if (form.getFieldState(field.name)?.value?.includes(tag)) {
+        return
+      }
+      form.mutators.insert(field.name, 0, tag)
+      setValue('')
+    },
+    [form, field]
```
ncphillips, line 39:
```suggestion
    [form, field.name]
```

kendallstrautman:
Would be great to get some rudimentary documentation on this.

kendallstrautman:
[Some docs](https://github.com/tinacms/tinacms.org/pull/424)

packages/@tinacms/fields/src/plugins/TagsFieldPlugin.tsx
spbyrne: Good catch 👍

packages/@tinacms/fields/src/plugins/TagsFieldPlugin.tsx
spbyrne: Removed

packages/@tinacms/fields/src/plugins/TagsFieldPlugin.tsx
spbyrne: Good call, removed.

PR #1158, opened 2020-05-21
Fix link text in next-tinacms-github README

grncdr:
Very small docs fix. I was confused because the link text is the page you're already on 😄


ncphillips:
@all-contributors please add @grncdr for docs

PR #1160, opened 2020-05-21
Inline Group

ncphillips:
- [x] Refactor `InlineBlockContext` to be a generic `InlineFieldContext`
- [x] Remove `BlockField` so there's only one set of InlineField
  - [x] Text
  - [x] Textarea
  - [x] Image
- [x] Add a new `InlineGroup`

### Misc. Fields Editable in Modal
A list of fields is provided. When hovering over the group an **Edit** icon will appear. That will open a modal that renders out the form with the given fields. All the fields are assumed to be referenced from the root of the form. This is an artificial grouping that doesn't reflect that data structure.

```tsx
// Data
{
  title: "Hello World",
  createdAt: new Date(),
  updatedAt: new Date(),
  description: "This is a new world."
}

// Component
<InlineGroup
  fields={[
    { name: "createdAt", component: "date" },
    { name: "updatedAt", component: "date" },
  ]}
>
  Created At: {post.createdAt}
  Updated At: {post.updatedAt}
</InlineGroup>
```

### Author Editable Inline
In this example the `name` is set to `"author"`. This makes it so all inline fields within the group are referenced from inside the `"author"`

```tsx
{
  title: "My New Post",
  author: {
    firstName: "Bob",
    lastName: "Ross",
  }
}

<InlineTextarea name="title" />
<InlineGroup name="author">
  First Name: <InlineTextarea name="firstName" />   // author.firstName
  Last Name: <InlineTextarea name="lastName" />    // author.lastName
</InlineGroup>
```

### Author Editable from Modal

```tsx
{
  author: {
    firstName: "Bob",
    lastName: "Ross",
  }
}

<InlineGroup name="author"
  fields={[
    { name: "firstName", component: "text" },
    { name: "lastName", component: "text" },
  ]}
>
  Written by {post.author.firstName}
</InlineGroup>
```

### An Image Field Editable from Modal

```tsx
const post = {
  image: {
    src: "Bob",
    alt: "Ross",
  }
}

<InlineGroup
  fields={[
    { name: "image.src", component: "image" },
    { name: "image.alt", component: "textarea" },
  ]}
>
  <img {...post.image} >
</InlineGroup>

```


spbyrne:
Screen Shot 2020-05-26 at 3 12 05 PM

I've updated the inline group UI to mirror the way blocks look, minus the additional actions. The focus is managed in a similar way, so it doesn't use the old focus ring which relies on input focus (doesn't always work). The way the settings cog is inset from the corner looks odd but it's consistent with blocks. This UI will likely be updated very soon.

spbyrne:
In this example I'm passing 'focusRing={false}` to the inline textarea field to prevent a double focus wrapper.

PR #1166, opened 2020-05-25
feat: creating plugin for mode toggle menu component

jpuri:
Create plugin for raw mode toggle menu option.


packages/react-tinacms-editor/src/components/WysiwygEditor/index.tsx
```diff
@@ -19,10 +19,9 @@ limitations under the License.
 import * as React from 'react'
 import { useEffect, useRef, useState } from 'react'
 import styled from 'styled-components'
-import { Plugin } from '@tinacms/core'
```
jpuri, line 22: Old plugin in props was not being used I am not sure what was intent for it.

Currently I am using it to pass custom plugins to editor.

DirtyF:
I get an error when I try to build the editor package from Lerna:

```
lerna ERR! npm run build stderr:
(typescript) Error: /Users/frank/tina/tinacms/packages/react-tinacms-editor/src/tinacms-inline/inline-wysiwyg.tsx(20,29): semantic error TS2307: Cannot find module 'react-tinacms-inline'.
Error: /Users/frank/tina/tinacms/packages/react-tinacms-editor/src/tinacms-inline/inline-wysiwyg.tsx(20,29): semantic error TS2307: Cannot find module 'react-tinacms-inline'.
```

ncphillips:
@DirtyF make sure to run `npm run bs`

DirtyF:
@ncphillips yeah I could compile, I still don't see the raw editor in the demo because I guess the demo needs to be updated accordingly. So I'll defer QA in this.

jpuri:
@DirtyF : yep demo needs to be updated for this, I will do that tomorrow.

jpuri:
Hey @DirtyF : raw mode toggle is not added to demo-gatsby.

packages/react-tinacms-editor/src/components/EditorModeMenu/index.test.tsx
```diff
@@ -0,0 +1,33 @@
...
+jest.mock('../../context/editorState')
+
+describe('EditorModeMenu', () => {
+  it('should render a menu option', () => {
+    const { findByTestId } = render(
+      <EditorModeMenu toggleEditorMode={() => {}} />
+    )
+    expect(findByTestId('markdown-toggle')).toBeDefined()
+  })
+})
```
jpuri, line 33: These test cases were intended for next PR, but anyways no harm here :)

ncphillips:
Hmm the `demo-gatsby` should not have to be updated in order for the raw editor to work. 

The components can be organized internally in whatever way, but the only exports for the package should be:

* `Wysiwyg`
* `InlineWysiwyg`
* `MarkdownFieldPlugin`
* `HtmlFieldPlugin`

That's to say, clients of this package shouldn't have to decide between `InlineRawModeWysiywg` and `InlineWysiwyg`.

If the raw mode is only available when using `markdown` then that should be figured out within the main exported component.

packages/react-tinacms-editor/src/types.ts
```diff
@@ -50,3 +51,15 @@ export type NodeViews = {
 export interface Command {
   (state: EditorState, ...options: any[]): void
 }
+
+export interface Plugin {
+  name: string
+  WysiwygMenu: ElementType
```
ncphillips, line 57: Could we switch this to `MenuItem`? The `Wsyiwyg` part feels redundant.

```suggestion
  MenuItem: ElementType
```

jpuri:
Hey @ncphillips : actually that was reason I was thinking of not creating new component `RawModeEditor` but include this behaviour within current `WysiwygEditor`.

Let me change what gets exported from the package and keep it same as it is right now, and each time mode is `markdown` we render `RawModeEditor` by default.

ncphillips:
I think `RawModeEditor` should always be rendered, not just when `format="markdown"`. Instead the `RawModeEditor` would only the toggle plugin to the `Wysiwyg` if format="markdown". 

Then you could make the change without moving that behaviour around:

* Make `RawModeEditorProps` the same as `WysiwygProps
* Rename `Wysiwyg` to `ProsemirroEditor` 
* Rename `RawModeEditor` to `Wysiwyg`

The benefit of this approach being the prosemirror code and raw-text code are isolated from each other and switching between them is handled outside of them both.

packages/react-tinacms-editor/src/types.ts
ncphillips: actually, the RFC says `menu`

jpuri: I was thinking of possible support for different menus like `WysiwygMenu` for wysiwyg mode, `MarkdownMenu` for raw mode.

But I think this will be ok:

**MenuItem - wysiwyg mode
MarkdownMenuItem - markdown mode**

Markdown menu support is future think.

ncphillips: Right. An alternative could be for the `MenuItem` to be passed the `mode` and make the decision from there

jpuri:
@ncphillips : I have updated the PR with changes as discussed, I plan to do some more cleanup tomorrow.

packages/react-tinacms-editor/src/components/Wysiwyg/index.tsx
```diff
@@ -57,16 +57,15 @@ export const RawModeEditor = ({
           toggleEditorMode={() => setMode('wysiwyg')}
         />
       ) : (
-        <WysiwygEditor
+        <ProsemirrorEditor
           input={{
             value,
             onChange: handleChange,
           }}
```
ncphillips, line 64: Do you remember the reason for the indirection of the `value` and `onChange`?

I'm pretty sure that keeping the value in state will introduce the bug. 

```suggestion
          input={input}
```

The editor should not be keeping it's own state, that's the job of the form object outside of it. By keeping track of the value here the editor component can get out of sync with the form's data.

jpuri: Recent values is needed here so that it can be kept in sync between markdown and wysiwyg editors.

Values here is not creating issues, reason being that editors in library behave as un-controlled components. Making editors controlled can be bit expensive as it requires conversion from markdown to prosemirror json for each change in content.

jpuri:
Hey @ncphillips : PR is updated with hybrid controlled behaviour as discussed.

jpuri:
Fixes https://github.com/tinacms/tinacms/issues/1179

ncphillips:
Awesome! It is indeed working but it's a bit too narrow on what "losing focus" means. It should "lose focus" when the window loses focus.

[Here's a loom video showing what I'm talking about](https://www.loom.com/share/a39835593f6e4fea87cb9268935b593c)

jpuri:
> Awesome! It is indeed working but it's a bit too narrow on what "losing focus" means. It should "lose focus" when the window loses focus.
> 
> [Here's a loom video showing what I'm talking about](https://www.loom.com/share/a39835593f6e4fea87cb9268935b593c)

@ncphillips : I see, I will fix this. I used prosemirror `view.hasFocus()` for check, I need change this.

jpuri:
@ncphillips : PR is updated in both raw and wysiwyg mode it will update content if either browser window is not focused or within the window editor is not focused.

DirtyF:
Gave it a quick test, this is functional that's cool.

I just noticed a few UI glitches that can be adressed in another PR: https://www.loom.com/share/530241871e6043c9aee5fc84dc788999

ncphillips:
Works great! 

I just noticed one issue, it looks like the styling was not being applied for the sidebar:

image

I just pushed up a fix for that though 👍 

I'll approve once the checks pass

PR #1170, opened 2020-05-26
feat(cms): add new first-class enabled state

ncphillips:
this is THE source of truth for whether or not the CMS content should be editable

_Note: this is a prototype_


dwalkr:
should CMS disabled state should override sidebar/toolbar visibility?

ncphillips:
Yes I would assume so. Those would be used to choose your UI not turn off editing

Issue #1140, opened 2020-05-14
Tina shouldn't write to disk before 'real' edits are made


kendallstrautman:
## Summary

Relating to #1098 — Tina makes formatting changes to source files even if no form changes have been make. It would be a nice enhancement if Tina did not write to disk until an initial form change is instigated. 

### Motivation

To not let the form get dirty until edits are made by the user.

PR #1174, opened 2020-05-26
feat: form({ onChange })

ncphillips:
This PR makes it possible to pass an `onChange` callback to a form construction.

This change makes it possible to override default behaviour of `useGitForm`:

```js
const cms = useCMS()

useRemarkForm(node, {
  onSubmit(data) {
    return cms.api.git.writeToDisk({
      fileRelativePath: values.fileRelativePath,
      content: toMarkdown(values),
    })
  }, 
  onChange: null,
})
```

This makes it so:

1. The form does not write to disk `onChange` 
1. Pressing `Save` writes to disk

Effectively disabling the "live" preview mode which causes problems for users with slow builds.

### DIFF
--- a/packages/@tinacms/forms/src/form.ts
+++ b/packages/@tinacms/forms/src/form.ts
@@ -17,7 +17,7 @@ limitations under the License.
 */
 
 import arrayMutators from 'final-form-arrays'
-import { FormApi, createForm, Config } from 'final-form'
+import { FormApi, createForm, Config, FormState } from 'final-form'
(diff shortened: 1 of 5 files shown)


kendallstrautman:
Addresses #1140

ncphillips:
Really? Did you test it?

As far as I can tell, this is essentially a refactor

kendallstrautman:
lol didn't test, just saw the description and got excited to link the issue 🤠

PR #1184, opened 2020-05-28
Feat: Adds StyleProvider to handle font loader 

kendallstrautman:
_Merging into the **dynamic font loading PR**_ @spbyrne 

### Renames some things in `@tinacms/styles`
- `GlobalStyles` 👉 `TinaGlobalStyles`
- `TinaResetStyles` 👉 `tina_reset_styles` (since this is just plain css, the name shouldn't look like a component)
- `TinaReset` 👉 `StyleReset`

---

### Introduces `StyleProvider` component into `TinaProvider`. 

`TinaProvider` now accepts a `styled : boolean` prop that gets passed to `StyledProvider` which then renders the `FontLoader` & `TinaGlobalStyles` or returns null. 

```js
interface StyleProviderProps {
  active: boolean
}

export function StyleProvider({ active }: StyleProviderProps) {
  if (!active) return null

  return (
    <>
      <FontLoader />
      <TinaGlobalStyles />
    </>
  )
}
```

This allows people to prevent styles (and the inter font) from being loaded when they don't need or want it, but it will default to `true` if no `styled` prop is passed.
```js
// Example: Tina styles and font don't load in production
<TinaProvider cms={this.cms} styled={process.env.NODE_ENV === 'production'}>
  <Component {...pageProps} />
</TinaProvider>
```

This also allows people to pick and choose components from `@tinacms/styles` to load. For example, they may not want to load the 'Inter' font, so they could do something like this: 
```js
import { TinaGlobalStyles } from '@tinacms/styles'

// Example: this config doesn't load external 'Inter' Font
<TinaProvider cms={this.cms} styled={false}>
  <TinaGlobalStyles />
  <Component {...pageProps} />
</TinaProvider>
```


kendallstrautman:
Apologies for the commit history, this merges master as well 🤦‍♀️

ncphillips:
I think the `TinaProvider` could adhere to the new `cms.enabled` concepts:

```
 <StyleProvider active={cms.enabled && styled} />
```

kendallstrautman:
Yeah that sounds great!

spbyrne:
Everything looks good to me. I think this is a great way to handle styling.

PR #1189, opened 2020-05-29
fix: circular dependencies

jpuri:
Fix circular dependencies.


jpuri:
Fixes https://github.com/tinacms/tinacms/issues/1177

PR #1202, opened 2020-06-02
fix: github client fetch file works on any branch

logan-anderson:
Added ref to the fetch file method in the github client. before this it would only fetch from the default branch.


ncphillips:
Is the `branch` being passed in `data` not required?

logan-anderson:
@ncphillips I am not sure if branch is required or not. The reason I left it in is because I thought the [proxy request ](https://github.com/tinacms/tinacms/blob/672fd9a16b265a4311541ca4204573e3f5c51f51/packages/next-tinacms-github/src/github/proxy.ts#L22)might be using it but in terms on the github API I dont think it is needed as it is passed in the URL.

PR #1203, opened 2020-06-02
Style: Revised Blocks Controls

spbyrne:
- Changes "Add Block" button to "+" buttons positioned at the top & bottom of the focus ring.
- Updates the block actions to reflect the design of the WYSIWYG toolbar.
- Replaces the settings cog with a pencil icon
- Updates inline groups to use the new components

Screen Shot 2020-06-02 at 11 59 43 AM


spbyrne:
I just need to update the empty state add block menu.

packages/react-tinacms-inline/src/blocks/add-block-menu.tsx
```diff
@@ -19,18 +19,45 @@ limitations under the License.
...
+  const addBlockButtonRef = React.useRef<HTMLButtonElement>(null)
+  const [openTop, setOpenTop] = React.useState(false)
 
   const handleOpenBlockMenu = () => {
+    const addBlockButtonElem = addBlockButtonRef.current
+
+    if (addBlockButtonElem !== null) {
+      const menuBounding = addBlockButtonElem.getBoundingClientRect()
+      const halfWindowHeight =
+        (window.innerHeight || document.documentElement.clientHeight) / 2
```
kendallstrautman, line 51: I'm guessing this is related to the error. We should check whether `window` or `document` is available before running this. Maybe put it in a componentDidMount-style useEffect?

packages/react-tinacms-inline/src/blocks/add-block-menu.tsx
```diff
@@ -86,24 +132,59 @@ const AddBlockButton = styled(Button)<AddMenuProps>`
...
+  offset?: number
+  position?: 'top' | 'bottom'
+  isOpen: boolean
+}
+
+const AddBlockWrapper = styled.div<AddBlockWrapperProps>(
+  p => css`
+  position: absolute;
+  left: 50%;
+  z-index: calc(var(--tina-z-index-1) - ${p.index ? p.index : 0});
```
kendallstrautman, line 158: Should be be more specific here? 
```suggestion
  z-index: calc(var(--tina-z-index-1) - ${p.index !== undefined ? p.index : 0});
```

kendallstrautman: or is it checking that its not 0?

packages/react-tinacms-inline/src/blocks/add-block-menu.tsx
spbyrne: Looks like it's from the dismissible component itself. You can provide it with a document if you're inside an iFrame (I believe that's what it's for?) but in this case it didn't seem necessary. Strange.

spbyrne: Yeah I can't find another solution, I just removed dismissible for now. It's functionally almost identical.

PR #1204, opened 2020-06-02
Inline settings — cancel & confirm

### LINKED ISSUE #1003: Block Settings Modal Exit Options
## Description

The block settings modal doesn't provide a way to confirm your changes, and the 'cancel' button doesn't undo any changes you've made.

### Steps to reproduce

On TinaCMS.org, enter edit mode on the homepage. Under the demo video, there's three blocks with settings cogs. Click the cog to open the settings modal.

### Expected result

I'd expect there's a way to confirm my changes to dismiss the modal, and I'd expect that clicking 'cancel' would undo my changes.

### Actual result

No 'save' or 'confirm' option is presented, so you have to click the 'x' close button or 'cancel' to get out of the modal.

Screen Shot 2020-04-14 at 10 46 08 AM

kendallstrautman:
Closes #1003 

## What this does:

- Hitting the 'Cancel' button resets any changes made in the modal form to the initial values when the settings modal was opened
- Adds a confirm button that essentially just closes the modal but makes for better UX. 

Screen Shot 2020-06-02 at 1 29 31 PM

My questions @spbyrne : 

- Do we add an alert or confirmation to this action? - Waiting until it's needed
- Should we call this 'reset' or does 'cancel' feel good? 👉 Cancel
- ~Should the type be red to indicate a delete action?~ Nay
- Should 'cancel' also close the modal? ✅

### DIFF
--- a/packages/@tinacms/styles/src/Button.tsx
+++ b/packages/@tinacms/styles/src/Button.tsx
@@ -24,6 +24,7 @@ export interface ButtonProps {
   grow?: boolean
   open?: boolean
   busy?: boolean
+  disabled?: boolean
 }
 
 export const Button = styled.button<ButtonProps>`
--- a/packages/demo-next/pages/info.js
+++ b/packages/demo-next/pages/info.js
@@ -133,6 +133,7 @@ function Info(props) {
               flex-direction: column;
               text-align: center;
               padding: 3rem;
+              margin-top: 4rem;
             }
           `}
         </style>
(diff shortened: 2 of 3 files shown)


spbyrne:
- Personally I like 'cancel' given it's in a modal and that feels fairly standard. We're using 'discard' in the toolbar but it doesn't feel appropriate here, IMO.
- I don't think red text is necessary.
- I think it should also close the modal.
- Not sure about a confirmation.

kendallstrautman:
Let's just keep it simple and add an alert later if we feel its necessary

kendallstrautman:
I'm just gonna add the confirm button on this PR too

kendallstrautman:
Right now this 'Confirm' button will only show up if the form is dirty...I don't love this. Maybe it is in a disabled stated until the form is dirty?

kendallstrautman:
@spbyrne I'll let you take it from here to make any style changes you see fit! Feel free to push to this branch

PR #1205, opened 2020-06-02
Inline form focus

ncphillips:
This PR makes it so only one block/group can have focus at a time

### Clicking for Focus
Focus is "outside-in". For example with this structure

```
<A>
  <B>
    <C />
  </B>
  <D>
    <E />
  </D>
</A>
```

 Clicking on element `<C>` does not immediately put the focus on `<C>`. 

* First it will put focus on the entire group `<A>`
* A second click will put the focuse on `<B>`
* The third click will put the focus on `<C>
* A fourth click on `<E>` will shift the focus to `<D>`.
* If at any point somewhere outside of the form entirely is clicked then all fields will lose focus.

### Hovering

The same pattern applies to hovering.

When there is no focus, only `<A>` will be outlined when hovering over `<C>`

When focus is on `<C>` then the hover pattern doesn't show for `<A>`

 ## TODO

- [x] This change currently prevents clicking anything


spbyrne:
This is looking good. It seems to expose a positioning bug for the blocks controls when the offset is set to 0, but I'll address that when I work on the blocks controls positioning.

kendallstrautman:
I'm seeing an issue on the _Info_ page, where if you manually add the controls, they don't render and you can't focus / edit on the child fields. Its the second group on that page, the one with the image and the wysiwyg.

kendallstrautman:
Seeing an issue where if you don't provide a name to `InlineGroup`, focus is never set and controls don't render.

spbyrne:
I think 'name' is required now.

kendallstrautman:
It should be optional for groups

PR #1208, opened 2020-06-03
Column Block Support

spbyrne:
- Adds `direction` property to `InlineBlocks` component which accepts `row` or `column` (defaults to column)
- Repositions add block buttons for row blocks
- Replaces up/down reorder arrows with left/right for row blocks
- Adds chevron left & right icons


packages/demo-next/pages/nesting.tsx
```diff
@@ -186,6 +186,7 @@ const ROW = {
         <div style={{ display: 'flex', flexDirection: 'row' }}>
           <InlineBlocks
             name="items"
+            layout="columns"
```
ncphillips, line 189: This feels backwards to me, I would expect it to be `layout: row` (singular), which matches the CSS of the parent div `flexDirection: row`.

spbyrne: Ah, that's a good point. I'll make that change.

spbyrne: I'm also going to change 'layout' to 'direction' to make it more clear.

spbyrne: Changed.

PR #1210, opened 2020-06-03
Add Inset Option For Blocks Controls

spbyrne:
This allows you to set `insetControls={true}` on BlocksControls & InlineGroups. I'm not sure if it would make sense to also have the ability to set this at the `InlineBlocks` or not since it depends on how the individual block is set up.

Screen Shot 2020-06-03 at 2 59 39 PM


kendallstrautman:
I feel like it makes sense to just have this on the controls for now.

PR #1211, opened 2020-06-03
Adds focus ring prop to inline groups and blocks controls

kendallstrautman:
```js
<InlineGroup
  name="frontmatter"
  fields={[
    { label: 'Name', name: 'name', component: 'text' },
    { label: 'Hometown', name: 'hometown', component: 'text' },
  ]}
  offset={10}
  borderRadius={5}
  focusRing={false}
>
 ...
</InlineGroup>
```
or 

```js
<BlocksControls index={index} focusRing={false}>
  ...
</BlocksControls>
```

### DIFF
--- a/packages/demo-next/pages/info.js
+++ b/packages/demo-next/pages/info.js
@@ -73,8 +73,9 @@ function Info(props) {
                 { label: 'Name', name: 'name', component: 'text' },
                 { label: 'Hometown', name: 'hometown', component: 'text' },
               ]}
-              offset={10}
-              borderRadius={5}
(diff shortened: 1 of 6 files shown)


ncphillips:
What if it was like:
```
<InlineGroup
  name="frontmatter"
  focusRing={{
    borderRadius: 5,
    offset: 0,
  }}
>
 ...
</InlineGroup>
```

and then

```
<InlineGroup
  name="frontmatter"
  focusRing={false}
>
 ...
</InlineGroup>
```

Issue #1155, opened 2020-05-20
`loadInitialValues` doesn't work with fallback values


dwalkr:
## Description

When using `loadInitialValues` to populate form values asynchronously, its return value only updates the form values when `initialValues` and `watch.values` are not set.

I'm unsure if this is intended behavior, but it becomes very unwieldy to try and use `loadInitialValues` without fallback values. Based on this comment inside the `useForm` hook, I'm guessing this is not intended behavior:

```
  /**
   * `initialValues` will be usually be undefined if `loadInitialValues` is used.
   *
   * If the form helper is using `watch.values`, which would contain
   * the current state of the form, then we set that to the `initialValues`
   * so the form is initialized with some state.
   *
   * This is beneficial for SSR and will hopefully not be noticeable
   * when editing the site as the actual `initialValues` will be set
   * behind the scenes.
   */
```

### Steps to reproduce

```
const [values, form] = useForm({
  id: 'form',
  label: 'Form',
  fields: [
    {
        name: 'title',
        label: 'title',
        component: 'text'
    }
  ]
  loadInitialValues: async () => {
    return { title: 'Foo' }
  }
}, { values: { title: 'Bar' } })
```

### Expected result

after `loadInitialValues` resolves, title == 'Foo'

### Actual result

after `loadInitialValues` resolves, title == 'Bar'

### Environment

latest (working from monorepo)

## Triage

It appears that the problem is in [Form#updateInitialValues](https://github.com/tinacms/tinacms/blob/master/packages/%40tinacms/forms/src/form.ts#L201), where the form is re-populated with its old values after re-initializing.

ncphillips:
This sounds like expected behaviour to me.

`loadInitialValues` loads the values you see in the `pristine` state

the `values` object sets the values for the `dirty` state

dwalkr:
What would be the expected behavior in this case?

```
const [values, form] = useForm({
  id: 'form',
  label: 'Form',
  fields: [
    {
        name: 'title',
        label: 'title',
        component: 'text'
    }
  ]
  initialValues: { title: 'Bar' },
  loadInitialValues: async () => {
    return { title: 'Foo' }
  }
})
```

ncphillips:
I'm guessing once it was done you would have:
```
values: { title: 'Bar' },
initialValues: { title: 'Foo' }
```
I don't expect people to use both of those

dwalkr:
I'm interested in a way to

1. bootstrap form data asynchronously from an API source
[...]

PR #1220, opened 2020-06-04
feat(form): values can be passed to constructor

### LINKED ISSUE #1219: useForm: add a way to pass fallback values
## Summary

When using `loadInitialValues` with `useForm`, the values returned by the hook are empty until `loadInitialValues` completes.

We should provide a way to pass fallback values to `useForm` that the hook will return until `loadIntialValues` completes.

### Basic example

```
const [values] = useForm({
  //...
  loadInitialValues: async () => { 
    const data = await loadGithubData
    return data
  },
  fallbackValues:  productionData
})
```

### Motivation

When using `loadInitialValues`, users have to perform some defensive programming against the initial null state until the values are loaded. Due to how Tina works, the developer will usually have a copy of "compatible" values (the production data) that could be used until `loadInitialValues` completes. Example of this sort of defensive programming: https://gist.github.com/dwalkr/d97a8495a33ffd766fdb726e0fae2fdf

I think this scenario will be so common to `loadInitialValues` usage that it should be handled within `useForm`

ncphillips:
This means a form can be initialized with both it's initialV and it's current values in one step

### DIFF
--- a/packages/@tinacms/forms/src/form.test.ts
+++ b/packages/@tinacms/forms/src/form.test.ts
@@ -36,6 +36,46 @@ describe('Form', () => {
         new Form(DEFAULTS)
       })
     })
+    describe('with initialValues', () => {
+      it('has initialValues for form.values', () => {
+        const initialValues = { title: 'Hello World' }
+
+        const form = new Form({
+          ...DEFAULTS,
+          initialValues,
+        })
+
+        expect(form.values).toEqual(initialValues)
+      })
+      describe('with values', () => {
+        it('has the values set for form.values', () => {
+          const initialValues = { title: 'Hello World' }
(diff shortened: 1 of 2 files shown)


dwalkr:
my goal with the fallback values is not that the fallback values replace the form state, but that they provide "dummy values" to be returned by `useForm` until `loadInitialValues` resolves

ncphillips:
Ah, well what you just pass both?

```
new Form({
  initialValues: DUMMY_VALUES,
  loadInitialValues: loadRealValues,
})
```

dwalkr:
Doesn't work that way currently: https://github.com/tinacms/tinacms/issues/1155#issuecomment-631544462

But I would be happy if it did!

PR #1221, opened 2020-06-05
feat(ImageFieldPlugin): support async previewSrc

### LINKED ISSUE #1214: Image field: allow for async previewSrc function
## Summary

The image field accepts a `previewSrc` function, allowing developers to derive the URL for displaying an image preview in the Tina UI.

This feature request is to allow developers to pass an async function / promise to `previewSrc`.

### Basic example

```js
useForm({
  fields: [
    {
      name: "header_image",
      label: "Header Image",
      previewSrc: async (formValues: any) => {
        const src = await cms.api.github.getMediaUri(formValues.header_image)
        return src
      }  
    }
  ]
})
```

### Motivation

It's possible to get the download URL of a media object from the GitHub API. While the `raw.githubusercontent` URL can conceivably be pieced together without a network request, when working with private repos, GitHub attaches a token to the download URL to authorize the request and thus we have to get the download URL from the API.

dwalkr:
Peek 2020-06-05 11-40

closes #1214

### DIFF
--- a/packages/@tinacms/fields/src/plugins/ImageFieldPlugin.tsx
+++ b/packages/@tinacms/fields/src/plugins/ImageFieldPlugin.tsx
@@ -39,14 +39,23 @@ export const ImageField = wrapFieldsWithMeta<InputProps, ImageProps>(props => {
   useEffect(() => {
     ;(async () => {
       setSrcIsLoading(true)
-      const imageSrc = await props.field.previewSrc(
-        props.form.getState().values,
-        props
-      )
+      let imageSrc = ''
+      try {
+        imageSrc = await props.field.previewSrc(
+          props.form.getState().values,
+          props
+        )
+      } catch (e) {
+        // @ts-ignore cms.alerts
+        cms.alerts.error(
+          `Failed to generate preview for '${props.field.name}': ${e.message}`
+        )
+      }
       setSrc(imageSrc)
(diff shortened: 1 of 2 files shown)


packages/@tinacms/fields/src/plugins/ImageFieldPlugin.tsx
```diff
@@ -21,22 +21,38 @@ import { wrapFieldsWithMeta } from './wrapFieldWithMeta'
...
+  const [src, setSrc] = useState('')
+  useEffect(() => {
+    ;(async () => {
+      setSrcIsLoading(true)
+      const imageSrc = await props.field.previewSrc(
+        props.form.getState().values,
+        props
+      )
+      setSrc(imageSrc)
+      setSrcIsLoading(false)
```
ncphillips, line 47: Might want to do error handling?

dwalkr:
@ncphillips in its current state, `previewSrc` now only runs when the value of the image field changes. Since `previewSrc` receives other data, they _may_ need to be deps of the `useEffect`, but doing it by default means the `previewSrc` request is re-run every time any form value changes.

I wasn't sure the best way to deal with this. Practically speaking, only running `previewSrc` when the value of the image field changes is likely fine for most use cases. If you have a suggestion, let me know.

PR #1222, opened 2020-06-05
Form loading

ncphillips:


dwalkr:
#1219

packages/@tinacms/forms/src/form.test.ts
```diff
@@ -36,6 +37,67 @@ describe('Form', () => {
...
+      })
+      describe('after promise resolves', () => {
+        it('is not loading once resolved', async () => {
+          let resolve: () => void | null
+
+          const form = new Form({
+            ...DEFAULTS,
+            loadInitialValues: () =>
+              new Promise(r => {
+                resolve = r
```
dwalkr, line 74: :exploding_head:

ncphillips: Shocked by the crazy, resolve a promise outside of the promise shenanigans?

dwalkr: I just never would have thought of it. clever

PR #1228, opened 2020-06-08
refactor: cms.enabled is used to determine if inline forms are active

ncphillips:


packages/demo-gatsby/src/templates/blog-post.js
```diff
@@ -42,14 +42,14 @@ import { useCMS } from "tinacms"
 const get = require("lodash.get")
 
 function InlineWysiwyg(props) {
-  const { status } = useInlineForm()
+  const cms = useCMS()
   const [{ InlineWysiwyg }, setEditor] = React.useState({})
 
   React.useEffect(() => {
     if (!InlineWysiwyg && status === "active") {
```
dwalkr, line 49:
```suggestion
    if (!InlineWysiwyg && cms.enabled) {
```

ncphillips: Weird, I thought I made that change

kendallstrautman:
Should the individual fields call `useCMS` or could they still get passed a 'status' prop from `InlineField` based on`cms.enabled`?

ncphillips:
They could but I'm not sure what the value of that indirection would be if the status is always going to be `active` when `cms.enabled`

kendallstrautman:
Yeah I was thinking it could just be a boolean or something, so that only `InlineForm` calls `useCMS` and all the child fields don't have to. But I'm down with either way.

packages/react-tinacms-inline/src/inline-form.tsx
```diff
@@ -40,29 +39,16 @@ export interface InlineFormState {
...
   const inlineFormState = React.useMemo(() => {
     return {
       form,
-      status,
       focussedField,
       setFocussedField,
-      activate: () => setStatus('active'),
-      deactivate: () => setStatus('inactive'),
     }
   }, [form, status, focussedField])
```
kendallstrautman, line 53:
```suggestion
  }, [form, focussedField])
```

PR #1232, opened 2020-06-09
Code mark fix

jpuri:
Code mark should not co exist with other marks. Currently it is possible to create code mark using keyboard shortcut or input rules and have it co-exist with other marks.


ncphillips:
What prompted this change? 

Code can mix with other marks in CommonMark, that’s why the tests are failing

jpuri:
I have updated the PR to exclude link mark as that is breaking scenario like this 
**[`commonmark.js`](https://github.com/jgm/commonmark.js)** which is valid according to commonmark.

I will create a separate issue for more detailed handling of links.

ncphillips:
Ahh yess that's what I was missing. Inline code should work with links but not bold or italic. 👍

PR #1238, opened 2020-06-10
fix: code block node view

jpuri:
Fix for code block not using code mirror.

This appears to be regression from recent refactoring.


jpuri:
Hey @ncphillips, @DirtyF  : I just added fix for hybrid controlled behaviour.

Issue #228, opened 2019-10-03
Provide ability to customize the sidebar-empty state


ncphillips:
The empty-state ui is great for a developer setting up the site, but not appropriate afterwards.

<img width="476" alt="image" src="https://user-images.githubusercontent.com/824015/66169166-d8f2c000-e615-11e9-9960-474ce2ee1cb9.png">

 I want to be able to somehow make my own empty sidebar so I can add links to other services i.e. hosting, analytics, github repo, etc.

kendallstrautman:
It would be cool to perhaps pass an `EmptyState` component via the `sidebar` config.

PR #1243, opened 2020-06-10
Custom empty state

kendallstrautman:
[Original PR](https://github.com/tinacms/tinacms/pull/1242) by @DaniTulp 

_Notes from Original:_

Addresses #228. I'm not completely sure about the naming of the prop. I've added an example to the demo-next example, if you go to the blocks page you can see it renders it correctly.


ncphillips:
Looks good! I have a couple nit-picky comments around naming and default values but I'll post those tomorrow morning.

DaniTulp:
I'll wait on the nitpicks before adding a pull-request for the docs.

ncphillips:
Thanks for the patience! Here's my thoughts:

I think defaulting to `NoFormsPlaceholder` should be done in `SidebarState` instead of in the `FormsView`. The `FormView` will just not render anything if there's no `CustomEmptyState`. That means the `emptyState` will be still be optional when creating, but the default component will be `NoFormsPlaceholder`:

```ts
  constructor(private events: EventBus, options: SidebarStateOptions = {}) {
    this.position = options.position || 'displace'
    this._hidden = !!options.hidden
    this.emptyState = options.emptyState || NoFormsPlaceholder
```

That said, I'm also not big on the names `emptyState` and `CustomEmptyState`. They're a bit misleading as the usual meaning of "state" in React is about data not components.

I'm not really sure what I think that should be called, but an idea I have is `sidebar.body`. Instead of passing the component as a prop:

```jsx
<FormsView CustomEmptyState={sidebar.body} />
```

The component could then be the child:

```jsx
<FormsView>
  <sidebar.body />
</FormsView>
```

I like that but not totally sold on it, so other thoughts are welcomed!

DaniTulp:
I definitely agree with pulling out the empty component outside of the `<FormsView>` and into the Sidebar component. 

Where should the `NoFormsPlaceHolder` component live? Its own file?

Naming these things is indeed pretty hard, `emptySidebarBody` comes to mind or `emptySidebar` I feel like having the word empty in it does add some significance. Another option could be to mirror the `NoFormsPlaceholder` component as a prop.

ncphillips:
> Where should the NoFormsPlaceHolder component live? Its own file?

Yeah I think so. In it's own file. Exact place in package matters a bit less.

>  I feel like having the word empty in it does add some significance.

Yeah that's a good point. I don't think the `sidebar` is required in the name, since it's placed on the component. What about `sidebar.emptyBody` or `sidebar.placeholder`?

```jsx
<FormsView>
  <sidebar.emptyBody />
</FormsView>
```

or

```jsx
<FormsView>
  <sidebar.placeholder />
</FormsView>
```

DaniTulp:
I'm nudging towards placeholder for consistency with the default option.

ncphillips:
Cool. Agreed 👍

PR #1249, opened 2020-06-11
refactor: changed way internal logic works and the exposed api

DaniTulp:
#1243

### DIFF
--- a/packages/@tinacms/react-sidebar/src/sidebar.ts
+++ b/packages/@tinacms/react-sidebar/src/sidebar.ts
@@ -17,13 +17,14 @@ limitations under the License.
 */
 
 import { EventBus, Callback } from '@tinacms/core'
+import { NoFormsPlaceholder } from './components/NoFormsPlaceHolder'
 import * as React from 'react'
 
 export interface SidebarStateOptions {
   hidden?: boolean
   position?: SidebarPosition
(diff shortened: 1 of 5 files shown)


packages/@tinacms/react-sidebar/src/sidebar.ts
```diff
@@ -17,13 +17,14 @@ limitations under the License.
...
 import { EventBus, Callback } from '@tinacms/core'
+import { NoFormsPlaceholder } from './components/NoFormsPlaceHolder'
 import * as React from 'react'
 
 export interface SidebarStateOptions {
   hidden?: boolean
   position?: SidebarPosition
   buttons?: SidebarButtons
-  emptyState?: React.ElementType
+  placeholder?: React.ReactNode
```
ncphillips, line 27: I think I would rather pass the component than a node. Something like:

```suggestion
  placeholder?: React.FC
```

ncphillips: Besides that this great! 🙌

DaniTulp: Yep sounds good!

ncphillips:
@all-contributors please add @DaniTulp for docs

PR #1257, opened 2020-06-13
fix: prevents empty tag field registering

### LINKED ISSUE #1201: Empty Tags field
## Description

I can generate an empty tag field by pressing space.

tags-space

### Steps to reproduce

Focus on a tags field, press space.

### Expected result

It should only create tags when you type a least one char.

### Actual result

It creates an empty tag

#### Tinacms Package Versions

```text
tinacms 0.21
```

isabellabrookes:
Fixes #1201 

Added a simple check to return out of the function if tag field value is an empty string.

(I couldn't find the tests or I would have added one)


ncphillips:
Amazing! Thanks a lot @isabellabrookes 

@all-contributors please add @isabellabrookes for code

Issue #1264, opened 2020-06-15
GitHub API: Switching branches does not sync form state


dwalkr:
Using the GitHub API with a client-side editing flow, switching branches via the toolbar does not update the form state with the contents of that branch. We need to re-initialize the form (re-run `loadInitialValues`) after switching branches.

ncphillips:
Right now the Form `id` is just the file's relative path within the repo. If that was changed to be the branch + fileRelativePath then re-initializing would happen automatically.

dwalkr:
I just had that exact thought! The thing is, in some cases we will need to get the branch asynchronously

dwalkr:
Adding branch name to the ID would also make the localStorage behavior work seamlessly

ncphillips:
Why would tou have to get the branch name asynchronously?

dwalkr:
j/k that's for the SHA, looks like we'll always have a branch name to work from

dwalkr:
Updating the form ID works but it needs a signal from the branch switcher to know that the branch has been updated. In a moment I will submit a PR that dispatches an event on branch select

PR #1268, opened 2020-06-15
feat(react-tinacms-github): fire event on branch change

dwalkr:
This feels like a **feat** to me but let me know if it should be a **chore** instead.

By firing an event, developers can subscribe to branch change in their form-creating code.

This addresses the problem described in #1264, but we might also want to implement the desired behavior in `useGithubFileForm`


packages/react-tinacms-github/src/toolbar-plugins/BranchSwitcherPlugin.tsx
```diff
@@ -166,6 +166,10 @@ const BranchSwitcher = ({ onBranchChange }: BranchSwitcherProps) => {
             }
             setConfirmSwitchProps(null)
             setCreateBranchProps(null)
+            cms.events.dispatch({
+              type: 'github:branch:select',
```
ncphillips, line 170: Just had an idea. What if we went with the `git` terminology?

```suggestion
              type: 'github:branch:checkout',
```

PR #1274, opened 2020-06-17
feat: add a blockProps prop which spreads props to child elements

sakulstra:
this allows passing trough dynamic props

---

I've just gave tina-cms a spin after reading the next.js blogpost about it.
When experimenting a bit with the incredible cool inline features I faced some problems.

I wanted to make a collapsible list which roughly works like that:
```js
function InteractiveBlocks() {
  // ensure a single collapsible is open at a time
  const [expanded, setExpanded] = React.useState<number | false>(false);

  const handleChange = (panel: number) => (
    event: React.ChangeEvent<{}>,
    newExpanded: boolean
  ) => {
    setExpanded(newExpanded ? panel : false);
  };
  return (
   <InlineBlocks name="faqs" blocks={FAQ_BLOCKS} blockProps={{onChange: handleChange, expanded}} />
  )
}

const faq_item_template = {
  type: "faq_item",
  label: "Selling Point",
  defaultItem: {
    title: "How much is the fish",
    body: "8000 mark",
  },
};

const FAQ_BLOCKS = {
  faq_item: {
    Component: FaqItem,
    template: faq_item_template,
  },
};
```

### DIFF
--- a/packages/react-tinacms-inline/src/blocks/inline-field-blocks.tsx
+++ b/packages/react-tinacms-inline/src/blocks/inline-field-blocks.tsx
@@ -34,6 +34,12 @@ export interface InlineBlocksProps {
   }
   className?: string
   direction?: 'column' | 'row'
+  /**
+   * object will be spread to every block child element
+   */
+  blockProps?: {
+    [key: string]: any
+  }
 }
 
 export interface InlineBlocksActions {
@@ -68,6 +74,7 @@ export function InlineBlocks({
   blocks,
   className,
   direction = 'column',
+  blockProps,
 }: InlineBlocksProps) {
   const cms = useCMS()
   const [activeBlock, setActiveBlock] = useState(-1)
@@ -151,6 +158,7 @@ export function InlineBlocks({
 
                       return (
                         <InlineBlock
+                          {...blockProps}
                           key={index}
                           index={index}
                           name={blockName}
--- a/packages/react-tinacms-inline/src/blocks/inline-field-blocks.tsx
+++ b/packages/react-tinacms-inline/src/blocks/inline-field-blocks.tsx
@@ -37,7 +37,7 @@ export interface InlineBlocksProps {
   /**
    * object will be spread to every block child element
    */
-  blockProps?: {
+  itemProps?: {
     [key: string]: any
   }
 }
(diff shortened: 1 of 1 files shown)


sakulstra:
Did this as a quick local fix, but am not convinced that's the best solution overall :thinking:
The current api doesn't feel very "react"y to me :sweat_smile: - as i just started using tina i'm not sure if that's reasonable: Could `InlineBlocks` provide a renderProp api or simply allow passing child elements as child instead of going through the template hoops?

ncphillips:
## PR Feedback 
This looks good! Thanks a lot @sakulstra 

To better match the existing APIs I would rename it from `blocksProps` to `itemProps`.

I don't understand how the `onChange` works though. Is your Block component handling that itself?

## render props

> Could InlineBlocks provide a renderProp api or simply allow passing child elements as child instead of going through the template hoops?

A render prop could work but just passing children wouldn't. The `InlineBlocks` component loops over all the data and makes sure the correct blocks are rendered and handles some other stuff. It could possibly delegate part of that rendering to a `renderItem` prop. 

But I'm not tempted to add that right now, as I'm not sure what problem that would solve. I like the solution you presented here, and it is more in line with the API for the regular [BlocksFieldPlugin](https://tinacms.org/docs/fields/blocks).

Also I'm not worried about the current API being "reacty". This pattern solves the problem and requires less effort from the user of the API then adding the `renderItem` prop would. An optional `renderItem` could good, but I would like to put that off until we see more usecases that can't be solved using the current API.

## InlineGroupList

Looking more at your example though, I think I see another way to make your life easier. 

You're only using **one** Block type here. 

In the sidebar forms we have a separate field for data like that–the [GroupListPlugin](https://tinacms.org/docs/fields/group-list). This makes me think that a seperate feature request could be made to create an `InlineGroupList` which could have a simpler API:

```tsx
<InlineGroupList 
  name="faqs" 
  defaultItem={{
    title: "How much is the fish",
    body: "8000 mark",
  }}
  component={FaqItem}
/>
```

Or to fit the above example:

```tsx
<InlineGroupList 
  name="faqs" 
  defaultItem={{
    title: "How much is the fish",
    body: "8000 mark",
  }}
  renderItem={props => {
    return (
      <FaqItem 
        {...props} 
        onChange={handleChange} 
        expanded={expanded}
      />
    )
  }}
/>
```

sakulstra:
> To better match the existing APIs I would rename it from blocksProps to itemProps.

okay changed it

> I don't understand how the onChange works though. Is your Block component handling that itself?

yes exactly

PR #1278, opened 2020-06-18
changed the direction of inline blocks from row and col to vertical and horizontal

logan-anderson:
changed the direction of inline blocks from row and col to vertical and horizontal and also updated the next-demo

### DIFF
--- a/packages/react-tinacms-inline/src/blocks/inline-block-field-controls.tsx
+++ b/packages/react-tinacms-inline/src/blocks/inline-block-field-controls.tsx
@@ -67,9 +67,17 @@ export function BlocksControls({
   const blockMoveDownRef = React.useRef<HTMLButtonElement>(null)
 
   const addBeforePosition =
-    direction === 'column' ? 'top' : direction === 'row' ? 'left' : 'top'
+    direction === 'vertical'
+      ? 'top'
+      : direction === 'horizontal'
+      ? 'left'
+      : 'top'
   const addAfterPosition =
-    direction === 'column' ? 'bottom' : direction === 'row' ? 'right' : 'bottom'
+    direction === 'vertical'
+      ? 'bottom'
+      : direction === 'horizontal'
+      ? 'right'
+      : 'bottom'
 
   if (cms.disabled) {
     return children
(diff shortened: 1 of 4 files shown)


ncphillips:
Error message
```
(typescript) Error: /home/runner/work/tinacms/tinacms/packages/react-tinacms-inline/src/blocks/inline-block-field-controls.tsx(70,5): semantic error TS2367: This condition will always return 'false' since the types '"vertical" | "horizontal"' and '"column"' have no overlap.
```

logan-anderson:
> Error message
> 
> > (typescript) Error: /home/runner/work/tinacms/tinacms/packages/react-tinacms-inline/src/blocks/inline-block-field-controls.tsx(70,5): semantic error TS2367: This condition will always return 'false' since the types '"vertical" | "horizontal"' and '"column"' have no overlap.

Pushing up a fix now. Just testing it.

logan-anderson:
@ncphillips Fixed

packages/react-tinacms-inline/src/blocks/inline-block-field-controls.tsx
```diff
@@ -67,9 +67,17 @@ export function BlocksControls({
   const blockMoveDownRef = React.useRef<HTMLButtonElement>(null)
 
   const addBeforePosition =
-    direction === 'column' ? 'top' : direction === 'row' ? 'left' : 'top'
+    direction === 'vertical'
+      ? 'top'
+      : direction === 'horizontal'
+      ? 'left'
+      : 'top'
```
ncphillips, line 74: I think this can be simplified
```suggestion
    direction === 'horizontal'
      ? 'left'
      : 'top'
```

packages/react-tinacms-inline/src/blocks/inline-block-field-controls.tsx
```diff
@@ -67,9 +67,17 @@ export function BlocksControls({
...
   const addAfterPosition =
-    direction === 'column' ? 'bottom' : direction === 'row' ? 'right' : 'bottom'
+    direction === 'vertical'
+      ? 'bottom'
+      : direction === 'horizontal'
+      ? 'right'
+      : 'bottom'
```
ncphillips, line 80:
```suggestion
    direction === 'horizontal'
      ? 'right'
      : 'bottom'
```

PR #1281, opened 2020-06-19
Remove Date Field from Default Plugins

ncphillips:
## Problem

Tina adds a lot of Javascript to your site. 

As discussed in #771 a large portion of that comes from the `DateFieldPlugin` in `tinacms`. This plugin is just a wrapper around [react-datetime](https://github.com/YouCanBookMe/react-datetime) which requires [moment](https://github.com/moment/moment/) to operate

## Solution

In this branch I moved the `DateFieldPlugin` from `tinacms` to `react-tinacms-date`. Doing this means the date field will not be available by default. Instead, developers will have to add `react-tinacms-date` themselves if they want to be able to use it. This gives developers the option to exclude it entirely, or load these fields dynamically to reduce the initial bundle size javascript being transferred.

## Impact: Reduces Bundle Size by 720kb

The @next/bundle-analyzer plugin to measure the impact of tinacms on the bundle

### Loading Immediately. 

This is the simplest solution but it will increase your initial bundle size by 700kb

```tsx
import { DateFieldPlugin } from "react-tinacms-date"

cms.plugins.add(DateFieldPlugin)
```

### Loading Dynamically

```js
import { useCMS } from "tinacms"

const Layout = ({ children }) => {
  const cms = useCMS()

  React.useEffect(() => {
    import("react-tinacms-date").then(({ DateFieldPlugin }) => {
      cms.plugins.add(DateFieldPlugin)
    })
  }, [])

  return <div>{children}</div>
}
```

### Loading Dynamically & Only if the CMS is enabled

```js
import { useCMS } from "tinacms"

const Layout = ({ children }) => {
  const cms = useCMS()

  React.useEffect(() => {
    import("react-tinacms-date").then(({ DateFieldPlugin }) => {
      if (cms.enabled) {
        cms.plugins.add(DateFieldPlugin)
      }
    })
  }, [cms.enabled])

  return <div>{children}</div>
}
```

## Placeholder

image

---

- [x] Updated docs PR approved – https://github.com/tinacms/tinacms.org/pull/496


logan-anderson:
The PR link on the place holder takes you to https://github.com/tinacms/tinacms/pull/1134 which is the PR for markdown/html the could be confusing for people.
instead, you could change 
```ts
function createPlaceholder(name: string) {
//...
```
to 
```ts
 function createPlaceholder(name: string, PRlink: string ) {
//...
````
and
```ts
<PlaceholderParagraph>
          Visit the{' '}
          <a
            href="https://github.com/tinacms/tinacms/pull/1134"
            target="_blank"
            rel="noreferrer noopener"
          >
            Pull Request
          </a>{' '}
          to learn more about why this change was made.
   </PlaceholderParagraph>
```
to
```ts
<PlaceholderParagraph>
          Visit the{' '}
          <a
            href={PRlink}
            target="_blank"
            rel="noreferrer noopener"
          >
            Pull Request
          </a>{' '}
          to learn more about why this change was made.
   </PlaceholderParagraph>
```

ncphillips:
Good catch @logan-anderson 👍 

Made that change

ncphillips:
[Just opened a PR for the docs changes](https://github.com/tinacms/tinacms.org/pull/496)

djensen47:
I'm having a little trouble adding this plugin to Gatsby. After I install the plugin and add it to the gatsby-config, I still see the "deprecated" message.

I'm using tina-grande as my base. Here's the section in gatsby-config.

```
    {
      resolve: "gatsby-plugin-tinacms",
      options: {
        enabled: process.env.NODE_ENV !== "production",
        sidebar: {
          position: "displace",
          theme: {
            color: {
              primary: {
                light: theme.color.primary,
                medium: theme.color.primary,
                dark: theme.color.primary,
              },
            },
          },
        },
        plugins: ["gatsby-tinacms-git", "gatsby-tinacms-remark", "react-tinacms-date"],
      },
    },
```

Is it no longer compatible with Gatsby?

djensen47:
It took some digging but I found the answer here:

https://tina.io/docs/plugins/#adding-plugins-in-gatsby

PR #1282, opened 2020-06-19
Add Modal Portal & Fix Color Picker Popup

spbyrne:
This is a quick and dirty solution to the modal clipping issue, but it also improves the color picker popup in a number of other ways. This same approach will likely work with the date picker popup. This also prevents popup clipping in the sidebar.

Screen Shot 2020-06-19 at 2 51 41 PM

Screen Shot 2020-06-19 at 2 52 49 PM

* Adds a modal portal so that elements can portal out of the clipped modal content div
* Uses this modal portal for the color picker popup (even when not in a modal; might need to rethink this, though it's also beneficial for color popups in the sidebar)
* Adds styling to open the color popup to the top of the color swatch if needed (currently it opens on the top if it's more than halfway down the window)
* Tracks window resize events and repositions the popup as needed

Problems:
* Doesn't track sidebar for scroll events so the behaviour is a bit funky if the popup is open and you scroll
* Maybe having a generic top-level modal provider would make more sense since this is beneficial outside of things inside modals.


packages/@tinacms/fields/src/components/ColorPicker/ColorPicker.tsx
```diff
@@ -210,31 +317,41 @@ export const ColorPicker: React.FC<Props> = ({
...
-            />
-          </Dismissible>
-        </Popover>
+        <ModalPortal>
+          <Popover openTop={openTop} triggerBoundingBox={triggerBoundingBox}>
+            <Dismissible
+              click
+              escape
+              disabled={!displayColorPicker}
+              onDismiss={() => setDisplayColorPicker(false)}
```
kendallstrautman, line 342: Just use `toggleColorPicker` here?

packages/@tinacms/fields/src/components/ColorPicker/ColorPicker.tsx
```diff
@@ -210,31 +317,41 @@ export const ColorPicker: React.FC<Props> = ({
     )
   }
 
+  const toggleColorPicker = () => {
```
kendallstrautman, line 320: Could simplify? 

```
setDisplayColorPicker(!displayColorPicker)
if (!displayColorPicker) {
  //...
}
```

packages/@tinacms/fields/src/components/ColorPicker/ColorPicker.tsx
spbyrne: Good call.

packages/@tinacms/fields/src/components/ColorPicker/ColorPicker.tsx
spbyrne: I find that less intuitive since you want that conditional to run if the color picker is being displayed. Since it toggles first but the setter is async, you'd use the inverse value (as you have here), but it makes it look like the logic is reversed when it's really not.

kendallstrautman:
Well if we want to go with using the form portal instead, seems like we just need to remove the modal portal file and then I think this is ready for review?

spbyrne:
For sure. I also wanted to apply this to the date picker since it experiences all the same issues, but after playing around with that for a while it seems that should be another PR. Unfortunately for the datepicker we just get a 'ReactDatetime' component that renders out the input and the popup, so wrapping the popup in a portal isn't so easy.

PR #1286, opened 2020-06-23
feat: Adds list field plugin

jeffsee55:
## TODO

- [x] Docs Approved

## Description
There's no `list` component for when you have an array of values, for example in the shape of `bullet_points` shown below:
```json
{
  "title": "Pro",
  "description": "Pro plans give you exactly what to aim for on all runs and workouts with paces based on your target goal time.",
  "bullet_points": [
    "Everything in the Free plans plus...",
    "Goal-specific paces on every run & workout",
    "Over 40 specialized guides around strenth drills, nutrition and mindfulness tips",
    "Advanced guides for elite-level goals"
  ],
  "category": "Pro"
}
```
This PR adds a list field which allows you to define the type of field used for homogeneous array of items. I'm mainly putting this up for discussion, perhaps there's a better way to do this that I'm missing.

Usage:

```json
{
  "name": "bullet_points",
  "label": "Bullet Points",
  "component": "list",
  "itemField": {
    "label": "Bullet Points Item",
    "component": "text" 
  }
}
```

Which will submit a payload like:
```
"bullet_points": [
  "Everything in the Free plans plus...",
  "Goal-specific paces on every run & workout",
  "Over 40 specialized guides around strenth drills, nutrition and mindfulness tips",
  "Advanced guides for elite-level goals"
],
```


kendallstrautman:
Nice! Works in the demos as expected 👌 We should write docs for this before merging.

kendallstrautman:
I am curious though, since this is so similar to `group-list`, is the benefit here mainly config simplicity?

jeffsee55:
> Nice! Works in the demos as expected 👌 We should write docs for this before merging.

I'm happy to write this up myself, will do if we are all on board with this.

> I am curious though, since this is so similar to `group-list`, is the benefit here mainly config simplicity?

That's a good question, I could go either way - what would the ideal field definition look like? 
```json
{
  "name": "bullet_points",
  "label": "Bullet Points",
  "component": "list",
  "itemField": {                            // what does this part look like?
    "label": "Bullet Points Item",
    "component": "text" 
  }
}
```

ncphillips:
> I am curious though, since this is so similar to group-list, is the benefit here mainly config simplicity?

I don't think so since they represent fairly different data structures. i.e. arrays of primitives vs arrays of objects

**list's array of primitives**
```json
[
  "Red",
  "Blue"
]
```
**group-list's array of objects**
```json
[
  { 
     color: "Red",
     examples: [],
  },
  { 
     color: "Blue",
     examples: [
       { name: "Blueberry", category: "berry" },
       { name: "ocean", category: "water" },
  },
]
```

IMO the visual similarity between the fields is kind of a coincidence. This field could be designed as more of a multi-select or tags field.

spbyrne:
Works well for me, nice! 👍 

I feel like given the itemField definition I expected the 'label' to go next to the field, but it's used as a label for the item itself. Would it make sense have separate item label & field label definitions? Or maybe the field doesn't need any label since there's only one?

Screen Shot 2020-06-24 at 11 18 12 AM

It might be nice to be able to use a function for the item label so instead of "Bullet Points Item" for each you could point to the field data (or something similar).

kendallstrautman:
It'd be nice if you could edit the value from the first view and not have to go to the second panel.

ncphillips:
Might also make sense to put some limit on what the `component` can be. i.e. only primitive value field's such as `text`, `textarea`, `color`, `select`

spbyrne:
Screen Shot 2020-06-24 at 12 11 54 PM

Something like this could maybe work. The UI is already functional so maybe it makes sense to merge as-is an upgrade later?

packages/@tinacms/fields/src/plugins/ListFieldPlugin.tsx
```diff
@@ -0,0 +1,438 @@
...
+  TrashIcon,
+  LeftArrowIcon,
+} from '@tinacms/icons'
+import { GroupPanel, PanelHeader, PanelBody } from './GroupFieldPlugin'
+import { useFormPortal } from '@tinacms/react-forms'
+import { FieldDescription } from './wrapFieldWithMeta'
+
+interface ListFieldDefinititon extends Field {
+  component: 'list'
+  defaultItem?: object | (() => object)
```
kendallstrautman, line 38: I wonder if the default item should be a string or another primitive.

jeffsee55: Yeah that's a mistake on my part, copying and pasting laziness 🤦

jeffsee55:
> It'd be nice if you could edit the value from the first view and not have to go to the second panel. So there wouldn't be per-item labels it would just render the input with the value.

I agree, so I like @spbyrne's UI and we could eliminate the extra label since they're all the same, just not sure if we'd run out of horizontal space with the drag-n-drop handle, especially for other types of fields like textarea or images:
Screen Shot 2020-06-24 at 8 36 52 AM


So in hearing the feedback, is this a better API? (maybe `field` is more clear than `itemField`? I'm ok with either)
```js
{
  name: "bullet_points",
  label: "Bullet Points",
  component: "list",
  field: {
    component: "text",
  },
};
```
Or for an image:
```js
{
  name: "bullet_points",
  label: "Bullet Points",
  component: "list",
  field: {
    component: "image",
    parse: () => {},
    previewSrc: () => {},
    uploadDir: () => {},
  },
};
```
Any other thoughts a more clear API?

> Might also make sense to put some limit on what the component can be. i.e. only primitive value field's such as text, textarea, color, select

From what I can tell the only fields it **shouldn't** support are: `blocks`, `group`, `group-list`

Then there are some where it's probably a strange choice and might be hard for UI decisions but it's at least a valid data shape: `HTML`, `Markdown`, `Tags`

Then there are the ones that output a single value and make the most sense: `text`, `textarea`, `number`, `date & time`, `image`, `color`, `toggle`, `select`. Thoughts?

spbyrne:
Screen Shot 2020-06-24 at 1 34 41 PM

Here's three options with the label removed. I think option three provides the most room for the field while still being fairly consistent with blocks/group lists.

ncphillips:
Like the `field` as the prop name.

The `field.component` prop feels too open to me. I don't think it should use the Field plugins. Maybe it could define it's own "private" set of components that don't have the field meta data wrapped around them. Limit the available inputs to something like `textarea`, `number`, and `select`.

I have 3 reasons for that

* I can't see a usecase for `boolean[]` or `date[]` so I would prefer to not to make them available by default. [YAGNI](https://www.martinfowler.com/bliki/Yagni.html) rules code and APIs.
* A list of images is definitely useful but I would follow the Forestry example and make that a separate `GalleryFieldPlugin`. There's probably more features we would want to add on that. 
* The existing fields aren't designed to fit in that little rectangle

We could still make it extensible though. Like with Field, the `component` prop could accept a component or a string:

```js
const bulletPointsField = {
  name: "bullet_points",
  label: "Bullet Points",
  component: "list",
  field: {
    component: "image",
    parse: () => {},
  },
}

import { Toggle } from "@tinacms/fields"

const togglesField = {
  name: "toggles",
  label: "name-less-toggles",
  component: "list",
  field: {
    component: ({ input }) => <Toggle {...input} />,
    parse: (value) => !value,
  },
}
```

jeffsee55:
Ok sounds good, I'll get around to implementing the rest of this with a scoped down version of what we accept and following option #3 from what @spbyrne provided.

spbyrne:
I styled it to match option #3 from above, works well for me. Only tested and working with text, textarea, number, and select. I tried throwing a color picker in but it didn't work.

Jun-30-2020 12-30-00

ncphillips:
This is awesome 🙌 

Once the docs are ready we can merge

kendallstrautman:
I implemented some changes based on the feedback.

- changed `itemField` --> `field`
- removed `label` from the individual field item
- when no default item is set, an empty string is returned instead of an object.
- made the component type more specific -- `'text' | 'textarea' | 'number' | 'select'`

I made a test case in `demo-next` in the sidebar form on the Info page (data isn't rendered on the page).

PR #1289, opened 2020-06-24
fix: update readme to correct links

logan-anderson:
# update readme
I updated the readme so that the "community" link went to the community page instead of the doc pages. Not sure why it was linked to the docs was kinda confusing for me at least.

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,7 +16,7 @@
 - [Website](https://tinacms.org/)
   - [Blog](https://tinacms.org/docs/)
   - [Docs](https://tinacms.org/docs/)
-- [Community](https://community.tinacms.org/)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -15,8 +16,7 @@
 - [Website](https://tinacms.org/)
   - [Blog](https://tinacms.org/docs/)
   - [Docs](https://tinacms.org/docs/)
-- [Community](https://tinacms.org/docs/)\
-  - [Forum](https://community.tinacms.org/)
+- [Community](https://community.tinacms.org/)
```
kendallstrautman: Should we just call this Forum?

logan-anderson: I updated it to be "Community Forum" not sure if that is better or if just "Forum" would be better

PR #1291, opened 2020-06-24
feat: Block limits

kendallstrautman:
Pretty simple implementation. The add block and trash icons are hidden / shown based on the limits. Checkout the blocks page in `demo-next` for an example. When the range is limited (i.e. min === max), neither add block or trash icon will show. 

**Example**
```jsx
<InlineBlocks
  name="blocks"
  blocks={PAGE_BUILDER_BLOCKS}
  min= {1}
  max={4}
/>
```

---
- [x] Docs updates PR approved

_Note: I will write docs if we like this API_

### DIFF
--- a/packages/demo-next/pages/blocks.tsx
+++ b/packages/demo-next/pages/blocks.tsx
@@ -62,10 +62,8 @@ export default function BlocksExample({ jsonFile }) {
               itemProps={{
                 style: { backgroundColor: 'red' },
               }}
-              limits={{
-                min: 2,
-                max: 4,
-              }}
+              min={2}
+              max={4}
             />
           </Wrap>
         </InlineForm>
--- a/packages/demo-next/pages/blocks.tsx
+++ b/packages/demo-next/pages/blocks.tsx
@@ -62,6 +62,10 @@ export default function BlocksExample({ jsonFile }) {
               itemProps={{
                 style: { backgroundColor: 'red' },
               }}
+              limits={{
+                min: 2,
+                max: 4,
+              }}
             />
           </Wrap>
         </InlineForm>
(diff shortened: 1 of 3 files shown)


packages/react-tinacms-inline/src/blocks/inline-field-blocks.tsx
```diff
@@ -53,6 +57,10 @@ export interface InlineBlocksActions {
   activeBlock: number | null
   setActiveBlock: any
   direction: 'vertical' | 'horizontal'
+  limits?: {
+    min: number
+    max: number
```
ncphillips, line 62:
```suggestion
    min?: number
    max?: number
```

ncphillips:
Hmm looking at it again, instead of `limits` what about just putting `min` and `max` at the root?

```jsx
<InlineBlocks
  name="blocks"
  blocks={PAGE_BUILDER_BLOCKS}
  min={1}
  max={4}
/>
```

kendallstrautman:
good call @ncphillips, that simplified things.

packages/demo-next/pages/blocks.tsx
```diff
@@ -62,6 +62,10 @@ export default function BlocksExample({ jsonFile }) {
               itemProps={{
                 style: { backgroundColor: 'red' },
               }}
+              limits={{
+                min: 2,
+                max: 4,
+              }}
```
ncphillips, line 68:
```suggestion
              min={2}
              max={4}
```

packages/react-tinacms-inline/src/blocks/inline-block-field-controls.tsx
```diff
@@ -120,6 +123,12 @@ export function BlocksControls({
     .slice(0, -1)
     .join('.')
 
+  function withinLimit(limit: number | undefined) {
+    if (!limit) return true
+
+    return !(limit === count || (max === count && min === count))
+  }
```
ncphillips, line 130: Nice changeup! I like this fn

PR #1292, opened 2020-06-25
fix: extra paragraphs created while inserting table

jpuri:
The PR fixes issue with extra paragraphs being inserted when table is created.

May be we can reduce margins also above and below the table:

Currently its `32px`:

---
Screenshot 2020-06-25 at 7 44 52 PM
---

It can be reduced to `20px` or lesser by moving also icons can be moved closer to table:

---
Screenshot 2020-06-25 at 7 46 06 PM
---

@spbyrne , @DirtyF : plz suggest.

### DIFF
--- a/packages/react-tinacms-editor/src/plugins/Image/Popups/Edit.tsx
+++ b/packages/react-tinacms-editor/src/plugins/Image/Popups/Edit.tsx
@@ -285,13 +285,25 @@ const SaveLink = styled.button`
   }
 `
 
-const CancelLink = styled(SaveLink)`
-  background-color: white;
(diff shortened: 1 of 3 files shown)


packages/react-tinacms-editor/src/plugins/Image/Popups/Edit.tsx
```diff
@@ -285,14 +285,26 @@ const SaveLink = styled.button`
   }
 `
```
jpuri, line 287: Extending another styled component was somehow creating circular dependency issue.
Today it broke build, I guess some latest build changes thus I refactored it.

PR #1296, opened 2020-06-26
Minor Style Fixes

spbyrne:
* Fixes toolbar menu panel styles. On sites where a box sizing reset wasn't used the padding was being added to the calculated width.
* Bumps minimum Tina font size from 11 to 12px. This size is used for field descriptions.
* Prevents field description text from inheriting site <p> styles.


kendallstrautman:
Fixes[ issue with alpaca](https://github.com/tinacms/tina-starter-alpaca/issues/52)

PR #1298, opened 2020-06-26
GithubClient improvements

### LINKED ISSUE #1250: GitHub Media Store: Handle 409 error
Right now the GitHub media store assumes an image doesn't exist when uploading, so it doesn't pass a SHA. Trying to overwrite a file without passing the SHA of the previous revision results in HTTP 409.

We should handle 409 response by either (1) automatically fetching the latest SHA and retrying, or (2) prompting the user with an overwrite notice, which upon confirmation, fetches the latest file SHA and re-uploads.

dwalkr:
usability improvements for GitClient

closes #1250


packages/react-tinacms-github/src/github-client/GithubFile.ts
```diff
@@ -0,0 +1,83 @@
...
+  fetchFile = async () => {
+    const res = await this.cms.api.github.fetchFile(this.path)
+    this.sha = res.sha
+    return this.parse ? this.parse(res.content) : res.content
+  }
+
+  commit = async (
+    data: any,
+    message: string = 'Update from TinaCMS',
+    retryOnConflict = true
```
ncphillips, line 44: When would you not want to retryOnConflict?

dwalkr: If there are irreconcilable conflicts with upstream changes it would just retry forever. We only want to retry when the conflict is due to a stale SHA, so retrying exactly once seems suitable.

dwalkr: To put it a different way: a conflict that can't be resolved with a single retry will not be resolved by this system.

I suppose if we directly called commit from the registered API on line 62 instead of recursively calling this method again, we wouldn't need this guard parameter

packages/react-tinacms-github/src/github-client/GithubFile.ts
```diff
@@ -0,0 +1,83 @@
...
+      const response = await this.cms.api.github.commit(
+        this.path,
+        this.sha,
+        serializedContent,
+        message
+      )
+      this.sha = response.content.sha
+      this.cms.alerts.success(
+        `Saved Successfully: Changes committed to ${this.cms.api.github.workingRepoFullName}`
+      )
```
ncphillips, line 57: I don't think this belongs here. By putting the alert in the github client we're preventing people from choosing the message themselves. This message would probably be fine for most English speakers, but i18n is the main concern. If someone from Italy is using Tina they can't change the text to Italian. 

I would prefer to push this to the client. If they want an alert, then instead of doing:

```js
useForm({
  onSave: file.commit,
})
```
they would do
```js
useForm({
  onSubmit: data => file.commit(data).then(() => cms.alert.success("Bueno!")
})
```

dwalkr: changed it to fire an event instead of an alert. Developers then have the option to add the alert in `onSubmit` or just subscribe to all commit events

```
  cms.events.subscribe('github:commit', (response) => {
    cms.alerts.success('Changes saved successfully.')
  })
```

dwalkr: I'm fine with this change, but I would also note here that there are lots of hard-coded console and alert messages elsewhere in Tina. If we want to tackle i18n holistically we'll want to think about some kind of translation solution; I don't think we should put the onus on Tina consumers to wire up alerts for every message that comes from the system

ncphillips: >  I don't think we should put the onus on Tina consumers to wire up alerts for every message that comes from the system

I somewhat agree. I think we'll eventually get there, but It's something we should be slow to put an official stamp on until we see all the ways i18n will bite us.

I like you're idea of using plain events and triggering alerts from somewhere's else. My gut says that would make it easier to deal with i18n, because there could likely be a single source of truth for those messages. i.e.

```js
cms.events.subscribe("*", dispatchAlerts(cms))
```

And then maybe you can override:

```js
import { alerts: githubAlerts } from "react-tinacms-github"

cms.events.subscribe('*', alertDispatcher(cms, {
  ...githubAlerts,
  "event:name": { status: "error", message: "It broke", timeout: 4000 },
})
```

Where the function would look something like:

```js
interface Alerts {
  [type: string]: {
    message: string
    status: AlertStatus
    timeout?: number
  }
}

function alertDispatcher(cms: CMS, alerts: Alerts) {
  return (event) => {
    const alert = alerts[event.type]

    if (alert). cms.alerts.dispatch(alert)
  }
}
```

PR #1299, opened 2020-06-29
feat: added GitFile class to make changing individual files easier

ncphillips:
This PR introduces a new `GitFile` class to the `@tinacms/git-client`. It's purpose is to make working with individual files much easier. A benefit of this, is it makes customizing forms much easier. 

It is inspired by the work being done in [react-tinacms-github](https://github.com/tinacms/tinacms/pull/1298)


packages/gatsby-tinacms-git/src/use-git-form.ts
```diff
@@ -33,42 +35,40 @@ export function useGitForm<N extends GitNode>(
...
-      id: node.fileRelativePath,
+      ...defaultConfig,
+      ...config,
+      id: gitFile.relativePath,
     },
     watch
   )
 }
+
+function useGitFile(
```
dwalkr, line 62: I know @tinacms/git-client is react-agnostic, but should we consider exporting `useGitFile` from there so it can be used in other contexts?

ncphillips: Yeah I was thinking that 🤔 

I kind of regret the name now. I wish we just had `react-tinacms-git`. 

I guess we can put it there

PR #1301, opened 2020-06-30
fix: adjust readme to new signature

sakulstra:
when trying to update packages i didn't immediately see why my build was failing - this was the reason

### DIFF
--- a/packages/next-tinacms-github/README.md
+++ b/packages/next-tinacms-github/README.md
@@ -31,13 +31,18 @@ import { createAuthHandler } from 'next-tinacms-github'
 
 export default createAuthHandler(
   process.env.GITHUB_CLIENT_ID,
-  process.env.GITHUB_CLIENT_SECRET
+  process.env.GITHUB_CLIENT_SECRET,
+  process.env.SIGNING_KEY
 )
 ```
 
 _See [Next's documentation](https://nextjs.org/docs/api-reference/next.config.js/environment-variables) for adding environment variables_
 
 [See below](#github-oauth-app) for instructions on creating a Github OAuth App to generate these **Client ID** & **Client Secret** variables.
+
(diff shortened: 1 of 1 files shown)


sakulstra:
> Just needs _process.env.GITHUB_CLIENT_SECRET_ added to the proxy and previewHandler, e.g.: `export default apiProxy(process.env.GITHUB_CLIENT_SECRET)` and `export default previewHandler(process.env.GITHUB_CLIENT_SECRET)`

Could you elaborate :sweat_smile:, not exactly sure what changes you are requesting.
The changes I applied here are to align the readme with the function signature in https://github.com/tinacms/tinacms/blob/master/packages/next-tinacms-github/src/github/create-auth-handler.ts#L29 as the currently stated usage will error due to the missing parameter.

jhuggett:
@sakulstra Apologies for not being clear, and I wrote the totally wrong thing (should be process.env.SIGNING_KEY not process.env.GITHUB_CLIENT_SECRET), what you've got is perfect. I'm suggesting in addition to show in the readme that the signing key needs to be passed into the apiProxy and previewHandler functions.  

For the apiProxy, from this:
```
// pages/api/proxy-github.ts

import { apiProxy } from 'next-tinacms-github'

export default apiProxy
```
to be turned into this:
```
// pages/api/proxy-github.ts

import { apiProxy } from 'next-tinacms-github'

export default apiProxy(process.env.SIGNING_KEY)
```

and for previewHandler, from this:
```
// pages/api/preview.ts

import { previewHandler } from 'next-tinacms-github'

export default previewHandler
```
to this: 
```
// pages/api/preview.ts

import { previewHandler } from 'next-tinacms-github'

export default previewHandler(process.env.SIGNING_KEY)
```

PR #1314, opened 2020-07-07
Fix: FormView buttons have default value

### LINKED ISSUE #1305: useFormScreenPlugin with disabled sidebar and enabled toolbar
## Description

When having a form using `useFormScreenPlugin` and having the sidebar disabled and the toolbar enabled. It show the toggle to open the global form drawer. When we then click the form it throws an error about buttons being undefined. I think [this line](https://github.com/tinacms/tinacms/blob/b377ccd7fbbb8260636d91ddcf8f8317add82bb1/packages/%40tinacms/react-forms/src/FormView.tsx#L38) is the culprit.

### Steps to reproduce

1. Disable the sidebar and enable the toolbar

1. Create form using the `useFormScreenPlugin`

1. Open the form from the global form drawer

### Expected result
Would be able to use the form or not show the global form drawer when the sidebar is disabled. The first option has my preference.

### Actual result

It throws a runtime error about buttons not being defined.

#### Tinacms Package Versions

```text
├─┬ react-tinacms-editor@0.8.3
├─┬ react-tinacms-inline@0.10.0
├─┬ tinacms@0.22.0
│ ├── @tinacms/alerts@0.1.5
│ ├── @tinacms/core@0.10.0
│ ├── @tinacms/icons@0.14.0
│ ├── @tinacms/media@0.1.0
│ ├── @tinacms/react-alerts@0.1.11
│ ├─┬ @tinacms/react-core@0.3.2
│ ├─┬ @tinacms/react-forms@0.2.2
│ ├── @tinacms/react-modals@0.1.11
│ ├── @tinacms/react-screens@0.1.10
│ ├─┬ @tinacms/react-sidebar@0.3.0
│ ├─┬ @tinacms/react-toolbar@0.3.0
│ ├─┬ @tinacms/styles@0.5.2
```

#### Tools

| Tool    | Version |
| ------- | ------- |
| browser |   Chrome 83.0.4103.116      |
| os      |      osX 10.15.5   |
| node    |      12.16.3   |
| npm     |   6.14.5      |

kendallstrautman:
Should fix #1305 

To test: 

- Start up `demo-gatsby` and try to open the 'Authors' form. Before the changes, the site would break. Now default button values are being provided.

I wonder if there should be some cms-level 'buttons' extracted from the sidebar, given that various modals and screens use these save and reset strings.


ncphillips:
> I wonder if there should be some cms-level 'buttons' extracted from the sidebar, given that various modals and screens use these save and reset strings.

I'm thinking it's in the wrong place altogether. The text isn't for the sidebar or even the toolbar in general, it's for the Form.

```js
new Form({
  buttons: {
    save: "Commit",
  }.
  actions: [DeleteAction]
})
```

Issue #1316, opened 2020-07-07
api-git: Improve error communication on failed push


kendallstrautman:
### Summary

Reposting from [the Forum](https://community.tinacms.org/t/no-error-message-when-github-push-fails/252).

TL:DR, when a commit or push fails, we should better alert the editor beyond just posting error details to the console.

### Basic example

Seems like adding a `cms.alerts.error()` to these catch statements would be good? Or should we allow the developer to configure this with a callback in the `GitRouterConfig`?

### Motivation

To make sure editors get feedback when their attempts to save content fails.

ncphillips:
This came up with [GithubFile](https://github.com/tinacms/tinacms/blob/github-file-commit/packages/react-tinacms-github/src/github-client/GithubFile.ts#L62) recently. I think the direction we're going is to dispatch plain CMS events rather then alerts. The develope could then setup a listener to dispatch a CMS alert in their language of choice, pop a modal, or do whatever else they have in mind.

```js
cms.events.subscribe("git:error", (event) => {
  if (isGitPushFailure(event)) cms.alerts.error("Error al empujar al repositorio")
})
```

PR #1318, opened 2020-07-07
feat: Adds git:commit event to git client

kendallstrautman:
Starts to address #1316. 

You can optionally pass `cms` when you instantiate the git client.
```
const client = new GitClient('http://localhost:3000/___tina', this.cms)
```
If the cms is passed, a `git:commit` event will dispatch when `commit` is invoked, returning the response. Devs can then parse the response to handle various alerts. 

```js
  React.useEffect(() => {
    cms.events.subscribe("git:commit", function handleCommitAlerts(event) {
      if (!event.response.ok) {
        cms.alerts.error("Something went wrong! Changes weren't saved")
      } else {
        cms.alerts.info("Content saved successfully!")
      }
    })
  }, [])
```
Not sure if this is the best way to go about it, **looking for feedback** before adding events to other actions (push etc.). Checkout `demo-gatsby/src/components/bio.js` for an example.


kendallstrautman:
Ooo I just looked at the `GitFile` PR and maybe this event stuff would be better there

kendallstrautman:
Closing to implement on `GitFile` instead of the client

PR #1319, opened 2020-07-08
feat: added delete action for markdown files

logan-anderson:
I am unsure whether this is the best way of doing things but factory function seemed the easiest way to abstract how to get the file name and title

example usage

```ts
//...
import { createMarkdownDeleteAction } from "react-tinacms-github";

  const deleteAction = createMarkdownDeleteAction()
  const formOptions = {
    actions:[deleteAction],
    fields: [
//...
}
const [data, form] = useGithubMarkdownForm(props.file, formOptions);
  usePlugin(form);

```

createMarkdownDeleteAction takes in an object  that contains to hook functions for getting the path of the file and the display name of  the file

ex
```ts
 const deleteAction = createMarkdownDeleteAction({
    getTitle: (form)=>{return form.values.frontmatter.title},
  })
```

## To-do
- [ ] Docs Approved


logan-anderson:
https://www.loom.com/share/2f26b23b1b114c3da6de6f956c398e74

chrisdmacrae:
@logan-anderson you beat me to the punch! :)

@ncphillips @spbyrne is a dropdown button possible, bound to the edge of save? I feel like discoverability would be better there.

E.g.,: => https://www.loom.com/share/5bf9d294eb384445a203f7bfdf47ff37

spbyrne:
I don't think this would be a good case for a dropdown button, this is exactly the kind of action the 'more actions' menu was designed for. We should add a tooltip and possible improve the icon, though.

DirtyF:
Grouping "Delete document" and "Exit CMS" on one end and "Save Document" on the other feels like we are mixing unrelated actions.

spbyrne:
The main action of the editor is 'Save', so it's a stand-alone action alongside resetting the form. Doing something like exiting the editor or deleting a file aren't as important, so they're grouped into an additional actions menu. All actions are grouped on the same end of the toolbar (or footer of the sidebar) so that the user doesn't have to remember what action goes where.

chrisdmacrae:
> I don't think this would be a good case for a dropdown button, this is exactly the kind of action the 'more actions' menu was designed for. We should add a tooltip at some point and possibly improve the icon, though.

Yeah -- a tooltip on the actions menu would get the job done. 💪

I might also increase the affordance in that menu by making the "delete" button / text red or by adding a correlated icon.

kendallstrautman:
Is this still WIP @logan-anderson?

logan-anderson:
> Is this still WIP @logan-anderson?

@kendallstrautman  It's ready for review. Id love to know what you guys think. May need tweaking in terms of how it looks

packages/react-tinacms-github/src/toolbar-plugins/createMarkdownDeleteAction.tsx
```diff
@@ -0,0 +1,93 @@
...
+    getFilePath: getTitleDefault,
+  }
+) => {
+  const DeleteAction = ({ form }: { form: Form }) => {
+    const cms = useCMS()
+    const [active, setActive] = React.useState(false)
+    const open = () => setActive(true)
+    const close = () => setActive(false)
+    const title = getTitle!(form)
+    const filePath = getFilePath!(form)
```
ncphillips, line 53: I think I would have just exported DeleteAction instead of making it configurable. In the case of `DeleteAction` in `gatsby-tinacms-git` we we just export the component. In that package the filepath and the title are consistent: `filepath=form.id` and `title=form.label`. I went with this because I know that a form of that type is always going to have the fileRelativePath as the id.

To be clear, I'm not saying you _should_ switch it, I'm just explaining the choice I would have made. 

Did you have a use-case in mind for taking this approach?

packages/react-tinacms-github/src/toolbar-plugins/createMarkdownDeleteAction.tsx
```diff
@@ -0,0 +1,93 @@
...
+
+    return (
+      <div>
```
ncphillips, line 56: What happens if this is just a fragment? `<>`

packages/react-tinacms-github/src/toolbar-plugins/createMarkdownDeleteAction.tsx
```diff
@@ -0,0 +1,93 @@
...
+import {
+  ActionButton,
+  useCMS,
+  Form,
+  Modal,
+  ModalPopup,
+  ModalHeader,
+  ModalBody,
+  ModalActions,
+} from 'tinacms'
```
ncphillips, line 29: I would prefer to remove the `tinacms` dependency and load these from the core packages directly. Something like this:

```
import { Modal, ... } from "@tinacms/modals"
import { Form, ActionButton } from "@tinacms/react-forms"
import { useCMS } from "@tinacms/react-core
```

packages/react-tinacms-github/src/toolbar-plugins/createMarkdownDeleteAction.tsx
```diff
@@ -0,0 +1,93 @@
...
+              <ModalActions>
+                <Button
+                  onClick={async () => {
+                    try {
+                      close()
+                      await cms.api.github.delete!(filePath)
+                      await cms.alerts.info(`${filePath} was deleted`)
+                    } catch (error) {
+                      close()
+                      cms.alerts.error(`Error in deleting ${filePath}`)
```
ncphillips, line 74: Do you know if there's any more information on the error that would be useful? Or would it be too low-level/networky?

packages/react-tinacms-github/src/toolbar-plugins/createMarkdownDeleteAction.tsx
logan-anderson: The reason I want this route was for two reasons
1. I wasn't sure if we are guaranteed 100% of the time that `form.name` would always have the path to the file.
2. I wanted to make sure that folks had a way to set how the title was retrieved
I good use case of this would be if I wasn't the title to be "are you sure you want to delete <title of blog post>"
someone could say
```ts
 const deleteAction = createMarkdownDeleteAction({
    getTitle: (form)=>{return form.values.frontmatter.some.nested.title},
  })
```
and for some reason, if the `form.name was` not not the name of the file then 
```ts
 const deleteAction = createMarkdownDeleteAction({
    getFilePath: (form)=>{return form.values.path.to.file},
  })
we might not need this one but I think the first one would be useful for sure.

packages/react-tinacms-github/src/toolbar-plugins/createMarkdownDeleteAction.tsx
logan-anderson: It totally can be a fragment and probably should be

packages/react-tinacms-github/src/toolbar-plugins/createMarkdownDeleteAction.tsx
ncphillips: > I wasn't sure if we are guaranteed 100% of the time that form.name would always have the path to the file.

That's fair. I think it's an okay assumption to make for now.

> I wanted to make sure that folks had a way to set how the title was retrieved

`form.label` would be the best bet there. When using the Sidebar, that's what is displayed in the list of forms. It defaults to the ID but it can be set manually.

packages/react-tinacms-github/src/toolbar-plugins/createMarkdownDeleteAction.tsx
ncphillips: Might makes styling simpler for @spbyrne?

packages/react-tinacms-github/src/toolbar-plugins/createMarkdownDeleteAction.tsx
logan-anderson: There may be a problem with `form.label`.  One may have a blog page with a form called `Edit blog page`. If we use form.label the prompt will say `do you want to delete Edit blog page` but if we allow the user to define how the label gets set they can point it to a custom name for the file so the prompt would say like `Do you want to delete <Blog title>`

ncphillips: Interesting. That's not really how the form label was intended to be used, but I see your point. Let's keep this for now.

dwalkr:
This looks like it could be used for other file formats as well. What about this makes it markdown-specific?

ncphillips:
Great catch @dwalkr. This should just be createGithubDeleteAction or something

packages/react-tinacms-github/src/toolbar-plugins/createMarkdownDeleteAction.tsx
dwalkr: the form label also appears at the top of the sidebar when editing the form. When only one form is used on a page, using a label of "Edit post" or similar feels more natural than putting the title of the post there.

logan-anderson:
> This looks like it could be used for other file formats as well. What about this makes it markdown-specific?

@dwalkr @ncphillips  Yeah I totally agree.

packages/react-tinacms-github/src/toolbar-plugins/createMarkdownDeleteAction.tsx
logan-anderson: I don't think there would be anything useful here.

packages/react-tinacms-github/src/toolbar-plugins/createMarkdownDeleteAction.tsx
logan-anderson: Should be fixed now

logan-anderson:
@ncphillips  added docs here https://github.com/tinacms/tinacms.org/pull/544

PR #1321, opened 2020-07-08
feat: adds git:commit event to GitFile

### LINKED ISSUE #1316: api-git: Improve error communication on failed push

kendallstrautman:
Continuation of work started in #1318 

---

Should provide ability to fix #1316. 

When using the git client, a `git:commit` event will dispatch when `commit` is invoked from `GitFile`, returning the [response](https://developer.mozilla.org/en-US/docs/Web/API/Response). Devs can then parse the response to handle various alerts. 

```js
  React.useEffect(() => {
    cms.events.subscribe("git:commit", function handleCommitAlerts(event) {
      if (!event.response.ok) {
        cms.alerts.error("Something went wrong! Changes weren't saved")
      } else {
        cms.alerts.info("Content saved successfully!")
      }
    })
  }, [])
```
Checkout `demo-gatsby/src/components/bio.js` for an example. 

Also, I noticed that the helpers in `next-tinacms-markdown` & `next-tinacms-json` don't use the `GitFile` yet in their `onSubmit` functions...we should do that...As of now this event will only _just work_ with `useGitForm` in the Gatsby packages.


packages/@tinacms/git-client/src/git-file.ts
```diff
@@ -36,17 +36,19 @@ export class GitFile {
...
+    )
   }
 
   reset = () => {
     this.cms.api.git.reset({ files: [this.relativePath] })
   }
 
   write = (values: any) => {
-    this.git.writeToDisk!({
+    this.cms.api.git.writeToDisk!({
```
ncphillips, line 51: What prompted this switch?

kendallstrautman: Oh I was having trouble testing with `this.git`...commits weren't working so I thought it may have been a bug, but after switching back and rebuilding it seems to work fine. I'll change that.

PR #1324, opened 2020-07-10
feat: custom buttons configured on form

kendallstrautman:
Right now, you can only configure buttons on the sidebar. Based on [this convo](https://github.com/tinacms/tinacms/pull/1314#issuecomment-655562569), these changes allow for setting button text values in the form options object. This way, the buttons will be consistent in the form views (sidebar, modal) and the toolbar. You can reference `blog-post` in the `demo-gatsby` or `index` in `demo-next` for an example.

**example**
```js
  const [author, authorForm] = useJsonForm(data.dataJson, {
    label: "Author",
    fields,
    buttons: {
      save: "Commit",
      reset: "Discard",
    },
  })
```

_Note:_ This **changes the hard coded toolbar buttons text** to the default 'Save' & 'Reset' values. If there is no form registered, these defaults will show in the Toolbar, then when a form is registered, the overridden values will show. I could see this being weird for some CMS's if they have lots of pages where forms are registered and the toolbar button values are changing... Something to think about.

Also, if `sidebar.buttons` is set, that will **override the form buttons**, so folks sidebar configs don't break. But we should note that this is deprecated.

- [x] Documentation PR Approved


ncphillips:
One thing we could add to the docs would be a description of how someone could use composition so that they don't have to name the buttons every time:

**src/cms/use-form.ts**
```ts
import { useForm as baseUseForm, FormOptions, WatchableFormValues } from "tinacms"
export const useForm = (options: FormOptions, watch?: WatchableFormValues) => useForm({
  buttons: {
    save: "Banana",
    reset: "Peeling",
  }
}, watch)

**src/pages/index.js**
```js
import { useForm } from "../cms/use-form"

  useForm(...)
}
```

PR #1336, opened 2020-07-20
fix: hide title input from link modal

jpuri:
Prevents confusion when adding a link, as most of the time the `title` attribute should not be used on links anyway for accessibility concerns (see https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/title#Accessibility_concerns).

Screenshot 2020-07-20 at 5 29 11 PM

### DIFF
--- a/packages/react-tinacms-editor/src/plugins/Link/Popups/Form.tsx
+++ b/packages/react-tinacms-editor/src/plugins/Link/Popups/Form.tsx
@@ -85,11 +85,11 @@ export const LinkForm = () => {
   const { arrowOffset, left, top } = position || {}
   const { state, dispatch } = view
   let href = ''
-  let title = ''
+  // let title = ''
   const linkMark = getMarkPresent(state, state.schema.marks.link)
   if (linkMark) {
     href = linkMark.attrs.href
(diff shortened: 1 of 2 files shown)


jpuri:
It can be removed, I left it there for just in case we want to later enable it with more intuitive designs.

jpuri:
Hey @DirtyF : in case you do not see it being added back let me remove it from codebase.

ncphillips:
I get that there's usability issues with the title input, but removing it entirely feels like the wrong solution to me. The `title` field is part of links [to help people browsing](https://www.nngroup.com/articles/title-attribute/) the website. I wouldn't be at all surprised if we start getting requests to support editing the title field.

DirtyF:
`title` will only help if you hover a link with a mouse. You can provide a similar function with a more accessible tooltip, else [this content is not accessible to a lot more people](https://developer.paciellogroup.com/blog/2013/01/using-the-html-title-attribute-updated/), the one using touch screen devices for instance.

ncphillips:
I guess so. I'll merge anyway, but we're not providing the ability to do those things.

PR #1338, opened 2020-07-22
Add react-tinacms-strapi

ncphillips:


packages/react-tinacms-strapi/src/api.ts
```diff
@@ -0,0 +1,20 @@
+import Cookies from 'js-cookie'
+import { STRAPI_JWT } from './tina-strapi-client'
+
+export async function fetchGraphql(query: string, variables = {}) {
```
ncphillips, line 4: - [ ] Move to `StrapiClient`

packages/react-tinacms-strapi/src/tina-strapi-client.ts
```diff
@@ -0,0 +1,11 @@
+export const STRAPI_JWT = 'tina_strapi_jwt'
+
+export class TinaStrapiClient {
+  async authenticate(username: string, password: string) {
+    return fetch(`${process.env.STRAPI_URL}/auth/local`, {
```
ncphillips, line 5: - [ ] Accept `STRAPI_URL` as a constructor argument so this package doesn't rely on a specific environment variable

packages/react-tinacms-strapi/src/strapi-provider.tsx
```diff
@@ -0,0 +1,53 @@
...
+  }
+
+  const onClose = async () => {
+    setActiveModal('none')
+  }
+
+  const onAuthSuccess = async () => {
+    onLogin()
+    setActiveModal('none')
+  }
```
ncphillips, line 30: These don't need to be async

```suggestion
  const beginAuth = () => {
    setActiveModal('authenticate')
  }

  const onClose = () => {
    setActiveModal('none')
  }

  const onAuthSuccess = () => {
    onLogin()
    setActiveModal('none')
  }
```

packages/react-tinacms-strapi/src/strapi-media.store.ts
```diff
@@ -0,0 +1,54 @@
...
+    return process.env.STRAPI_URL + fileUrl
+  }
+}
+
+export async function uploadFile(file: File) {
+  const authToken = Cookies.get(STRAPI_JWT)
+  const formData = new FormData()
+  formData.append('files', file)
+  const uploadResponse: Response = await fetch(
+    process.env.STRAPI_URL + '/upload',
```
ncphillips, line 40: Is there a reason why this function is not a method on `StrapiMediaStore`? 

I would lean towards 
1. Make it a method
1. Remove the `process.env.STRAPI_URL` access in favor of an instance variable set on media store at initialization.

packages/react-tinacms-strapi/src/strapi-auth-modal.tsx
```diff
@@ -0,0 +1,190 @@
...
+
+  // poll the cookie value for a change. close the auth window on change
+  // there are no native JS events that support this behaviour
+  window.setInterval(() => {
+    const currentCookie = Cookies.get(STRAPI_JWT)
+    if (currentCookie && currentCookie != previousCookie) {
+      if (authTab) authTab.close()
+      onAuthSuccess()
+    }
+  }, 1000)
```
ncphillips, line 158: Should this clear the interval at some point?

packages/react-tinacms-strapi/src/strapi-auth-modal.tsx
```diff
@@ -0,0 +1,190 @@
...
+        close={close}
+        onAuthSuccess={onAuthSuccess}
+        error={error}
+        onSubmit={async (values: LoginFormFieldProps) => {
+          strapi
+            .authenticate(values.username, values.password)
+            .then(async (response: Response) => {
+              if (response.status != 200) {
+                cms.events.dispatch({ type: 'strapi:error' })
+                console.log(response)
```
ncphillips, line 49:
```suggestion
```

DirtyF:
@ncphillips How can we test this package locally, I went through the Strapi Guide and tried to use the webpack-helpers package and amened the project' config file as described in https://github.com/tinacms/tinacms/tree/master/packages/%40tinacms/webpack-helpers#nextjs but when running `yarn` in the Strapi blog project, I  get: 

```
error Received malformed response from registry for "react-tinacms-strapi". The registry may be down.
``ù

ncphillips:
It's a bit funky to test because it's not published at all. We were testing locally by copyin the folder to an external node_modules. All is good, I'm going to merge and do a pre-release so we can test it out.

PR #1343, opened 2020-07-22
docs: update next package docs

kendallstrautman:
Copies over package docs from the .org repo


ncphillips:
As an aside, I think we should deprecate `next-tinacms-markdown` and `next-tinacms-json` and move all of their code into `next-tinacms-git`. That's how we did it with the Github domain

PR #1346, opened 2020-07-23
fix(react-tinacms-github): github  form re-created when file path cha…

logan-anderson:
…nges
Without this change, the changing of routes doesn't result in the form changing.

### DIFF
--- a/packages/react-tinacms-github/src/form/useGithubFileForm.ts
+++ b/packages/react-tinacms-github/src/form/useGithubFileForm.ts
@@ -35,26 +35,29 @@ export const useGithubFileForm = <T = any>(
     serialize: options.serialize,
   })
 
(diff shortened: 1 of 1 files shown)


packages/react-tinacms-github/src/form/useGithubFileForm.ts
```diff
@@ -35,26 +35,29 @@ export const useGithubFileForm = <T = any>(
...
+              `Saved Successfully: Changes committed to ${github.workingRepoFullName}`
+            )
+          })
+          .catch((error: any) => {
+            return { [FORM_ERROR]: error }
+          })
+      },
     },
-  })
+    [file.fileRelativePath]
```
ncphillips, line 59: I was totally wrong about this. This won't fix the issue

PR #1351, opened 2020-07-27
feat: focus ring on editor

### LINKED ISSUE #1347: Focus ring for inline wysiwyg field
**Issue**
Help editors understand when they can start typing in the editor, this issue was spotted during user testing session, it's easy to think you type in the editor, when you haven't, because it's hard to tell.

**Demo**
https://www.loom.com/share/349ca847002843dba5bafdaa7127cc0a

**Expected**

> It would probably make sense to at least have the same focus ring API on the wysiwyg inline field, even if we decide to default it to 'false'. The current inline field focus ring is out of date and it needs to be updated to match the 'new' focus ring used on inline blocks. — @spbyrne

jpuri:
Focus ring works well except the scenario when code block or link modal is focused, I am working to fix those.

Fix #1347

### DIFF
--- a/packages/react-tinacms-editor/src/components/MarkdownEditor/index.tsx
+++ b/packages/react-tinacms-editor/src/components/MarkdownEditor/index.tsx
@@ -100,7 +100,10 @@ const EditingSection = styled.textarea`
   margin: 10px 0;
   resize: none;
   width: 100%;
+  border: 1px solid transparent;
   &:focus {
     outline: none;
+    border: 1px solid #2296fe;
+    border-radius: 10px;
   }
 `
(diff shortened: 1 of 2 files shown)


ncphillips:
There's a [FocusRing](https://github.com/tinacms/tinacms/blob/master/packages/react-tinacms-inline/src/styles/focus-ring.tsx#L32) component in the `react-tinacms-inline` package. @spbyrne do you think that the `InlineWysiwyg` should be using that component to add a focus ring?

spbyrne:
Probably, but it may have to be adapted. I'll need to take a look.

spbyrne:
@jpuri I've added the focusRing on a new branch, I'll create a PR today. The focus ring component wraps inline fields, their code doesn't need to modified in any way to accommodate it.

PR #1356, opened 2020-07-28
Use form load initial values

ncphillips:
The `useForm` hook was making a weird assumption that `loadInitialValues` should never execute if the CMS is disabled. This PR undoes that assumption and pushes the responsibility to the client. If a developer does not want to make a request in prod, then it's the their responsibility to make sure the `loadInitialValues` they define works that way.


kendallstrautman:
Related to #1155

PR #1358, opened 2020-07-29
Focus Ring Fixes

### LINKED ISSUE #1347: Focus ring for inline wysiwyg field

spbyrne:
This standardizes all inline fields with the same focus ring component/API and removes the old focus ring. 

closes #1347 

- [x] Docs PR approved


ncphillips:
There's a ton of duplication here. 

We could at least extract a common hook, or better yet we could push all that behaviour into the FocusRing component itself

spbyrne:
Working on moving all the logic inside focusRing now.

packages/react-tinacms-inline/src/styles/focus-ring.tsx
```diff
@@ -19,7 +19,7 @@ import styled, { css } from 'styled-components'
 import { BlocksEmptyState } from '../blocks/inline-field-blocks'
 
 export interface FocusRingStyleProps {
-  offset?: number | { x: number; y: number }
+  offset?: number | { x: number; y: number } | undefined
```
kendallstrautman, line 22: `undefined` is implied here by the optional `?`

packages/react-tinacms-editor/src/tinacms-inline/inline-wysiwyg.tsx
```diff
@@ -18,28 +18,63 @@ limitations under the License.
...
-import { InlineField } from 'react-tinacms-inline'
+import { InlineField, FocusRing, useInlineForm } from 'react-tinacms-inline'
+import { FocusRingStyleProps } from 'react-tinacms-inline/src/styles'
 import { Wysiwyg } from '../components/Wysiwyg'
 import { EditorProps } from '../types'
 
 export interface InlineWysiwygFieldProps extends Omit<EditorProps, 'input'> {
   name: string
   children: any
+  focusRing?: boolean | FocusRingStyleProps
```
kendallstrautman, line 29: Does it work if you pass `true`? Curious just cause the block interface is `false | FocusRingStyleProps` so we should update that one as well if were going to have it accept any boolean

kendallstrautman: I can test this also if you're looking for another pair of eyes

packages/react-tinacms-editor/src/tinacms-inline/inline-wysiwyg.tsx
```diff
@@ -18,28 +18,63 @@ limitations under the License.
...
 }
 
 export function InlineWysiwyg({
   name,
   children,
+  focusRing = false,
```
kendallstrautman, line 35: Should the focus ring render by default? As it does with blocks

ncphillips: Personally I think defaulting to false makes more sense but I'm not too concerned

packages/react-tinacms-editor/src/tinacms-inline/inline-wysiwyg.tsx
spbyrne: Yeah I'm working on standardizing this on a boolean since you may want to enable a focusRing on something without specifying any custom options. Though that was designed around the WYSIWYG having the focusRing disabled by default, which I may change.

packages/react-tinacms-editor/src/tinacms-inline/inline-wysiwyg.tsx
spbyrne: I suppose everything could have it enabled by default.

packages/react-tinacms-inline/src/styles/focus-ring.tsx
spbyrne: Yeah I can try changing this. I think I was getting a type error and this fixed it but it shouldn't have so I'll try reversing.

packages/react-tinacms-editor/src/tinacms-inline/inline-wysiwyg.tsx
spbyrne: I also want to make the WYSIWYG toolbar hidden when the focus isn't active since it will get a hover focus ring to indicate editing is possible, but that's another PR.

packages/react-tinacms-inline/src/styles/focus-ring.tsx
```diff
@@ -15,21 +15,70 @@ See the License for the specific language governing permissions and
...
   active: boolean
   disableHover?: boolean
   disableChildren?: boolean
 }
 
-export const FocusRing = styled.div<FocusRingProps>(p => {
+export interface FocusRingProps {
+  name?: string
+  children?: any
+  options?: boolean | FocusRingOptions
```
ncphillips, line 40: My gut says that this `options` shouldn't be on `FocusRingProps`, but that it should be `FocusRingProps extends FocusRingOptions` but I know this situation is kind of weird so that's fine I guess

jpuri:
Hey @spbyrne : this looks good 👍 

I have a doubt about wysiwym. Wysiwym has complex structure, can you plz check and ensure that it works as expected even when link / image modal inputs are focused in wysiwyg or code block is active ?

spbyrne:
There are a lot of events that can trigger a focus loss given how that's handled (Dismissible onDismiss function on a wrapper) including interacting with a settings modal, clicking an item within a group that's not editable, interacting with links or images in the WYSIWYG..etc. In all cases it doesn't interrupt editing, the focus ring just goes away. The WYSIWYG does seem the most prone to losing focus, even double clicking to edit will remove the focus.

@ncphillips I'm not sure if this is an issue to solve now or not.

packages/react-tinacms-inline/src/styles/focus-ring.tsx
spbyrne: Right, I think it was using the 'extends' syntax before I made my changes. I was just confused so I did what made sense to me at the time but I figured I'd need help improving the interfaces.

ncphillips:
Yeah this is really weird I'll take a look

DirtyF:
On a side note, to make sure we're on the same page @jpuri: It's perfectly OK to put the focus on the current dialog being used, that's the whole point of focus. When you open a modal for image upload, or to add a link URL the focus is on those, the wysiwyg editor gets the focus back when user is editing the text inside the editor.

kendallstrautman:
I added a checkbox up top but we'll need to make docs changes

ncphillips:
This is totally unrelated:

@all-contributors please add @laurenclark for docs

jpuri:
> Yeah this is really weird I'll take a look

Losing focus is actually not weird for wysiwyg. Wysiwyg is not single input component but it is group of complex input components, main editing area, code block, modal inputs are all focusable.

Thus though focus ring works well in markdown mode it disappears as main editing are loses focus in wysiwyg more.

The [PR](https://github.com/tinacms/tinacms/pull/1351/files#diff-4146d4e56c2941ffc14c5208f0d06077R92) I created did not depended on browser's focused state to show border but did its own check. We need something similar and more complex for wysiwyg mode of editing.

ncphillips:
Yeah I meant the behaviour is weird for the end user. It's not weird that it's happening. 

Right now the FocusRing and the Wysiwyg have their own focus detection logic. A future change should allow the FocusRing to delegate that responsibility to it's child somehow. Maybe something like:

```tsx
<FocusRing {...etc} >
  {({ onFocus, onBlur}) => <Wysiwyg onFocus={onFocus} onBlur={onBlur} />
</FocusRing>
```

FYI the `FocusRing` does a lot more then just add the visual-interface around the component, that's why we don't want that behaviour to be duplicated inside the Wysiwyg itself.

PR #1373, opened 2020-08-04
fix: enabling CMS does not remount children

### LINKED ISSUE #1366: `TinaProvider` remounts children on enabled / disabled
## Description

Noticed this during the getting started tutorial set up and wanted to make an issue so we don't forget. When toggling enabled states, the entire app will be remounted since the `TinaProvider` is rendering it without the UI and theme etc.

This may not necessarily be a bug, but it's not ideal. Not sure how we'd get around this — something to think about.

### Steps to reproduce

Add a console.log on a page component on initial mount. Toggle edit mode and you'll see it remounts.

kendallstrautman:
closes #1366

### DIFF
--- a/packages/@tinacms/react-sidebar/src/components/Sidebar.tsx
+++ b/packages/@tinacms/react-sidebar/src/components/Sidebar.tsx
@@ -170,6 +170,7 @@ const SidebarToggle = ({ sidebar }: { sidebar: SidebarState }) => {
     <SidebarToggleButton
       onClick={() => (sidebar.isOpen = !sidebar.isOpen)}
       open={sidebar.isOpen}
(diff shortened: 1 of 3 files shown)


packages/@tinacms/react-sidebar/src/components/Sidebar.tsx
```diff
@@ -170,6 +171,7 @@ const SidebarToggle = ({ sidebar }: { sidebar: SidebarState }) => {
     <SidebarToggleButton
       onClick={() => (sidebar.isOpen = !sidebar.isOpen)}
       open={sidebar.isOpen}
+      aria-label="toggles cms sidebar"
```
DirtyF, line 174: ❤️

PR #1376, opened 2020-08-05
feat(react-tinacms-inline): InlineText now accepts a placeholder prop

logan-anderson:
example

```html
 <h1>
          <InlineText name="thisIsnotHEre" placeholder="text" />
  </h1>
```


chrisdmacrae:
This should also be added to `InlineTextarea` @ncphillips @logan-anderson

kendallstrautman:
Can we update the docs pleez

PR #1380, opened 2020-08-06
feat: InlineWysiwyg default image upload function

ncphillips:


packages/react-tinacms-editor/README.md
```diff
@@ -135,7 +136,27 @@ export function Page(props) {
 }
 ```
 
-### Example: Dynamic Imports
+### Sticky Menu
+
+When editing long content it is likely that the editor will scroll past the wysiwyg menu.
+
+By adding the `sticky` property the menu will follow the user as they scroll through the page.
```
DirtyF, line 143: According to the docs sticky is the default, right?

DirtyF:
Just tried to test on the demo-gatsby package and got an image/undefined error after drag and dropping the image:

tinacms 2020-08-06 à 20 59 29

packages/react-tinacms-editor/README.md
ncphillips: Nope, default is non-sticky

ncphillips:
I can't reproduce that 🤔

packages/react-tinacms-editor/src/tinacms-inline/inline-wysiwyg.tsx
```diff
@@ -33,10 +33,27 @@ export function InlineWysiwyg({
...
+            directory: passedInImageProps?.directory || '',
+            file,
+          }))
+        )
+
+        return allMedia.map(media => `${media.directory}${media.filename}`)
+      },
+      ...passedInImageProps,
+    }
+  }, [cms.media.store])
```
dwalkr, line 55:
```suggestion
  }, [cms.media.store, passedInImageProps])
```
should we add this here?

DirtyF:
@ncphillips OK I get what happened here, when you try to upload an image that already exists, you get the undefined.

https://www.loom.com/share/98717e9589a64689908402a685e6ed30

ncphillips:
Oh! That's interesting. Unrelated to this PR but let's make note

packages/react-tinacms-editor/src/tinacms-inline/inline-wysiwyg.tsx
ncphillips: Maybe. It would probably make the hook rerun on every render

dwalkr: maybe we're memoizing the wrong thing then

```ts
const defaultUpload = React.useCallback((files: File[]) => {
  //...
}, [cms.media.store])

const imageProps = {
  upload: defaultUpload,
  ...passedInImageProps
}
```

Issue #1307, opened 2020-07-01
State of image preview in sidebar out of step with what displays on page


aratner:
## Description

- _Originally posted_ [here](https://community.tinacms.org/t/image-preview-out-of-step-with-what-displays-on-page/208/6)
- _Link to example repo_ [here](https://github.com/alexanderratner/template-tests/tree/master/gatsby-tina-test)

When I upload a new file the image on the page (in the content area) updates, but the preview stays the previous image/state. 

When I refresh the browser both load to the correct state.

I’m also getting the following error in the console, not sure if they are related?

```
Warning: Can't perform a React state update on an unmounted component. This is a no-op, but it indicates a memory leak in your application. To fix, cancel all subscriptions and asynchronous tasks in a useEffect cleanup function.
```

### Steps to reproduce
1. Click on Header section in sidebar
2. Click on image preview in sidebar to upload new image
3. Image in main content area will update to the new uploaded image (json source file points to correct file as well), but the image preview in the sidebar stays stuck in the previous image/state.
4. Click on image preview in sidebar to upload another image
5. Image in main content area will update to this new image, but the image preview in the sidebar will now show the image you tried to upload from step 2. 
6. Reload page, image preview in sidebar will now matches image in content area.

### Expected result

Preview image in sidebar and image in content area should update to same image (without page refresh)

#### Tinacms Package Versions

```
├─┬ gatsby-plugin-tinacms@0.2.2
│ └─┬ tinacms@0.21.2
│   ├── @tinacms/alerts@0.1.4
│   ├── @tinacms/core@0.9.0
│   ├─┬ @tinacms/fields@0.12.0
│   ├─┬ @tinacms/form-builder@0.3.5
│   ├─┬ @tinacms/forms@0.6.0
│   ├── @tinacms/icons@0.14.0
│   ├── @tinacms/media@0.1.0
│   ├── @tinacms/react-alerts@0.1.9
│   ├─┬ @tinacms/react-core@0.3.0
│   ├─┬ @tinacms/react-forms@0.2.0
│   ├── @tinacms/react-modals@0.1.9
│   ├── @tinacms/react-screens@0.1.8
│   ├─┬ @tinacms/react-sidebar@0.2.0
│   ├─┬ @tinacms/react-toolbar@0.2.0
│   ├─┬ @tinacms/styles@0.5.1
├─┬ gatsby-tinacms-git@0.5.4
│ ├─┬ @tinacms/api-git@0.7.0
[...]

PR #1386, opened 2020-08-10
MediaStore#previewSrc

ncphillips:
This RFC introduces `MediaStore#previewSrc`. This allows `MediaStores` to provide a default method that can be overridden on a field-by-field basis

Todo:

Done
- [x] Implement `StrapiMediaStore#previewSrc`
- [x] `InlineWysiwyg` should expect `imageProps.parse` 
- [x] `InlineImage`
  - Needs to work with async previewSrc
- [x] Introduces `MediaStore#previewSrc`
- [x] `InlineWysiwyg` 
- [x] `ImageFieldPlugin`
- [x] Implement `GithubMediaStore#previewSrc`
- [x] Implement `GitMediaStore#previewSrc`

Note: READMEs don't really require any updates from this PR

### Test with GithubMediaStore on Tinacms.org
Setup
* Add the `parse` function to the `InlineWysiwyg` wrapper.
* Run with local TINA version
* Turn on the sidebar and add a dummy `image` field to the sidebar
* Make sure to create a new testing branch
* Add an Inline Image Field somewhere

Tests
- [ ] Dragging images onto the Editor
- [ ] Using the image menu in the Editor 
- [ ] Sidebar image field
- [ ] Inline Image field

### Test GitMediaStore in `demo-gatsby`

- [x] Editor on a Blog
- [x] Inline Image on above and right of the blog body
- [x] Sidebar image

### Test StrapiMediaStore a site built by going through the guide
- [ ] Setup images on the wysiwyg
- [ ] Remove `previewSrc` from the InlineImage example

### DIFF
--- a/packages/react-tinacms-github/src/github-media-store/index.ts
+++ b/packages/react-tinacms-github/src/github-media-store/index.ts
@@ -50,4 +50,12 @@ export class GithubMediaStore implements MediaStore {
 
     return uploaded
   }
+
+  async previewSrc(src: string) {
+    try {
+      return this.githubClient.getDownloadUrl(src)
+    } catch {
+      return src
+    }
+  }
 }
--- a/packages/react-tinacms-inline/src/fields/inline-image-field.tsx
+++ b/packages/react-tinacms-inline/src/fields/inline-image-field.tsx
@@ -25,9 +25,9 @@ import { useState, useEffect } from 'react'
 
 export interface InlineImageProps {
   name: string
-  previewSrc(formValues: any): string
   parse(filename: string): string
   uploadDir(form: Form): string
+  previewSrc?(formValues: any): string
   focusRing?: boolean | FocusRingOptions
   children?: any
 }
--- a/packages/react-tinacms-inline/src/fields/inline-image-field.tsx
+++ b/packages/react-tinacms-inline/src/fields/inline-image-field.tsx
@@ -37,60 +37,72 @@ export interface InlineImageProps {
  */
 export const InlineImageField = InlineImage
 
-export function InlineImage({
+export function InlineImage(props: InlineImageProps) {
+  const cms = useCMS()
+
+  return (
+    <InlineField name={props.name}>
+      {({ input, form }) => {
+        if (cms.enabled) {
+          return <EditableImage {...props} input={input} form={form} />
+        }
+
+        return props.children ? props.children() : <img src={input.value} />
+      }}
+    </InlineField>
+  )
+}
+
+interface EditableImageProps extends InlineImageProps {
+  input: any
+  form: any
+}
+
+function EditableImage({
+  form,
+  input,
   name,
   previewSrc,
   uploadDir,
   parse,
   children,
   focusRing = true,
-}: InlineImageProps) {
+}: EditableImageProps) {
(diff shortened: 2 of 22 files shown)


packages/react-tinacms-github/src/github-media-store/index.ts
```diff
@@ -50,4 +50,12 @@ export class GithubMediaStore implements MediaStore {
 
     return uploaded
   }
+
+  async previewSrc(src: string) {
+    try {
+      return this.githubClient.getDownloadUrl(src)
+    } catch {
+      return src
```
dwalkr, line 58: Not sure we should smother this error. I'd say either catch it higher in the call stack or give media store access to the event bus like we're planning with API clients. That doesn't have to be addressed in this PR tho.

In the short term I would prefer to either remove the try-catch or, if we're trying to avoid blowing up someone's whole app, at least returning null / `''`. Returning the original src could cause some unexpected behavior

packages/react-tinacms-inline/src/fields/inline-image-field.tsx
```diff
@@ -21,12 +21,13 @@ import { InlineField } from '../inline-field'
...
 import { useDropzone } from 'react-dropzone'
 import { FocusRing, FocusRingOptions } from '../styles'
+import { useState, useEffect } from 'react'
 
 export interface InlineImageProps {
   name: string
-  previewSrc(formValues: any): string
   parse(filename: string): string
   uploadDir(form: Form): string
+  previewSrc?(formValues: any): string
```
dwalkr, line 30:
```suggestion
  previewSrc?(formValues: any): string | Promise<string>
```

packages/react-tinacms-github/src/github-media-store/index.ts
ncphillips: My thought was that you might have an image URL that references a whole URL i.e.

```markdown
![](http://i.ytimg.com/vi/ypG3yBxgDEs/maxresdefault.jpg
```

GithubMediaStore would fail to find that in the repo and just default ot the src

dwalkr: The only time this would really fall over is if you are overwriting an existing prod image, `getDownloadUrl` fails, and a root-relative path to the image is returned. User would then see the prod image in the preview, with the appearance that everything is working normally.

I see your point though. IMO, this is actually a significant pitfall when creating a media store. I would be OK with leaving this as-is for now, but it's something to chew on.

kendallstrautman:
Image stuff isn't working in `demo-gatsby`

kendallstrautman:
Oh nvm I think it was just a duplicate image issue. Man we need to fix that or give better feedback. For some reason I thought we fixed the duplicate stuff?

kendallstrautman:
I do see a discrepancy with the sidebar preview image not being updated immediately, even though it properly updates on the page. I see that was [an issue](https://github.com/tinacms/tinacms/issues/1307) made a while back though. We should fix that.

ncphillips:
We fixed the duplicate thing for GitHub, but not Git

The sidebar discrepancy is likely a Gatsby-only problem but not sure where it originates

PR #1391, opened 2020-08-11
Inline text children

ncphillips:
Backwards compatible change.

When inside a form with `title = "Hello World"`

**Default Usage**

```jsx
<InlineText name="title" />
```

Output when `cms.disabled` will be:
```html
Hello World
```

**With Children**

```jsx
<InlineText name="title">
  Not Editing
</InlineText>
```

Output when `cms.disabled` will be:

```html
Not Editing
```
---
- [ ] Docs


chrisdmacrae:
@ncphillips would like to see this on `InlineTextarea` as well

ncphillips:
It is

chrisdmacrae:
> It is

I can't read apparently

PR #1404, opened 2020-08-12
Allow configuration of react dnd draggable div with InlineBlocks

chrisdmacrae:
This PR allows `InlineBlocks` to take a renderChild that allows you to configure the `<div>` with `react-dnd`s `providedRef`.

It also logs when a missing block template is found, to notify the developer.

This is because a WIP inline field we're using requires two instances of InlineBlocks to be direct DOM dependencies of each other, and currently `InlineBlocks` places a div between them.


chrisdmacrae:
@tinacms/maintainers this is good to go.

packages/react-tinacms-inline/README.md
```diff
@@ -492,5 +492,46 @@ const PAGE_BLOCKS = {
...
 
+**Configuring the drag and drop wrapper**
+
+`InlineBlocks` wraps your blocks with a `<div>` element that informs the drag and drop functionality of what can be dragged and dropped.
+
+This can be an issue if your styles require direct inheritance, such as a flexbox grid:
+
+```html
+<div class="row">
+  <div class="column>
```
dwalkr, line 503:
```suggestion
  <div class="column">
```

packages/react-tinacms-inline/src/blocks/inline-field-blocks.tsx
```diff
@@ -42,8 +42,18 @@ export interface InlineBlocksProps {
   }
   min?: number
   max?: number
+  components?: {
```
kendallstrautman, line 45: We need to update the `InlineBlocks` interface on the README to include this

PR #1407, opened 2020-08-13
chore: fallback if there's no parse function

ncphillips:


packages/react-tinacms-editor/src/tinacms-inline/inline-wysiwyg.tsx
```diff
@@ -42,14 +42,20 @@ export function InlineWysiwyg({
...
+        }))
+
+        const allMedia = await cms.media.store.persist(filesToUpload)
 
-        return allMedia.map(media => passedInImageProps.parse(media.filename))
+        return allMedia.map(media => {
+          if (passedInImageProps.parse) {
+            return passedInImageProps.parse(media.filename)
+          } else {
+            return media.filename
```
dwalkr, line 56: We are essentially creating a default parse function here. What if we prepended the directory here by default?

PR #1415, opened 2020-08-18
gatsby-plugin-tinacms: pass enabled option

DirtyF:
According to the docs: https://tinacms.org/docs/releases/2020-08-03#upgrading-from-v0250,  you have to pass `enabled` since 0.26.0.

Context: being able to upgrade tina-starter-grande to latest Tina

### DIFF
--- a/packages/gatsby-plugin-tinacms/gatsby-ssr.tsx
+++ b/packages/gatsby-plugin-tinacms/gatsby-ssr.tsx
@@ -25,7 +25,7 @@ exports.wrapRootElement = (
   options: GatsbyPluginTinacmsOptions
 ) => {
   const cms = new TinaCMS({
-    enabled: true,
+    enabled: options.enabled,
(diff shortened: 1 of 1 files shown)


packages/gatsby-plugin-tinacms/gatsby-ssr.tsx
```diff
@@ -25,6 +25,7 @@ exports.wrapRootElement = (
   options: GatsbyPluginTinacmsOptions
 ) => {
   const cms = new TinaCMS({
+    enabled: true,
```
ncphillips, line 28:
```suggestion
    enabled: options.enabled,
```

DirtyF: Right, even better.

ncphillips:
This is only in `gatsby-ssr.js` does it also need to change in `gastby-node.js`?

PR #1430, opened 2020-08-24
Updates `demo-next`

kendallstrautman:
This demo was using outdated form helpers


kendallstrautman:
The blocks page on this demo is broken. React DND is throwing an error about `innerRef`...I'm assuming its related to changes from this pr #1404

PR #1431, opened 2020-08-24
chore(react-tinacms-inline): html element sets innerRef

kendallstrautman:
According to [their docs](https://github.com/atlassian/react-beautiful-dnd/blob/master/docs/guides/using-inner-ref.md#a-common-error-), we need to set `provider.innerRef` directly on a DOM Node or else our 'application will explode', which is what I was seeing in `demo-next`.


kendallstrautman:
Closing, duplicate fix from #1432

PR #1446, opened 2020-09-05
fix: table and link focus

jpuri:
@DirtyF : the pR fixes 2 issues that we discussed:

1. as table is created focus is set to first cell in it
2. link modal does not open as link is clicked but only after link menu option is clicked

record


DirtyF (changes requested):
✅  The table focus is fixed. 
❌  CMD-K Keyboard do not open the link modal anymore

jpuri:
Hi @DirtyF : the issue with link popup shortcut is fixed now.

DirtyF (approved):
Thank you @jpuri this works great now, tested with Firefox, Chrome, and Safari on macOS 👌 

https://www.loom.com/share/9e3b9923f66f44e6b76811dae9db3abf

PR #1448, opened 2020-09-05
feat: keyboard shortcut for toggle editor mode

jpuri:
Adding keyboard shortcut `Cmd-/` or `Ctrl-/` for mode toggling in editor.

### DIFF
--- a/packages/react-tinacms-editor/src/context/editorMode.tsx
+++ b/packages/react-tinacms-editor/src/context/editorMode.tsx
@@ -17,7 +17,7 @@ limitations under the License.
 */
 
 import * as React from 'react'
-import { useState } from 'react'
+import { useEffect, useState } from 'react'
 import { createContext, useContext } from 'react'
 
 const EditorModeContext = createContext<{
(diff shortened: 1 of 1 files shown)


DirtyF (commented):
We need to change the shortcut as it's already taken under Chrome and Firefox (fire help)

Maybe `Option-CMD-/` ?

jpuri:
I checked other editors and found that `Cmd-/` is mostly used to open help. For raw mode https://wordpress.org/gutenberg/ uses `alt-cmd-shift-M`. We can use check `alt-cmd-s` or `alt-cmd-t`.

DirtyF:
Why not the letter M (for Markdown)?

jpuri:
`alt-cmd-m`, `cmd-m` seems to be already used by chrome to  minimise window.

DirtyF:
@jpuri yes that's why going with `alt-cmd-shift-M` seems to be the way to go with the letterr M.

jpuri:
@DirtyF : PR is updated with use of key combination `alt-cmd-shift-M`.

DirtyF (approved):
This is so cool, thanks @jpuri

PR #1475, opened 2020-09-15
Previewsrc refactor

ncphillips:
The `previewSrc` signature for `MediaStore`, `InlineImageField`, and `ImageFieldPlugin` are all identical now.

### DIFF
--- a/packages/react-tinacms-inline/src/fields/inline-image-field.tsx
+++ b/packages/react-tinacms-inline/src/fields/inline-image-field.tsx
@@ -18,16 +18,15 @@ limitations under the License.
 
 import * as React from 'react'
 import { InlineField } from '../inline-field'
-import { useCMS, Form, Media } from 'tinacms'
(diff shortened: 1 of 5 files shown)


packages/react-tinacms-inline/src/fields/inline-image-field.tsx
```diff
@@ -70,34 +69,12 @@ function EditableImage({
...
-    return () => {
-      canceled = true
-    }
-  }, [input.value])
+  const [_previewSrc] = usePreviewSrc(
+    input.value,
+    name,
+    form.values,
+    previewSrc
+  )
```
kendallstrautman, line 77: 🙌

kendallstrautman (approved):

kendallstrautman:
Just so I remember, the new api is:

```
{
  input.value,
  input.name,
  formValues
}
```

PR #1476, opened 2020-09-15
Refactors & styles media manager

kendallstrautman:
Extracted UI components and implemented basic theme styles.


packages/tinacms/src/components/media-manager.tsx
```diff
@@ -68,131 +71,234 @@ export function MediaPicker({ onSelect, close, ...props }: MediaRequest) {
   )
   const [offset, setOffset] = useState(0)
   const [limit, setLimit] = useState(props.limit || 50)
```
ncphillips, line 73:
```suggestion
  const [limit] = useState(props.limit || 50)
```

dwalkr (approved):

PR #1494, opened 2020-09-23
2nd and 3rd previewSrc args are optionsal

ncphillips:


dwalkr (approved):

packages/react-tinacms-editor/src/tinacms-inline/inline-wysiwyg.tsx
```diff
@@ -59,7 +59,7 @@ export function InlineWysiwyg({
       },
       previewSrc(src) {
         // TODO: Implement formValues correctly
```
kendallstrautman, line 61:
```suggestion
```

PR #1524, opened 2020-10-06
Wysiwyg images: refactor imageProps, parse, uploadDir

kendallstrautman:
- Refactors the image props generation from `InlineWysiwyg` to `Wysiwyg`
- Adjusts `parse` to handle media object
- Replaces `directory` string with `uploadDir` function

**Todo**
---
- [x] merge master
- [x] update README if needed

PR #1530, opened 2020-10-07
feat(react-tinacms-editor): wysiwyg opens media manager

### LINKED ISSUE #1472: Wysiwyg should integrate with the Media Manager

kendallstrautman:
## What this does 
- Refactors the image props generation from `InlineWysiwyg` to `Wysiwyg`
- Adjusts `parse` to handle [media object ](https://tinacms.org/docs/media/#media)— _**BREAKING CHANGE**_
- Replaces `directory` string with `uploadDir` function — _**BREAKING CHANGE**_
- When the image upload popup drop is clicked, the media manager opens (at the root) and images can be inserted from there. 

## Next steps: 
- The media manager should open from `uploadDir`, similar to the image fields.
- The `previewSrc` function still does not match the media store interface, it would need to be passed `path` and `formValues`. The team discussed this and is okay leaving it _for now._
- The image menu popup could be refactored to use react dropzone.

---
To Test: head to demo-next, and somewhat hidden `/blog/apple` route...there may be some broken links since I removed the new image upload commits, but try to add new ones or select from the media manager. 

_Note_: this was reopened from #1524 and this closes #1472


packages/react-tinacms-editor/src/plugins/Image/Menu/ProsemirrorMenu.tsx
```diff
@@ -60,13 +63,22 @@ export const ProsemirrorMenu = ({ uploadImages }: MenuProps) => {
     setImageUrl('')
   }
 
-  const uploadSelectedImage = (event: ChangeEvent) => {
-    const { files } = event.target as any
-    if (files[0]) {
-      uploadImageFile(files[0])
+  function onMediaSelect(media?: Media) {
+    console.log({ media })
```
kendallstrautman, line 67:
```suggestion
```

packages/react-tinacms-editor/src/tinacms-inline/inline-wysiwyg.tsx
```diff
@@ -51,13 +51,20 @@ export function InlineWysiwyg({
     <InlineField name={name}>
       {({ input, form }: InlineWysiwygRenderProps) => {
         return (
-          <FocusRing name={name} options={focusRing}>
+          <FocusRing name={input.name} options={focusRing}>
```
kendallstrautman, line 54:
```suggestion
          /**
           * Note: We use `input.name` not `name` here because
           * the given name is only relative to the block, not
           * the absolute path in the form.
           */
          <FocusRing name={input.name} options={focusRing}>
```

ncphillips (approved):

PR #1534, opened 2020-10-08
build: fix package-lock

ncphillips:


dwalkr:
lockfiles inside of packages don't really serve a purpose and aren't automatically updated by our lerna scripts, so I think we're better off just removing them all

PR #1540, opened 2020-10-08
chore(demo-gatsby): media updates

kendallstrautman:
Updates the demo to work with latest media changes. I had to adjust the way the selected media preview src was being generated with wysiwyg images...I'm not sure if this is the best way to handle the `previewSrc`. The previous code didn't account for promises. 

Also, a small style fix in the media manager.

### DIFF
--- a/packages/demo-gatsby/src/templates/blog-post.js
+++ b/packages/demo-gatsby/src/templates/blog-post.js
@@ -203,7 +203,7 @@ function BlogPostTemplate(props) {
                   return postDirectory
                 },
                 previewSrc(src) {
-                  const formattedSrc = src.replace(/\/public|\/static/, "")
+                  const formattedSrc = src.replace(/^\/public|\/static/, "")
                   return formattedSrc
                 },
               }}
--- a/packages/demo-gatsby/src/templates/blog-post.js
+++ b/packages/demo-gatsby/src/templates/blog-post.js
@@ -18,6 +18,7 @@ limitations under the License.
 
 import React from "react"
 import { Link, graphql } from "gatsby"
+import path from "path"
 
 import Bio from "../components/bio"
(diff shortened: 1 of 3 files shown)


kendallstrautman:
@ncphillips just needed the extra .3rem 😅

kendallstrautman:
@all-contributors please add @kud for docs

packages/demo-gatsby/src/templates/blog-post.js
```diff
@@ -189,24 +189,22 @@ function BlogPostTemplate(props) {
...
 
-                  let media = await cms.media.store.persist(
-                    files.map(file => ({
-                      directory,
-                      file,
-                    }))
-                  )
+                // Decide the file upload directory for the post
+                uploadDir: blogPost => {
+                  const postPathParts = blogPost.fileRelativePath.split("/")
```
dwalkr, line 197: could replace this whole function w/ a call to `path.dirname` https://nodejs.org/api/path.html#path_path_dirname_path

packages/demo-gatsby/src/templates/blog-post.js
```diff
@@ -189,24 +189,22 @@ function BlogPostTemplate(props) {
...
+                  return postDirectory
                 },
-                previewUrl(src) {
-                  return new Promise(resolve => {
-                    setTimeout(() => {
-                      resolve(src)
-                    }, 3000)
-                  })
+                previewSrc(src) {
+                  const formattedSrc = src.replace(/\/public|\/static/, "")
```
dwalkr, line 206:
```suggestion
                  const formattedSrc = src.replace(/^\/public|\/static/, "")
```

dwalkr (changes requested):

dwalkr (approved):

PR #1542, opened 2020-10-10
(@tinacms/git-client): git-media-store#list doesn't set previewSrc

kendallstrautman:
With this change, all media files will render with the `File` icon in the media manager list. This default prevents broken preview src states for media files. 

The assumption is that for preview src's to work in the media manager with any project using the git-client, they will need to extend the media store to tailor the `previewSrc` and `list` methods according to their project / framework.

### DIFF
--- a/packages/demo-next/next-git-media-store.js
+++ b/packages/demo-next/next-git-media-store.js
@@ -16,12 +16,30 @@ limitations under the License.
 
 */
 
-import { GitMediaStore } from '@tinacms/git-client'
+import { GitMediaStore, nextOffset } from '@tinacms/git-client'
 
 export class NextGitMediaStore extends GitMediaStore {
   previewSrc(src) {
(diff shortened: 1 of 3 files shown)


packages/demo-next/next-git-media-store.js
```diff
@@ -16,12 +16,30 @@ limitations under the License.
...
-import { GitMediaStore } from '@tinacms/git-client'
+import { GitMediaStore, nextOffset } from '@tinacms/git-client'
 
 export class NextGitMediaStore extends GitMediaStore {
   previewSrc(src) {
     return /jpg|jpeg|png|svg|gif$/.test(src.toLowerCase())
       ? src.replace('/public', '')
       : null
   }
+  async list(options) {
```
dwalkr, line 27: what if we just called `super.list` and filtered the results?

kendallstrautman: good call!

dwalkr (approved):

PR #1544, opened 2020-10-12
feat(react-tinacms-editor): media mgr opens from mediaDir

kendallstrautman:
A new `mediaDir` prop was added to the imageProps interface; this is where the media manager will open from. It defaults to the return value from `uploadDir` or the root directory if `uploadDir` & `mediaDir` aren't set.

### DIFF
--- a/packages/react-tinacms-editor/README.md
+++ b/packages/react-tinacms-editor/README.md
@@ -103,6 +103,7 @@ interface ImageProps {
   uploadDir?(form: Form): string
   upload?: (files: File[]) => Promise<string[]>
   previewSrc?: (url: string) => string | Promise<string>
+  mediaDir?: string
 }
 
 interface FocusRingProps {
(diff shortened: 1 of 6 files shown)


packages/react-tinacms-editor/README.md
```diff
@@ -154,6 +155,7 @@ To upload and manage images in the Wysiwyg, you'll need to configure `imageProps
 | `uploadDir?`    | Defines the upload directory. This function is passed the current form values.                                                                  |
 | `upload?`     | An asynchronous function that handles image upload. By default, this calls the `persist` function on the [media store](https://tinacms.org/docs/media). |
 | `previewSrc?`     | 	An asynchronous function that returns the path or url for the image `src` in preview or edit mode. By default, this calls the `previewSrc` function on the [media store](https://tinacms.org/docs/media)_.                             |
+| `mediaDir?`    | Defines where the media manager opens from. Defaults to the return value from `uploadDir`. If `uploadDir` isn't defined, falls back to the root directory.                                                                  |
```
dwalkr, line 158: Do you mean if `uploadDir` and `mediaDir` are both undefined, it falls back to the root?

dwalkr: I re-read it and I understand. I suggest we be annoyingly explicit to avoid confusion here:

```suggestion
| `mediaDir?`    | Defines the directory that the media manager opens to. Defaults to the return value from `uploadDir`. If neither `mediaDir` nor `uploadDir` are defined, falls back to the root directory.                                                                  |
```

PR #1546, opened 2020-10-12
feat(react-tinacms-editor): media mgr opens from uploadDir

kendallstrautman:


dwalkr (dismissed):

dwalkr (approved):

dwalkr (approved):

DirtyF:
Did a quick test and the media manager does open from the root folder in the Gatsby demo.
I couldn't insert an image though, thumbnails are not displayed and the "Upload" label is not displayed properly.

https://www.loom.com/share/581012d2357144398a342d37411a6352

dwalkr:
@DirtyF We don't have a good solution for Gatsby yet.

kendallstrautman:
@DirtyF the `uploadDir` isn't set on that demo (on the blog) so it will always open from the root. And it seems like you need to rebuilt because we fixes those styles issues — upload button and and broken images.

dwalkr (approved):

PR #1549, opened 2020-10-13
fix(react-tinacms-editor): fixes menu jumping to top on scroll

### LINKED ISSUE #1495: Sticky WYSIWYM menu aligns with top too soon
## Description

Screen Shot 2020-09-22 at 11 17 01 AM

There's some bugginess in how and when the WYSIWYM menu aligns in its _sticky_ top position. There's some visual glitching and in some cases the 'jump' to the top occurs too soon, causing the UI to seem broken until the user scrolls down.

### Steps to reproduce

Head to tinacms.org/community, enable the cms and scroll. You should see the above behavior. 

### Expected result

The menu aligns with the top when the bounds of the WYSIWYM component reach the _top_ of the window, without jumping or glitching.

### Actual result

The menu jumps / glitches to the top before it makes sense for it to do so .

kendallstrautman:
closes #1495


packages/react-tinacms-editor/src/components/BaseMenubar/index.tsx
```diff
@@ -63,17 +64,20 @@ export const BaseMenubar = ({
...
-      const endPosition = wysiwygWrapper
-        ? startPosition + wysiwygWrapper.offsetHeight
-        : 0
-
-      if (window.scrollY > startPosition && window.scrollY < endPosition) {
-        setMenuFixed(true)
-      } else {
-        setMenuFixed(false)
+      if (wysiwygWrapper && menuStartPos === null) {
+        setMenuStartPos(wysiwygWrapper?.getBoundingClientRect().top)
```
kendallstrautman, line 70:
```suggestion
        setMenuStartPos(wysiwygWrapper.getBoundingClientRect().top)
```

DirtyF (dismissed):
It's not jumping anymore, yeah!

Bonus points: Could we achieve a smoother transition here but make the toolbar sticky a bit earlier? 
https://www.loom.com/share/c738699dcc1847c3a38b55a45f208f48

kendallstrautman:
Oh yeah it's not taking into account the toolbar offset. I'll look into that

packages/react-tinacms-editor/src/components/BaseMenubar/index.tsx
```diff
@@ -47,11 +47,28 @@ export const BaseMenubar = ({
...
   const { editorView } = useEditorStateContext()
   const { mode } = useEditorModeContext()
 
+  function getOffsetTop(el: any) {
+    let distance = 0
+    /**
+     * traverses the dom up to get the true distance
+     * of an element from the top of the document
+     */
+
```
kendallstrautman, line 61:
```suggestion
```

kendallstrautman:
@all-contributors please add @franknoirot for docs

packages/react-tinacms-editor/src/components/BaseMenubar/index.tsx
```diff
@@ -47,10 +47,28 @@ export const BaseMenubar = ({
...
+    if (el.offsetParent) {
```
dwalkr, line 62:
```
while (el.offsetParent) {
//...
}
```

packages/react-tinacms-editor/src/components/BaseMenubar/index.tsx
```diff
@@ -47,10 +47,28 @@ export const BaseMenubar = ({
   const isBrowser = typeof window !== `undefined`
   const menuRef = useRef<HTMLDivElement>(null)
   const [menuBoundingBox, setMenuBoundingBox] = useState<any>(null)
+  const [menuOffsetTop, setMenuOffsetTop] = useState<number | null>(null)
   const menuFixedTopOffset = typeof sticky === 'string' ? sticky : '0'
   const { editorView } = useEditorStateContext()
   const { mode } = useEditorModeContext()
 
+  function getOffsetTop(el: any) {
```
dwalkr, line 55: we should move this outside of the component body

packages/react-tinacms-editor/src/components/BaseMenubar/index.tsx
```diff
@@ -62,18 +80,22 @@ export const BaseMenubar = ({
...
-      const endPosition = wysiwygWrapper
-        ? startPosition + wysiwygWrapper.offsetHeight
-        : 0
-
-      if (window.scrollY > startPosition && window.scrollY < endPosition) {
-        setMenuFixed(true)
-      } else {
-        setMenuFixed(false)
+    const wysiwygWrapper = menuRef.current!.parentElement
+    const stickyTopOffset = parseInt(menuFixedTopOffset.replace(/px$/, ''))
```
dwalkr, line 84:
```suggestion
    const stickyTopOffset = parseInt(menuFixedTopOffset, 10)
```

DirtyF (approved):
Soooo much better, awesome work @kendallstrautman 👍

dwalkr:
I'm still seeing the title bar flicker in the next demo on Chrome

DirtyF:
It's almost OK on my end when I add the sticky prop 
https://www.loom.com/share/4aeae82a474d41a49b267580e6e54f96

There's one glitch when you scroll down past the editor. Maybe a border / focus ring thing?

dwalkr:
it's not working at all for me.

editmodebrowserext

DirtyF:
Right, when the WYSIWYG is positioned in the bottom it does not follow the scroll, and I don't manage to set sticky `true` 🤷

kendallstrautman:
yeah something went wrong. I've been testing on /blog/apple but i'll try to see whats up

DirtyF:
@kendallstrautman Agreed,  it's weird to have to pass the CSS variables of the toolbar height to the sticky prop. If the toolbar is here, sticky true should be visible by default.

Do we have a way to define a low priority to load the toolbar, like can it be done only when you cick in the wysiwyg field? I feel this issue is opening some rabbit hole on how to better smart load that toolbar.

kendallstrautman:
That's a great point @DirtyF. I feel like there's a broader UX discussion here of _how_ we want the wysiwyg menubar (toolbar) to behave. I like the idea of it only showing up on click. First I need to suss out this sticky `offset` logic, but I think we could investigate your other points in another PR.

kendallstrautman:
This is ready for another look @DirtyF & @dwalkr. I debounced the scroll event handlers to try and smooth things out and contained the layout calculations for whether the menu should stick or not in a `requestAnimationFrame`. I also delayed the offset calculation until the the whole page (with images) finished loading. I tested on the info page and it seems to work as expected, but please do some digging. 

One last thing that would be nice is to programmatically handle the toolbar height in the offset calculation. That would require us knowing whether the toolbar was mounted and it felt like another PR.

DirtyF (approved):
- This is better on demo-next package 🎉 
- Didn't manage to see the update on demo-gatsby 🤷 (pebkac,)
- `Sticky` does not seem to work in Safari 🧭 (https://caniuse.com/css-sticky: A parent with overflow set to auto will prevent position: sticky from working in Safari )

dwalkr (changes requested):
@kendallstrautman while testing this out on the next demo I found the sizing of the floating menu to be a little glitchy - it would often show up at first as too small.

While looking into the problem I ended up with a fix that cuts out a lot of code, so I made a separate PR into this branch that I'd like you to look at https://github.com/tinacms/tinacms/pull/1569

packages/react-tinacms-editor/src/components/BaseMenubar/index.tsx
```diff
@@ -37,6 +43,24 @@ interface Props {
   popups?: ReactElement[]
 }
```
jpuri, line 45: Function to calculate offsetTop is already in utils: https://github.com/tinacms/tinacms/blob/master/packages/react-tinacms-editor/src/utils.ts#L21

jpuri:
Hey @kendallstrautman : changes look good, but as I checked these in demo gatsby, they are still not addressing all scenario.

sticky_menu

@spbyrne : implemented previous version of sticky menu and has also context about it.

kendallstrautman:
Thanks @DirtyF & @jpuri I'll check this out on demo-gatsby and use that offsetTop from the utils!

jpuri (approved):
Good work 👍

dwalkr:
For the record, I am still seeing the sizing issue, but it's not shrinking as much as it was before:

menubar-sizing

With that said, this looks pretty minor now and is a significant improvement over the previous buggy jumping. I don't think we need to flog this horse any more :smile:

dwalkr (approved):

PR #1554, opened 2020-10-15
chore(react-tinacms-editor): don't insert previewSrc into content body

dwalkr:
previewSrc is generated by the nodeView and should not mutate content

### DIFF
--- a/packages/react-tinacms-editor/src/plugins/Image/Menu/ProsemirrorMenu.tsx
+++ b/packages/react-tinacms-editor/src/plugins/Image/Menu/ProsemirrorMenu.tsx
@@ -31,7 +31,7 @@ import { insertImage } from '../commands'
 import { ImageProps } from '../../../types'
 
 export interface MenuProps {
-  imageProps: ImageProps | undefined
+  imageProps?: ImageProps
 }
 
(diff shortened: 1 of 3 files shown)


packages/react-tinacms-editor/src/plugins/Image/Menu/ProsemirrorMenu.tsx
```diff
@@ -65,8 +65,7 @@ export const ProsemirrorMenu = ({ uploadImages, mediaDir }: MenuProps) => {
 
   async function onMediaSelect(media?: Media) {
     if (media) {
-      const previewSrc = await cms.media.previewSrc(media.id)
-      insertImageInEditor(previewSrc)
+      insertImageInEditor(media.id)
```
dwalkr, line 68: this should call the parse method instead of passing the id thru directly

kendallstrautman: I think that mayyybe the parse is handled in that `useImageProps` hook on `upload`?

dwalkr:
@kendallstrautman I refactored this to pass the full `imageProps` object down to prosemirrorMenu. could use some extra scrutiny.

kendallstrautman (approved):

PR #1555, opened 2020-10-15
feat(react-tinacms-editor): Sidebar wysiwyg images, removes media popup

kendallstrautman:
Enables `imageProps` to be passed to the sidebar wysiwyg. Also removes the image popup from the menu entirely, so when you hit the media icon in the wysiwyg menu, the media manager opens immediately. 

Would love help testing this out to make sure it's all good and see if we actually like it better without the popup. If we want to keep the popup, we'll need to adjust some things in the sidebar.

### DIFF
--- a/packages/demo-next/pages/info.js
+++ b/packages/demo-next/pages/info.js
@@ -15,7 +15,7 @@ import matter from 'gray-matter'
 import styled from 'styled-components'
 import { useMarkdownForm } from 'next-tinacms-markdown'
 import ReactMarkdown from 'react-markdown'
-import { useCMS } from 'tinacms'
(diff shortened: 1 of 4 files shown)


kendallstrautman:
There will be a merge conflict with #1554 but let's get that one in first since it fixes a higher priority issue

DirtyF (dismissed):
I can insert images from the sidebar through the WYSIWYG image menu 👍

dwalkr:
> There will be a merge conflict with #1554 but let's get that one in first since it fixes a higher priority issue

merged + resolved on this branch

dwalkr (approved):

PR #1569, opened 2020-10-23
chore: refactor floating menu sizing

dwalkr:


packages/react-tinacms-editor/src/components/BaseMenubar/index.tsx
```diff
@@ -144,15 +121,13 @@ export const BaseMenubar = ({
       setMenuOffsetTop(getOffsetTop(wysiwygWrapper, stickyOffset))
     window.addEventListener('scroll', handleScrollStart)
     window.addEventListener('scroll', handleScrollStop)
-    window.addEventListener('resize', handleResize)
 
     return () => {
       window.removeEventListener('scroll', handleScrollStart)
       window.removeEventListener('scroll', handleScrollStop)
-      window.removeEventListener('resize', handleResize)
```
kendallstrautman, line 152: Without this, the menubar width won't be recalculated on window resize when the menu is sticky

kendallstrautman:
Could you clarify what this PR is fixing so I can see the base case issue in the other sticky-menu PR?

packages/react-tinacms-editor/src/components/BaseMenubar/index.tsx
DirtyF: Is that really an issue as long as the menu is floating as expected?

dwalkr: @kendallstrautman what browser are you testing in? It's funny, this PR is intended to fix that exact problem!

DirtyF:
For what it's worth the menu is not floating anymore on the demos for me 🤷

packages/react-tinacms-editor/src/components/BaseMenubar/index.tsx
kendallstrautman: lol weird! I'm in firefox. I only see these weird states on window resize.

dwalkr: right, I realized after posting the previous message that I was experiencing this sizing issue on initial page load. makes sense that removing the resize handler wouldn't recalculate it on resize

kendallstrautman: I haven't been able to recreate the width sizing issue on page load in chrome that you've described....from what I just experienced in fixing the sticky behavior for chrome, it seems like chrome shows a greater lag than other browsers between `document.readyState === 'complete'` and the window's 'load' event. 

My guess is that maybe that initial `setMenuBoundingBox` call is happening before the content is actually loading. Could we try using `useLayoutEffect` instead or perhaps check if document.readyState is actually complete before we run that set state?

PR #1581, opened 2020-10-29
feat(tinacms): Video support for media manager, inline forms and fields

jc-delvalle:
This days we also need to be able to edit videos just as we do with images. 

In this PR I added similar functionality as inline images so you are able to add an InlineVideo component and edit its content using the media manager and the fields components. 

Changes to the media manager include to be able to upload video/* and filter the list depending on what component or upload field you are coming from ( Image / Video ).

Also added UploadVideo form field so you can change a video source via media manager / upload.


DirtyF:
@jc-delvalle Could you share more of your context use here? What do you wish to accomplish with videos?

jc-delvalle:
Sure, we need to be able to edit videos just as images. So what I added was same functionality as inline images so you are able to add an InlineVideo component and edit its content using the media manager and the fields components. The only changes to the media manager are to be able to upload video/* and filter the list depending on what component you are coming from ( Image/Video ). 

Some content may also have video as background, you should be able to change that video source from the Edit form.

kendallstrautman:
Hey @jc-delvalle! Thanks so much for making this PR, this is awesome. 🎉 I'm sure lots of folks will find these fields useful.

In order to make it easier to review this work, could you rebase this into smaller, more defined commits using the [conventional commit naming](https://github.com/tinacms/tinacms/blob/master/CONTRIBUTING.md#making-commits)? 

We do this so we can generate changelogs for each package accordingly and have a better track of how different packages are affected by each change. Additionally, it helps us keep track of when we need to update package READMEs or add new documentation. 

Another thing that would really help us test this PR would be adding a test case in `demo-next` and `demo-gatsby`, as we need to verify that this works with both. Thanks again!

kendallstrautman (changes requested):
Hey @jc-delvalle! Thanks so much for making this PR, this is awesome. 🎉 I'm sure lots of folks will find these fields useful.

In order to make it easier to review this work, could you rebase this into smaller, more defined commits using the [conventional commit naming](https://github.com/tinacms/tinacms/blob/master/CONTRIBUTING.md#making-commits)?

We do this so we can generate changelogs for each package accordingly and have a better track of how different packages are affected by each change. Additionally, it helps us keep track of when we need to update package READMEs or add new documentation.

Another thing that would really help us test this PR would be adding test cases in demo-next and demo-gatsby, as we need to verify that this works with both. And if you could provide some general explanation and upfront examples of how the new changes being introduced work in your PR description, that makes the whole process smoother. Thanks again!

jc-delvalle:
Ok, will try to do that this weekend and break down code into more commits. Will comment here when its done.

jc-delvalle:
@kendallstrautman @DirtyF I added more detailed commits and a test implementation on demo-next. You'r feedback will be appreciated.

DirtyF:
This is breaking existing media-stores:

```
Error: /Users/frank/github/tina/tinacms/packages/@tinacms/git-client/src/git-media-store.ts(48,9): semantic error TS2322: Type '"file"' is not assignable to type '"image" | "video" | "dir"'.
```

chrisdmacrae:
> This is breaking existing media-stores:
> 
> ```
> Error: /Users/frank/github/tina/tinacms/packages/@tinacms/git-client/src/git-media-store.ts(48,9): semantic error TS2322: Type '"file"' is not assignable to type '"image" | "video" | "dir"'.
> ```

@jc-delvalle IMHO, don't do a breaking change here. Type should be:

```
"file" | "image" | "video" | "dir"
```

File being a "generic" fallback for any non-special format.

jc-delvalle:
Ok @chrisdmacrae, that makes sense, I'll do the change and update the PR.

jc-delvalle:
@chrisdmacrae, I added the generic 'file' type to media manager so it doesn't break current media sources

packages/react-tinacms-inline/src/blocks/inline-block-video.tsx
```diff
@@ -0,0 +1,27 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+
+*/
+
+import * as React from 'react'
+import { InlineVideoProps, InlineVideo } from '../fields/inline-video'
+
+/**
+ * @deprecated
```
kendallstrautman, line 23: This looks like a copy / paste of other deprecated block field files. We can delete this file and just recommend folks use `InlineVideo` from the start.

packages/tinacms/src/components/media/media-item.tsx
```diff
@@ -61,6 +55,17 @@ export function MediaItem({
   )
 }
 
+function renderPreview(item: Media) {
+  if (item.previewSrc) {
+    if (item.type === 'image' || item.type === 'file') {
+      return <img src={item.previewSrc} alt={item.filename} />
+    } else if (item.type === 'video') {
+      return <video src={item.previewSrc} />
```
kendallstrautman, line 63: I'm wondering if its a good idea to try and use the actual videos as the preview src. We may just want to use an icon.

PR #1586, opened 2020-10-30
feat(react-tinacms-inline): Export SettingsModal

### LINKED ISSUE #1583: Expose SettingsModal in react-tinacms-inline
## Summary

It would be very useful to have the `SettingsModal` component in `react-tinacms-inline` exported and available for global usage. The `Screens` and `Content-Creator` plugins let you create "Modals", but they add menu items in the sidebar.

However, for certain use cases, such as a custom `InlineField`, it would be useful to be able to use custom modals such as the SettingsModal directly, without going through `InlineGroup` or `InlineBlocks`.

My use case is that I have an `Icon` component in my website that I want to make editable through a custom `InlineField`. When the user clicks on it, I want to show the `SettingsModal` directly, as an `InlineGroup` would be too restrained by the small space available on this component. The code of `SettingsModal` can easily be copy-pasted in my code and is working, but it would be nice to directly get it from the `tinacms` or `react-tinacms-inline` package.

### Basic example

Usage would be the same as in the `react-tinacms-inline` package:

```tsx
{selectorOpen && (
  <SettingsModal
      title="Icon Settings"
      fields={[
        {
          name: input.name,
          label: 'Icon',
          description: 'Choose an icon to display',
          component: 'text',
        },
      ]}
      close={() => setSelectorOpen(false)}
  />
)}
```

Also, maybe also add a `title` prop to be able to customize the Modal title.

### Motivation

This would answer several use cases such as this one and ensure a consistent editing UX for users, in case of future modifications to this `SettingsModal` component.

_Additional Note: Just started playing with TinaCMS and inline editing a week back and I really enjoy it! Very cool library and easy to use API :tada:_

alexbchr:
Export `SettingsModal` outside package so we can use it for various use cases. Also add `title` prop, optional and falling back to `Settings` text.

Fixes #1583

### DIFF
--- a/packages/react-tinacms-inline/src/inline-settings.tsx
+++ b/packages/react-tinacms-inline/src/inline-settings.tsx
@@ -63,12 +63,17 @@ export function InlineSettings({ fields }: InlineSettingsProps) {
   )
 }
 
-interface SettingsModalProps {
+export interface SettingsModalProps {
+  title?: string
   fields: Field[]
   close(): void
 }
 
-function SettingsModal({ fields, close }: SettingsModalProps) {
+export function SettingsModal({
+  title = 'Settings',
+  fields,
+  close,
+}: SettingsModalProps) {
   const { form } = useInlineForm()
   const { name } = React.useContext(InlineFieldContext)
   const [initialValues] = React.useState(form.values)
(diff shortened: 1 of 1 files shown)


kendallstrautman (approved):
This looks great! Thank you 👏

kendallstrautman:
I'm going to merge this into a new branch directly on the repo so the checks can run.

PR #1582, opened 2020-10-30
chore(deps): React 17

DirtyF:
Testing tinacms against React. 17...

Ran `npm run bs && npm run build`, and demos are running smoothly 🎉

PR #1590, opened 2020-11-02
Upgrade react 17

kendallstrautman:
[Original PR](https://github.com/tinacms/tinacms/pull/1582)


kendallstrautman:
Is there any need to update the peer dependency ranges in the individual packages?

kendallstrautman:
We need to check if any of these [breaking changes](https://reactjs.org/blog/2020/08/10/react-v17-rc.html#fixing-potential-issues) will interfere.

dwalkr:
> Is there any need to update the peer dependency ranges in the individual packages?

I haven't checked but if they're all using the notation `>=16.8` then no, 17 satisfies that constraint. If any are using the notation `^16.8` they should be updated to `>=16.8` or React 17 won't satisfy it.

kendallstrautman:
cool yeah I think they're all using the `>=`

kendallstrautman:
After reading all of those breaking changes, I think the event listener one had the greatest potential to cause issues. I did some minor testing and I don't _think_ there are issues.

But I also think there may be some [ref.current cleanup](https://reactjs.org/blog/2020/08/10/react-v17-rc.html#potential-issues) instances we'll need to adjust. Particularly with some of the recent sticky menu code changes.

DirtyF:
When  "react-hooks/exhaustive-deps": "warn" is set on warn, running yarn lint displays a lot of warnings, some are related to current:

```
  299:6  warning  React Hook React.useEffect has an unnecessary dependency: 'triggerRef.current'. Either exclude it or remove the dependency array. Mutable values like 'triggerRef.current' aren't valid dependencies because mutating them doesn't re-render the component  react-hooks/exhaustive-deps
  79:5  warning  React Hook React.useEffect has missing dependencies: 'form.id' and 'options'. Either include them or remove the dependency array. If 'setForm' needs the current value of 'options', you can also switch to useReducer instead of useState and read 'options' in the reducer  react-hooks/exhaustive-deps
  43:7  warning  React Hook React.useCallback has an unnecessary dependency: 'wrapperRef.current'. Either exclude it or remove the dependency array. Mutable values like 'wrapperRef.current' aren't valid dependencies because mutating them doesn't re-render the component  react-hooks/exhaustive-deps
  36:8  warning  React Hook React.useEffect has a missing dependency: 'triggerRef'. Either include it or remove the dependency array. Mutable values like 'triggerRef.current' aren't valid dependencies because mutating them doesn't re-render the component  react-hooks/exhaustive-deps
  133:6  warning  React Hook React.useMemo has missing dependencies: 'cms.media', 'form', and 'passedInImageProps'. Either include them or remove the dependency array. If 'setImageProps' needs the current value of 'passedInImageProps', you can also switch to useReducer instead of useState and read 'passedInImageProps' in the reducer  react-hooks/exhaustive-deps
  41:5  warning  React Hook React.useCallback has an unnecessary dependency: 'wrapperRef.current'. Either exclude it or remove the dependency array. Mutable values like 'wrapperRef.current' aren't valid dependencies because mutating them doesn't re-render the component  react-hooks/exhaustive-deps
  52:6  warning  React Hook useEffect has missing dependencies: 'cms.enabled' and 'cms.events'. Either include them or remove the dependency array. If 'setEnabled' needs the current value of 'cms.enabled', you can also switch to useReducer instead of useState and read 'cms.enabled' in the reducer  react-hooks/exhaustive-deps
```

Do we need to fix all those?

dwalkr:
> cool yeah I think they're all using `>=`, but I'm just not sure if we should adjust that range to start at 17.

Nah, peer deps should be as flexible as possible. We're not _requiring_ that users use React 17+ at this time

kendallstrautman:
> Do we need to fix all those?

Not sure if they're all affected by this upgrade but in general we should def look into those.

DirtyF:
React 17 will be required in Next.js 11: https://github.com/vercel/next.js/blob/master/errors/react-version.md

PR #1594, opened 2020-11-02
Fix(@tinacms/fields): onClear resets the image input

### LINKED ISSUE #1593: image field plugin: 'clearing' the image does not work
## Description

If I set `clearable: true` on an image field, the trash icon shows up as expected, but clicking it does nothing. Should also note that [in the code](https://github.com/tinacms/tinacms/blob/master/packages/%40tinacms/fields/src/plugins/ImageFieldPlugin.tsx#L31) it says clearable should be true by default, and that is not the case. 

### Steps to reproduce

Add `clearable: true` to any sidebar form image config. This is easily done with the thumbnail image in the blog post template in `demo-gatsby`.

### Expected result

On click, the image field should reset to the blank or empty field state. 

### Actual result

Nothing happens. The field remains exactly the same with the currently populated image.

#### Tinacms Package Versions

Reproducible currently in master —   
  "name": "tinacms",
  "version": "0.32.0",

kendallstrautman:
Should close #1593

### DIFF
--- a/packages/@tinacms/fields/src/plugins/ImageFieldPlugin.tsx
+++ b/packages/@tinacms/fields/src/plugins/ImageFieldPlugin.tsx
@@ -28,7 +28,7 @@ interface ImageProps {
   path: string
   previewSrc?: MediaStore['previewSrc']
   uploadDir?(formValues: any): string
-  clearable?: boolean // defaults to true
+  clearable?: boolean
 }
 
 export function usePreviewSrc(
@@ -81,7 +81,7 @@ export const ImageField = wrapFieldsWithMeta<InputProps, ImageProps>(props => {
 
   let onClear: any
   if (props.field.clearable) {
-    onClear = () => onChange()
+    onClear = () => props.input.onChange('')
   }
 
   function onChange(media?: Media) {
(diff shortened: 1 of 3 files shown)


DirtyF (commented):
Rebuilt the @tinacms/fieds plugin and tested on the Gatsby demo and it does not work for me:

https://www.loom.com/share/133567c6b2cb4c70916424ff79bb3246

When i click on the trash icon, the image is cleared from within the page, but not the sidebar
When I refresh, the image is cleared from the sidebar, but I still get the trash icon not the default 
And I get ./undefined in the front matter and I don't see the default image placeholder
When I insert a new image it does not show up in the page

When I click on the image in the page, I get an error `uploadDir` is undefined.

kendallstrautman:
@DirtyF — I think what you're seeing is [this issue](https://github.com/tinacms/tinacms/issues/1307).

kendallstrautman:
We noticed that the return value is being formatted as a string so instead of the actual placeholder, we get a blank purple background with a broken image path (e.g. './undefined')

kendallstrautman:
After testing in `demo-next`, things work as expected. I think that behavior 👆🏻is related to the gatsby 'form out of step' bug.

DirtyF:
yay, works perfectly fine on the `demo-next` package 😅

dwalkr (approved):

PR #1606, opened 2020-11-05
style: fix font color

spbyrne:
Fixing two more cases where global styles were interfering with Tina styles. When a Tina element applies padding or border it relies on the `box-sizing: border-box;` model, which was previously inherited from global Tina styles. Going forward this will have to be applied manually at the component level as needed, after `all: unset;` is used. This still feels like a patchwork solution but any elements updated to include 'all: unset;` should be safe moving forward.


DirtyF (approved):
🙆‍♂️

dwalkr:
Maybe we should compose reset styles together into a variable since we're interpolating anyway. Even if it's just `all:unset` + `box-sizing:border-box`

dwalkr (approved):

PR #1611, opened 2020-11-08
Allow custom actions to be added to media manager

chrisdmacrae:
The PR adds the ability to add custom actions to a media store.

It adds a new property to the media store interface and media manager:

```
interface MediaStore {
  ...
  actions: MediaAction[]
  ...
}
```

Actions have the following interface (similar to custom form actions) but with an added `type`:

```
export type MediaAction<Type = MediaActionType> = {
  type: MediaActionType
  action: Type extends 'single'
    ? React.FC<{ media: Media }>
    : React.FC<{ media: Media[] }>
}
```

**Global Actions**

Always display alongside the "upload" button in a dropdown menu, and receives the current media items listed as a prop.

Screen Shot 2020-11-08 at 4 32 30 PM

**Single Actions**

Display alongside "single" media items, and receives that media item as a prop.

Screen Shot 2020-11-08 at 4 33 09 PM

**Multi actions**

Currently unimplemented, but these are intended for bulk actions, when we add affordance to select multiple media items and perform an action on them.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -8210,6 +8210,11 @@
         "@types/react": "*"
       }
     },
+    "@types/resize-observer-browser": {
+      "version": "0.1.4",
+      "resolved": "https://registry.npmjs.org/@types/resize-observer-browser/-/resize-observer-browser-0.1.4.tgz",
+      "integrity": "sha512-rPvqs+1hL/5hbES/0HTdUu4lvNmneiwKwccbWe7HGLWbnsLdqKnQHyWLg4Pj0AMO7PLHCwBM1Cs8orChdkDONg=="
+    },
     "@types/resolve": {
       "version": "0.0.8",
       "resolved": "https://registry.npmjs.org/@types/resolve/-/resolve-0.0.8.tgz",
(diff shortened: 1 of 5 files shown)


chrisdmacrae:
Some context: I'm refactoring a media implementation to the new media manager that doesn't have the concept of directories.

Because of this, a custom "filter" screen plugin was created to apply search query filters on the entire store. This will allow us to add this filter plugin as a global action to the new media manager.

Also, this store has the concept of "archiving" versus "deleting", so this allows us to enable that as well.

kendallstrautman:
As an aside, could you make sure to include the [package name](https://github.com/tinacms/tinacms/blob/master/CONTRIBUTING.md#making-commits) where the change occurs in your commit scope. For example:

`chore(react-tinacms-inline): some commit`

So the changelogs can be generated properly 😊

chrisdmacrae:
@dwalkr @kendallstrautman this needs more discussion, specifically, _how should custom actions be added?_.

Should they work as a plugin like in #1615, or should they be an implementation detail of a media store?

My gut leans towards the latter, as a custom action will likely almost always depend on the media store implementation and related api client, and we may want to mount multiple media stores in future and thus the actions should be scoped to the media store you're using.

However, we probably want a pattern that:

- Allows the store to define actions
- Allows the end-user of the store to also define custom actions

chrisdmacrae:
Closing this in favour of a spike to an RFC

PR #1612, opened 2020-11-08
 Added advanced pagination UI for media stores that return dozens, hundreds, or thousands of pages

chrisdmacrae:
Improves the media manager pagination UI, as it is currently totally unstyled, and for media stores that return dozens of pages, it looks terrible (and breaks).

In fact, listing pages doesn't work passed around 10 pages, and feels ineffective. Considering, the media manager has been updated to have two behaviours...

For media stores that return 10 pages or less, it renders this:

Screen Shot 2020-11-08 at 6 50 21 PM

For media stores that render more than 10 pages, it renders this:

Screen Shot 2020-11-08 at 6 49 23 PM

Which allows the user to quickly jump to a specific page as follows:

Screen Shot 2020-11-08 at 6 52 32 PM

The behaviour is as follows:

- The first button is disabled on the first page
  - It goes back to page 1
- The prev button is disabled on the first page
  - It goes back 1 page
- The next button is disabled on the last page
  - It goes forward 1 page
- The last button is disabled on the last page
  - It goes to the last page
- On change of the page dropdown, the media manager changes to that page

### DIFF
--- a/packages/tinacms/src/components/media/pagination.tsx
+++ b/packages/tinacms/src/components/media/pagination.tsx
@@ -16,10 +16,9 @@ limitations under the License.
 
 */
 
-import React from 'react'
+import React, { ChangeEvent } from 'react'
 import styled, { css } from 'styled-components'
 import { MediaList } from '@tinacms/core'
-import { Media } from 'tinacms'
 
 interface PageLinksProps {
   list: MediaList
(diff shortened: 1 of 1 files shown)


DirtyF:
@chrisdmacrae rather than a tedious pattern to browse media (who uses this?), could we rely on search + progressive loading with interactions observer here?

chrisdmacrae:
> @chrisdmacrae rather than a tedious pattern to browse media (who uses this?), could we rely on search + lazy loading with an interaction observer?

@DirtyF search is a premature abstraction. I plan on adding a custom search to my store, but this store makes a clear assumption: media will be stored in folders.

I am experiencing one of the _few_ where that is not true. So the advanced pagination is more of an "escape hatch" of the folder-based UI.

In 80% of scenarios, I'm guessing any directory of media will have less than 10 pages. (That's 300 files with the default logic)

dwalkr:
After talking with @chrisdmacrae I think we're instead going to make the relevant parts of this UI pluggable so this can be solved in userland (at least for now until we have time to work on an official solution)

dwalkr:
> After talking with @chrisdmacrae I think we're instead going to make the relevant parts of this UI pluggable so this can be solved in userland (at least for now until we have time to work on an official solution)

#1615

chrisdmacrae:
Closing this as it is handled by #1615

PR #1613, opened 2020-11-09
Performance improvements for inline blocks

dwalkr:
inline-block-perf
_Left: v0.32.0; Right: build of this branch_

### 1. Fixes an event bubbling bug causing WYSIWYG editor to lose focus when nested inside of a block
When an inline field is active, the focus-setting callback would return before `event.stopPropagation` is called, yielding to the parent element which would assign itself focus. This led to a state where focus was constantly toggling back and forth between parent and child fields.

### 2. Skips rendering block menu for blocks that don't have focus
This seems to give a pretty significant performance boost when there are many blocks in a single inline blocks field.

### 3. **Skips rendering ProseMirror editor for `InlineWysiwyg` fields that don't currently have focus**

This PR adds a render-child option to `FocusRing`, enabling conditional rendering of children depending on if the focus ring is currently focused.

```jsx
<FocusRing>
  {active => {
    if (active) {
      return <ComplicatedEditableComponent />
    }
    return <SimpleDisplayComponent />
  }}
</FocusRing>
```

This feature is now used by `InlineWysiwyg` to only render the `Wysiwyg` component when its `FocusRing` is focused, falling back to the component's children when not. In addition to boosting runtime performance when multiple `InlineWysiwyg` components are used on a page, this gives editors a way to preview their content without the WYSIWYG chrome by clicking off of the field to remove focus.

### DIFF
--- a/packages/react-tinacms-inline/src/styles/focus-ring.tsx
+++ b/packages/react-tinacms-inline/src/styles/focus-ring.tsx
@@ -59,10 +59,10 @@ export const FocusRing = ({ name, options, children }: FocusRingProps) => {
   }, [name, options, focussedField])
 
   const updateFocusedField = (event: React.MouseEvent<HTMLElement>) => {
-    if (active || !name) return
-    setFocussedField(name)
     event.stopPropagation()
     event.preventDefault()
+    if (active || !name) return
+    setFocussedField(name)
   }
 
   return (
--- a/packages/react-tinacms-inline/src/styles/focus-ring.tsx
+++ b/packages/react-tinacms-inline/src/styles/focus-ring.tsx
@@ -35,10 +35,7 @@ export interface StyledFocusRingProps
 
 export interface FocusRingProps {
   name?: string
-  children:
-    | React.ReactChild
-    | React.ReactChild[]
-    | ((active: boolean) => React.ReactChildren)
+  children: React.ReactNode | ((active: boolean) => React.ReactNode)
   options?: boolean | FocusRingOptions
 }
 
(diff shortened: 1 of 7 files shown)


dwalkr:
I'm done with the code part of this. leaving as draft until I update the package docs.

DirtyF:
@dwalkr nice https://www.loom.com/share/43143c2d7e164289933635f888ad64c4

Maybe not specific to this issue but for the record, I don't manage to edit content from the sidebar on Safari. Works in Firefox.

packages/react-tinacms-inline/src/styles/focus-ring.tsx
```diff
@@ -35,7 +35,10 @@ export interface StyledFocusRingProps
 
 export interface FocusRingProps {
   name?: string
-  children: React.ReactChild | React.ReactChild[]
+  children:
+    | React.ReactChild
```
kendallstrautman, line 39: We could consider changing the children type to 
```
React.ReactNode | ((active: boolean) => React.ReactNode)
```
ReactNode allows multiple elements, strings, numbers, fragments, portals and probably a few more things. ReactChild only allows strings, numbers, and elements.

kendallstrautman (approved):
Well done! I think there are opportunities to fine-tune the UX of how the toolbar mounts / dismounts but this is a great start.

DirtyF:
Unintended Side-effect, you can't focus anymore into an empty Markdown body: https://www.loom.com/share/3cb721e1ab0446319464ebb634af5504

I _think_ we need to [init the editor with an empty paragraph](https://github.com/tinacms/tinacms/blob/master/packages/react-tinacms-editor/stories/wysiwyg.stories.tsx#L40).

PR #1615, opened 2020-11-09
Pluggable media pagination

dwalkr:
Experimental.

## Usage Example

`custom-paginator.tsx`:
```tsx
import { MediaPaginatorPlugin } from 'tinacms'

const CustomPaginatorComponent = ({ list, setOffset }) => {
  const numPerPage = list.limit
  const currentPage = Math.floor(list.offset / numPerPage)
  const numPages = Math.ceil(list.totalCount / numPerPage)

  return (
    <ol className="paginator">
      {[...Array(numPages)].map((_, index) => {
        return (
          <li key={index} className={index == currentPage ? 'current' : ''}>
            <button onClick={() => setOffset(numPerPage * index)}>
              {index + 1}
            </button>
          </li>
        )
      })}
    </ol>
  )
}

export const CustomPaginatorPlugin: MediaPaginatorPlugin = {
  __type: 'media:ui',
  name: 'paginator',
  Component: CustomPaginatorComponent,
}
```

`_app.tsx`:
```tsx
import { CustomPaginatorPlugin } from './custom-paginator'

const cms = new TinaCMS({
  //...
  plugins: [CustomPaginatorPlugin]
})

```

### DIFF
--- a/packages/demo-next/pages/_app.js
+++ b/packages/demo-next/pages/_app.js
@@ -23,6 +23,7 @@ import { GitClient, GitMediaStore } from '@tinacms/git-client'
 import { GlobalStyles as TinaCustomStyles } from '@tinacms/styles'
 import { NextGitMediaStore } from '../next-git-media-store'
 import { MarkdownFieldPlugin } from 'react-tinacms-editor'
+//import { CustomPaginatorPlugin } from '../plugins/CustomPaginator'
 
 function Empty() {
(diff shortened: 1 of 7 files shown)


kendallstrautman (approved):

dwalkr:
> Should we release docs on this or since its experimental keep it under the radar?

I think we should keep it undocumented for now

chrisdmacrae (approved):

PR #1619, opened 2020-11-10
feat(react-tinacms-inline): InlineGroup field names are relative

### LINKED ISSUE #1617: InlineGroup is not aware of its context in the form
> Disclaimer: I dug into this last week and I know what the solution is but I'm riffing on this problem from memory, just trying to jot this issue down quickly. This specific test case might be wrong in places but the core issue is accurate.

This becomes confusing when using InlineGroup with InlineBlocks. For example:
```js
function Page() {
  //...
  return (
    <InlineForm>
        <InlineBlocks name="blocks-root" blocks={{ test: { component: MyInlineBlock }} />
    </InlineForm>
  )
}

function MyInlineBlock({data, index, name}) {
  <InlineGroup name="group-of-fields">
    <InlineBlocks name="blocks-within-blocks" />
  </InlineGroup> 
}
```

In the above example, `group-of-fields` doesn't save to the correct part of the data, but `blocks-within-blocks` knows it's nested inside of `group-of-fields`:

```
{
  "blocks-root": [
    {
      "blocks-within-blocks": [
        ...
      ]
    }
  ]
  "group-of-fields": {}
}
```

Currently, to make this work as expected, `MyInlineBlock` should be written as follows:

```
function MyInlineBlock({data, index, name}) {
  <InlineGroup name={`${name}.group-of-fields`}>
    <InlineBlocks name="blocks-within-blocks" />
  </InlineGroup> 
}
```

It's confusing that one field in this hierarchy is context-aware, and the other isn't. If you changed the name of the nested inline blocks to `${name}.blocks-within-blocks` it would also behave incorrectly.

Since building an inline form involves nesting components into an explicit hierarchy, the data hierarchy should be inferred by default, and an inline field's name is assumed to be relative to its parent field.

dwalkr:
fixes #1617


dwalkr:
This might be a good thing to \*gasp\* write some tests for

kendallstrautman (approved):

PR #1621, opened 2020-11-12
chore(react-tinacms-inline): add min-height to focus ring

### LINKED ISSUE #1620: Unable to edit wysiwyg content when body is empty
https://github.com/tinacms/tinacms/pull/1613#issuecomment-726044365

I suspect this is somewhat dependent on site's styles, but we need to make sure that, in edit mode, WYSIWYG areas with the WYSIWYG disabled still create a large enough click target to obtain focus

dwalkr:
fixes #1620

### DIFF
--- a/packages/react-tinacms-inline/src/styles/focus-ring.tsx
+++ b/packages/react-tinacms-inline/src/styles/focus-ring.tsx
@@ -84,6 +84,7 @@ export const StyledFocusRing = styled.div<StyledFocusRingProps>(p => {
     position: relative;
     width: 100%;
     height: 100%;
+    min-height: var(--tina-font-size-0);
 
     ${!p.disableHover &&
       css`


packages/react-tinacms-inline/src/styles/focus-ring.tsx
```diff
@@ -84,6 +84,7 @@ export const StyledFocusRing = styled.div<StyledFocusRingProps>(p => {
     position: relative;
     width: 100%;
     height: 100%;
+    min-height: var(--tina-font-size-0);
```
kendallstrautman, line 87:
```suggestion
    min-height: var(--tina-padding-small);
```

dwalkr: Those variables are the same value: https://github.com/tinacms/tinacms/blob/master/packages/%40tinacms/styles/src/Styles.tsx#L61

IMO font-size is more appropriate than padding here, since we're defining the height of a content area (and not its surrounding padding.) Ideally a line-height variable would be most correct, but we don't have that

kendallstrautman (approved):

spbyrne (approved):
👌

Issue #1589, opened 2020-10-31
MediaStore should accept multiple media items for upload, non-image media types should be accepted


unicornware:
## Description

After reading through the [Media documentation](https://tinacms.org/docs/media/#adding-a-media-store), I implemented a custom media store. I have no trouble listing and deleting files, but I've encountered a few issues:

- I cannot upload multiple files at once via the upload button or dropzone
- Despite setting the input accept string value, I'm only able to upload images
- A `Warning: Each child in a list should have a unique "key" prop` error is thrown when media items are listed

### Steps to reproduce

This project is for a client, so unfortunately I can't share a demo project.

1. Add the media store below

```typescript
import {
  Media,
  MediaList,
  MediaListOptions,
  MediaStore,
  MediaUploadOptions
} from 'tinacms'

export class FirebaseMediaStore implements MediaStore {
  accept = '*'

  async persist(files: MediaUploadOptions[]): Promise<Media[]> {
    console.debug(files)
    return []
  }

  async delete(media: Media): Promise<void> {
    return
  }

  async list(options: MediaListOptions = {}): Promise<MediaList> {
    return {
      items: [],
      limit: 0,
      offset: 0,
      totalCount: 0
    }
  }

  previewSrc(src: string): string | Promise<string> {
    return src
  }
}
```

2. Open the Media Manager.

3. Open the console to see `Warning: Each child in a list should have a unique "key" prop` error.

4. Attempt to upload a `PDF` file or video. Notice that only images can be selected.

5. Close the file explorer and re-open it. Attempt to upload at least two images at once.

6. Open the console to see that the `files` array, logged by `FirebaseMediaStore.persist`, only contains one `MediaUploadOptions` object

### Expected result

What should happen?

- The console should be clear of "unique key prop" errors
- **All** of the files dragged into the dropzone or selected through the file explorer should be passed to `FirebaseMediaStore.persist`
- Files of any type should be accepted by the media store

### Actual result

What actually happened?

I do not have access to all of the files that need to be persisted. After selecting two images to upload, the array logged by the `FirebaseMediaStore.persist` method shows that there is only **one** file in the array. Additionally, despite setting the input accept string value, I'm only able to upload images. Documents, videos, and other file types are greyed out in my file explorer. I believe these issues are caused by the `useDropzone` configuration within the [`MediaManager` component](https://github.com/tinacms/tinacms/blob/v0.32.0/packages/tinacms/src/components/media/media-manager.tsx):

```typescript
[...]

PR #1622, opened 2020-11-12
feat(tinacms): media manager dropzone accepts multiple files for upload

kendallstrautman:
Addresses some concerns from #1589

### DIFF
--- a/packages/tinacms/src/components/media/breadcrumb.tsx
+++ b/packages/tinacms/src/components/media/breadcrumb.tsx
@@ -19,6 +19,7 @@ limitations under the License.
 import React from 'react'
 import styled, { css } from 'styled-components'
 import { LeftArrowIcon } from '@tinacms/icons'
+import path from 'path'
 
 interface BreadcrumbProps {
   directory?: string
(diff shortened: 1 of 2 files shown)


packages/tinacms/src/components/media/breadcrumb.tsx
```diff
@@ -41,6 +41,11 @@ export function Breadcrumb({ directory = '', setDirectory }: BreadcrumbProps) {
       {directory &&
         directory.split('/').map((part, index, parts) => (
           <button
+            /** since these can't be reordered, using
+             * index as a part of key in case of dirs
+             * with the same name
+             * */
+            key={`${part}-${index}`}
```
kendallstrautman, line 48: Let me know if there's a better way to handle this this key.

DirtyF (approved):
🙆‍♂️  I can drag and drop multiple PNG files in the media manager, they're showing fine when uploaded in the public folder.

kendallstrautman:
Reminder to pair on the key issue @dwalkr

DirtyF:
When you upload many images at the root folder, thumbnails are not shown

Demo: https://www.loom.com/share/195acac33d484175b74ec59d05523a8f

dwalkr:
> When you upload many images in the default root folder, thumbnails are not shown, it seems to work only when they're in public (which is the root of the NextGitHubMediaStore)
> 
> Demo: https://www.loom.com/share/195acac33d484175b74ec59d05523a8f

I don't believe this issue to be related to this PR; it's due to how NextGitMedaStore is implemented (which is not a published package) and is, essentially, expected behavior in demo-next

dwalkr (approved):
👏👏👏👏

PR #1623, opened 2020-11-16
feat: Use custom actions in BlocksControls

alexbchr:
## Description
Make it possible to use custom Actions in BlocksControls. Usage looks like this:

``` typescript
<BlocksControls
  index={index}
  customActions={[
    {
      icon: <TinaIcon />,
      onClick: () => alert('Hello from a custom Block Action!'),
    },
  ]}
/>
``` 

## Use Cases
This opens a variety of use cases, such as:
- Opening a custom Screen when a custom action is clicked.
- Toggle a `boolean` field value when clicked.

### DIFF
--- a/packages/demo-next/pages/nesting.tsx
+++ b/packages/demo-next/pages/nesting.tsx
@@ -24,6 +24,7 @@ import {
   InlineBlocks,
   BlocksControls,
 } from 'react-tinacms-inline'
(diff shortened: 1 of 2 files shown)


kendallstrautman:
This looks great! I'm going to merge it into a branch on the repo to test further.

PR #1625, opened 2020-11-17
feat(react-tinacms-inline): Custom block actions

kendallstrautman:
This opens a variety of use cases, such as:

- Opening a custom screen when a custom action is clicked.
- Toggle a boolean field value when clicked.

[Original PR](https://github.com/tinacms/tinacms/pull/1623) by @alexbchr 
---
- [x] Docs


kendallstrautman:
@DirtyF would love help testing this. There is an example on the nesting page of `demo-next`.

kendallstrautman:
This looks good to me. I updated the blocks controls interface, but we could use a better explanation / example in the docs.

DirtyF:
What would be a good sample case for this feature?

kendallstrautman:
There's a sample case on the nesting page of `demo-next` on the color block

DirtyF:
@kendallstrautman the example works fine, but it's a bit light to my taste, I wish we would come with a more realistic example to better convey when this could be useful.
https://www.loom.com/share/c6c1a10357ef4d0c9f97c0ae19bb1072

@alexbchr Why do you need this feature at the moment? Could you share a concrete example?

DirtyF (commented):
@kendallstrautman the example works fine, but it's a bit light to my taste, I wish we would come with a more realistic example to better convey when this could be useful.
https://www.loom.com/share/c6c1a10357ef4d0c9f97c0ae19bb1072

@alexbchr What for do you need this feature at the moment? Could you share a concrete example?

alexbchr:
> @alexbchr What for do you need this feature at the moment? Could you share a concrete example?

@DirtyF I didn't have time to implement it yet, but this would be to use as a "toggle" field. For example, I have 3 tabs but only one needs to be selected at all time. Having the `selected` status as a field on each block is kind of weird, while putting it in a `InlineGroup` which contains the blocks adds a focus level which complexify usage and is not very intuitive, especially to show only a toggle field.

So ideally, my idea was to toggle this field directly from the `BlocksControls`, without having to open the fields modal. I think this feature is very useful to toggle between 2 states, like showing an image for a `Card` component or not. Of course, it should be used sparingly with editor UX in mind.

DirtyF:
I see, in this case, it can wait until you have the time to implement a more realistic example and then @spbyrne might weight in on this feature.

kendallstrautman:
To me, I don't mind that it's simple. It's the first iteration of custom block actions. Eventually we will probably create plugins for this.

dwalkr:
This looks pretty good!

Do you envision a need to receive the mouse event in the control's `onClick`? I'm thinking we likely don't, in which case we should explicitly call the passed-in `onClick` with 0 arguments. Since we're not sure of all the ways this might be used, passing 0 args now makes it simpler to update later (without breaking BC) if we decide there is something useful we could pass into those callbacks.

alexbchr:
> This looks pretty good!
> 
> Do you envision a need to receive the mouse event in the control's `onClick`? I'm thinking we likely don't, in which case we should explicitly call the passed-in `onClick` with 0 arguments. Since we're not sure of all the ways this might be used, passing 0 args now makes it simpler to update later (without breaking BC) if we decide there is something useful we could pass into those callbacks.

No I don't need it! I agree with you, it could be removed then to simplify the code. Also I don't think of any use-case needing it.

packages/react-tinacms-inline/README.md
```diff
@@ -283,13 +283,19 @@ interface BlocksControlsProps {
...
 }
 
 interface FocusRingProps {
   offset?: number | { x: number; y: number }
   borderRadius?: number
 }
+
+export interface BlocksControlActionItem {
+  icon: React.ReactNode
+  onClick: React.MouseEventHandler<HTMLElement>
```
alexbchr, line 297:
```suggestion
  onClick: () => void
```

packages/react-tinacms-inline/src/blocks/inline-block-field-controls.tsx
```diff
@@ -188,6 +195,11 @@ export function BlocksControls({
                       {direction === 'vertical' && <ReorderIcon />}
                       {direction === 'horizontal' && <ReorderRowIcon />}
                     </BlockAction>
+                    {customActions.map((x, i) => (
+                      <BlockAction key={i} onClick={x.onClick}>
```
dwalkr, line 199:
```suggestion
                      <BlockAction key={i} onClick={() => x.onClick()}>
```

dwalkr (approved):

PR #1628, opened 2020-11-19
Wrapper & Initializer plugins

dwalkr:
Spiking https://github.com/tinacms/rfcs/pull/13


packages/tinacms/src/components/TinaProvider.tsx
```diff
@@ -55,6 +61,20 @@ export const TinaProvider: React.FC<TinaProviderProps> = ({
     throw new Error(INVALID_CMS_ERROR)
   }
 
+  const [wrappers, setWrappers] = React.useState(() =>
+    cms.plugins.getType<WrapperPlugin>('unstable:wrapper').all()
+  )
+  React.useEffect(() => {
+    return cms.events.subscribe('plugin:add:unstable:wrapper', () => {
```
dwalkr, line 68: also need to subscribe to the corresponding `plugin:remove`

PR #1642, opened 2020-11-27
Support `defaultItem` as a function inside `add-block-menu.tsx`

### LINKED ISSUE #1640: Function supplied to `defaultItem` is not executed (using inline-blocks-demo)
## Description

[The docs](https://tina.io/docs/ui/inline-editing/inline-blocks/#part-2-block-template) say that the `defaultItem` key on the `BlockTemplate` interface can be either an object or a function that returns an object. 

image

However, when trying to supply a function to `defaultItem`, the function is not executed and the created block is not populated with any data.

### Steps to reproduce

1. Clone and checkout the `final-state` branch of the `inline-blocks-demo`
```bash
git clone https://github.com/tinacms/inline-blocks-demo.git
cd inline-blocks-demo
git checkout final-state
yarn install
yarn dev
```

2. Open `inline-blocks-demo\src\components\FeatureList.js` and modify `defaultItem` to be a function
```javascript
function FeatureList({ index }) { ... }

const FEATURE_BLOCKS = { ... };

export const featureListBlock = {
  ...
  template: {
    ...
    defaultItem: () => {
      console.log("This should print to console")
      return { ... }
    },
    fields: [],
  },
};
```

3. Save the file
4. Add a new feature list block at the bottom of the page

### Expected result

- The feature list is created and populated with default values 
- "This should print to console" appears in developer console

### Actual result

- The feature list is created but is not populated with any values
- Nothing is printed to developer console

#### Tinacms Package Versions
See package.json of https://github.com/tinacms/inline-blocks-demo

#### Tools

| Tool    | Version |
| ------- | ------- |
| browser |   Chrome      |
| os      |  Windows       |
| node    |   12.18.2      |
| npm     |   6.14.5      |

sneyed:
Fixes #1640

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -21,7 +21,7 @@ npm run build
 
 # Start Next.js Demo
 cd packages/demo-next
-npm run dev-next
+npm run dev
 ```
 
 **WARNING: Do not run `npm install` from inside the `packages` directory**
@@ -120,6 +120,7 @@ There are two reasons this error might occur:
 1. **`some-tinacms-package` was not built.**
 
    This is likely the problem if: the `build` directory is missing; there are no `.d.ts` or `.js` files. To fix this issue simply run `npm run build` from the root of the repository.
+
 
 ## Release Process
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -21,7 +21,7 @@ npm run build
 
 # Start Next.js Demo
 cd packages/demo-next
-npm run dev
+npm run dev-next
 ```
 
 **WARNING: Do not run `npm install` from inside the `packages` directory**
@@ -120,7 +120,6 @@ There are two reasons this error might occur:
 1. **`some-tinacms-package` was not built.**
 
    This is likely the problem if: the `build` directory is missing; there are no `.d.ts` or `.js` files. To fix this issue simply run `npm run build` from the root of the repository.
-
 
 ## Release Process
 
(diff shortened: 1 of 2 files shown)


CONTRIBUTING.md
```diff
@@ -21,7 +21,7 @@ npm run build
 
 # Start Next.js Demo
 cd packages/demo-next
-npm run dev
+npm run dev-next
```
spbyrne, line 24: Hey, was this change intentional?

packages/react-tinacms-inline/src/blocks/add-block-menu.tsx
```diff
@@ -43,6 +43,10 @@ export function AddBlockMenu({
   const addBlockButtonRef = React.useRef<HTMLButtonElement>(null)
   const [openTop, setOpenTop] = React.useState(false)
 
+  const getDefaultProps = (defaultItem: any) => {
+    return typeof defaultItem === 'function' ? defaultItem() : defaultItem
+  }
```
spbyrne, line 48: Looks good! This is exactly what we needed.

spbyrne (changes requested):
The defaultItem  specific code looks great, just wondering about the `CONTRIBUTING.md` changes.

CONTRIBUTING.md
kendallstrautman: Good catch, it could be this or `develop` I forgot we removed the `dev` script

sneyed: Yes, it was intentional, although not strictly related to fixing #1640 😄 
Just noticed the run command was not correct while working on this change

sneyed: Reverted 910f1411632773b70f2078da006e7ad8f2888ff4 to exclude this change

spbyrne (approved):

CONTRIBUTING.md
spbyrne: Ah, my bad. Thanks!

PR #1646, opened 2020-11-30
feat(react-tinacms-inline): Adds Block Label

spbyrne:
Adds a block label for inline blocks which appears opposite the blocks controls. This label uses the 'label' from the block template. `BlocksControls` now accepts a 'label' prop (boolean) to disable the label when it's not needed. The label is displayed by default.

I'd like to see what people think of this label design. I also tried a smaller blue label joined to the focus ring and also putting the label alongside the other blocks controls. This seemed to be the better option. 

It probably makes sense to add this to inline groups but they don't have an easy label option like `template.label` used by inline blocks.

[Docs PR](https://github.com/tinacms/tinacms.org/pull/761)

Inline Block Label

Here's what happens on blocks that are smaller than the width of the controls:

Screen Shot 2020-12-01 at 10 45 00 AM

Thanks @chrisdmacrae for coming up with and trying out this solution.

### DIFF
--- a/packages/react-tinacms-inline/src/blocks/inline-block-field-controls.tsx
+++ b/packages/react-tinacms-inline/src/blocks/inline-block-field-controls.tsx
@@ -42,6 +42,7 @@ export interface BlocksControlsProps {
   children: React.ReactChild | React.ReactChild[]
   index: number
   insetControls?: boolean
+  label?: boolean
   focusRing?: boolean | FocusRingOptions
 }
 
@@ -49,6 +50,7 @@ export function BlocksControls({
   children,
   index,
   insetControls,
+  label = true,
   focusRing = {},
 }: BlocksControlsProps) {
   const cms = useCMS()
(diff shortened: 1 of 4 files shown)


packages/react-tinacms-inline/src/blocks/inline-block-field-controls.tsx
```diff
@@ -261,6 +272,51 @@ export const BlockMenuWrapper = styled.div<BlockMenuWrapperProps>(p => {
   `
 })
 
+interface BlockLabelProps {
```
kendallstrautman, line 275: Could we separate this into a new file?

kendallstrautman: I guess since the other components are there, it makes sense. But it may be time to break this file up into smaller bits in a block controls directory, considering the size.

spbyrne: For sure, we should refactor this soon.

spbyrne: One of the changes I'd like to make soon is the ability to disable the directional controls and/or the add block buttons. I think that will be a more extensive change and maybe a good time to reorganize.

DirtyF (commented):
Should we display the label when you hover a block?

https://www.loom.com/share/4125796c32b6401ead84b311ed773a91

chrisdmacrae:
> Should we display the label when you hover a block?
> 
> https://www.loom.com/share/4125796c32b6401ead84b311ed773a91

Yup -- this was requested by Telus

kendallstrautman (approved):

PR #1673, opened 2020-12-18
fix(react-tinacms-github): Fix types for form hooks

zenflow:
Before this commit, the return type of `useGithubFileForm` (and transitively of `useGithubJsonForm` & `useGithubMarkdownForm`) was `[any, any]`. 
Now it is `[T, Form]`, where `T` is the generic type parameter describing the shape of the parsed file data.

Also, `useGithubJsonForm` & `useGithubMardownForm` didn't **have** that type parameter `T` (describing the shape of the parsed file data), and where it could be used, an implicit `any` was used instead. 
Now they have that type parameter, defaulted to `any`, so that it may be something more specific than `any`, either by explicit definition, or by inference from argument types.
This same problem existed in options interfaces for all 3 hooks, and I applied the same solution.

Thanks for your work on this awesome project :)

### DIFF
--- a/packages/react-tinacms-github/src/form/useGithubFileForm.ts
+++ b/packages/react-tinacms-github/src/form/useGithubFileForm.ts
@@ -19,21 +19,22 @@ limitations under the License.
 import { GitFile } from './useGitFileSha'
 import { useForm, FormOptions } from 'tinacms'
 import { useGithubFile } from '../github-client'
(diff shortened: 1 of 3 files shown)


dwalkr:
@all-contributors add @zenflow for code

dwalkr (approved):

PR #1687, opened 2021-01-13
Suggested trick for pushing menu below toolbar

joeinnes:
By default, Tina's toolbar will hide the WYSIWYG menu bar. As the `sticky` prop accepts a string, it also accepts CSS variables, such as the one used by Tina to determine toolbar height.

This means that editors can pass `var(--tina-toolbar-height)` to `sticky` to ensure the WYSIWYG menu is always visible, even in an environment where Tina's toolbar is displaying. It might be worth considering this as default behaviour whenever the toolbar is enabled in case sticky is not set explicitly. If this is seen as a good idea, I can create a PR for it.

### DIFF
--- a/packages/react-tinacms-editor/README.md
+++ b/packages/react-tinacms-editor/README.md
@@ -191,6 +191,14 @@ Alternatively you can pass a string to set the exact offset of the menu.
 </InlineWysiwyg>
 ```
 
+If you're using Tina's toolbar, you can pass 'var(--tina-toolbar-height)' to ensure the toolbar does not cover the WYSIWYG menu.
+
+```jsx
+<InlineWysiwyg name="markdownBody" format="markdown" sticky="var(--tina-toolbar-height)">
+  <ReactMarkdown source={data.markdownBody} />
+</InlineWysiwyg>
+```
+
 ### Dynamic Imports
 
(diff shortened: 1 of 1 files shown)


DirtyF (approved):
Yeah, we use this or a hardcoded pixel value to work around this issue.

spbyrne:
Good tip!

I've considered defaulting the offset to `var(--tina-toolbar-height)` and making that default to 0 _(or undefined)_. Right now the toolbar height is always set, even when the toolbar isn't being used, so it would require a couple changes.

Maybe we could add a new property, `--tina-sticky-offset`, which is abstracted away from the toolbar height (which can be left as-is) but set to `var(--tina-toolbar-height)` if the toolbar is active. This could be used by any UI that needs to stick but below the toolbar.

joeinnes:
That makes sense. 

I was thinking it might also be possible to do something around [here](https://github.com/tinacms/tinacms/blob/e7615bb6593e922c934b8dd9599161c4b378f528/packages/react-tinacms-editor/src/components/BaseMenubar/index.tsx#L52) which only returns 0 if there's no `toolbar` available on `cms` (I couldn't find a specific 'toolbarEnabled' bool to check for, but seems that the `toolbar` key only exists when it's enabled, so that should do the trick I think).

Alternatively, I guess you could set the sticky offset variable in [this if block](https://github.com/tinacms/tinacms/blob/e7615bb6593e922c934b8dd9599161c4b378f528/packages/tinacms/src/tina-cms.ts#L102)?

Issue #1610, opened 2020-11-06
Bug: Dragging Blocks in a wrapping container does not work well


alexbchr:
## Description

When dragging Blocks in a wrapping container, the blocks are ordered weirdly and blocks are moved quite strangely. This is caused by `react-beautiful-dnd` not supporting wrapping containers (see [this issue](https://github.com/atlassian/react-beautiful-dnd/issues/316#issuecomment-364734987))

**NOTE: This should be fixed in a near future version of `react-draggable-dnd` with [this PR](https://github.com/atlassian/react-beautiful-dnd/pull/1992). Tina should update to this once it is released. Another option could be to let users disable the drag'n'drop behavior on `InlineBlocks` or `BlocksControls`.**

### Steps to reproduce

Sorry I don't have a demo project for now, but this is quite simple to achieve, use a wrapping component (using flexbox or grid) as `InlineBlocks` `components.Container`, have enough items to wrap them on at least 2 lines and try to drag'n'drop the items.

### Expected result

When you drop an item somewhere, it should remain there.

### Actual result

When you drop the item, it will actually render at an unexpected index.

#### Tinacms Package Versions

```text
│  ├─ react-tinacms-inline@^0.32.0
│  ├─ tinacms@^0.32.0
├─ @tinacms/alerts@0.32.0
├─ @tinacms/core@0.32.0
├─ @tinacms/fields@0.32.0
├─ @tinacms/form-builder@0.32.0
│  ├─ @tinacms/forms@^0.32.0
│  ├─ @tinacms/react-core@^0.32.0
├─ @tinacms/forms@0.32.0
├─ @tinacms/icons@0.31.0
├─ @tinacms/react-alerts@0.32.0
├─ @tinacms/react-core@0.32.0
│  ├─ @tinacms/core@^0.32.0
│  └─ @tinacms/forms@^0.32.0
├─ @tinacms/react-forms@0.32.0
│  ├─ @tinacms/form-builder@^0.32.0
├─ @tinacms/react-modals@0.32.0
├─ @tinacms/react-screens@0.32.0
├─ @tinacms/react-sidebar@0.32.0
│  ├─ @tinacms/core@^0.32.0
│  ├─ @tinacms/forms@^0.32.0
├─ @tinacms/react-toolbar@0.32.0
│  ├─ @tinacms/core@^0.32.0
│  ├─ @tinacms/fields@^0.32.0
│  ├─ @tinacms/forms@^0.32.0
[...]

PR #1702, opened 2021-01-21
refactor(react-tinacms-inline): remove drag + drop for blocks

dwalkr:
The drag + drop feature of inline blocks adds a lot of complexity to the code. The incoming [planned changes to inline editing](https://tina.io/blog/more-changes-coming-to-inline-editing/) do not yet have a plan for working with drag + drop functionality.

The feature doesn't work in all cases (https://github.com/tinacms/tinacms/issues/1610), and even at its best, it is of limited utility when the layout is composed of tall, full-width blocks (our most common pattern right now.)

Because of this, we believe that, at least for now, it makes sense to remove drag + drop capabilities for inline blocks. Blocks can still be reordered using the chevron buttons on block controls. This UI is simpler and less buggy than drag + drop.


jamespohalloran (approved):

chrisdmacrae (commented):
LGTM

PR #1703, opened 2021-01-23
Some options for GlobalFormPlugin can't be passed through useFormScreenPlugin

joeinnes:
Adds an optional second argument to 'useFormScreenPlugin' to allow an icon to be passed to GlobalFormPlugin when instantiating a new Form Screen.

### DIFF
--- a/packages/tinacms/src/react-tinacms/use-form.ts
+++ b/packages/tinacms/src/react-tinacms/use-form.ts
@@ -46,11 +46,11 @@ export function useGlobalForm<FormShape = any>(
 /**
  * Creates and registers ScreenPlugin that renders the given Form.
  */
-export function useFormScreenPlugin(form: Form, icon?: any) {
+export function useFormScreenPlugin(form: Form, icon?: any, layout?: "fullscreen" | "popup") {
   const GlobalForm = useMemo(() => {
     if (!form) return
 
-    return new GlobalFormPlugin(form, icon)
+    return new GlobalFormPlugin(form, icon, layout)
   }, [form])
 
   usePlugins(GlobalForm)
--- a/packages/tinacms/src/react-tinacms/use-form.ts
+++ b/packages/tinacms/src/react-tinacms/use-form.ts
@@ -51,7 +51,7 @@ export function useFormScreenPlugin(form: Form, icon?: any, layout?: "fullscreen
     if (!form) return
 
     return new GlobalFormPlugin(form, icon, layout)
-  }, [form])
+  }, [form, icon, layout])
 
   usePlugins(GlobalForm)
 }
--- a/packages/tinacms/src/react-tinacms/use-form.ts
+++ b/packages/tinacms/src/react-tinacms/use-form.ts
@@ -46,11 +46,11 @@ export function useGlobalForm<FormShape = any>(
 /**
  * Creates and registers ScreenPlugin that renders the given Form.
  */
-export function useFormScreenPlugin(form: Form) {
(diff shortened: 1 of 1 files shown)


zenflow:
@joeinnes The `GlobalFormPlugin` constructor also has a third argument: `layout?: "fullscreen" | "popup"`.

I think it would make sense to, at the same time, add this argument to `useFormScreenPlugin` too.

I would frame the root issue as "some options for `GlobalFormPlugin` can't be passed through `useFormScreenPlugin`".

In my case I really needed to change the layout from 'popup' to 'fullscreen' (because of an issue where the height of the popup was too small) and it took me a while to realize (by digging into TinaCMS code) that instead of using `useFormScreenPlugin` I could use `GlobalFormPlugin` directly (e.g. `const plugin = useMemo(() => new GlobalFormPlugin(form, SettingsIcon, "fullscreen"), [form]);`).

packages/tinacms/src/react-tinacms/use-form.ts
```diff
@@ -46,11 +46,11 @@ export function useGlobalForm<FormShape = any>(
 /**
  * Creates and registers ScreenPlugin that renders the given Form.
  */
-export function useFormScreenPlugin(form: Form) {
+export function useFormScreenPlugin(form: Form, icon?: any) {
```
zenflow, line 49:
```suggestion
export function useFormScreenPlugin(form: Form, icon?: any, layout?: "fullscreen" | "popup")) {
```

packages/tinacms/src/react-tinacms/use-form.ts
```diff
@@ -46,11 +46,11 @@ export function useGlobalForm<FormShape = any>(
...
   const GlobalForm = useMemo(() => {
     if (!form) return
 
-    return new GlobalFormPlugin(form)
+    return new GlobalFormPlugin(form, icon)
```
zenflow, line 53:
```suggestion
    return new GlobalFormPlugin(form, icon, layout)
```

joeinnes:
Absolutely right, I hadn't considered the layout option.

zenflow:
@joeinnes Awesome! Don't forget to update the PR title. 😃

joeinnes:
> @joeinnes Awesome! Don't forget to update the PR title. 😃

Done. Thanks!

packages/tinacms/src/react-tinacms/use-form.ts
```diff
@@ -46,11 +46,11 @@ export function useGlobalForm<FormShape = any>(
...
  * Creates and registers ScreenPlugin that renders the given Form.
  */
-export function useFormScreenPlugin(form: Form) {
+export function useFormScreenPlugin(form: Form, icon?: any, layout?: "fullscreen" | "popup") {
   const GlobalForm = useMemo(() => {
     if (!form) return
 
-    return new GlobalFormPlugin(form)
+    return new GlobalFormPlugin(form, icon, layout)
   }, [form])
```
dwalkr, line 54:
```suggestion
  }, [form, icon, layout])
```

dwalkr (changes requested):

DirtyF (commented):
Docs will need to be updated:
https://tina.io/docs/plugins/screens/#useformscreenplugin

joeinnes:
tinacms/tinacms.org#786 opened for docs.

dwalkr (approved):

dwalkr:
@all-contributors add @joeinnes for code

PR #1704, opened 2021-01-24
feat: radio group field

austincondiff:
## Summary
Added a new radio group field that is configurable to display as radios or buttons eithor horizontally or vertically.

## Usage
```js
interface RadioGroupFieldProps {
  label?: string
  name: string
  component: string
  options: (Option | string)[]
  direction?: 'horizontal' | 'vertical'
  variant?: 'radio' | 'button'
}
```

### Horizontal Radios 
```js
{
  label: 'Page Width',
    name: 'frontmatter.pagewidth',
    component: 'radio-group',
    options: [
      { value: 'xs', label: 'XS' },
      { value: 'sm', label: 'SM' },
      { value: 'md', label: 'MD' },
      { value: 'lg', label: 'LG' },
      { value: 'xl', label: 'XL' },
  ],
}
```
https://user-images.githubusercontent.com/806104/105625589-ed049700-5def-11eb-971f-cbc30cedb8a1.mov

### Vertical Radios
```js
{
  label: 'Page Width',
    name: 'frontmatter.pagewidth',
    component: 'radio-group',
    direction: 'vertical',
    options: [
      { value: 'xs', label: 'Extra Small' },
      { value: 'sm', label: 'Small' },
      { value: 'md', label: 'Medium' },
      { value: 'lg', label: 'Large' },
      { value: 'xl', label: 'Extra Large' },
  ],
}
```
https://user-images.githubusercontent.com/806104/105625586-ebd36a00-5def-11eb-80a2-e09586c6693c.mov

### Horizontal Buttons

```js
{
  label: 'Page Width',
    name: 'frontmatter.pagewidth',
    component: 'radio-group',
    variant: 'button',
    options: [
      { value: 'xs', label: 'XS' },
      { value: 'sm', label: 'SM' },
      { value: 'md', label: 'MD' },
      { value: 'lg', label: 'LG' },
      { value: 'xl', label: 'XL' },
  ],
}
```
https://user-images.githubusercontent.com/806104/105625591-ed9d2d80-5def-11eb-99ba-8c3d63b1389a.mov

### Vertical Buttons

```js
{
  label: 'Page Width',
    name: 'frontmatter.pagewidth',
    component: 'radio-group',
    variant: 'button',
    direction: 'vertical',
    options: [
      { value: 'xs', label: 'Extra Small' },
      { value: 'sm', label: 'Small' },
      { value: 'md', label: 'Medium' },
      { value: 'lg', label: 'Large' },
      { value: 'xl', label: 'Extra Large' },
  ],
}
```
https://user-images.githubusercontent.com/806104/105625590-ed9d2d80-5def-11eb-9d21-842a9ca09f09.mov

## Motivation

Select fields require more interaction than is necessary for fields with a limited set of options. This offers developers a field to use in forms to provide their users with a quicker method of input when there are only a limited set of options.


DirtyF:
Very nice work here, appreciate the little animations on the buttons 👌 
https://www.loom.com/share/410ccc4e7b8744168853abe51142fce3

austincondiff:
Tab should focus the selected radio in a group, but pressing tab again will blur the group and focus the next focusable item on the page. Once you have. the radio group focussed, you can press the directional keys to navigate within the radio group as documented [here](https://www.w3.org/wiki/RadioButton). I did forget to include the `name` attribute on each radio input. This tells in the radio is part of a group. I've added that so now it should behave as expected when there are multiple groups in a single form.

spbyrne (approved):
This looks great, @austincondiff. This will be a very useful feature.

dwalkr:
We'll want to document this new field type on our website, added an issue to track https://github.com/tinacms/tinacms.org/issues/787

you're welcome to take a stab at it @austincondiff but it's OK if you can't

packages/@tinacms/fields/src/components/RadioGroup.tsx
```diff
@@ -0,0 +1,269 @@
...
+}) => {
+  const [
+    activeRadioRef,
+    setActiveRadioRef,
+  ] = React.useState<HTMLDivElement | null>(null)
+  const radioOptions = options || field.options
+  const radioRefs: RadioRefsInterface = {}
+
+  React.useEffect(() => {
+    console.log({ activeRadioRef: radioRefs[`radio_${input.value}`] })
```
dwalkr, line 61:
```suggestion
```

packages/@tinacms/fields/src/components/RadioGroup.tsx
```diff
@@ -0,0 +1,269 @@
...
+    return (
+      <RadioOptionWrap
+        variant={field.variant}
+        ref={ref => {
+          radioRefs[`radio_${option.value}`] = ref
+        }}
+      >
+        <input
+          type="radio"
+          key={option.value}
```
dwalkr, line 83: I think we want to move `key` to the parent component

dwalkr (changes requested):

dwalkr:
@all-contributors add @austincondiff for ideas

austincondiff:
@DirtyF @dwalkr @spbyrne As I was writing the docs for this, I had a thought. We can continue with this implemented API, but was curious to get your thoughts on an alternative API (which would require a slight refactor).  Currently the `component` is `radio-group` and an additional `variant` property can either be set to `radio` or `button` to determine if the field looks like a traditional radio group or a button group respectively.  An alternative API would be to remove the `variant` property and use the same component for an additional `button-group` field. So to summarize, this are our two options.

Option 1 _(currently in this PR as of this comment)_
```js
{
  component: 'radio-group' 
  variant:  'radio' | 'button'
  ...
}
```

Option 2
```js
{
  component: 'radio-group' | 'button-group'
  ...
}
```

Which of these two do y'all prefer? Why?

dwalkr:
@austincondiff since the two fields do the exact same thing (and indeed are just visual variations on the same control,) it makes sense to keep them as the same field type. It makes the individual field's API slightly more complex, but avoids adding unnecessary complexity to the Tina API overall. Your second option adds two fields to the mental model, when there only needs to be one.

austincondiff:
@dwalkr I was leaning this direction as well for that very reason. I just figured I'd give y'all the option in case you thought otherwise.

dwalkr (approved):

PR #1712, opened 2021-01-26
Run DangerJS workflow on pull_request_target instead of pull_request

dwalkr:
## The Problem

We've been wrangling with the problem of how to run DangerJS against PRs that originate from forks. DangerJS works by leaving comments on PRs with warnings/errors triggered by the script, and the read-only `GITHUB_TOKEN` provided to GItHub actions that run against forks can't comment on PRs.

Previously, I attempted to resolve this by using a machine user and generating a personal access token for them, which we then send to the script. The idea is that this user can comment on PRs, but doesn't have access to anything important since there is potential for this access token to be extracted by a malicious PR. Creating a build environment to provide this token, and requiring builds to first be approved before running, seemed like adequate mitigation against this threat. Unfortunately, I misunderstood the documentation around secrets provided to GitHub actions, as even when a build is authorized to run, environment secrets are still not provided to fork PRs. Presumably, using a machine user with DangerJS on fork PRs would require the access token to be hard-coded into the script. For a public repo, that means publishing this token. Even if the machine user has no important access, that solution is not one I am OK with.

## The pull_request_target event

The only practical solution, then, is to use the `pull_request_target` event instead of `pull_request`. Per the [GitHub docs](https://docs.github.com/en/actions/reference/events-that-trigger-workflows#pull_request_target):

> This event runs in the context of the base of the pull request, rather than in the merge commit as the pull_request event does. This prevents executing unsafe workflow code from the head of the pull request that could alter your repository or steal any secrets you use in your workflow. This event allows you to do things like create workflows that label and comment on pull requests based on the contents of the event payload.

Since `pull_request_target` will run in the context of `tinacms/tinacms`, even when executed on behalf of a fork PR, its provided `GITHUB_TOKEN` will be able to insert Danger's comments on PRs.

## Caveat

There is a caveat to this, which the DangerJS maintainer points out in [this issue](https://github.com/danger/danger-js/issues/1060): performing static analysis on code files fetched via the local file system will end up examining that code as it exists on the _target_ branch, and not the new code being proposed in the pull request. Thus, any static analysis needs to first fetch the file contents over the GItHub API, against the appropriate ref.

## Solution

Since this duality between locally-sourced content and content fetched from remotes is something we've frequently grappled with ourselves working on TinaCMS, I wasn't that intimidated by the problem. There are only two static analysis routines we perform in our danger script, and it seemed reasonable to swap out the local file access for API access. 

The results of these changes (running on `pull_request`, not `pull_request_target`) can be seen in [this PR](https://github.com/tinacms/tinacms/pull/1711), where I've removed a couple licenses to verify that danger will fail the build correctly

### DIFF
--- a/.github/workflows/danger.yml
+++ b/.github/workflows/danger.yml
@@ -4,7 +4,6 @@ on: [pull_request]
 jobs:
   danger:
     runs-on: ubuntu-latest
-    environment: test
 
     steps:
       - uses: actions/checkout@v1
@@ -14,4 +13,3 @@ jobs:
         uses: danger/danger-js@10.5.4
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-          DANGER_GITHUB_API_TOKEN: ${{ secrets.DANGER_GITHUB_API_TOKEN }}
--- a/.github/workflows/danger.yml
+++ b/.github/workflows/danger.yml
@@ -1,5 +1,5 @@
 name: 'Danger JS'
-on: [pull_request]
+on: [pull_request_target]
 
 jobs:
   danger:
--- a/dangerfile.ts
+++ b/dangerfile.ts
@@ -20,6 +20,7 @@ import { markdown, danger, warn, fail, message } from 'danger'
 import depcheck from 'depcheck'
 import * as fs from 'fs'
 import * as path from 'path'
+import { Buffer } from 'buffer'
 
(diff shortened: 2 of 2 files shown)


jeffsee55 (approved):
Looks great! We're just looking into Danger JS for the other repo, we'll probably want to borrow all of this when it becomes OSS 👍

chrisdmacrae (commented):
This is legit

PR #1714, opened 2021-01-27
Revert "fix(@tinacms/react-sidebar): fixed site elements relative to SiteWrapper"

### LINKED ISSUE #1713: Bug: `position: fixed` elements within SiteWrapper are no longer behaving as fixed
## Description

After upgrading to the latest release from earlier today (v0.36.0), my layout, including a `position: fixed` navbar at the top, is broken, since it is within the SiteWrapper, the styles of which were changed by PR #1690.

Instead of the navbar being fixed to the top of the viewport when you scroll down the page (as expected and desired when using `position: fixed; top: 0;`), it's now fixed to the top of the SiteWrapper element, and moves up and out of the viewport when you scroll down the page. In other words, it doesn't appear "fixed" at all.

Interestingly, the description on PR #1690 even states the above-mentioned problematic fact:

> As a result, the fixed element becomes fixed to the transformed element as opposed to the viewport and the underlying content remains visible.

This change will break any layout that makes use of `position: fixed`, and will make it impossible to have any actually fixed elements (without some complicated workaround, like creating another wrapping element within SiteWrapper that has styles something like `height: 100%; overflow-y: scroll`).

*Please please please revert this change! 
Please don't "un-fix fixed elements" (as the blog article linked in the PR describes it).*

It does fix the described problem with `position: fixed` elements, but at the same time it basically disables the `position: fixed` style altogether. If this is the effect that you (as a developer or designer) want, just don't use `position: fixed`! Use `position: absolute` instead!

### Steps to reproduce

1. Have a TinaCMS website that has an element (within SiteWrapper) with styles `position: fixed; top: 0;`
2. Make sure your browser viewport width is maximized ( > 840px ) since the styles in question (`transform: translate(0);`) are in a media query.
3. Make sure you're on a page with enough content that there's a vertical scrollbar.
4. Scroll down

It doesn't matter if the CMS is enabled or disabled.

### Expected result

As you scroll down, the fixed element should stay at the top of the viewport.

### Actual result

As you scroll down, the "fixed" element moves up and out of the viewport with the other non-fixed elements.

#### Tinacms Package Versions

Sorry I can't run the bash command as instructed because I'm on Windows, but all the tinacms packages I'm using are at the latest version, and I've already identified where the problem was introduced: PR #1690

#### Tools
Observed with Chrome 88.0.4324.104 but I'm pretty sure this will be an issue in any browser.

/cc @austincondiff @dwalkr

DirtyF:
Reverts tinacms/tinacms#1690

Fix #1713

### DIFF
--- a/packages/@tinacms/react-sidebar/src/components/Sidebar.tsx
+++ b/packages/@tinacms/react-sidebar/src/components/Sidebar.tsx
@@ -138,9 +138,8 @@ const SidebarGlobalStyles = createGlobalStyle`
 
 const SiteWrapper = styled.div<{ open: boolean }>`
   @media (min-width: 840px) {
-    margin-left: ${props => (props.open ? 'var(--tina-sidebar-width)' : '0')};
-    transition: margin-left 150ms ease-out;
-    transform: translate(0);
+    padding-left: ${props => (props.open ? 'var(--tina-sidebar-width)' : '0')};
+    transition: padding-left 150ms ease-out;
   }
 `
 


dwalkr:
thanks @DirtyF !

dwalkr (approved):

PR #1719, opened 2021-01-31
Stop event propagation

pristas-peter:
This fixes a bug, when color picker is nested inside form and does not open up. The problem is that the click event is fired twice (one for Swatch and one for SwatchInner) and the color picker then stays always closed.

Example field configuration:
```
const formOptions = {
    label: 'Header',
    fields: [
      {
        label: 'Banner',
        name: 'banner',
        component: 'group',
        fields: [
          {
            label: 'Background color',
            name: 'backgroundColor',
            component: 'color',
          },
          {
            label: 'Text',
            name: 'text',
            component: 'text',
          },
          {
            label: 'Text color',
            name: 'textColor',
            component: 'color',
          },
        ],
      },
    ],
  }

  const [data, form] = useGithubJsonForm(file, formOptions)
  useFormScreenPlugin(form)

...
```

### DIFF
--- a/packages/@tinacms/fields/src/components/ColorPicker/ColorPicker.tsx
+++ b/packages/@tinacms/fields/src/components/ColorPicker/ColorPicker.tsx
@@ -16,7 +16,7 @@ limitations under the License.
 
 */
 
-import * as React from 'react'
+import * as React, { SyntheticEvent } from 'react'
 import { useState } from 'react'
 import { Dismissible } from 'react-dismissible'
 import { SketchPicker, BlockPicker } from 'react-color'
@@ -317,7 +317,8 @@ export const ColorPicker: React.FC<Props> = ({
     )
   }
 
-  const toggleColorPicker = () => {
+  const toggleColorPicker = (event: SyntheticEvent) => {
+    event.stopPropagation()
     const display = !displayColorPicker
     setDisplayColorPicker(display)
     if (display) {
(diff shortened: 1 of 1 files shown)


packages/@tinacms/fields/src/components/ColorPicker/ColorPicker.tsx
```diff
@@ -16,7 +16,7 @@ limitations under the License.
 
 */
 
-import * as React from 'react'
+import * as React, { SyntheticEvent } from 'react'
```
dwalkr, line 19:
```suggestion
import * as React from 'react'
```
destructuring syntax doesn't work w/ `import * as`

packages/@tinacms/fields/src/components/ColorPicker/ColorPicker.tsx
```diff
@@ -317,7 +317,8 @@ export const ColorPicker: React.FC<Props> = ({
     )
   }
 
-  const toggleColorPicker = () => {
+  const toggleColorPicker = (event: SyntheticEvent) => {
```
dwalkr, line 320:
```suggestion
  const toggleColorPicker = (event: React.SyntheticEvent) => {
```

dwalkr (changes requested):

dwalkr:
@all-contributors add @pristas-peter for bugs and code

dwalkr (approved):

Issue #1664, opened 2020-12-14
InlineForm adds two non-targetable divs to DOM layout


chrisdmacrae:
## Description

Not sure if this is a "bug", but it's breaking behaviour for me. `InlineForm` appears to be rendering two classless divs when added to a page since updating to `tinacms@0.34.0` and `react-tinacms-inline@0.34.0`.

### Steps to reproduce

Add an inline form. See the divs.

### Expected result

Inline form shouldn't require any DOM elements?

### Actual result

Inline form is adding DOM elements, and they can't be classed or managed, so they mess up my design.

#### Tinacms Package Versions

```text
N/A
```

#### Tools

| Tool    | Version |
| ------- | ------- |
| browser |   safari@latest      |
| os      |    macos     |
| node    |     LTS    |
| npm     |     LTS    |

dwalkr:
Are you upgrading from `0.33` or an earlier version?

chrisdmacrae:
@dwalkr went from 0.30.0 => 0.34.0 => 0.35.0

dwalkr:
I checked out v0.30.0 and ran _demo-next_ in the monorepo, and still have the two wrapping divs. I can see how that's could be a problem, but doesn't seem to be a regression from previous behavior.

If you think it's a regression, I think we'll need more info and, ideally, a reproduction of the problem.

chrisdmacrae:
@dwalkr irregardless of it being a regression or not, it _is_ a problem, because the divs have no reliable way of targeting with CSS.

Ideally they wouldn't be there at all. But if that's a big challenge, then making them easier to target to fix styling would be ideal.

dwalkr:
I completely agree, but there is a difference between "my existing Tina setup suddenly broke after an update" and "Tina is hard to work with because X" in terms of how we prioritize an issue like this.

zenflow:
> Inline form shouldn't require any DOM elements?

You can see from PR #1726 (and issue #1725) why those divs are required.
They're used for capturing events that will deselect any selected inline form field.

Perhaps though there is another way to do this though, which doesn't require extra divs.
 
Note that the `<Dismissible>` component (which renders the other div) lives in another repo: https://github.com/tinacms/react-dismissible/blob/5bf1d004505a4ede9c8fab72b5b5057b94346821/src/Dismissible.tsx#L57-L75

zenflow:
[...]

PR #1726, opened 2021-02-02
fix(react-tinacms-inline): Fix ability to clear field selection

### LINKED ISSUE #1725: Bug: Inline form fields can no longer be deselected by clicking empty space in InlineForm
## Description

After upgrading to the latest release (0.36.1) from the previous (0.36.0) I noticed this regression in the editing experience, and linked it to PR #1702

Before this change, the user could deselect (or "unfocus") any/all fields (i.e. "clear the selection") by just clicking anywhere else where there wasn't another field.

After this change, that can only be done by clicking somewhere _outside_ the `<InlineForm>`. 
If the user clicks _inside_ the `<InlineForm>` (without selecting a different field) nothing happens.

For example, in the screenshot below (which has a header, a footer, and an inline form with a single inline block field), I can clear the selection by clicking in spot 1 or 2, but not spot 3 (since it's inside the `<InlineForm>`).
image

### Steps to reproduce

1. Select an inline form field that has a visual indication that it's selected (e.g. an inline block field, as in the example above)
2. Click any place inside `<InlineForm>` that won't cause another field to be selected

### Expected result

The visual indication that the field is selected should disappear.

### Actual result

Nothing happens. The visual indication that the field is selected stays.

zenflow:
Closes #1725

This adds back the `<div onClick={() => setFocussedField('')}>` element that was removed in #1702.

@dwalkr Did this element need to be removed for some reason? I'm guessing it was just forgotten when refactoring?

Thanks!

### DIFF
--- a/packages/react-tinacms-inline/src/inline-form.tsx
+++ b/packages/react-tinacms-inline/src/inline-form.tsx
@@ -70,18 +70,20 @@ export function InlineForm({ form, children }: InlineFormProps) {
           setFocussedField('')
         }}
       >
-        <FormBuilder form={form}>
-          {({ form, ...formProps }) => {
-            if (typeof children !== 'function') {
-              return children
-            }
+        <div onClick={() => setFocussedField('')}>
+          <FormBuilder form={form}>
+            {({ form, ...formProps }) => {
+              if (typeof children !== 'function') {
+                return children
+              }
 
-            return children({
-              ...formProps,
-              ...inlineFormState,
-            })
-          }}
-        </FormBuilder>
+              return children({
+                ...formProps,
+                ...inlineFormState,
+              })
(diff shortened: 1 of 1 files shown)


dwalkr:
@zenflow yeah, this was a clumsy attempt to address #1664 but it produces some undesirable side effects. Thanks for the fix!

dwalkr (approved):

dwalkr:
@all-contributors add @zenflow for code and bugs

zenflow:
@dwalkr Thanks for the quick merge! I just realized you're publishing prereleases on npm (just not on GitHub /w tags) so now I have everything working with the latest prerelease :D Really awesome! Thanks and keep up the great work!

PR #1741, opened 2021-02-10
useFieldRef: imperative field hover/focus

dwalkr:
This is a different approach to `useFieldRef` that doesn't require every field to have a hover/focus target component placed over it. Instead, it subscribes to DOM events of the element referenced by the ref. The userland API has not changed.

The biggest drawback with this is that we lose a lot of customizability with the hover state. We're now limited by the capabilities of the CSS `outline` property. Since we're indicating the hover state by modifying the DOM element directly, I recommend sticking with `outline` because:
1. It's designed for accessibility
2. it doesn't impact layout flow
3. it is unlikely to be a mission-critical style attribute on the user's component, so it's relatively safe to modify it

### DIFF
--- a/packages/react-tinacms-inline/src/inline-form.tsx
+++ b/packages/react-tinacms-inline/src/inline-form.tsx
@@ -21,7 +21,7 @@ import { FormRenderProps } from 'react-final-form'
 import { FormBuilder, Form, Field } from 'tinacms'
 import { Dismissible } from 'react-dismissible'
 import { useMap } from 'react-use'
-import { FieldOverlay, FieldTarget, FieldRefType } from './ref-fields'
+import { FieldOverlay, FieldRefType } from './ref-fields'
 
 type useMapObject<T> = { [key: string]: T }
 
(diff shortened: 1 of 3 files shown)


jeffsee55:
So from what I'm understanding, this does:

- Attach our ref to the node which will render your piece of data
- Via the ref, attach events to the real DOM node, which do a couple of things
  - a hover/mouseover will animate that node with raw styles added
  - a click will set that field to be focused
- `<InlineForm>` -> render all children
  - if there's a focused/current field, render `<FieldOverlay>` in the position of the ref that we attached earlier
    - inside the overlay, render the provided `inlineComponent`

> The biggest drawback with this is that we lose a lot of customizability with the hover state
And the reason `FieldOverlay` can't take on the job of hover state is that it's only rendered for the active/focused field, right?

So why can't `FieldOverlay` be rendered over all refs that have been registered, focused or not? In that case, picking up from the list above...
This would remain true:
- if there's a focused/current field, render `<FieldOverlay>` in the position of the ref that we attached earlier
    - inside the overlay, render the provided `inlineComponent`
But then we could also do this (I think):
- for unfocused fields, render `<FieldOverlay>` in the position of the ref that we attached earlier
  - make it responsible for hover states and animation, as well as rendering any controls we might have (removing the need for the real DOM node's hover events and styling).
  - set `pointer-events:none` so the underlying DOM node can still receive onclick events

dwalkr:
The impetus for this refactor was your comment here: https://github.com/tinacms/tinacms/pull/1729#discussion_r571200980

Manipulating the referenced node directly seemed like the most foolproof way to avoid a scenario where we overlay a click target somewhere we shouldn't.

Honestly, I'm not that bothered by the design limitations of `outline`, but I'm not a designer 😜 

I'm going to experiment with your idea to capture hover events on the source node and use that data elsewhere, though

jeffsee55:
FWIW, the RFC I did a while back was doing something like this after I sort of expanded on it and it worked pretty well if you want to peak at a possible implementation. It's admittedly a very simplistic one - but one thing that's jumping out to me is that the current inline editing API nearly gets in the way more than it helps since this is such fundamental shift.

PR #1749, opened 2021-02-18
event-based ref-based inline editing

dwalkr:
Implements `useFieldRef` by attaching handlers to whatever DOM element the ref is mounted to, and sending events over the event bus.


jeffsee55:
This is working nicely for me, so we're really getting at 2 things here, right?:

1. Continuing to push the form system into event-based architecture. I really like the idea of abstracting away final form's API behind our event system. The one thing that's bugging me about it is the lack of types, I realize that's part of a larger conversation, and I love the 2-way editing this enables with the sidebar. For the larger conversation - I think we could mimic a typed event system like you can find in [xstate](https://xstate.js.org/docs/). The `type` of the event is static, while the event payload determines the details (like which form/field the event is for), for example. In this machine, the `CHANGE` event must receive a value https://codesandbox.io/s/beautiful-kirch-vcj11?file=/src/index.ts

2. The inline API feels good to me, I think the biggest takeaway is the `useInlineRef` and `<InlineForm>` could be way higher up in the tree from where the `ref` is being attached, to me this is a good thing. Meaning your component needs 2 things to be inline-friendly, a `ref` and the value, and that's it. So it's more friendly to a design system integration. I know there's a lot more to discuss regarding field groups and blocks, etc. But I like it so far

jeffsee55 (approved):

PR #1770, opened 2021-03-24
style: scrolling add block menu

spbyrne:
This sets a `max-height` on the inline add block menu and adds a visual cue when the list is overflowing in a given direction. 

https://user-images.githubusercontent.com/5075484/112364056-e772e280-8cb4-11eb-9afa-695741bbc9d6.mp4


Enigmatical (approved):
👍

PR #1771, opened 2021-03-24
Fix Inline Settings Cancel Behaviour

spbyrne:
The inline settings modal 'cancel' button would revert changes, but the 'x' button in the modal header was just closing the modal. Both buttons now trigger the cancel behaviour.


Enigmatical (approved):
Makes total sense that both the "X" and "Cancel" would perform the same functions 👍

PR #1772, opened 2021-03-25
Inline Add Block Menu Filter

spbyrne:
This takes the filter pattern from the git branch switcher and applies it to the inline add block menu, with a few tweaks. This search input is only displayed if there are 10+ blocks, so it will hopefully make the current block selector more usable in cases with many blocks.

Here's how I padded out the existing demo-next blocks page with more blocks to hit the minimum required:

```
const HeaderBlock = () => {}

const header_template: BlockTemplate = {
  label: 'Header',
  fields: [],
}

const TestimonialBlock = () => {}

const testimonial_template: BlockTemplate = {
  label: 'Testimonial',
  fields: [],
}

const NavBlock = () => {}

const nav_template: BlockTemplate = {
  label: 'Nav',
  fields: [],
}

const VideoBlock = () => {}

const video_template: BlockTemplate = {
  label: 'Video',
  fields: [],
}

const AuthorBlock = () => {}

const author_template: BlockTemplate = {
  label: 'Author',
  fields: [],
}

const GalleryBlock = () => {}

const gallery_template: BlockTemplate = {
  label: 'Gallery',
  fields: [],
}


const PAGE_BUILDER_BLOCKS = {
  /* ....existing blocks */
  header: {
    Component: HeaderBlock,
    template: header_template,
  },
  testimonial: {
    Component: TestimonialBlock,
    template: testimonial_template,
  },
  nav: {
    Component: NavBlock,
    template: nav_template,
  },
  video: {
    Component: VideoBlock,
    template: video_template,
  },
  author: {
    Component: AuthorBlock,
    template: author_template,
  },
  gallery: {
    Component: GalleryBlock,
    template: gallery_template,
  },
}
```

https://user-images.githubusercontent.com/5075484/112492801-578a7280-8d60-11eb-97d4-91f680ab33d0.mp4

PR #1786, opened 2021-03-30
fix: close on click outside

spbyrne:
This reintroduces the 'click outside' code removed in #1772 but with added logic to handle the filter input.


DirtyF:
This solves the potential overlap as it prevent to click on two "add block" action in a row.

CleanShot 2021-03-30 at 16 10 30

Note: it only close when clicking outside, not when you cllick on the "X" sign (often means close)

spbyrne:
Fixing that issue now, thanks!

DirtyF (approved):
Working as expected 👌 

CleanShot 2021-03-30 at 16 43 41

Issue #999, opened 2020-04-08
react-tinacms-inline should depend directly on react-textarea-autosize


ncphillips:
this package is currently a `peerDependency` it should be a `dependency`

PR #1789, opened 2021-03-31
Fix grey 10

spbyrne:
`--tina-color-grey-10` had the wrong hue, this fixes it.


DirtyF (approved):
Great shade of gray.

Enigmatical (approved):
Always been a fan of simpler greys like #DDD or #999, but this is a pretty good grey, too.

Issue #1780, opened 2021-03-28
Inline settings modal field/submission issues


ezg27:
## Description

First of all I really love what you guys are building with Tina, it's a great editing experience and a real win for making interactive content management accessible to non-devs in our team 🎉

However I've run into a few bugs with the inline editing experience, specifically using the Inline Settings modal and nested fields inside Group, Group List and Blocks fields. I would love to use Inline Blocks to their fullest in my current project, but I'm having to fall back to the sidebar and standalone Inline elements for now.

All of these issues are reproducible in the demo linked from the Tina homepage (https://tina-demo-two.vercel.app), and in the tinacms repo's demo projects that use the above elements.

1. When clicking into a group/group-list/block, the first interaction with any field inside it (e.g. the first keypress inside a focused input/textarea, clicking a toggle) will retrigger the enter animation from the first form page and cause the active field to lose focus.
2. When clicking into a group/group-list/block, if there is another group/group-list/block nested inside it these items won't respond to clicks and you will not be able to proceed deeper into the form. However, if you click the back button and re-enter the group/group-list/block, the previously inaccessible fields will now work as expected. Furthermore, attempting to interact with any fields inside the newly accessible group/group-list/block will retrigger the enter animation mentioned in point 1 above, except you are returned to the previous page and must perform the 'go-forward-to-go-back' steps again in order to get back to where you were.
3. Specifically with group-lists (and sometimes groups, see 2nd video below), editing them in any way (either by dragging to reorder, deleting an item, editing any of the fields nested within a group-list item) causes the `Form Status` field in the toolbar to show `Unsaved changes` in the background as expected. But if `Confirm` is clicked without editing any other fields, the status returns to `No changes` and the changes cannot be saved, even though they can still be seen on the page. If you re-open the inline settings menu the `Form Status` shows in the background that there are unsaved changes, but this does not persist when closing the settings modal.

None of these issues appear to exist when using the same fields in the sidebar, everything functions seamlessly there. My suspicion is this is caused by stale form state/re-rendering issues related to React Final Form in the inline package, but I haven't been able to diagnose so far when running the tinacms repo demo's locally.

### Steps to reproduce

1.

https://user-images.githubusercontent.com/40216869/112761787-cc75da80-8ff4-11eb-9986-52d9ed942f26.mp4

2.

https://user-images.githubusercontent.com/40216869/112761799-da2b6000-8ff4-11eb-9caf-4e82e0de96b6.mp4

3.

[...]

PR #1790, opened 2021-04-10
Fix/do not recreate FormPortal on a ref value change

c58:
Connects to https://github.com/tinacms/tinacms/issues/1780

### The Cause
I had a nice time trying to find the cause of this annoying bug. It turned out that FormPortal function changed after the first rerender (when a user changes something in a form of a group) because the `wrapperRef.current` is null on the first render and not null on the second render. Since the FormPortal reference is changed (basically a new function), the context will trigger the rerendering (because its value changed) and since we use FormPortal as a component the tree will be completely recalculated and recreated, which recreates a Panel element from GroupListFieldPlugin which triggers the animation to start over.

### The Fix
So the fix is quite simple – do not recreate the function on a change of the ref value. Instead, just track changes of the ref itself and that is it. I found the same problem with the MenuPortal, so I fixed it there as well


spbyrne (approved):

spbyrne (approved):

DirtyF:
@all-contributors please add @c58 for code

DirtyF:
@all-contributors please add @Jaza  for docs

PR #1791, opened 2021-04-10
Fix/form portal zindex management

c58:
Connects to #1780

### What changed
In one FormPortal might be rendered multiple panels. Each new panel should be shown on screen above the previous one, no matter in what order they was rendered in the DOM. This PR adds zindex management to the FormPortal by switching children prop to a function that accepts an object with zIndexShift. The value of that shift can be used to set a final zIndex of a panel or whatever wants to be rendered inside the portal


spbyrne (approved):
Clever and simple solution, looks good to me.

DirtyF:
Thanks @c58 could you please [sign the CLA](https://cla-assistant.io/tinacms/tinacms?pullRequest=1791) for us to merge this change?

c58:
@DirtyF done!

spbyrne:
Thank you! 👍

PR #1795, opened 2021-04-14
feat: drag to resize sidebar

### LINKED ISSUE #558: Customizable width of the sidebar

spbyrne:
You can now click and drag the right edge of the sidebar to resize it!  🎉

- Sidebar width will still default to `--tina-sidebar-width`, which the user can customize in their CSS.
- There's a 250px minimum width and it can't extend beyond the width of the window viewport. 
- All the logic is contained in a new `ResizeHandle` component.
- There's a 4px clickable area at the right edge of the sidebar instead of a visible handle. Hovering this area will change your cursor.
- Text selection is blocked with a universal selector while dragging.

- [ ] Needs docs 

[Thanks @weibenfalk for your initial work on this](https://github.com/tinacms/tinacms/pull/565)

https://user-images.githubusercontent.com/5075484/114748808-a6587600-9d28-11eb-95d3-b5f7df98c5d8.mp4

Closes #558

### DIFF
--- a/packages/@tinacms/react-sidebar/src/components/ResizeHandle.tsx
+++ b/packages/@tinacms/react-sidebar/src/components/ResizeHandle.tsx
@@ -1,0 +1,115 @@
+/**
+
+Copyright 2021 Forestry.io Holdings, Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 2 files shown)


DirtyF (approved):
👏 This will make writing long form content from the sidebar so much easier. 🙏

dwalkr (approved):

PR #1796, opened 2021-04-14
fix: alert container click blocking

spbyrne:
The container alerts pop up in was blocking click events, so in order to open the sidebar you had to wait for any alerts to go away given the toggle is underneath this container. This is now fixed so that the container itself doesn't capture any click events, just the alerts.


DirtyF (approved):
Nice catch!

PR #1805, opened 2021-04-22
WIP - Single build command

jeffsee55:
WIP


DirtyF:
Just curious, are we hiring Vite and esbuild to provide a better development experience here?

Issue #1806, opened 2021-04-27
Add Date Field to Default Plugins


Enigmatical:
Originally, `date` field was removed from Tina's default plugins to address both performance and bundle size:
https://github.com/tinacms/tinacms/pull/1281

Today, we feel that having a `datetime` field is a core functionality of Tina and have made the decision to add the field back to the core of Tina.

For now, we will be adding the plugin as-is from `react-tinacms-date` (including the dependencies of both `react-datetime` and `momentjs`) and looking to deprecate that package.

In the very near future, we would like to replace `react-datetime` (and any dependency on `momentjs`) with a leaner, more performant package with ideally no visual or functional changes to the field itself.

PR #1807, opened 2021-04-27
feat(@tinacms/fields): Adds date field to default plugins

Enigmatical:
Moves the contents of `react-tinacms-date` back to `@tinacms/fields` and `tinacms`, making `DateFieldPlugin` a default plugin once more.

Closes https://github.com/tinacms/tinacms/issues/1806


packages/@tinacms/fields/package.json
```diff
@@ -59,7 +59,9 @@
     "color-string": "^1.5.3",
     "react-color": "^2.17.3",
     "react-dismissible": "^1.3.0",
-    "react-dropzone": "10.1.8"
+    "react-dropzone": "10.1.8",
+    "moment": ">=2 <=3",
```
DirtyF, line 63: https://day.js.org/ seems to be much lighter and have the same moment API

DirtyF (commented):
In big favor of this change, to avoid getting an invalid date if you defined a `datetime` in your GraphQL Schema when you want to work with Tina Cloud.

packages/@tinacms/fields/package.json
Enigmatical: I'll have to take a look at that; not heard of `day.js`.  Been looking at `date-fns` (which seems to be where most of the community went with their `datepicker` solutions) and `luxom` (which shares one of the creators/maintainers of Moment.js).

Ideally, we'd choose a `datepicker` solution that uses a minimal `date` library and use that same `date` library in the gateway so that parsing/formating is the same on the frontend and backend (right now, it isn't).

Enigmatical:
> In big favor of this change, to avoid getting an invalid date if you defined a `datetime` in your GraphQL Schema when you want to work with Tina Cloud.

That's the initial goal.  Just get it so `datetime` works and then we'll revisit in a later phase with a replacement for `react-datepicker` that hopefully reduces the size impact of bringing this Plugin back.

packages/@tinacms/fields/package.json
DirtyF: I saw `day.js` is used on https://github.com/calendso/calendso an open source Calendly alternative who is built with Next.js that have to rely on a good date picker with timezone support since it's core to their product.

Enigmatical: Never heard of it, but it definitely checks all the boxes of something we might want for an alternative.  @spbyrne Have you heard of https://github.com/calendso/calendso in your circles?  I'm going to take a closer look and see what functionality it offers vs. `react-datepicker`.

jeffsee55 (approved):

logan-anderson:
Looks good to me as well!

PR #1819, opened 2021-05-12
feat(next-tinacms-s3): Add S3 media store

Jaza:


Jaza:
But `package.json` is there! Any idea why that "danger" build is failing, anyone?

DirtyF:
@Jaza Thank you so much from submitting a media store plugin, we'll look into the Danger errror 🙄

DirtyF:
@all-contributors please add @Jaza its contribution to code and plugin

dwalkr:
> But `package.json` is there! Any idea why that "danger" build is failing, anyone?

@Jaza this is a problem with our danger script. `depcheck` is looking at `master` instead of your fork. I'll look into it this week.

Jaza:
Trying and seeing if upgrading danger-js and depcheck helps, if so, then it would be great to get #1822 merged first.

dwalkr:
@Jaza the issue is 100% on our end, due to the way we're using github actions. I'm working on a fix.

dwalkr:
We've decided to drop `depcheck` for now because it doesn't work correctly when running on `pull_request_target`. This isn't really `depcheck`'s fault; see #1712 for more info on why we're doing this.

Once the `depcheck` change is merged to master, the check on this PR should run correctly (NOTE: in my experience I'll have to close and reopen this PR instead of re-running the check, just FYI.)

Jaza:
Thanks @dwalkr ! I've rebased this PR against latest master, let's see if the danger-js build finally passes.

Jaza:
Great that this has finally passed CI checks, only took a week :grin: 

Feedback welcome.

I know, still no tests. Hoping to write them soon.

Re: paging of results. This will be a bit tricky to implement, because you need to pass the `NextContinuationToken` (string value) from the previous page's response, as a `continuation-token` param to S3's `ListObjectsV2` method, to get the next page; whereas the TinaCMS media manager only currently supports paging via an "offset" integer value. So the media manager would need to support that "bookmark" style of paging.

Also, the current auth implementation is totally insecure. Currently, the API function remembers a single IAM access key and secret (passed in via env vars), and it dishes out an STS token using those credentials (which are expected to be linked to an IAM policy allowing full write access to the S3 bucket in question), to potentially anyone (especially if github open authoring is enabled), no questions asked. And calls to S3 (including calls to delete any object!) then can't be tracked to any individual user, because we're just using that one set of "universal" credentials.

So, at the least, I should change it so that an IAM access key and secret need to be entered in a form, and then passed to the API function. A better option would be to have each user authenticate via Cognito (which can in turn go through a federated identity provider, e.g. log in with Google), and to then get a token for S3 calls from Cognito. That way, calls to S3 get tracked per individual user. And it's possible to restrict Cognito sign-up to a whitelist of email addresses.

jeffsee55:
Thanks for the thorough write-up @Jaza

> TinaCMS media manager only currently supports paging via an "offset" integer value

We'll need to improve this for many other media APIs anyway. I know for GraphQL-based APIs we'd like to support `cursor` which I believe would be similar to what you're suggesting for S3.

> I should change it so that an IAM access key and secret need to be entered in a form

Agreed that we don't want to get into a situation where we encourage AWS credentials being sent to the frontend in any manner. Would your suggestion mean we just use the form to make a client-side request to get the STS token? That would feel the most secure to me, with the downside of forcing non-technical people to maintain their access token and secrets. We'll have to think about potential attach vectors if we need to store things somewhere.

In general I think you're right that the best solution would be to provide an OAuth workflow, but that feels like a lot of extra baggage to carry. The nice thing about S3 for media is its simplicity and pulling in Cognito as a requirement feels like a bit too much. 

As an aside - we're currently looking into ways to support media in an OAuth workflow for any media provider, so this might be a good ticket to hang onto as a first-class example of how to do that.

Jaza:
Closing this, have opened #1838 instead. Turns out that if we make the AWS credentials get passed in via a form, then we don't need any back-end API endpoint. So, instead of `next-tinacms-s3` , this plugin can just be `react-tinacms-s3` . I figured that a new PR would be easier, because all the file paths have changed, and because the code is significantly different (taken out the next.js API, and added the modal form).

PR #1821, opened 2021-05-13
feat(next-tinacms-s3): Add S3 media store metadata

Jaza:
Trying out adding just the metadata files first (i.e. everything except the files in `src/`) from #1819 , see if danger-js passes.


Jaza:
Nope ... first tried adding all the files except `src/` , then tried adding just `package.json` and `.gitignore` , same result, danger-js is not happy. Bah. So much for this idea.

PR #1829, opened 2021-05-17
chore(build): remove depcheck from danger script

dwalkr:
Because we're running against pull_request_target, local scans don't work here. We've decided to remove it for now and investigate other options for maintaining dependency hygiene.


dwalkr:
blocking #1821

Enigmatical (approved):

PR #1831, opened 2021-05-19
fix(refactor): hoist CreateContentMenu into react-forms

jbevis:
Closes #1807

This PR will hoist ```CreateContentMenu``` into ```react-forms``` so that we are not duplicating it in ```react-sidebar``` and ```react-toolbar```


jbevis:
@dwalkr this PR addresses the component refactor, but not yet the issue with the form button not being disabled with failing validation on any field.

I'm seeing that behavior across Tina, not only within the content creator plugin, so curious I should address it here, or via a different PR.

dwalkr:
> I'm seeing that behavior across Tina, not only within the content creator plugin, so curious I should address it here, or via a different PR.

Separate PR is fine. Minor nitpick but I recommend rebasing/amending this to a `chore` instead of a `fix`; refactors don't need to show up in the changelog

PR #1832, opened 2021-05-20
chore: hoist CreateContentMenu into react-forms

jbevis:
This PR will refactor CreateContentMenu.tsx to be a component inside of ```react-forms``` instead of duplicated inside of ```react-sidebar``` and ```react-toolbar```


spbyrne:
Hey, looks good to me. I'm going to do a build & test locally before approving.

spbyrne:
There's a slight UI issue I'll fix. The dropdown menu opens in the wrong direction now for the sidebar.

spbyrne (dismissed):
Looks good 👍

dwalkr (commented):
Looks good. Noticed there aren't any dependency changes - `@tinacms/react-forms` is a peer dep of both the sidebar & toolbar packages. I'm wondering if it should now be a direct dependency.

DirtyF:
Given I run the next demo locally 
When I add a new post from /blog/apple and click on Create
Then I get a `rejected` error message and no console output

CleanShot 2021-05-26 at 15 55 08@2x

packages/demo-next/pages/blog/[slug].tsx
```diff
@@ -38,6 +38,13 @@ const addBlogPlugin: ContentCreatorPlugin<any> = {
     {
       name: 'title',
       component: 'text',
+      validate: value => {
```
Comment (Enigmatical) [inline packages/demo-next/pages/blog/[slug].tsx:41]: I'm not sure we want to commit these changes; looks like they are strictly the result of debugging.

Enigmatical (changes requested):
Rest of the PR looks good; barring the comments from @dwalkr.

packages/demo-next/pages/blog/[slug].tsx
Comment (DirtyF) [reply packages/demo-next/pages/blog/[slug].tsx:41]: Down with Chris, either it works for the demo prupose or we don't ship it.

jbevis:
@dwalkr so are you suggesting that ```@tinacms/react-forms``` be removed from ```peerDependencies``` in both the sidebar and toolbar packages, and then added as a dependency as well or just remove it as ```@tinacms/react-forms``` is also a devDependency for both. 

The clarification is more for my own understanding as to how the all the packages within the monorepo are wired together.

packages/demo-next/pages/blog/[slug].tsx
Comment (jbevis) [reply packages/demo-next/pages/blog/[slug].tsx:41]: Agreed, I'll prune the debugging changes from the next demo.

dwalkr (dismissed):

dwalkr (approved):

packages/demo-next/pages/blog/[slug].tsx
Comment (jbevis) [reply packages/demo-next/pages/blog/[slug].tsx:41]: @Enigmatical those changes have been pruned out.

jbevis:
I'll be honest, this PR has gotten a bit messy...

The new lock file also has me feeling a bit queasy. I'm going to close this to clean up the branch locally and then try out a new PR hopefully with less issues.

PR #1838, opened 2021-06-01
feat(react-tinacms-s3): Add S3 media store

Jaza:


Jaza:
Tests are still TODO. Paging of results is still TODO.

The form workflow could be nicer. My use case is to enable this plugin together with github open authoring, and currently, the github auth and the S3 auth are both triggered on CMS enable, and they interfere with each other. Would probably better to make the S3 auth just get triggered when opening the media manager.

Anyway, it works in its current state. And it's much more secure than my original PR #1819 - no more hard-coded credentials, now a given user has to enter AWS credentials in the form, in order to CRUD the S3 bucket.

Feedback welcome.

Jaza:
Credit where credit's due: most of the form code I shamelessly borrowed from `react-tinacms-strapi` .

nstolmaker:
@Jaza Supporting s3 as a media store is on the roadmap for my application this month, so I'm actively interested in this and very likely will contribute!

Our use-case seems to differ in one way though: _we don't allow editing for non-authenticated users_. Specifically, our Next.js app has some API endpoints which are guarded by requiring a session. Sessions are managed with JWT tokens via next-auth.js.org and our federated identity provider (OIDC). 

While I agree that a client-side-only solution would be ideal, I have my doubts that asking anyone but a developer to wrangle AWS credentials is viable. So, the solution I intend to bring will probably require next-auth (although could failover to a form with credentials). I will write some logic in the API that choses from several s3 credentials, which are then returned to the front-end client. 

Let me know if you like this idea and think it would be useful, because otherwise I might just do the s3 PUT directly from the API/backend, and expose a new endpoint there for our client to use instead of talking to s3 itself.

Jaza:
@nstolmaker sounds like you'd be happier with my original PR #1819 , that doesn't require the user to enter AWS credentials his/herself , but that instead has one set of credentials defined in the back-end via environment variables. You can read the comments in that PR for background, but in a nutshell, I decided that that approach was too insecure (for a great many setups - and if this is to get merged into tinacms proper, then it can't assume a particular setup, such as yours of auth protected by a federated identity provider).

It gets complicated. If you define one or more S3 credentials in the back-end, then ideally you need to protect the endpoint that returns the credentials, behind an auth check. But the different tinacms auth plugins (github, strapi, custom) have different auth systems, with different session cookies, different use of CSRF tokens, and so on. You'd need to make the S3 credentials endpoint specifically support one of more of those auth systems, and it would be impossible to support every auth system out there (because people could have their own custom ones, as it sounds like you do).

And "authenticated" in tinacms doesn't necessarily mean "trusted". With github open authoring (which I assume is the most popular setup, it's the one used by tinacms.org itself), anyone with a github account can log in and start editing. That's fine for github interaction, because untrusted github user joebloggs can only submit a PR from a branch / fork, he/she can't merge to the main branch / repo without the requisite privileges. But it's not fine for S3 interaction, because we don't want that same joebloggs to go and add malicious / inappropriate files to the media bucket, nor to delete existing files that live content depends on.

Shared S3 credentials in general is also problematic, because then we have no audit trail in AWS of who touched what in the bucket - those shared IAM credentials aren't linked to any sessions / identities in the given tinacms auth system.

So, feel free to work on those thorny issues, and you're welcome to contribute in any other way too.

Jaza:
@nstolmaker currently this PR just checks for a populated set of cookies, that contain the STS credentials (access key id, secret access key, session token) needed to talk to S3. The modal form populates these cookies, but potentially, some other custom thing could too. So you could write custom code to run alongside this, that populates the STS credentials from a back-end call to retrieve them, on login.

jeffsee55:
Thanks @Jaza for this PR, we've moved into a focus towards "backend" media integrations, which allows us to shift the burden of secrets to env vars rather than needing to deal with them on the frontend, it will likely make this work a fair bit easier to deal with in the future. We'll have some updates to media integrations soon!

Closing this as it'll be difficult to maintain given the more abbreviated API we're after. Please feel free to reopen for more discussion

PR #1847, opened 2021-06-09
feat(@tinacms/form-builder): Combines FormView and FormBuilder

### LINKED ISSUE #1840: Combine `FormView` and `FormBuilder`
Full pitch: https://github.com/tinacms/product/discussions/8

> ...Tina Forms display and behave differently depending on where they are built and this functionality is at the whim of how they are plugged into the CMS.

> Referencing the diagram once more, only Sidebar and Screen Forms use <FormView /> which contains <DragDropContext />, <FormPortalContext /> and support for <FormActionMenu />.

> `Content Creator Plugins` (and `BranchSwitcher`) only utilize `<FormBuilder />` which does not have `<DragDropContext />` and offers limited customization of the `<FormHeader />` and `<FormFooter />` (more than likely because the form is split into parts for the `<Modal />`).

### Objectives

Combine `<FormView />` and `<FormBuilder />`

Enigmatical:
In an effort to both simplify and address inconsistencies in forms, we've decided to merge `FormView` (which was exclusively used by the Sidebar and Screen Forms) and `FormBuilder` (which was used by `Content Creator` and other modal forms).

Initially, this should result in very few changes in userland.  Those will come later as we seek to change how `<FormActionMenu />` behaves and more.

Closes #1840


Enigmatical:
Emphasis on **Draft** here. 

I've got a handful of things I still want to do:

* `<FormMaker />` is not the final name.  I'm leaning towards claiming `<FormBuilder />` to go alongside `<FieldsBuilder />`.  
* I also need to do a thorough check that I can replace all instances of `<FormBuilder />|<FormView />` - the only two I've done is Sidebar and Content Creator.
* We want to use `labels: { ... }` instead of `buttons: { ... }` in the `FormOptions` to better identify _what_ you are changing.
* We'd also like to use plugins for `FormActionMenu` (rather than `actions: { ... }`), but I might make that a separate issue.

There's also a "hack" of sorts with using `<ModalProvider />` in `<FormMaker />`.  For some reason, triggering a `<Modal />` (like `<ResetForm />`) in `<FormMaker />` results in a `No Modal Context` error - despite clearly being rendered in one (according to `react-dev-tools`).  @dwalkr and I took a look at it and I'll probably keep poking at it.

packages/@tinacms/react-forms/src/CreateContentMenu.tsx
```diff
@@ -116,44 +115,14 @@ const FormModal = ({ plugin, close }: any) => {
...
-                </Button>
-              </ModalActions>
-            </ModalPopup>
-          )
-        }}
-      </FormBuilder>
+      <ModalPopup>
+        <ModalHeader close={close}>{plugin.name}</ModalHeader>
+        <ModalBody>
+          <FormMaker form={form} />
```
Enigmatical, line 123: In a Zoom Meeting on Tuesday, June 8th, @dwalkr, @spbyrne, and myself decided that _every_ TinaForm should have a standard set of actions - specifically, Save/Submit and Reset.

For a while, TinaCMS has been mixing controls for the Form with controls for the Modal.  In an effort to unify our Forms, we're going to try and not have a dedicated "Close/Cancel" button and instead we'll rely on the 'X' (cross) in the top-right corner.

What this means is that Submit, Reset, and any additional `actions:[]`* should appear with a `Form` regardless of where it is rendered.

* Once we move `actions` to utilize the plugin system.

jeffsee55 (approved):

dwalkr (approved):

PR #1851, opened 2021-06-15
explicitly declare missing deps

dwalkr:
The root `package.json` pins an older version of `tinacms` which, in addition to causing some TypeScript errors, appears to be obscuring some undeclared dependency problems elsewhere in the monorepo (the missing deps are transitive dependencies of `tinacms` and thus can be loosely satisfied when shared deps are hoisted.)

This PR removes the `tinacms` reference from the root, and aims to reconcile any lingering build issues by explicitly declaring any missing dependencies.


dwalkr:
@Enigmatical unfortunately it seems this has unearthed a circular dependency in the recently-merged form builder stuff

jeffsee55 (approved):

Enigmatical (approved):

PR #1852, opened 2021-06-15
feat(react-tinacms-inline): InlineBlocks children

Jaza:
Solves: https://community.tinacms.org/t/get-block-count-from-outside-inlineblocks/596

### DIFF
--- a/packages/react-tinacms-inline/README.md
+++ b/packages/react-tinacms-inline/README.md
@@ -230,6 +230,7 @@ export interface InlineBlocksProps {
   components?: {
     Container?: React.FunctionComponent<BlocksContainerProps>
   }
+  children?: React.ReactNode | null
 }
 ```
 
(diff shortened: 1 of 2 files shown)


DirtyF (approved):
Thanks for the edit 🙏

dwalkr:
Hey @Jaza! Thanks for the contribution! This looks good to me, and we should be able to get this into the next release of `react-tinacms-inline`.

As Logan mentioned in that forum thread, we plan to revisit Inline Editing wholesale in the (hopefully near) future, though we don't have an exact timeline for that just yet. We are generally avoiding any feature development on `react-tinacms-inline` right now, and this specific feature _may or may not_ make it into that future revision. But this is far off enough and not super clear just yet, so if this feature solves your problem there's no real harm in putting it out.

We definitely appreciate the contribution and discussion of your problem - this are things we will absolutely want to keep in mind!

dwalkr (approved):

dwalkr:
@Jaza this is now available in prerelease `react-tinacms-inline@0.43.0-alpha.0`

Jaza:
Cool, thanks for the run-down @dwalkr , and thanks for merging.

PR #1855, opened 2021-06-16
1841 - Unify `FormOptions` for `useForm(...)` and its variations

### LINKED ISSUE #1841: Unify `FormOptions` interface
Full Pitch: https://github.com/tinacms/product/discussions/8

> Delving deeper into the code, useForm() is our predominant way to create a Tina Form and accepts an object of FormOptions. However, only a subset of those FormOptions (known as Options) are respected by forms outside of the Sidebar and Global (Screen) Forms.

> Content Creator Plugins use new Form(...) behind the scenes, but provide very little in the way of configuration (FormOptions is predetermined by the Plugin). Additionally, while it accepts fields[], certain field types (like list) are not supported (more on this below).

> Additionally, useForm() derivatives, useJsonForm() and useMarkdownForm(), completely ignore certain FormOptions providing a developer no way to add additional logic to loadInitialValues(), onSubmit(), onChange(), or reset().

### Objective

Implement consistent support for `FormOptions` across `useForm()`, `useJsonForm()`, `useMarkdownForm()`, and `Content Creator Plugins`. Remove `Options` and ensure every `FormOptions` value is respected.

Enigmatical:
**This PR is dependent on #1840 and should be merged afterwards.**

Discovered during research into Form Inconsistencies, our various versions of `useForm(...)` and its wrappers (i.e. `useGitHubFileForm(...), `ContentCreator` Plugin) accepted and respected a certain subset of `FormOptions` - usually identified as `Options` (that sometimes extended `FormOptions`).

This has the unintended consequence of unpredictable behavior when using a `Form`.  Can I provide `reset`?  `onChange`?  `buttons`?  `actions`?  Will `label` actually be used?

This PR seeks to smooth over those rough edges by (to some degree) allow `FormOptions` to be directly passed and each of the options to be respected.  There are still places where `overrides` or `default` options are used (namely, `loadInitialValues()` is defined on most of the `useForm(...)` wrappers, so passing `loadInitialValues(...)` will be ignored).

But _most_ of the other options will, at least, affect the underlying `Form`.

Closes #1841

### DIFF
--- a/packages/gatsby-tinacms-git/src/use-git-form.ts
+++ b/packages/gatsby-tinacms-git/src/use-git-form.ts
@@ -46,8 +46,16 @@ export function useGitForm<N extends GitNode>(
       return gitFile.show()
     },
     onSubmit: gitFile.commit,
-    reset: gitFile.reset,
+    reset: () => {
+      if (options.reset) {
+        options.reset()
+      }
+      return gitFile.reset()
+    },
     onChange({ values }: any) {
+      if (options.onChange) {
+        options.onChange(values)
+      }
       return gitFile.write(values)
     },
   }
--- a/packages/gatsby-tinacms-mdx/src/useMdxForm.tsx
+++ b/packages/gatsby-tinacms-mdx/src/useMdxForm.tsx
@@ -38,7 +38,7 @@ const matter = require('gray-matter')
 
 export function useMdxForm(
   _mdx: MdxNode | null | undefined,
-  formOverrrides: Partial<FormOptions<any>> = {}
+  formOverrides: Partial<FormOptions<any>> = {}
 ): [MdxNode | null | undefined, Form | null | undefined] {
   const mdx = usePersistentValue(_mdx)
 
@@ -56,9 +56,9 @@ export function useMdxForm(
 
   /* eslint-disable-next-line react-hooks/rules-of-hooks */
   const cms = useCMS()
-  const label = formOverrrides.label || mdx.frontmatter.title
+  const label = formOverrides.label || mdx.frontmatter.title
   const id = mdx.fileRelativePath
-  const actions = formOverrrides.actions
+  const actions = formOverrides.actions
 
   /**
    * The state of the MdxForm, generated from the contents of the
@@ -81,7 +81,7 @@ export function useMdxForm(
    */
   /* eslint-disable-next-line react-hooks/rules-of-hooks */
   const fields = React.useMemo(() => {
-    let fields = formOverrrides.fields || generateFields(valuesOnDisk)
+    let fields = formOverrides.fields || generateFields(valuesOnDisk)
     fields = fields.map(field => {
       /**
        * Treat the field.name prefix `frontmatter` as an alias to
@@ -100,7 +100,7 @@ export function useMdxForm(
     })
 
     return fields
-  }, [formOverrrides.fields])
+  }, [formOverrides.fields])
 
   /* eslint-disable-next-line react-hooks/rules-of-hooks */
   const [, form] = useForm(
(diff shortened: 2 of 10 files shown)


Enigmatical:
More a question for @dwalkr , but I don't know to what extent I need to update the documentation for these packages.  Many of them suggest passing a `FormOptions` object... so the documentation is _more_ correct now that I'm actually using all the parts with this PR?

Any suggestions would be greatly appreciated.

DirtyF:
@Enigmatical Given how forms are central to Tina, what do we need to communicate with the community? Will it impact existing websites? Are there some changes to apply? What could be affected?

packages/gatsby-tinacms-mdx/src/useMdxForm.tsx
```diff
@@ -38,7 +38,7 @@ const matter = require('gray-matter')
 
 export function useMdxForm(
   _mdx: MdxNode | null | undefined,
-  formOverrrides: Partial<FormOptions<any>> = {}
+  formOverrides: Partial<FormOptions<any>> = {}
```
Enigmatical, line 41: No real changes here other than fixing this typo consistently >.>;

packages/gatsby-tinacms-remark/src/use-remark-form.tsx
```diff
@@ -121,9 +121,9 @@ function fromMarkdownString(content: string) {
  */
 export function useLocalRemarkForm(
   markdownRemark: RemarkNode | null | undefined,
-  formOverrrides: Partial<FormOptions<any>> = {}
+  formOverrides: Partial<FormOptions<any>> = {}
```
Enigmatical, line 124: Likewise, just wanted to fix this typo and saw an opportunity >.>;

packages/gatsby-tinacms-git/src/use-git-form.ts
```diff
@@ -46,8 +46,16 @@ export function useGitForm<N extends GitNode>(
...
+    reset: () => {
+      if (options.reset) {
+        options.reset()
+      }
+      return gitFile.reset()
+    },
     onChange({ values }: any) {
+      if (options.onChange) {
+        options.onChange(values)
+      }
```
Enigmatical, line 58: While `<FormOptions />` allowed for `reset` and `onChange` to be passed in, they were ignored.  This alters the behavior to more closely match standard `useForm()`.

packages/next-tinacms-json/src/use-json-form.ts
```diff
@@ -72,9 +65,15 @@ export function useJsonForm<T = any>(
...
       reset() {
+        if (options.reset) {
+          options.reset()
+        }
         return cms.api.git.reset({ files: [id] })
       },
       onChange: formState => {
+        if (options.onChange) {
+          options.onChange(formState.values)
+        }
```
Enigmatical, line 76: Same story as `use-git-form` above.

packages/next-tinacms-markdown/src/use-markdown-form.ts
```diff
@@ -90,9 +82,15 @@ export function useMarkdownForm(
...
       reset() {
+        if (options.reset) {
+          options.reset()
+        }
         return cms.api.git.reset({ files: [id] })
       },
       onChange(formState) {
+        if (options.onChange) {
+          options.onChange(formState.values)
+        }
```
Enigmatical, line 93: And again.

Enigmatical:
> @Enigmatical Given how forms are central to Tina, what do we need to communicate with the community? Will it impact existing websites? Are there some changes to apply? What could be affected?

Apologies, @DirtyF , for not responding to this immediately.  I had to get the parent to this PR ready to go first.

Looking through the code, I've outlined the changes I made.  I think the only piece that requires documentation is the changes I made to `ContentCreator` which I've outlined in this issue:

https://github.com/tinacms/tinacms.org/issues/911

The other packages that are affected do not need to have their documentation changed.  Instead, as I was alluding to, with these changes, they are actually _more_ accurate.

For example, `next-tinacms-markdown` has a section for `Use Markdown Form`:

https://github.com/tinacms/tinacms/tree/master/packages/next-tinacms-markdown#usemarkdownform

Which states that it (optionally) took a `FormOptions` object, but not _entirely_.  Before my changes, even if you passed a `reset` or `onChange`, the callback would never be called:

https://github.com/tinacms/tinacms/blob/master/packages/next-tinacms-markdown/src/use-markdown-form.ts

So, with my changes, the documentation is _more_ accurate, does that make sense?

jeffsee55 (approved):

packages/gatsby-tinacms-mdx/src/useMdxForm.tsx
dwalkr: I think a pirate wrote this initially

dwalkr (approved):
el gee tee em

PR #1859, opened 2021-06-21
feat(tinacms): use cursor-based pagination in media manager

dwalkr:
This is essential to support Cloudinary and likely other DAMs in the media manager. The API changes reflect a relaxing of the needed pagination data from Media Stores, moving from a "SQL-style" `limit` + `offset` calculation to a less-opinionated `prevOffset` / `nextOffset` approach.

- [x] Update Media types to support cursor-based pagination
- [x] Update Media Manager UI to use cursor-based pagination instead of numbered
- [x] Update all media stores in monorepo to support new Media API
- [x] Polish pagination styles
- [x] Remove support for custom pagination plugins
    - *This was an experimental feature added to work around UI bugs with page numbers in large media libraries. Existing paginator plugins are expected to break due to the API changes anyway. I think the best course of action is to remove support for these plugins until the needs are better understood.*
- [x] Prepare documentation for release

### DIFF
--- a/packages/react-tinacms-strapi/README.md
+++ b/packages/react-tinacms-strapi/README.md
@@ -26,9 +26,7 @@ export default function MyApp({ Component, pageProps }) {
         apis: {
           strapi: new StrapiClient("http://localhost:1337/"),
         },
-        media: {
-          store: new StrapiMediaStore("http://localhost:1337/"),
-        },
+        media: new StrapiMediaStore("http://localhost:1337/"),
       }),
     []
   );
--- a/packages/@tinacms/core/src/cms.ts
+++ b/packages/@tinacms/core/src/cms.ts
@@ -75,6 +75,13 @@ export interface CMSConfig {
   plugins?: Array<Plugin>
   apis?: { [key: string]: any }
   media?: MediaStore
+  /**
+   * Might potentially make more sense to consolidate both media.store and mediaOptions.pageSize
+   * under the 'media' config namespace, but that's a breaking change.
+   */
+  mediaOptions?: {
+    pageSize?: number
+  }
 }
 
 export class CMS {
@@ -129,6 +136,9 @@ export class CMS {
 
     if (config.media) {
       this.media.store = config.media
+    }
+    if (config.mediaOptions && config.mediaOptions.pageSize) {
+      this.media.pageSize = config.mediaOptions.pageSize
     }
 
     if (config.plugins) {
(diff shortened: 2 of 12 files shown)


dwalkr:
Relevant docs are in https://github.com/tinacms/tinacms.org , will work on a PR to go out when these changes are **released**. Will want to merge + prerelease before that happens.

dwalkr:
docs PR is here: https://github.com/tinacms/tinacms.org/pull/910

I am going to write an upgrade guide to attach to the changelog when this is published. It will appear in the GitHub release notes as well as the auto-generated changelogs on the website.

jbevis (approved):
@dwalkr this all looks good to me. I like the simplification of the APIs and in general the move towards a less opinionated approach to work better with a variety of asset managers.

packages/react-tinacms-strapi/README.md
```diff
@@ -26,9 +26,7 @@ export default function MyApp({ Component, pageProps }) {
         apis: {
           strapi: new StrapiClient("http://localhost:1337/"),
         },
-        media: {
-          store: new StrapiMediaStore("http://localhost:1337/"),
-        },
+        media: new StrapiMediaStore("http://localhost:1337/"),
```
jeffsee55, line 29: I think we had some confusion about what `store` meant when implementing the `image` field in the GraphQL form [builder](https://github.com/tinacms/tina-graphql-gateway/pull/268/files#diff-084e602bd8601f418ba3b371e7f3d985817c0b70a1a2ceb70e64d8cedd8492b2R175). What's the effect of moving this up? Seems like we'll want to standardize that

jeffsee55 (approved):

packages/react-tinacms-strapi/README.md
dwalkr: This config change was made long ago, the docs here are just out of date and I updated them while looking over the various media store docs.

IMO this is really a bad change that necessitate the creation of the `mediaOptions` namespace in CMS config (I actually noted that here: https://github.com/tinacms/tinacms/pull/1859/files#diff-dd2f5d315632b09ae87ea8636cca1060d2e72b0136520d43aacc2166f19c0ba8R79 ) but reverting back to the "old way" would actually be a breaking change now.

dwalkr: The `cms.media` situation is a little confusing; the object available at `cms.media` implements `MediaStore` but actually does a little more than what is denoted by that interface, and it wraps the userland Media Store.

The Media Store (as denoted by the `MediaStore` interface) is meant to describe the methods that need to be implemented in userland (or "plugin-land") to interface with whichever media provider is being used

PR #1868, opened 2021-07-08
feat(tinacms): enable custom media list error via plugin

dwalkr:


packages/tinacms/src/components/media/media-manager.tsx
```diff
@@ -80,6 +95,10 @@ export function MediaPicker({
     if (cms.media.isConfigured) return 'loading'
     return 'not-configured'
   })
+  const [listError, setListError] = useState<Error>(
+    () => new Error('Error fetching media')
```
jeffsee55, line 99: This seems strange but actually kind of ok 🤔 . Would it make it more straightforward to just store the error message string in state and pass that down to the `ErrorComponent`? On the flip side, I suppose having access to the entire error might allow for a richer component.

jeffsee55 (approved):
Looks good, curious about the error function being stored in state but if it works 🤷

dwalkr:
opted to solve this via a more opinionated abstraction in https://github.com/tinacms/tinacms/pull/1869

PR #1869, opened 2021-07-09
feat(tinacms): customize list error message by throwing a MediaListError

dwalkr:
alternative to https://github.com/tinacms/tinacms/pull/1868

this enables customizing error feedback by throwing an instance of `MediaListError` which enables you to customize the title, explanation message, and link to documentation.

This feels simpler, and allows Media Stores to include troubleshooting instructions directly.

```
import { MediaListError } from 'tinacms'

class MyCustomMediaStore {
  // ...
  async list(options) {
    // ...
    throw new MediaListError({
      title: 'Missing API Key'
      message: 'We couldn't find the API key',
      docsLink: 'https://example.com/my-media-store-docs',
    })
  }
}
```

### DIFF
--- a/packages/@tinacms/core/src/media.ts
+++ b/packages/@tinacms/core/src/media.ts
@@ -249,3 +249,21 @@ export interface SelectMediaOptions {
   directory?: string
   onSelect?(media: Media): void
 }
+
+interface MediaListErrorConfig {
+  title: string
+  message: string
+  docsLink: string
+}
+
+export class MediaListError extends Error {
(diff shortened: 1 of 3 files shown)


Enigmatical (approved):
I think I prefer this over the `Plugin` approach.  Personally, defining the `MediaListError` within the scope of the `class` feels a lot better than defining a `MediaListErrorPlugin` and then having to register it with the CMS.  Feels more contained?

jeffsee55 (approved):

PR #1882, opened 2021-07-20
add rest of packages to workspaces

dwalkr:


packages/@tinacms/toolkit/src/components/media/media-manager.tsx
```diff
@@ -19,7 +19,7 @@ limitations under the License.
 import React from 'react'
 import { useEffect, useState } from 'react'
 import styled, { css } from 'styled-components'
-import { useCMS } from '../../react-tinacms'
+import { useCMS } from '../../react-tinacms/use-cms'
```
Enigmatical, line 22: Can you help me understand the difference between importing `useCMS` from `../../react-tinacms/use-cms` versus other places where we import from `tinacms`?

(Example being packages/react-tinacms-github/src/github-client/GithubFile.ts)

dwalkr: this is to fix a circular dependency internally within this package. external packages will just be importing from `tinacms`.

the existence of both is likely vestigial at this point now that everything is merged, but that will need to wait for a future refactor.

Enigmatical (commented):
Aside from my one question about the use of `useCMS()`, this looks good 👍 

I also appreciate the extra effort of cleaning up a few things too - breaking out `BlockAction` into its own component, `getOffset` becoming its own import, etc.  Very nice!

packages/@tinacms/toolkit/src/components/media/media-manager.tsx
Enigmatical: Ah, that makes sense. 👍

Enigmatical (approved):

PR #1885, opened 2021-07-20
parse media item from within the ImageField itself

jeffsee55:
Previously the Cloudinary image parsing happened from a `parse` hook we were hijacking from the `useGraphqlForms` API. This would make it more "core", though I'm admittedly a little lost on what `media` vs `media.store` represents. And we don't have an optional type on the `parse` function from the `MediaStore` interface. Didn't want to add that until I get some more info on how this might be problematic.

EDIT: select "Ignore Whitespace" to see actual diff

### DIFF
--- a/packages/@tinacms/toolkit/src/packages/fields/plugins/ImageFieldPlugin.tsx
+++ b/packages/@tinacms/toolkit/src/packages/fields/plugins/ImageFieldPlugin.tsx
@@ -68,69 +68,69 @@ export function usePreviewSrc(
   return [src, loading]
 }
 
-export const ImageField = wrapFieldsWithMeta<InputProps, ImageProps>(props => {
-  const cms = useCMS()
-  const { form, field } = props
-  const { name, value } = props.input
-  const [src, srcIsLoading] = usePreviewSrc(
-    value,
-    name,
-    form.getState().values,
-    field.previewSrc
-  )
+export const ImageField = wrapFieldsWithMeta<InputProps, ImageProps>(
+  (props) => {
+    const cms = useCMS()
+    const { form, field } = props
+    const { name, value } = props.input
+    const [src, srcIsLoading] = usePreviewSrc(
+      value,
+      name,
+      form.getState().values,
+      field.previewSrc
+    )
 
-  let onClear: any
-  if (props.field.clearable) {
-    onClear = () => props.input.onChange('')
-  }
-
-  function onChange(media?: Media) {
-    if (media) {
-      const mediaParse = cms.media.store
-
-      const parsedValue =
-        // @ts-ignore
-        typeof cms?.media?.store?.parse === 'function'
-          ? // @ts-ignore
-            cms.media.store.parse(media)
-          : // @ts-ignore
-            mediaParse.parse(media)
-      props.input.onChange('')
-      props.input.onChange(parsedValue)
+    let onClear: any
+    if (props.field.clearable) {
+      onClear = () => props.input.onChange('')
     }
-  }
 
-  const uploadDir = props.field.uploadDir || (() => '')
+    function onChange(media?: Media) {
+      if (media) {
+        const parsedValue =
+          // @ts-ignore
+          typeof cms?.media?.store?.parse === 'function'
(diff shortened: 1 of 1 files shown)


packages/@tinacms/toolkit/src/packages/fields/plugins/ImageFieldPlugin.tsx
```diff
@@ -68,62 +68,69 @@ export function usePreviewSrc(
...
+  const uploadDir = props.field.uploadDir || (() => '')
+
+  return (
+    <ImageUpload
+      value={value}
+      previewSrc={src}
+      loading={srcIsLoading}
+      onClick={() => {
+        const directory = uploadDir(props.form.getState().values)
+        cms.media.open({
```
Enigmatical, line 112: Definitely see what you mean about `media` vs. `media.store`.  Are the APIs available to `media` more global?  As in, the act of `open` is universal, regardless of the `store`?

packages/@tinacms/toolkit/src/packages/fields/plugins/ImageFieldPlugin.tsx
```diff
@@ -68,62 +68,69 @@ export function usePreviewSrc(
...
+    if (media) {
+      const mediaParse = cms.media.store
+
+      const parsedValue =
+        // @ts-ignore
+        typeof cms?.media?.store?.parse === 'function'
+          ? // @ts-ignore
+            cms.media.store.parse(media)
+          : // @ts-ignore
+            mediaParse.parse(media)
```
Enigmatical, line 97: Trying to make sense of this... if `cms.media.store.parse` is a `function`, then use it.  If not... use `cms.media.store.parse`?  Might be missing the point here, but this almost reads like we are sourcing the same `parse()` method.

Enigmatical (commented):
Not as familiar with the Cloudinary approach, but I left some comments.

packages/@tinacms/toolkit/src/packages/fields/plugins/ImageFieldPlugin.tsx
jeffsee55: This wasn't a change I made, just a whitespace diff, I probably mixed up the prettier formatter during migration

packages/@tinacms/toolkit/src/packages/fields/plugins/ImageFieldPlugin.tsx
jeffsee55: That's a mistake, I meant to just pass `media` for the fallback condition. Will fix

Enigmatical (approved):
I'm fine with this.  I still wouldn't mind having a discussion on `media` vs. `media.store`, but that doesn't have to happen with this field.

jbevis:
Been reviewing this and revisiting our ```Media``` and ```MediaStore``` interfaces just to try and solidify what we're trying to do.

as far as ```media``` vs ```media.store``` is concerned, it would appear that the `parse` would be tied to the store instance and specifically how we're changing the data for a given Media file. The `media` key on the cms appears to be a way to scope the functionality and logic of the store under media, while also providing a developer the ability to control access to the store outside of the baked in screen plugin.

I'm a bit confused on where the `parse` function even comes from. @jeffsee55 you mentioned above how the MediaStore interface doesn't allow for an optional parse, but I don't even see where the MediaStore provides for any parse. Could just be I'm looking at outdated code though.

Overall, I agree with Chris, I think this is a fine solution for now. It does speak to the broader issue of maybe do a more wholesale reexamination of Media in core.

jbevis (approved):

jbevis:
@jeffsee55 is this PR still relevant? Image field/type appears to be working, and wasn't sure if we still need this one.

jeffsee55:
It is? 🤔 

Do you have it working with all the latest stuff on the starter?

jbevis:
Latest in `new-design` branch of starter is working.

```
...
     {
          type: "image",
          name: "heroImg",
          label: "Hero Image",
        },
...
```

PR #1897, opened 2021-07-26
Fix Editor plugin

jeffsee55:
Closes https://github.com/tinacms/tinacms/issues/1896

This PR is in 2 parts:
1. Get the editor working by commenting out the Image popup, not sure why it's not working
2. Set final-form react-final-form and tinacms-react-inline as direct dependencies so you don't need to add them all when using the editor. It seems to be working for me although I'm not sure if `useField` hooks will get mixed up and if we're using them here.

At the very least part 1 will unblock us

### DIFF
--- a/examples/tina-cloud-starter/.tina/__generated__/_schema.json
+++ b/examples/tina-cloud-starter/.tina/__generated__/_schema.json
@@ -65,7 +65,7 @@
           "name": "body",
           "isBody": true,
           "ui": {
-            "component": "textarea"
+            "component": "markdown"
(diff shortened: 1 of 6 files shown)


Enigmatical (approved):

logan-anderson:
> Get the editor working by commenting out the Image popup, not sure why it's not working
I am guessing this might have something to do with it. https://github.com/tinacms/tinacms/commit/cbd632ea5e3b0f119a0b7d7cb76e75b0d8384097

Issue #1901, opened 2021-07-26
Make sure media works on starter in monorepo


jeffsee55:
The image field `onChange` is throwing an error. There was an attempt to fix it but I wasn't really sure what I was doing. This needs to be revisited with someone more knowledgable about what it should do https://github.com/tinacms/tinacms/pull/1885

logan-anderson:
Error is `TypeError: n.src.includes is not a function`. Have to add `mediaStore={TinaCloudCloudinaryMediaStore}` to tina-wrapper.

PR #1906, opened 2021-07-26
fix media in the starter

logan-anderson:
I think this is more of a band aid solution more than anything.

Fixes https://github.com/tinacms/tinacms/issues/1901


jeffsee55:
Yeah agreed, this doesn't feel right but if it works it'll have to do. 

@dwalkr curious on your insight here.

dwalkr:
seems like the call to `onChange('')` could be causing it (https://github.com/tinacms/tinacms/pull/1906/files#diff-c8e402f2922aa38679245802380c0cc425af158c94104e7f022f99657f321c72R96)

there is maybe a bit of a discrepancy to the basic `onChange` sending the field value, but media preferring to work with `Media` objects as much as possible. but this fix seems fine for now.

dwalkr (approved):

jeffsee55 (approved):

PR #1907, opened 2021-07-26
Use new apis and patterns for adding Tina to your site

jeffsee55:


packages/@tinacms/cli/src/cmds/init/index.ts
```diff
@@ -93,12 +87,12 @@ export async function tinaSetup(ctx: any, next: () => void, options) {
     fs.writeFileSync(blogPostPath, blogPost)
   }
 
-  // 2. Create a Tina Wrapper
-  if (!fs.pathExistsSync(TinaWrapperPath)) {
-    logger.info(logText('Adding a tina-wrapper...'))
-    fs.mkdirpSync(TinaWrapperPathDir)
-    fs.writeFileSync(TinaWrapperPath, TinaWrapper)
-  }
+  // // 2. Create a Tina Wrapper
```
jeffsee55, line 90: @logan-anderson I don't think we need a wrapper of any kind anymore, any reason you can think of?

packages/@tinacms/cli/src/cmds/init/index.ts
```diff
@@ -127,7 +121,8 @@ export async function tinaSetup(ctx: any, next: () => void, options) {
       wrapper = true
       logger.info(
         dangerText(
-          'you will have to wrap your app in a tina-wrapper component. For more information see generated file'
+          `Heads up, to enable live-editing you'll need to wrap your page or site in Tina:\n`,
+          warnText(AppJsContent)
```
jeffsee55, line 125: This prints out the component for easy copy/paste, 🤔  seems ok. It only shows if they choose not to override their app.tsx

packages/@tinacms/cli/src/cmds/init/index.ts
logan-anderson: If a components/tina-wrapper.tsx is not needed then this is not needed. So yes

packages/@tinacms/cli/src/cmds/init/index.ts
logan-anderson: I believe the AppJsContent is also in the getting started I frame.

logan-anderson: > if they choose not to override their app.tsx
Yeah this only happens when they choose not to override there App.js or App.tsx

jeffsee55: > I believe the AppJsContent is also in the getting started I frame.

I added it in there because it's pretty small now, they'd be able to copy and paste it quickly

packages/@tinacms/cli/src/cmds/init/index.ts
```diff
@@ -139,7 +134,7 @@ export async function tinaSetup(ctx: any, next: () => void, options) {
   const tinaBlogPagePathFile = p.join(tinaBlogPagePath, '[filename].tsx')
   if (!fs.pathExistsSync(tinaBlogPagePathFile)) {
     fs.mkdirpSync(tinaBlogPagePath)
-    fs.writeFileSync(tinaBlogPagePathFile, nextPostPage({ wrapper }))
+    fs.writeFileSync(tinaBlogPagePathFile, nextPostPage())
```
logan-anderson, line 137: Should we still pass wrapper here? That way we can render the edit state in the generated file if they choose not to use a _app.js ?

jeffsee55: You mean dump it out into `/components/tina-wrapper`? I'm not sure we should give them anything else to worry about. We'd like them to just put it in `_app.js` I think

logan-anderson: Previously if the user did not want to update there _app.js the CLI would wrap /demo/blog/[filename].tsx with the edit provider and Tina Provider so that edit would still work on that page (it would not work on the rest of their site). We dont have to worry about maintaining that behaviour for this PR I can add it back latter if needed.

logan-anderson: I think it would be ideal to merge and test this before tomorrow. Probably better to merge and release it before we worry about other features.

dwalkr (approved):

packages/@tinacms/cli/src/cmds/init/index.ts
jeffsee55: > Previously if the user did not want to update there _app.js the CLI would wrap /demo/blog/[filename].tsx

Ok I see, instead of adding it to a separate file would it be better to just wrap the default export in `/demo/blog/[filename].tsx`? That way it's all housed in one file.

jeffsee55:
@logan-anderson merging this to keep things unblocked but if I'll either circle back to this to do as you suggest or if you get around to it feel free to tackle

PR #1909, opened 2021-07-26
chore: Excludes examples folder when determining files needing license

Enigmatical:
Trying to change the dangerfile to exclude files in the `examples/` folder.  I believe it needs to be a separate PR to work.

### DIFF
--- a/dangerfile.ts
+++ b/dangerfile.ts
@@ -198,7 +198,7 @@ function checkForLicense({ packageJson }: TinaPackage) {
  *
  */
 function fileNeedsLicense(filepath: string) {
-  return new RegExp(/^(?!examples\/).+\.(js|tsx?)$/).test(filepath)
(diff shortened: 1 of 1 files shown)


logan-anderson (approved):

dangerfile.ts
```diff
@@ -198,7 +198,7 @@ function checkForLicense({ packageJson }: TinaPackage) {
  *
  */
 function fileNeedsLicense(filepath: string) {
-  return new RegExp(/\.(js|tsx?)$/).test(filepath)
+  return new RegExp(/^(?!examples\/).+\.(js|tsx?)$/).test(filepath)
```
dwalkr, line 201: likely not that relevant since our source files should be typescript, but might as well include `jsx` files

```suggestion
  return new RegExp(/^(?!examples\/).+\.(jsx?|tsx?)$/).test(filepath)
```

dwalkr (approved):

PR #1926, opened 2021-07-28
move InlineWysiwyg to react-tinacms-inline

dwalkr:
this removes the dependency on react-tinacms-inline from react-tinacms-editor


jeffsee55 (approved):

DirtyF (approved):
I approve this message

PR #1936, opened 2021-07-28
parse date string in date field

### LINKED ISSUE #1903: Datepicker causes error when saving without touching the given field
The datepicker `parse` function doesn't seem to fire unless the user touches it, causing an invalid date to be sent through. 

`unexpected error - invalid date`. 

The backend may be expecting a format which has been setup by the schema. It may be the case that if you provided a specific format that things will work untouched. Need to do some discovery on if this is an easy fix. 

If this can't be fixed in the next day or so we should just remove the `date` field from the Tina Cloud Starter

jbevis:
This PR addresses the bug with the date field where an invalid Date error is thrown if the date field is untouched on a save.

Closes #1903

### DIFF
--- a/packages/@tinacms/graphql/src/primitives/resolver/index.ts
+++ b/packages/@tinacms/graphql/src/primitives/resolver/index.ts
@@ -636,7 +636,6 @@ export class Resolver {
   }
 }
 
-const DEFAULT_DATE_FORMAT = 'MMM dd, yyyy'
 const resolveDateInput = (
   value: string,
   field: { dateFormat?: string; timeFormat?: string }
@@ -644,43 +643,13 @@ const resolveDateInput = (
   /**
    * Convert string to `new Date()`
    */
-  const parsedValue = new Date(value).toISOString()
-  const date = parseISO(parsedValue)
+  const date = parseISO(value)
   
   if (!isValid(date)) {
     throw 'Invalid Date'
   }
-
-  /**
-   * Remove any local timezone offset (putting the date back in UTC)
-   * https://stackoverflow.com/questions/48172772/time-zone-issue-involving-date-fns-format
-   */
-  const dateUTC = new Date(
-    date.valueOf() + date.getTimezoneOffset() * 60 * 1000
-  )
-
-  /**
-   * Determine dateFormat
-   * This involves fixing inconsistencies between `moment.js` (that Tina uses to format dates)
-   * and `date-fns` (that Gateway uses to format dates).
-   * They are explained here:
-   * https://github.com/date-fns/date-fns/blob/master/docs/unicodeTokens.md
-   */
-  const dateFormat = field.dateFormat || DEFAULT_DATE_FORMAT
-  const fixedDateFormat = dateFormat.replace(/D/g, 'd').replace(/Y/g, 'y')
-
-  /**
-   * Determine timeFormat, if any
-   */
-  const timeFormat = field.timeFormat || false
-
-  /**
-   * Format `date` and `time` parts
-   */
-  const datePart = format(dateUTC, fixedDateFormat)
-  const timePart = timeFormat ? ` ${format(dateUTC, timeFormat)}` : ''
-
-  return `${datePart}${timePart}`
+  
+  return value
 }
 
 type FieldParams = {
--- a/packages/@tinacms/graphql/src/primitives/resolver/index.ts
+++ b/packages/@tinacms/graphql/src/primitives/resolver/index.ts
@@ -649,7 +649,11 @@ const resolveDateInput = (
     throw 'Invalid Date'
   }
   
-  return value
+  const dateUTC = new Date(
+    date.valueOf() + date.getTimezoneOffset() * 60 * 1000
+  )
+  
+  return dateUTC.toUTCString()
 }
 
 type FieldParams = {
--- a/packages/@tinacms/graphql/src/primitives/resolver/index.ts
+++ b/packages/@tinacms/graphql/src/primitives/resolver/index.ts
@@ -529,9 +529,8 @@ export class Resolver {
       case 'string':
         if (field.options) {
           if (field.list) {
-            // FIXME: this is awaiting checkbox suppport
             return {
-              component: 'checkbox',
+              component: 'checkbox-group',
               ...field,
               ...extraFields,
               options: field.options,
--- a/packages/@tinacms/graphql/src/primitives/resolver/index.ts
+++ b/packages/@tinacms/graphql/src/primitives/resolver/index.ts
@@ -201,7 +201,6 @@ export class Resolver {
             await this.database.put(realPath, {
               _template: lastItem(template.namespace),
             })
-            return this.getDocument(realPath)
         }
       }
       const templateInfo =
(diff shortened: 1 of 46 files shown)


DirtyF (dismissed):
Thanks @jbevis

packages/@tinacms/graphql/src/primitives/resolver/index.ts
```diff
@@ -644,7 +644,9 @@ const resolveDateInput = (
   /**
    * Convert string to `new Date()`
    */
-  const date = parseISO(value)
+  const parsedValue = new Date(value).toISOString()
```
dwalkr, line 647: This seems redundant, as `parseISO` returns a date object as well. So here we're creating a date object, converting it to ISO, and then parsing it from ISO back into a date.

If we want to proceed with this change, we could accomplish the same with `const date = new Date(value)` but I'm concerned that this could just be masking the problem by using a more generous parser

dwalkr (changes requested):

packages/@tinacms/graphql/src/primitives/resolver/index.ts
jbevis: I was thinking about this more last night after I pushed this up.

That check was the first fix to handle the untouched date field, since the value in that situation would come through as `"July 12, 2021"` for example, thus hitting the invalid date error.

Probably not the best solution but this would at least take that string, and convert to a date object before going through the ISO parse.

I'm sure there is a better way to address.

packages/@tinacms/graphql/src/primitives/resolver/index.ts
```diff
@@ -644,7 +644,9 @@ const resolveDateInput = (
...
+  const date = parseISO(parsedValue)
```
Enigmatical, line 648: I only have one concern here.  If `value` is `invalid` for any reason, I think `parsedValue` will fail and could cause the whole resolver to fail?  It is why I originally had the check for `isValid()`.  I would just want to make sure `parsedValue` can fail gracefully (maybe just a `try... catch...` around that)?

Enigmatical (dismissed):
I have a concern around using `new Date()` on a potentially bad `value`, but if that has been thoroughly tested, I've no issues giving this a 👍

packages/@tinacms/graphql/src/primitives/resolver/index.ts
dwalkr: The purpose of lines 647 & 648 here are to create a date object. The proposed change parses the passed-in value using the Date constructor's parsing method (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/parse), converts the resulting `Date` object into an ISO string, and then parses that ISO string back into a `Date` object. That's what I meant about it being redundant.

If we think that using `Date.parse` instead of `parseISO` is the right solution to the problem, then we can do that more explicitly without passing through the arbitrary ISO conversion. But that has the feel of trying to address a symptom rather than the cause, since the frontend should be able to pass whatever format the backend needs

packages/@tinacms/graphql/src/primitives/resolver/index.ts
dwalkr: the problem with a try-catch when the format is bad is, what do you do to recover? It honestly makes sense for the backend to error out here to avoid data corruption

packages/@tinacms/graphql/src/primitives/resolver/index.ts
jeffsee55: And this is a problem because we're literally sending in the string `Jul 2, 2020` right? And we can't enforce it to be UTC because final form doesn't call parse on that field unless it's touched. Do I have that right? If that's the case, the best course of action may be to require UTC format from the UI and not allow another format

packages/@tinacms/graphql/src/primitives/resolver/index.ts
jbevis: So what kind of concern should the parsing be though?

Right now, the DateFieldPlugin is handling the formatting of the date, and then we're essentially passing bad data to the backend. So should the backend have a check to convert say a string to a Date, or should the backend fail, and then it shifts to a client end concern for converting the data.

packages/@tinacms/graphql/src/primitives/resolver/index.ts
jeffsee55: It sounds like the proper fix is for the input the field to have its own local state of the _formatted_ version of the time, and final form only knows about the UTC version.

packages/@tinacms/graphql/src/primitives/resolver/index.ts
dwalkr: The backend is already parsing a string into a Date, via `parseISO`. What is currently in question is whether we want to be more generous about which date formats the backend accepts (relying on `Date.parse` to figure it out.) If the backend receives a date string that `Date.parse` doesn't know how to parse, we're still left figuring out how to deal with the error.

packages/@tinacms/graphql/src/primitives/resolver/index.ts
dwalkr: this seems related https://github.com/final-form/react-final-form/issues/431

packages/@tinacms/graphql/src/primitives/resolver/index.ts
jeffsee55: I feel like the backend should only accept UTC, and we need to deal with it on the frontend. I was probably worried too much about how the data is stored when Chris and I originally discussed this

packages/@tinacms/graphql/src/primitives/resolver/index.ts
jeffsee55: Yeah, that issue seems to indicate that they don't have a concept of presentation vs value, which leads me to think we need to manage the presentation ourselves with local state

packages/@tinacms/graphql/src/primitives/resolver/index.ts
jbevis: @jeffsee55 I agree that the backend should only be accepting UTC, but when talking about addressing this on the frontend are you meaning this transformation occur in local state in `DateFieldPlugin.tsx` or should that be a concern on a per app basis. e.g. a developer is addressing this on a per component basis?

jbevis:
@jeffsee55 and discussed more in depth and decided on removing the burden of formatting the date time from the backend. 

Latest changes update `resolveDateInput` to expect the incoming value to be a UTC date string, and saves this as the date value to the fmt. The DateFieldPlugin field can still be configured to format the display date in the schema, but the formatting of the date value now shifts to a concern of the react component rendering the data. 

Added a quick a dirty formatting in `examples/tina-cloud-starter`

This removes some of the confusion and pain points regarding "presentation" vs "value". Developers can control UI formatting via the schema, and format the date as needed. "value" remains consistent through the dataflow.

jbevis:
This has fallen a bit too far behind `main`

I'm going to close it in favor of a new pr https://github.com/tinacms/tinacms/pull/1965

PR #1950, opened 2021-07-30
GitHub extendable

jamespohalloran:
We're looking to extend the GithubBridge Class in the cloud repo.
We can't currently override the instance methods, as they are currently properties
https://stackoverflow.com/questions/53997107/allow-subclass-to-override-with-either-a-function-property-or-a-method

I also added a few extra types to the export.


jamespohalloran:
Could I get a hand cutting a canary release of this to test before marking ready for review? @tinacms/maintainers (I should remember how to properly do this myself, but always forget :) )

jeffsee55 (approved):

PR #1963, opened 2021-08-02
expose a createGlobalForm function in formify args

dwalkr:
the formify API can provide a simple way to create global forms in its callback, similar to the existing createForm


Enigmatical (approved):
:chefskiss:

PR #1965, opened 2021-08-03
Fix/date format

### LINKED ISSUE #1903: Datepicker causes error when saving without touching the given field

jbevis:
Closes #1903 

Discussion surrounding Date and this work covered in this PR, which has become a bit stale.
https://github.com/tinacms/tinacms/pull/1936

Opening a new one for more streamlined review.

Updates the backend to expect and store dates as UTC timestamps. This moves onus to format dates to the developer and outside the concerns of Tina within the sidebar.

### DIFF
--- a/packages/@tinacms/graphql/src/primitives/resolver/index.ts
+++ b/packages/@tinacms/graphql/src/primitives/resolver/index.ts
@@ -656,7 +656,8 @@ const resolveDateInput = (
   const dateUTC = new Date(
     date.valueOf() + date.getTimezoneOffset() * 60 * 1000
   )
-
+  
+  return dateUTC.toUTCString()
   /**
    * Determine dateFormat
    * This involves fixing inconsistencies between `moment.js` (that Tina uses to format dates)
--- a/packages/@tinacms/graphql/src/primitives/resolver/index.ts
+++ b/packages/@tinacms/graphql/src/primitives/resolver/index.ts
@@ -201,6 +201,7 @@ export class Resolver {
             await this.database.put(realPath, {
               _template: lastItem(template.namespace),
             })
+            return this.getDocument(realPath)
         }
       }
       const templateInfo =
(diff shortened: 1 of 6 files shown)


dwalkr (approved):

packages/@tinacms/graphql/src/primitives/resolver/index.ts
```diff
@@ -530,8 +529,9 @@ export class Resolver {
       case 'string':
         if (field.options) {
           if (field.list) {
+            // FIXME: this is awaiting checkbox suppport
             return {
-              component: 'checkbox-group',
+              component: 'checkbox',
```
Enigmatical, line 534: I'm not quite sure this PR is up-to-date with the latest changes to `tinacms`?  I believe my PR to convert this to `checkbox-group` was merged in last week?

packages/@tinacms/graphql/src/primitives/resolver/index.ts
```diff
@@ -201,7 +201,6 @@ export class Resolver {
             await this.database.put(realPath, {
               _template: lastItem(template.namespace),
             })
-            return this.getDocument(realPath)
```
Enigmatical, line 204: And this was to fix the bug with `useDocumentCreator()` and `templates` instead of a `fields`.

Enigmatical (changes requested):
Seems to be based on an out-of-date `tinacms`?  Check comments.

packages/@tinacms/graphql/src/primitives/resolver/index.ts
jbevis: Good eye @Enigmatical.

There must have been some oddities locally that made their way into that file, because I do not recall touching those parts of the file. 

Either way, those changes have been reverted.

Enigmatical (approved):

PR #1966, opened 2021-08-03
Extendable GitHub w/ interface

jamespohalloran:
Continuation of https://github.com/tinacms/tinacms/pull/1950

I forgot to update the Bridge interface (and other backends) to use methods instead of props, so that they can be overridden.


logan-anderson:
@jamespohalloran Looks good to me. It looks like there is a danger issue and it looks like the build is stuck. I am guessing neither one of these is because of this PR. Maybe you need to update the branch to reflect the current main? We have been doing some changes around our CI builds. I am guessing this branch got caught in a weird state.

logan-anderson (approved):

PR #1985, opened 2021-08-03
use collection label as form label

dwalkr:


Enigmatical (approved):
Simple enough.  I assume `label` is a required field for a `Collection`.

PR #1986, opened 2021-08-03
added regex to add css files to generated file

logan-anderson:
This PR looks at the users `_app.{js,tsx}` file and adds all css imports that where in it to the generate file.


spbyrne:
Big if true

perkinsjr (dismissed):
LGTM as discussed.

jeffsee55 (approved):

PR #1988, opened 2021-08-03
fix(tinacms): Test if currentFields is an Array

Enigmatical:
This cropped up when adding a new NavItem to the Starter.  The error message:

`currentFields.find() is not a function.`

Debugging this, it appears that `currentFields` is not always a function of `fields` and can sometimes be a singular `field`.

The fix appears to be simply to test if `currentFields` is an Array, otherwise, the value of `currentFields` is already a singular `field`.

Fixes https://github.com/tinacms/tina-cloud-starter/issues/141
Fixes https://github.com/tinacms/tina-cloud-starter/issues/143

### DIFF
--- a/.changeset/stale-tables-heal.md
+++ b/.changeset/stale-tables-heal.md
@@ -1,0 +1,6 @@
+---
+'tina-cloud-starter': patch
+'tinacms': patch
+---
+
+Handles situations where `currentFields` is not an Array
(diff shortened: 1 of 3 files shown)


DirtyF (approved):
LGTM

dwalkr (approved):

PR #1989, opened 2021-08-03
Fix issue with _body being written to frontmatter

jeffsee55:
Fix issue where the `isBody` field wasn't properly removing that value from frontmatter. Ensure that the field is not treating any differently for JSON format

Closes https://github.com/tinacms/tinacms/issues/1955


logan-anderson (approved):
Tested with local client and worked for me!

Enigmatical (approved):
Looks good 👍

PR #1990, opened 2021-08-04
added a better error message for childProcess

logan-anderson:
The issues we are seeing with windows are only happening when installed from NPM and not happening on the monorepo. This PR is to add a better error message for more context into what is happening.

### DIFF
--- a/.changeset/mighty-moons-divide.md
+++ b/.changeset/mighty-moons-divide.md
@@ -1,0 +1,5 @@
+---
+'@tinacms/cli': minor
+---
+
+Added better error handleing for childProcess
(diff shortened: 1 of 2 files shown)


logan-anderson:
Just did some digging and this. https://stackoverflow.com/questions/60386867/node-spawn-child-process-not-working-in-windows may be the cause of our issues. Not sure why it would work in the mono repo thought. I am  going to try adding it and see what happens.

dwalkr (approved):

jeffsee55:
> Not sure why it would work in the mono repo thought

@logan-anderson yarn v2 handles inconsistencies between shells, which explains why it only works over there. Do we need to add fsevents? https://www.npmjs.com/package/fsevents

PR #1999, opened 2021-08-04
Feat/set image on drag n drop

### LINKED ISSUE #1945: Media: drag-and-drop is not reflecting changes on the image field in the sidebar
I added my cloudinary credentials to my env file but I'm getting the following error when trying to add an image 

```Unhandled Runtime Error

TypeError: n.src.includes is not a function
```

jbevis:
Closes #1945

This PR addresses the above issue in that when a user drags and drops an image file, it will upload to Cloudinary, but does not set the value of the field properly.

This was due to an issue with the `persist` method in the next-tinacms-cloudinary package not returning an appropriate Media object.

### DIFF
--- a/.changeset/friendly-bananas-brake.md
+++ b/.changeset/friendly-bananas-brake.md
@@ -2,4 +2,4 @@
 'next-tinacms-cloudinary': patch
 ---
 
-The `persist` method on the Cloudinary Media Store was not returning a properly parsed response from the Cloudinary API. As such, the ImageFieldPlugin could not properly set the image field on drag and drop. The latest changes add new formatting logic to `persist`.
+Fix drag & drop behavior in ImageFieldPlugin
--- a/.changeset/friendly-bananas-brake.md
+++ b/.changeset/friendly-bananas-brake.md
@@ -1,5 +1,5 @@
 ---
-'next-tinacms-cloudinary': minor
+'next-tinacms-cloudinary': patch
 ---
 
 The `persist` method on the Cloudinary Media Store was not returning a properly parsed response from the Cloudinary API. As such, the ImageFieldPlugin could not properly set the image field on drag and drop. The latest changes add new formatting logic to `persist`.
--- a/.changeset/friendly-bananas-brake.md
+++ b/.changeset/friendly-bananas-brake.md
@@ -1,0 +1,5 @@
+---
+'next-tinacms-cloudinary': minor
+---
+
+The `persist` method on the Cloudinary Media Store was not returning a properly parsed response from the Cloudinary API. As such, the ImageFieldPlugin could not properly set the image field on drag and drop. The latest changes add new formatting logic to `persist`.
--- a/packages/@tinacms/toolkit/src/packages/fields/plugins/ImageFieldPlugin.tsx
+++ b/packages/@tinacms/toolkit/src/packages/fields/plugins/ImageFieldPlugin.tsx
@@ -79,7 +79,6 @@ export const ImageField = wrapFieldsWithMeta<InputProps, ImageProps>(
       form.getState().values,
       field.previewSrc
     )
-    const [imgLoading, setImgLoading] = useState(false)
     let onClear: any
     if (props.field.clearable) {
       onClear = () => props.input.onChange('')
@@ -93,7 +92,6 @@ export const ImageField = wrapFieldsWithMeta<InputProps, ImageProps>(
             ? // @ts-ignore
               cms.media.store.parse(media)
             : media
-        props.input.onChange('')
         props.input.onChange(parsedValue)
       }
     }
@@ -103,7 +101,7 @@ export const ImageField = wrapFieldsWithMeta<InputProps, ImageProps>(
       <ImageUpload
         value={value}
         previewSrc={src}
-        loading={imgLoading}
+        loading={srcIsLoading}
         onClick={() => {
           const directory = uploadDir(props.form.getState().values)
           cms.media.open({
@@ -113,7 +111,6 @@ export const ImageField = wrapFieldsWithMeta<InputProps, ImageProps>(
           })
         }}
         onDrop={async ([file]: File[]) => {
-          setImgLoading(true)
           const directory = uploadDir(props.form.getState().values)
           const [media] = await cms.media.persist([
             {
@@ -121,8 +118,8 @@ export const ImageField = wrapFieldsWithMeta<InputProps, ImageProps>(
               file,
             },
           ])
+          console.log({media})
           onChange(media)
-          setImgLoading(false)
         }}
         onClear={onClear}
       />
--- a/packages/@tinacms/toolkit/src/packages/fields/plugins/ImageFieldPlugin.tsx
+++ b/packages/@tinacms/toolkit/src/packages/fields/plugins/ImageFieldPlugin.tsx
@@ -118,7 +118,6 @@ export const ImageField = wrapFieldsWithMeta<InputProps, ImageProps>(
               file,
             },
           ])
-          console.log({media})
           onChange(media)
         }}
         onClear={onClear}
--- a/packages/@tinacms/toolkit/src/packages/fields/plugins/ImageFieldPlugin.tsx
+++ b/packages/@tinacms/toolkit/src/packages/fields/plugins/ImageFieldPlugin.tsx
@@ -79,6 +79,7 @@ export const ImageField = wrapFieldsWithMeta<InputProps, ImageProps>(
       form.getState().values,
       field.previewSrc
     )
+    const [isImgUploading, setIsImgUploading] = useState(false)
     let onClear: any
     if (props.field.clearable) {
       onClear = () => props.input.onChange('')
@@ -101,7 +102,7 @@ export const ImageField = wrapFieldsWithMeta<InputProps, ImageProps>(
       <ImageUpload
         value={value}
         previewSrc={src}
-        loading={srcIsLoading}
+        loading={isImgUploading || srcIsLoading}
         onClick={() => {
           const directory = uploadDir(props.form.getState().values)
           cms.media.open({
@@ -111,6 +112,7 @@ export const ImageField = wrapFieldsWithMeta<InputProps, ImageProps>(
           })
         }}
         onDrop={async ([file]: File[]) => {
+          setIsImgUploading(true)
           const directory = uploadDir(props.form.getState().values)
           const [media] = await cms.media.persist([
             {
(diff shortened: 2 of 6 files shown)


packages/@tinacms/toolkit/src/packages/fields/plugins/ImageFieldPlugin.tsx
```diff
@@ -97,14 +97,13 @@ export const ImageField = wrapFieldsWithMeta<InputProps, ImageProps>(
...
       }
     }
-
     const uploadDir = props.field.uploadDir || (() => '')
 
     return (
       <ImageUpload
         value={value}
         previewSrc={src}
-        loading={srcIsLoading}
```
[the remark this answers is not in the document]
jbevis: `srcIsLoading` doesn't appear to be evaluating to true during an image upload / update and so we're not getting the loading dots as any kind of visual affordance in the side bar.

Maybe it's not working as expected or my understanding of what it's doing is off. @dwalkr any insights on this one?

jbevis:
Tinkering with this a bit more. Seems like the imageSrc isn't updated in the form values consistently.

packages/@tinacms/toolkit/src/packages/fields/plugins/ImageFieldPlugin.tsx
```diff
@@ -93,11 +92,9 @@ export const ImageField = wrapFieldsWithMeta<InputProps, ImageProps>(
             ? // @ts-ignore
               cms.media.store.parse(media)
             : media
-        props.input.onChange('')
```
jbevis, line 96: I'm not 100% what this was doing, but what was occurring on the dnd upload / set is the `props.input.onChange(parsedValue)` wasn't making it through to values, just and empty string.

I removed this and everything seemed to be working smoothly.

Happy to bring it back if it's crucial.

packages/@tinacms/toolkit/src/packages/fields/plugins/ImageFieldPlugin.tsx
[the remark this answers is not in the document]
jbevis: Pushed a few changes that remove the extra `imgLoading` check in favor of the `srcIsLoading`. Able to consistently drag and drop files to upload and set as the image and see it reflected.

The boolean value isn't dynamically changing, and as such there is not visual affordance to indicate we're waiting on a file upload (it does take a few seconds).

In short, it works. But need to add that extra bit of polish and feedback for the user.

DirtyF:
How can we test this?

logan-anderson (dismissed):
Looking good to me! Tested it and everything seems to be working

logan-anderson:
@DirtyF this would have to be tested int he mono repo. Check out [this file](https://github.com/tinacms/tinacms/blob/main/CONTRIBUTING.md#contributing-to-tinacms) to get started

packages/@tinacms/toolkit/src/packages/fields/plugins/ImageFieldPlugin.tsx
dwalkr: from what I recall, uploading an image with the same filename (overwriting) was not triggering the `onChange` in some cases. I think this was from before the introduction of the media manager, and the fact that we now send objects through instead of strings may have changed this

packages/@tinacms/toolkit/src/packages/fields/plugins/ImageFieldPlugin.tsx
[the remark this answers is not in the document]
dwalkr: I agree w/ Jeff here, we likely need to keep track of two loading states:

1. retrieving the image URL asynchronously (use case: GithubClient + private repo), done by `usePreviewSrc`
2. uploading an image and waiting for it to be available on the server

we never really solved for 2 because I guess we weren't drag+droppers and were focused on using controls in the media manager interface

packages/next-tinacms-cloudinary/src/cloudinary-media-store.ts
```diff
@@ -44,14 +44,18 @@ export class CloudinaryMediaStore implements MediaStore {
...
+     */
+    const parsedRes: Media = {
+      type: fileRes.resource_type === 'image' ? 'file' : 'dir',
+      id: fileRes.public_id,
+      filename: fileRes.original_filename,
+      directory: '/',
+      previewSrc: fileRes.url
 
-    // TODO: be programmer
-    // NOTE: why do we need this?
```
dwalkr, line 49: just noting here, I believe images were taking a second to appear in the cloudinary API response after being uploaded, so this delay was just a bodge that waits a couple seconds before resolving `persist` to help ensure the new image is there. If removing this doesn't seem to be causing any problems then it's probably fine, but I'd keep an eye out for it

dwalkr: note that the scenario in which this problem will most likely appear is when uploading an image to the media manager from the media manager UI. In that case it does a re-fetch after the upload, and if the re-fetch happens too soon, the new image will not appear.

jbevis: Ah, I was definitely running into that. I'll keep testing it out to check for consistency, but at the very least we can bring that back.

.changeset/friendly-bananas-brake.md
```diff
@@ -0,0 +1,5 @@
+---
+'next-tinacms-cloudinary': minor
```
dwalkr, line 2:
```suggestion
'next-tinacms-cloudinary': patch
```

`patch` is preferred for bugfixes

.changeset/friendly-bananas-brake.md
```diff
@@ -0,0 +1,5 @@
...
+---
+
+The `persist` method on the Cloudinary Media Store was not returning a properly parsed response from the Cloudinary API. As such, the ImageFieldPlugin could not properly set the image field on drag and drop. The latest changes add new formatting logic to `persist`.
```
dwalkr, line 5:
```suggestion
Fix drag + drop behavior in ImageFieldPlugin
```

Here's my suggestion for the changelog entry. They don't always have to be this terse but IMO it should focus on what the impact of updating the package will be for the developer using it, and doesn't need to go deep into internal details

packages/next-tinacms-cloudinary/src/cloudinary-media-store.ts
```diff
@@ -44,14 +44,21 @@ export class CloudinaryMediaStore implements MediaStore {
       throw new Error(responseData.message)
     }
     const fileRes = await res.json()
-
-    // TODO: be programmer
-    // NOTE: why do we need this?
     await new Promise((resolve) => {
```
dwalkr, line 47:
```suggestion
    /**
    * Images uploaded to Cloudinary aren't instantly available via the API;
    * waiting a couple seconds here seems to ensure they show up in the next fetch.
    */
    await new Promise((resolve) => {
```
Might be good to throw in an explainer now that we've decided to keep it in

packages/next-tinacms-cloudinary/src/cloudinary-media-store.ts
```diff
@@ -44,14 +44,21 @@ export class CloudinaryMediaStore implements MediaStore {
...
       setTimeout(resolve, 2000)
     })
+    /**
+     * Format the response from Cloudinary to match Media interface
+     */
+    const parsedRes: Media = {
+      type: fileRes.resource_type === 'image' ? 'file' : 'dir',
```
dwalkr, line 54: is this correct? How do non-image assets (video, audio, text) come through the API? I wouldn't want to handle those as directories

packages/@tinacms/toolkit/src/packages/fields/plugins/ImageFieldPlugin.tsx
```diff
@@ -114,16 +112,18 @@ export const ImageField = wrapFieldsWithMeta<InputProps, ImageProps>(
...
             {
               directory,
               file,
             },
           ])
-
-          onChange(media)
+          if (media) {
+            await onChange(media)
+            setIsImgUploading(false)
```
dwalkr, line 125: consider adding a try-catch-finally here or something to deactivate the loading state in the event of an error

packages/@tinacms/toolkit/src/packages/fields/plugins/ImageFieldPlugin.tsx
dwalkr: I'm gonna hold up merging here because I want to test out of this `onChange` is still needed or not

dwalkr (changes requested):

packages/next-tinacms-cloudinary/src/cloudinary-media-store.ts
jbevis: Good catch, I'd need to check Cloudinary docs on how they handle that, but agree that check isn't robust enough

packages/@tinacms/toolkit/src/packages/fields/plugins/ImageFieldPlugin.tsx
jbevis: So, looks as though `props.input.onChange('')` does indeed prevent the parsedValue from being passed to the form. Essentially meaning this could be a whole rabbit hole.

Wondering if there is some kind of basic check we could throw in the interim...

e.g. 
```
  if (typeof media store ==== Cloudinary) => onChange(parsedValue)
  else => onChange('') .... onChange(parsedValue)
```

jbevis:
Latest on this effort:

@dwalkr I've implemented your suggestions on the changeset and adding some inline comments to explain some choices.

Digging into Cloudinary docs some more, looks as though the upload API only accepts `image` `video` `raw` and `auto`. Essentially we're only going to be sending files, so we'll explicitly set that on the response formatting.

Finally, I have not been successful with the two `onChange` calls. Removing the empty string check allows for things to work as expected. There is a rabbit hole to go down for sure to ensure how this can work with other Media Store implementations. For the sake of the immediate, since we're leaning into the Cloudinary solution we could proceed and come back to the issue as needed.

packages/@tinacms/toolkit/src/packages/fields/plugins/ImageFieldPlugin.tsx
```diff
@@ -121,8 +123,14 @@ export const ImageField = wrapFieldsWithMeta<InputProps, ImageProps>(
             },
           ])
           if (media) {
-            await onChange(media)
-            setIsImgUploading(false)
+            try {
+              await onChange(media)
+              setIsImgUploading(false)
```
dwalkr, line 128:
```suggestion
```
will be handled by the `finally` block

packages/@tinacms/toolkit/src/packages/fields/plugins/ImageFieldPlugin.tsx
dwalkr: That's not really feasible, unfortunately. I'm fine with eating this regression until we can find a better way around it

dwalkr (approved):

PR #2014, opened 2021-08-09
Changed media from only being images to more types

logan-anderson:


perkinsjr (dismissed):
LGTM

logan-anderson:
@perkinsjr I removed some console.logs but it should be good to go again.

perkinsjr (approved):

packages/next-tinacms-cloudinary/src/cloudinary-media-store.ts
```diff
@@ -25,7 +25,7 @@ export class CloudinaryMediaStore implements MediaStore {
   fetchFunction = (input: RequestInfo, init?: RequestInit) => {
     return fetch(input, init)
   }
-  accept = '*'
+  accept = 'text/*,  application/*, image/*'
```
DirtyF, line 28: what does it mean concretely for Cloudinary users, what type of files are now supported?

Enigmatical: These are MIME types:
https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types/Common_types

I couldn't find a list of _just_ `text`, `application`, or `image` types, but if you go through that list, you'll see the `MIME Type` column.  If the `MIME Type` begins with `text/`, `application/` or `image/`, we'll accept it.

I'm not quite sure there is an exhaustive list (especially for `application/` where that could be any application's file type).

DirtyF: So we accept things like zip files now, should we be concerned about security?

mystrdat: Could we get support for `audio/` MIME type?

PR #2015, opened 2021-08-09
Fix/loading state

logan-anderson:
Not sure why this PR changes deps. I pulled down mono repo and yarn `yarn install` and it changed a lot of stuff.

This should be a one line change to fix: https://github.com/tinacms/tina-cloud-starter/issues/149

This is the only change that needs to be made: https://github.com/tinacms/tinacms/pull/2015/files#diff-580eb812378c74b17309abdda8f39a9b133e433cc74d5289a80330335eb0454aR139

Edit: cloned a new version of the repo and the same thing happened. Running `yarn` after cloning yields this. 
Screen Shot 2021-08-09 at 11 01 36 AM


packages/tinacms/src/hooks/use-graphql-forms.tsx
```diff
@@ -136,6 +136,7 @@ export function useGraphqlForms<T extends object>({
   }, [pendingReset])
 
   React.useEffect(() => {
```
dwalkr, line 138: I can't make an inline suggestion this far away from changes apparently, but we should add a `setIsLoading(false)` into the `.catch` block as well

logan-anderson: 👍 Done

Enigmatical (approved):
I'm not 100% on the package situation, but I know from `gateway` that we usually just _trust_ `yarn install`.

dwalkr (approved):

PR #2021, opened 2021-08-09
Fix/datetime to iso

jbevis:
PR will update the datetime field so only save the date format as an ISO string vs UTC.

### DIFF
--- a/.changeset/smart-mayflies-change.md
+++ b/.changeset/smart-mayflies-change.md
@@ -1,0 +1,6 @@
+---
+'@tinacms/graphql': minor
+'@tinacms/toolkit': minor
+---
+
+Update datetime field to expect and receive ISO string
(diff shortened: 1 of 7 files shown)


Enigmatical (commented):
So, I want to start by saying that I've tested this locally and the changes here are behaving as we expected 👍 

That said, there appears to be an issue with the `DateFieldPlugin` itself.  I noticed that whenever I save a date, the timezone is UTC regardless.  So, if I use the datepicker, and choose 10:00 AM (CST), I would _expect_ to see it saved as 3:00 PM (UTC).  Instead, I'm seeing 10:00 AM (UTC) saved.

The reason is that the `DateFieldPlugin` has `utc` set to `true` by default - meaning all dates are saved in UTC regardless of the user's local timezone:

https://github.com/tinacms/tinacms/issues/256
https://github.com/tinacms/tinacms/pull/326

I understand the "bug" that Frank found - if I choose July 3, I meant July 3.  I didn't mean July 2 (UTC).  The core problem is around converting locale to UTC and then UTC back to locale.

I wonder if we should make an additional change to the way `DateFieldPlugin` works... or if we should leave it?  But then the user has to use mental math and know that when they choose a date, they are choosing the date in UTC (and it will be displayed in UTC).  That feels wrong?

Thoughts?

jeffsee55 (approved):

Enigmatical (approved):

PR #2031, opened 2021-08-11
Fix/cli css regex

logan-anderson:
Fixed issue where it would not find css imports that uses `'`

cc @scottgallant @perkinsjr


perkinsjr (approved):
LGTM I think this should work for both

PR #2046, opened 2021-08-18
Build with vite 2

jeffsee55:
https://www.loom.com/share/5c6b13ac43ea4af286831496e9999fd1

This uses `ebsuild` for node packages and `vite` for browser ones (or for packages with both, allows you to choose). The builds are super fast, it's a single process that watches the whole repo, which keeps things lighter weight (I hope). One of the bigger changes is that for the most part we don't need to run typescript builds. We just set the `index.d.ts` file to rexport the `src`. So the big remaining thing to do would be to replace those files with another build step which you only need to run during CI before publishing.

Also adds the `nuke` command back 

EDIT: This is now working as hoped.

### Some of the main things this PR does:
- Updates all packages to use the latest Typescript
- Commonizes tsconfig for all pacakges. Fixes and ignores a few errors as a result
- Uses ES6 as target, removing support for ES5. We were already doing this in newer packages, but now it's the behavior for all of them.
- Uses [sucrase ](https://github.com/alangpierce/sucrase/tree/main#readme) as a ts-jest replacement. Sucrase is kind of like esbuild, but has less issues when it comes to the strange things that Jest does with their mocking. [Read about some of the problems with esbuild and jest](https://github.com/aelbore/esbuild-jest/issues/21#issuecomment-809646016). Speeds up tests dramatically and doesn't require proper types before running.
- Splits the Typescript build from the regular build. Allows us to only build types before publishing (read "Building Types" below on how this is intended to work)

### Main concerns with this PR
- I think we probably want to just move this logic into @tinacms/scripts. I built it in a different package for simplicity, but no reason it can't be moved.
- I had to fix a lot of stuff, mostly just did `@ts-ignore`, I think this is mainly because rollup tsc didn't fully respect the tsconfig, and the raw `tsc` command does, so we probably need to take a look at the tsconfig.json to make sure we like everything it's doing.
- Only building for ES6. I don't _think_ we still need to support ES5, but we should make sure we communicate that.
- The disconnect of building types before publishing. In theory we could forget to build types and the published package would point to `src` files for it's types, which won't be there. We may want to build in some tooling to avoid this.

### Building types
When developing, the `build/watch` commands populate the `dist/index.d.ts` and any other entry point files with something like:
```ts
// dist/index.d.ts
export * from '../src/index'
// Note that `export type * from '../src/index'` is not supported
```
When building before publishing, we need to run `yarn types`, which will then run the plain `tsc` command, which overwrites the `d.ts` files generated by the build command.

This isn't a pattern I've seen before, but I don't see anything wrong with it, it saves us a ton of time during dev since there's no requirement for running a typescript server, and since VSCode runs one of it's own we get all the benefits. The downside would be if we _wanted_ a failing TS type to also fail a build, but I personally find that really disruptive. I could definitely see us throwing in a commit hook though, which would block pushes until types have been built, but that also might be really annoying. My thinking is that the only time we need to make sure types are there is before we publish on CI.


packages/next-tinacms-cloudinary/package.json
```diff
@@ -7,6 +7,15 @@
   ],
   "typings": "dist/index.d.ts",
   "license": "Apache-2.0",
+  "buildConfig": {
+    "entryPoints": [
+      "src/index.ts",
+      {
+        "name": "src/handlers.ts",
+        "target": "node"
+      }
```
jeffsee55, line 16: We can target different environments from the same package, by default the entrypoint is `src/index.ts`, if you specifiy `buildConfig.entryPoints`, you'll be in control. If you only give a string, we'll assume `browser` as the target.

packages/react-tinacms-inline/tsconfig.json
```diff
@@ -16,6 +16,7 @@
     "noImplicitThis": true,
     "alwaysStrict": true,
     "noUnusedLocals": true,
+    "isolatedModules": true,
```
jeffsee55, line 19: For `esbuild`, we're not building types, so we need to explicitly export types so the build knows it can ignore them:
```ts
// good
export type { MyInterface } from './other-file.ts'
// bad
export { MyInterface } from './other-file.ts'
```

packages/@tinacms/tina-build/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@tinacms/tina-build",
-  "version": "0.50.0",
+  "version": "0.2.51",
```
Enigmatical, line 3: This doesn't seem correct...?

packages/tinacms/src/hooks/use-content-creator.tsx
```diff
@@ -24,6 +24,7 @@ export type FilterCollections = (
 
 export type DocumentCreatorArgs = {
   onNewDocument?: OnNewDocument
+  testIt: boolean
```
Enigmatical, line 27: Is this just for debugging?

Enigmatical (dismissed):
Left a couple of questions, but otherwise, I don't see any problems with this.  I'm surprised we weren't using `import type` and `export type` before, honestly.

Enigmatical (commented):
Sorry, I meant to "Comment" not "Approve".

logan-anderson:
Looking really good! speedy fast! Did find one small issue. It seems like the packages are not being "swapped out" when being rebuilt. Here is a demo
https://www.loom.com/share/0e576bf440854ae8ada5efb79b2e5caa

jeffsee55:
@logan-anderson  the CLI doesn't hot-reload, the Graphql package should though

Enigmatical (approved):

DirtyF:
😲  *All* packages build in 19s on my machine ⚡  🚀

Issue #1969, opened 2021-07-05
feat: error when port 4001 is already in use


DirtyF:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

Given I already run a tina instance in another terminal
When I run yarn dev
Then I get an error: 

```
yarn run v1.22.10
$ yarn tina-gql server:start -c "next dev"
$ /Users/frank/github/tina/blog-starter-app/node_modules/.bin/tina-gql server:start -c 'next dev'

events.js:353
      throw er; // Unhandled 'error' event
      ^

Error: listen EADDRINUSE: address already in use :::4001
    at Server.setupListenHandle [as _listen2] (net.js:1318:16)
    at listenInCluster (net.js:1366:12)
    at Server.listen (net.js:1452:7)
    at start (/Users/frank/github/tina/blog-starter-app/node_modules/tina-graphql-gateway-cli/dist/index.js:660:18)
    at processTicksAndRejections (internal/process/task_queues.js:95:5)
Emitted 'error' event on Server instance at:
    at emitErrorNT (net.js:1345:8)
    at processTicksAndRejections (internal/process/task_queues.js:82:21) {
  code: 'EADDRINUSE',
  errno: -48,
  syscall: 'listen',
  address: '::',
  port: 4001
}
```

### Expected Behavior

A simpler message like: Port 4001 already in use.

### Steps To Reproduce

1. Run yarn dev in a tina cloud project
2. Run yarn dev in another tina cloud project

### Environment

```markdown
- Browser: Firefox
- OS: macOS
- Node: 14
- npm: 7
```

### Anything else?

Not critical.

CatalinCsnMaster:
I can take this issue 😁

logan-anderson:
@CatalinCsnMaster That would be awesome! The code for the CLI can be found [here](https://github.com/tinacms/tinacms/tree/main/packages/%40tinacms/cli). Please let us know if you have any issues contributing. 

The code for each command can be found here: https://github.com/tinacms/tinacms/tree/main/packages/%40tinacms/cli/src/cmds

I have a feeling the error is coming from the server start command.

CatalinCsnMaster:
@logan-anderson I got a little confused with the changeset. 
But I figured it out and next time the pr will be prettier 😃

DirtyF:
Fixed in #2047

PR #2047, opened 2021-08-18
feat(tinacms/cli): add simpler err message

CatalinCsnMaster:
Relatest to #1969

### DIFF
--- a/packages/@tinacms/cli/src/cmds/start-server/index.ts
+++ b/packages/@tinacms/cli/src/cmds/start-server/index.ts
@@ -42,14 +42,18 @@ export async function startServer(
         stdio: 'inherit',
         shell: true,
       })
-      ps.on("error", (code) =>{
-        logger.error(dangerText(`An error has occurred in the Next.js child process. Error message below`))
+      ps.on('error', (code) => {
+        logger.error(
+          dangerText(
+            `An error has occurred in the Next.js child process. Error message below`
+          )
+        )
         logger.error(`name: ${code.name}
 message: ${code.message}
 
-stack: ${code.stack || "No stack was provided"}`)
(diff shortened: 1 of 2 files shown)


packages/@tinacms/cli/src/cmds/start-server/index.ts
```diff
@@ -105,11 +109,16 @@ stack: ${code.stack || "No stack was provided"}`)
   const start = async () => {
     const s = require('./server')
     state.server = await s.default(experimental)
+
     state.server.listen(port, () => {
       logger.info(`Started Filesystem GraphQL server on port: ${port}`)
       logger.info(`Visit the playground at http://localhost:${port}/altair/`)
     })
-    state.server.on('connection', socket => {
+    state.server.on('error', function () {
```
logan-anderson, line 117: instead of doing this for all errors can we only display the error message on a more specific code.

```ts
state.server.on('error', function (e) {
      if (e.code === 'EADDRINUSE') {
        // If port is already in use warn the user and fail
        logger.error(
          dangerText(
            `ERROR: Can not start server. Port 4001 already in use. Please kill other processes and try again.`
          )
        )
        process.exit()
      }
      throw e
    })
```

logan-anderson (changes requested):
Looks good! Tested it and all is working. Thanks for your contribution. Just found one small detail that would be nice to have fixed before it is merged.

packages/@tinacms/cli/src/cmds/start-server/index.ts
CatalinCsnMaster: Yes, of course. I'm sorry, I don't know how I missed that 😅

logan-anderson (approved):
🎉

PR #2048, opened 2021-08-18
Active field indicator

dwalkr:
This PR aims to add visual feedback in the document when editing values in the sidebar by adding a dotted border to sections of the page that correspond to a blocks field in the sidebar.

https://user-images.githubusercontent.com/15221702/131728187-5815f9d6-e56e-4ff0-b8cc-0b093aa0c30e.mov

This has been experimented with previously, but this PR's approach is notably different because it doesn't rely on using the `ref` attribute.

## Pros and Cons of Refs

The hardest part of implementing this behavior is getting access to the right DOM element. It is assumed to be practically impossible to figure out which part of the DOM is being edited by a given form control without some guidance from the implementing developer, so the idea is to provide some API for the developer to provide that information to the system.

The experimental [ref-based inline editing](https://github.com/tinacms/tinacms/pull/1749) APIs involved requesting a ref object from the system and attaching it to a React component. From there the system has access to the underlying DOM elements for the purpose of getting its location and floating contextual information (like a dotted outline or some form controls) over the right part of the page. 

Much of the reasoning for this approach revolved around **breaking the dependency chain** - offering a "separation point" between Tina code and the code for the user's website. Because refs are a React concept, not a Tina concept, this approach felt like the right direction.

The main drawback of this approach is the somewhat unsolved story of **how and where to get these special refs in the first place**. `useFieldRef` can provide this special ref object to developers, but the specifics of how this would be done in the context of an entire project is currently left unexplored. I envisioned that, at the very least, component designers could accommodate the need to inject a ref into their components, but in practice this would be cumbersome.

## Bouncing off the DOM

Even while working on the ref-based approach, I had ultimately wanted a "dumber" way for developers to achieve this linking. My idea was to use a data-attribute with the name of the field as a plain string; this would allow components on the page to be "Tina-connected" without their source code needing any Tina libraries. 

The stumbling point always came back to trying to link up these data attributes to the system without jumping out of React and falling into imperative-DOM-land. However, it finally occurred to me that we can jump back into React's render flow from the DOM level with a single control point.

It works like this:

1. `setActiveField` fires off a custom DOM event containing the name of the field that should be activated.
2. The `ActiveFieldIndicator` component subscribes to this DOM event
3. When `setActiveField` fires, `ActiveFieldIndicator` searches the DOM for an element with a `data-tinafield` that matches the field name in the event payload.
4. `ActiveFieldIndicator` updates an internal state value with the results of this search.

After the state value is updated, we're back in React's component scope with a reference to the DOM element we need, as if we had fussed around with all that `ref` business.

## Things to be considered

The potential to link a part of your page to Tina's data model by just including an attribute is very appealing to me. I could see this potentially being incorporated at the design system level:

```jsx
export const Hero = ({title}) => (
  <div className="hero">
    <h1 data-tinafield="title">{title}</h1>
  </div>
)
```

Because there is already an implicit contract in the property name, it doesn't seem unreasonable that the same contract might be enforced by hard-coding a `data-tinafield` value. In this case, `data-tinafield` should be aware of the ability to nest fields within other fields. How this works in practice, I'm not certain yet.

Additionally, `data-tinafield` doesn't currently account for pages with multiple forms, but I imagine this would be relatively simple to incorporate.


packages/@tinacms/toolkit/src/packages/fields/plugins/BlocksFieldPlugin.tsx
```diff
@@ -232,7 +234,11 @@ const BlockListItem = ({
             {...provider.dragHandleProps}
           >
             <DragHandle />
-            <ItemClickTarget onClick={() => setExpanded(true)}>
+            <ItemClickTarget
+              onClick={() => setExpanded(true)}
+              onMouseOver={() => setActiveField(fieldName)}
```
jeffsee55, line 239: This alone will be super useful

jeffsee55:
I wholeheartedly support this approach. It's simpler than the ref-based one, and you're right that at the end of the day all we need to now is which DOM node belongs to which field.

I also think this is a perfect starting point that I could see us getting to production sooner than much of the stuff to follow since this is totally opt-in and very little effort.

> Because there is already an implicit contract in the property name, it doesn't seem unreasonable that the same contract might be enforced by hard-coding a data-tinafield value. In this case, data-tinafield should be aware of the ability to nest fields within other fields. How this works in practice, I'm not certain yet.

My feeling is these data attributes _will_ need to be fully namespaced to their respective field, though I'd love it if that weren't the case. Obviously as objects grow more complex it will get frustrating to have to do something like `getPostDocument.blocks[0].cta.actions[1].link`. That was the impetus of the refs coming _directly_ from a  hook we exposed, outlined in [this comment](https://github.com/tinacms/rfcs/pull/14#issuecomment-747711703). But I'm also hoping we can come up with something better. And this API is simple enough to grasp that I feel confident we can figure out some nice APIs to build out on top of it.

I agree that this is the ideal API:
```jsx
export const Hero = ({title}) => (
  <div className="hero">
    <h1 data-tinafield="<insert field name here>">{title}</h1>
  </div>
)
```
And getting that name in there does feel like the hardest part of the new inline editing direction. But if we get it right, it'll be a gamechanger. Either way, that's more of a DX concern, I feel like what you have here is the groundwork for actually making a viable solution.

jeffsee55:
Bonus feature that would be cool - if the active field is off the page, on hover we pop up an arrow pointing in the direction of the node (like in Super Smash Bros). On focusing the field we scroll to the node.

jbevis:
@dwalkr this looks awesome! I really think this elevates the editing experience. 

One knit pick though is a bit of a visual bug when clearing a text field. I created a quick loom to demonstrate.
https://www.loom.com/share/10d47372a1744e939abbf6c1145d38c7

I'm not even sure how to resolve, and I don't think it's necessarily a blocker.

Enigmatical (dismissed):
From a code standpoint, this looks good.  Definitely good for a first pass at this feature. 👍

dwalkr:
> @dwalkr this looks awesome! I really think this elevates the editing experience.
> 
> One knit pick though is a bit of a visual bug when clearing a text field. I created a quick loom to demonstrate.
> https://www.loom.com/share/10d47372a1744e939abbf6c1145d38c7
> 
> I'm not even sure how to resolve, and I don't think it's necessarily a blocker.

I'm glad you caught this. After experimenting with `MutationObserver` I've refactored this change in a way that fixes this bug, and I believe lays the groundwork for two-way communication between the sidebar and the (userland) DOM.

packages/@tinacms/toolkit/src/components/ActiveFieldIndicator.tsx
```diff
@@ -0,0 +1,184 @@
...
+  return (
+    <IndicatorWrap position="bottom">
+      <ArrowWrap>
+        <ChevronDownIcon />
+      </ArrowWrap>
+    </IndicatorWrap>
+  )
+}
+
+const useScrollToFocusedField = () => {
```
jeffsee55, line 70: [scrollIntoView](https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollIntoView) is a pretty slick API which may take care of all of this automatically. I remember there may be some footguns when there are nested scrollable elements, but I'm not sure either approach will handle those well.

dwalkr: I'm pretty happy w/ the scroll behavior. If `scrollIntoView` weren't experimental it might be worth looking into for the sake of simplifying this code, but as it is the current implementation is probably more future-proof.

packages/@tinacms/toolkit/src/components/MutationSignal.tsx
```diff
@@ -0,0 +1,49 @@
...
+    observer.observe(observerAreaRef.current as Node, {
+      childList: true,
+      subtree: true,
+      characterData: true,
+    })
+    return () => observer.disconnect()
+  }, [])
+  return (
+    <MutationSignalContext.Provider value={signal}>
+      <div ref={observerAreaRef as any}>{children}</div>
```
jeffsee55, line 41: I didn't expect us to be listening to everything from one observer like this, I was under the impression that each `data-tinaField` element would be `observed` individually rather than the entire DOM. What you have here seems simpler, but I wonder about performance. This does seem like the right place to start, though.

As I'm thinking about the more granular approach, it seems like it might be harder to know when a new `data-tinaField` element has entered the DOM, we'd have to have some sort of mechanism to trigger a DOM query, which then brings me back to the ref-based solution since it'd automatically register the DOM nodes as the enter the page. 🤔

jeffsee55 (approved):
This is going to be really nice, looking forward to using it. My two hesitations are:

1. Not having the form ID as part of the `data-tinaField` value may cause issues. I can see 2 forms both having a `title` field, for example
2. Listening to every subtree inside the TinaUI wrapper may be expensive. That seems like a premature concern for now though

dwalkr:
> This is going to be really nice, looking forward to using it. My two hesitations are:
> 
> 1. Not having the form ID as part of the `data-tinaField` value may cause issues. I can see 2 forms both having a `title` field, for example

100%, I've already found a minor bug in the cloud starter demo related to this. It will be implementation-dependent, but we'll definitely want to figure this out. I'd like to explore something a little more flexible than coding it directly into the field, but like you said this is functional enough to merge as-is and look into that later.

> 2. Listening to every subtree inside the TinaUI wrapper may be expensive. That seems like a premature concern for now though

Yeah that's definitely a concern I have. I'm not sure what alternatives we have if we want this to be really bulletproof though, particularly when it comes to components that don't render if their value is empty. It seemed heavy-handed at the time but I haven't seen it fall over yet.

PR #2052, opened 2021-08-20
Sample of using plugins for itemProps

jeffsee55:


DirtyF:
I don't know much about `itemProps`, if I understand correctly,  it generates the keys and labels for listed blocks, making it a plugin means less work for developers? Is that the goal here?

packages/@tinacms/toolkit/src/packages/fields/plugins/BlocksFieldPlugin.tsx
```diff
@@ -164,7 +166,16 @@ const Blocks = ({ tinaForm, form, field, input }: BlockFieldProps) => {
 
                   const itemProps = (item: object) => {
                     if (!template.itemProps) return {}
-                    return template.itemProps(item)
+                    if (typeof template.itemProps === 'function') {
+                      return template.itemProps(item)
+                    } else {
+                      const plugins = cms.plugins
+                        .all('itemProps')
+                        .filter(
```
dwalkr, line 174: `PluginTypeManager` has a more succinct way to get a specific named plugin:

`cms.plugins.getType('itemProps').find(template.itemProps.name)`

dwalkr:
This makes sense in terms of trying to make a serializable version of Tina config, but in practice I think it would feel kind of clunky. I'm wondering if it would be better expressed as a custom field plugin derived from the blocks/group field, to avoid introducing another plugin type here, or if `itemProps` would really be reusable enough to make sense as its own plugin.

jeffsee55:
>  I'm wondering if it would be better expressed as a custom field plugin derived from the blocks/group field, to avoid introducing another plugin type here, or if itemProps would really be reusable enough to make sense as its own plugin.

@dwalkr In the most aggressive terms about how I'm thinking about this: everything that Tina toolkit currently exposes as a _function_ that the user can provide, we should wrap up as a plugin _type_. 

I think the custom field logic is too broad as it stands. We kind of have this "you can do whatever you want with this" messaging for 'field' plugins, which I don't find very helpful because they're too broad and powerful. It's also very possible that you're concerns are more cross-cutting. If the example was for `validate` instead of `itemProps`, I wouldn't want to tell everyone that in order to provide a custom string validation function you need to supply your own custom field since you may want to validate a `text` and `textarea` UI field with the same string validation logic.

PR #2064, opened 2021-08-27
Improve the type definition

Ryuurock:
Thanks tinaCMS team
In the process of using tinaCMS, I found that the "data" type of BlockComponentProps was hard-coded to any. I suggest to modify it to generic
The type of Block is defined as `React.FC<BlockComponentProps>`, it will not be able to pass in class components, I suggest changing it to `React.ComponentType<BlockComponentProps>`

### DIFF
--- a/.changeset/violet-games-prove.md
+++ b/.changeset/violet-games-prove.md
@@ -1,0 +1,5 @@
+---
+'react-tinacms-inline': patch
+---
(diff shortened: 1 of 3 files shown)


logan-anderson (approved):
Thanks @Ryuurock For the contribution 🎉

PR #2075, opened 2021-09-01
Uses better default dateFormat for DateField

Enigmatical:
While dogfooding, I noticed some unusual behavior in the `react-datetime` picker when using `DEFAULT_DATE_FORMAT` (`MMM dd, yyyy`).

`dd` is actually not a supported token by `moment` (https://momentjs.com/docs/#/parsing/string-format/).  Swapping to `DD` resolves this.

We had not encountered this before because the `tina-cloud-starter` schema uses `DD` correctly for `dateFormat`.

I also made a change to better handle when a date is `invalid` (`parse` was failing).

### DIFF
--- a/.changeset/cyan-radios-decide.md
+++ b/.changeset/cyan-radios-decide.md
@@ -1,0 +1,6 @@
+---
+'tina-cloud-starter': patch
+'@tinacms/toolkit': patch
+---
+
+Uses a better default dateFormat for the DateField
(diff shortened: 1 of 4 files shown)


jbevis (approved):
Nice, this feels a lot better. Drier and some sounder checks on the date.

DirtyF (approved):

PR #2076, opened 2021-09-02
Fix/multiple image upload

jbevis:
Based on comments from folks about how editors / users in general would want to be able to upload many assets at once, I started poking around and realized the Cloudinary Media store was only handling the first media item in an array. Thus, if someone where to upload many files, or the first item was successfully posting to Cloudinary. 

This PR updates the `persist` method in the Cloudinary media store to handle multiple files, as well as adding in a loading state to the UI to provide more visual feedback.

@jeffsee55 I'll leave this in draft status for now since you mentioned we should discuss the issue I added to the UX improvements board, but this would at least be a first pass.

### DIFF
--- a/packages/@tinacms/toolkit/src/packages/fields/components/ImageUpload/ImageUpload.tsx
+++ b/packages/@tinacms/toolkit/src/packages/fields/components/ImageUpload/ImageUpload.tsx
@@ -147,9 +147,9 @@ export const ImageUpload = ({
             <ImageLoadingIndicator />
           ) : (
             <ImgPlaceholder>
-            Drag 'n' drop some files here,
+            Drag 'n' drop a file here,
             <br />
-            or click to select files
+            or click to select a file
           </ImgPlaceholder>
           )}
         </StyledImageContainer>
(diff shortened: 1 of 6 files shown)


packages/@tinacms/toolkit/src/packages/fields/components/ImageUpload/ImageUpload.tsx
```diff
@@ -147,9 +147,9 @@ export const ImageUpload = ({
             <ImageLoadingIndicator />
           ) : (
             <ImgPlaceholder>
-            Drag 'n' drop some files here,
+            Drag 'n' drop a file here,
             <br />
-            or click to select files
+            or click to select a file
           </ImgPlaceholder>
```
jbevis, line 153: Thinking about the image field, it really should only accept one file in this instance. I just updated the language to avoid confusion.

packages/react-tinacms-inline/src/fields/inline-image/dropzone-wrapper.tsx
```diff
@@ -38,6 +38,7 @@ export function DropzoneWrapper({
     accept: 'image/*',
     onDrop,
     noClick: !!onClick,
+    multiple: true
   })
```
jbevis, line 42: Ignore this.... shouldn't have been committed.

Enigmatical (approved):
Neat 👍

dwalkr (approved):

PR #2078, opened 2021-09-03
temporary fix for edit state context issues

logan-anderson:
There was an issue that was introduced with our new build set. For some reason the edit context is not getting shared. This caused the close button not to work. 

This PR is a temp fix that allows the close button to work once again.

### DIFF
--- a/packages/tinacms/src/auth/TinaCloudProvider.tsx
+++ b/packages/tinacms/src/auth/TinaCloudProvider.tsx
@@ -18,7 +18,7 @@ import { TinaCMS, TinaProvider, MediaStore } from '@tinacms/toolkit'
 import { Client } from '../client'
 import { useTinaAuthRedirect } from './useTinaAuthRedirect'
 import { CreateClientProps, createClient } from '../utils'
-import { useEditState } from '../edit-state'
+import { setEditing } from '../edit-state'
 
 type ModalNames = null | 'authenticate'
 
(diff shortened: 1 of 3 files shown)


perkinsjr (approved):
LGTM

packages/tinacms/src/auth/TinaCloudProvider.tsx
```diff
@@ -93,8 +92,14 @@ export const AuthWallInner = ({
...
               action: async () => {
-                setActiveModal(null)
-                setEdit(false)
+                // This does not work it looks like we have somehow getting two contexts
+                // console.log({ setEdit })
+                // setEdit(false)
+                // setActiveModal(null)
+
+                // This is a temp fix
+                setEditing(false) // set editing just sets the local storage
```
Enigmatical, line 101: Is there a separate bug card for tracking the actual issue?  Just want to make sure the "temp fix" doesn't become the "permanent fix" because we lost track of it.  Having two contexts should be a thing of the past...

Enigmatical (approved):

packages/tinacms/src/auth/TinaCloudProvider.tsx
logan-anderson: Good idea! https://github.com/tinacms/tinacms/issues/2081

PR #2079, opened 2021-09-08
remove demo-next

logan-anderson:
The demo-next example was not a good example of the current "happy path". This could be very confusing to someone who is learning tina and looks in our example folder.


perkinsjr (approved):
Nice! Approved.

Enigmatical (changes requested):
Given the size of this PR, I'd like for us to have a discussion about it before we merge.  While I don't disagree that `demo-next` wasn't the best example of Tina's "Happy Path," was it also a bad example of how to use the parts of the Toolkit, as well?

I might be speaking for myself here, but I used `demo-next` when studying how `useForm()` works and it also had a really good collection of `field` types and their configs.

If I'm alone on this, I'm fine with dropping `demo-next`, but just want to make sure we're all on the same page as a team.

logan-anderson:
> Given the size of this PR, I'd like for us to have a discussion about it before we merge. While I don't disagree that `demo-next` is not the best example of Tina's "Happy Path," was it also a bad example of how to use the certain parts of the Toolkit, as well?
> 
> I might be speaking for myself here, but I used `demo-next` when studying how `useForm()` works and it also had a really good collection of `field` types and their configs.
> 
> If I'm alone on this, I'm fine with dropping `demo-next`, but just want to make sure we're all on the same page as a team.

The main reason for removing this demo is that a few folks have stumbled upon it when looking for example code of tinaCMS. Since this is not really a path to using Tina Successfully they will be lead off track. 

There are a couple of other things we could do instead.
1. Update the readme to explain this
2. Move it to a sub-folder in the examples called "deprecated" or possibly "experimental"

What do you think of this @Enigmatical ?. Removing it will definitely insure that no one uses this to get started with Tina but I am also more than OK with this other idea.

Enigmatical (approved):
After a lengthy call with @jeffsee55 and a couple of days of discovery on my own, I think we've performed enough due diligence to agree that `demo-next` should be removed.

Appreciate @logan-anderson responding to my concerns here, ultimately, I agree that the patterns outlined in this example are not an ideal path going forward.

Enigmatical:
Closing this PR in favor of https://github.com/tinacms/tinacms/pull/2093

Ran into conflicts with `.pnp.js` and `yarn.lock` and after a couple of attempts resolving them, decided it would probably be best (and safer) to just remove `demo-next` off the latest `main`.

Issue #2056, opened 2021-08-25
Using `list: true` with `'type: "reference"`


logan-anderson:
Example field in schema.ts
```js
 {
          name: "featuredPosts",
          label: "Featured Posts",
          type: "reference",
          list: true,
          collections: ["posts"],
 },
```

Everything works as expected except the UI produced looks like this.

<img width="659" alt="Screen Shot 2021-08-24 at 3 12 16 PM" src="https://user-images.githubusercontent.com/43075109/130668204-d72695b7-e905-4222-8b76-778f0049fdb9.png">

I would expect that I would be able to add new posts to the list and delete posts from the list. Instead I am given a normal reference field. 

## Temporary work around

I can put this in my schema.ts file as a work around
```
        {
          type: "object",
          list: true,
          name: "featurePosts",
          label: "Featured Posts",
          fields: [
            {
              type: "reference",
              label: "Post",
              name: "post",
              collections: ["posts"],
            },
          ],
        },
```

DirtyF:
@logan-anderson what is the corresponding UI to the workaround you shared? Do you get a plus sign to add a new post?

eduoliveira85:
Hi! I've followed the workaround suggestion from @logan-anderson and what it returns from the reference is the path of the content file. Perhaps there's something else missing?

```
{
			type: "object",
			label: "Contatos de imprensa",
			name: "contacts",
			list: true,
			ui: {
				itemProps: (item) => {
					console.log('item', item)
					return {
						label: item?.name 
					}
				}
			},
			fields: [
				{
					name: 'name',
					label: 'nome',
					type: 'string'
					
				},
				{
					label: "Pessoa associada",
					name: "contactPerson",
					type: "reference",
					collections: ['pressContact'],
				},
				
			]
		},
```
Console return
```
{
  name: 'Luiza', 
  contactPerson: 'content/pressContact/luiza.md'
}

```

logan-anderson:
> and what it returns from the reference is the path of the content file. Perhaps there's something else missing?
> 
> ```
> {
> ```

@eduoliveira85  That is the expected behaviour. The reference field is string id (filename) that points to the file path of another collection item. When / If you query the data with the client or GQL API you can get the 'pressContact' data.

eduoliveira85:
Oh, I see. Makes sense. Thanks for clarifying that.

maxjvh:
so is this a wontfix or why was it closed exactly?

PR #2087, opened 2021-09-13
added new warning messages for list UI

logan-anderson:
Fixes https://github.com/tinacms/tinacms/issues/2056. It does not only fix it for the reference field but all fields that do not support list: true.

- [x] build time error message
- [x] Tina.io docs (https://github.com/tinacms/tinacms.org/pull/1038)

@jeffsee55 Should we provide a way to opt out as someone could provide their own UI and it would work as expected ?

EDIT: Maybe this change is too aggressive? We could also give a warning to the user that this UI is not currently supported and direct them to the docs for a work around (embed in object). 

EDIT again: I think I like the warning message better. I updated it to be a warning message now.

### DIFF
--- a/.changeset/real-pears-chew.md
+++ b/.changeset/real-pears-chew.md
@@ -1,0 +1,7 @@
+---
+'tina-cloud-starter': minor
+'@tinacms/cli': minor
+'@tinacms/graphql': minor
+---
+
+Added new error messages for list UI that we do not support by defualt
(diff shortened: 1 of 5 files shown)


jeffsee55 (approved):
This looks great, I think a warning is fine, it's something we plan to support as soon as possible. And the link to the docs is 🔥

PR #2091, opened 2021-09-14
add new branching endpoints for listing / creating

### LINKED ISSUE #2082: Add Github proxy endpoints for branch management
Provide endpoints for Github branching solution, export for use in Tina Cloud
- List branches
- Create branches

jbevis:
Closes #2082 

This PR will add basic endpoints for creating and listing branches for a given repo.

Still a WIP, but wanted to get it up for feedback.

Listing works, but need to figure out why create is firing as a GET call instead of a POST.

### DIFF
--- a/packages/@tinacms/cli/src/cmds/start-server/server.ts
+++ b/packages/@tinacms/cli/src/cmds/start-server/server.ts
@@ -19,6 +19,8 @@ import { altairExpress } from 'altair-express-middleware'
 // @ts-ignore
 import bodyParser from 'body-parser'
 
+const GITHUB_AUTH_TOKEN = process.env.GITHUB_ACCESS_TOKEN
+
 const gqlServer = async () => {
   // This is lazily required so we can update the module
   // without having to restart the server
@@ -66,11 +68,11 @@ const gqlServer = async () => {
       const { query } = req
       const { owner, repo } = query
       const result = await gqlPackage.listBranches({
-        auth: process.env.GITHUB_PERSONAL_ACCESS_TOKEN,
+        auth: GITHUB_AUTH_TOKEN,
         owner,
         repo
       })
-  
+
       return res.json(result.data)
     } catch(error) {
       console.error('There was a problem fetching the branches.', error)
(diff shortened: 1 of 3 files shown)


packages/@tinacms/cli/src/cmds/start-server/server.ts
```diff
@@ -18,6 +18,7 @@ import express from 'express'
 import { altairExpress } from 'altair-express-middleware'
 // @ts-ignore
 import bodyParser from 'body-parser'
+import { gql } from '@tinacms/graphql'
```
jeffsee55, line 21: This probably slipped in from an auto-import

jeffsee55 (approved):
Seems good to me, @dwalkr is this in line with what you're expecting?

Enigmatical (dismissed):
Looks good to me 👍 
At least, it is a good foundation for more GitHub endpoints in the future.

One improvement I might suggest (though not necessary right now) would be to break this code out into a separate file (sort of how `cache` and `primitives` are developed).

That keeps the `index` "clean" - only exporting what it imports and giving us a nice place for "all things GitHub".

Enigmatical:
Going to hold on merging this until @dwalkr has had a chance to weigh in.

logan-anderson:
> Going to hold on merging this until @dwalkr has had a chance to weigh in.

Looks good to me too! One thing that we _might_ consider changing is when creating a branch you mass in a query param called `branch`, would `baseBranch` make more sense?

packages/@tinacms/cli/src/cmds/start-server/server.ts
```diff
@@ -61,6 +61,39 @@ const gqlServer = async () => {
...
+      return res.json(result.data)
+    } catch(error) {
+      console.error('There was a problem fetching the branches.', error)
+    }
+  })
+
+  app.post('/create-branch', async (req, res) => {
+    try {
+      const { query } = req
+      const { owner, repo, name, branch } = query
```
dwalkr, line 83: agree that we should rename the `branch` parameter because it's a little ambiguous here. `baseBranch` sounds good to me.

packages/@tinacms/graphql/src/index.ts
```diff
@@ -40,6 +40,35 @@ import type {
...
+  const appOctoKit = new Octokit({ auth })
+  const branchList = await appOctoKit.repos.listBranches({
+    owner,
+    repo
+  })
+
+  return branchList
+}
+
+export const createBranch = async ({ auth, owner, repo, name, branch }) => {
```
dwalkr, line 55: I would suggest carrying over the `branch` -> `baseBranch` name change to this level as well.

jbevis: Makes sense to me, those changes just pushed up.

Issue #2083, opened 2021-09-09
Ensure demo-next works


jeffsee55:
Update `demo-next` to `legacy-demo-next` or something similar in the monorepo.

Inline editing and the Github open source APIs may not be working as expected for some, we need to revive the existing example app and ensure older APIs are working.

Let's timebox this at a few days, it's legacy APIs that we likely won't bring forward. We'll need to educate users inline APIs that they should pin their versions to TinaCMS `0.41.1`

PR #2093, opened 2021-09-15
Removes demo-next

Enigmatical:
Full credit to @logan-anderson who wrote the original PR to remove `demo-next` (#2079):

> The demo-next example was not a good example of the current "happy path". This could be very confusing to someone who is learning tina and looks in our example folder.

@jeffsee55 and I got on a call to try and salvage this demo mainly because we had some users encountering troubles with the old inline interfaces and were using `demo-next` as a starting point.  Unfortunately, due to recent changes in our build process, we'd have to do some significant work to get `demo-next` to run properly and even then it is not a direction we want to go.

Closes https://github.com/tinacms/tinacms/issues/2083


Enigmatical:
@jeffsee55 Just to confirm, I don't need to include a `changeset` when removing this folder, do I?

jeffsee55:
Correct, it's just a warning to remind you but doesn't make sense in this case.

dwalkr (approved):

Issue #2094, opened 2021-09-15
Improve error message for `staticRequest`


jeffsee55:
If `staticRequest` is used at runtime on a server, the localhost:4001/graphql endpoint wont be running. Right now the error message we see is:

```
ERROR    Unhandled Promise Rejection     {"errorType":"Runtime.UnhandledPromiseRejection","errorMessage":"FetchError: request to http://localhost:4001/graphql failed, reason: connect ECONNREFUSED 127.0.0.1:4001","reason":{"errorType":"FetchError","errorMessage":"request to http://localhos
```
We should catch that and show a more helpful message like:
```
The local server is not available outside of `getStaticProps` or `getStaticPaths` functions. 
```

PR #2096, opened 2021-09-16
added warning message when static request was being used outside a no…

logan-anderson:
Fixes https://github.com/tinacms/tinacms/issues/2094 but instead of an error message it adds a warning message. (You will still get the same error but will have this warning message directly above it). The warning message will always appear (local mode or prod).

How does this sound @perkinsjr and @jeffsee55  ?

### DIFF
--- a/.changeset/tall-cooks-battle.md
+++ b/.changeset/tall-cooks-battle.md
@@ -1,0 +1,5 @@
+---
+'tinacms': minor
+---
+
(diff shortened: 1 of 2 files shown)


perkinsjr (approved):

Enigmatical (approved):
Better than nothing.  We can always circle back if we feel this isn't sufficient enough.

Issue #2098, opened 2021-09-18
Can't resolve react-tinacms-strapi in versions after 0.50.5


FREEZX:
## Description

import 'react-tinacms-strapi' does not work with versions after react-tinacms-strapi 0.50.5. Upon inspecting the react-tinacms-strapi it seems like these releases don't contain a build directory and therefore the "main" reference cannot be resolved

### Steps to reproduce

Run `yarn add react-tinacms-strapi`

### Expected result

react-tinacms-strapi in node_modules should have a build directory.

### Actual result

react-tinacms-strapi in node_modules only has node_modules, CHANGELOG, README and package.json.

#### Tools

| Tool    | Version |
| ------- | ------- |
| browser |    edge     |
| os      |     Windows 10    |
| node    |    v14.17.6     |
| yarn  |     1.22.10    |

wiesys:
I had the same issue. I managed to fix it and just submitted a pull request #2109  to fix this issue. I think It was a bug introduced by the new build script.

jeffsee55:
@FREEZX marking this as closed, please reopen if the issue persists. Thanks @wiesys !

PR #2109, opened 2021-09-25
fix(react-tinacms-strapi): Main file in package.json

wiesys:
Fixes an issue #2098

Fixes a bug, that npm package is missing `dist` folder, because of the changed build path.


logan-anderson (approved):
Looks good to me 👍

PR #2114, opened 2021-09-27
Draft: Fix nested reference

jeffsee55:
@logan-anderson I think I've got it working, though it's very ugly. I added some notes but to be honest this is all so much more complex than it needs to be, we need to carve out some time to work on nested fields properly. Only marking this as WIP because the schema isn't something we probably want in our starter.

### DIFF
--- a/examples/tina-cloud-starter/.tina/schema.ts
+++ b/examples/tina-cloud-starter/.tina/schema.ts
@@ -389,6 +389,26 @@ export default defineSchema({
       path: "content/posts",
       fields: [
         {
+          type: "object",
+          name: "asdf",
+          label: "adsf",
+          list: true,
+          templates: [
+            {
+              name: "template1",
+              label: "template one",
+              fields: [
+                {
+                  name: "author",
+                  label: "Author",
+                  type: "reference",
+                  collections: ["authors"],
+                },
+              ],
+            },
+          ],
+        },
+        {
           type: "string",
           label: "Title",
           name: "title",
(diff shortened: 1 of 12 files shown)


logan-anderson:
@jeffsee55 Maybe I am missing something but it looks like the "avatar" property is not being resolved.

Screen Shot 2021-10-05 at 10 23 07 AM

logan-anderson:
> @jeffsee55 Maybe I am missing something but it looks like the "avatar" property is not being resolved.
> 
> Screen Shot 2021-10-05 at 10 23 07 AM

It was because the data was not in our graphql query. When I added it it worked as expected.

examples/tina-cloud-starter/.tina/schema.ts
```diff
@@ -388,6 +388,26 @@ export default defineSchema({
       name: "posts",
       path: "content/posts",
       fields: [
+        {
```
logan-anderson, line 391: We can probably remove this before this is merged.

examples/tina-cloud-starter/components/post.tsx
```diff
@@ -43,6 +43,7 @@ export const Post = ({ data }) => {
           </span>
         </h2>
 
+        <h3>{data.asdf && JSON.stringify(data.asdf, null, 2)}</h3>
```
logan-anderson, line 46: This too

examples/tina-cloud-starter/pages/posts/[filename].tsx
```diff
@@ -30,6 +30,18 @@ export const getStaticProps = async ({ params }) => {
                 }
               }
             }
+            asdf {
```
Comment (logan-anderson) [inline examples/tina-cloud-starter/pages/posts/[filename].tsx:33]: Probably can remove this as well

logan-anderson (approved):
Looks good to me! We might want to remove some of the stuff added for demo purposes but it's not a huge deal.

PR #2116, opened 2021-09-28
Branch switcher reboot

dwalkr:
This is a rewrite of the branch switcher work in two pieces:

1. a vendor-agnostic plugin in `@tinacms/toolkit`, and
2. an implementation of the plugin wrapped up in `TinaCloudProvider` (`tinacms` package)

My work thus far hasn't been tested and may not be working, so don't assume you've done something wrong if it doesn't.

The plugin is hidden behind a [feature flag](https://github.com/tinacms/tinacms/commit/9ce3419adcda1c4d7bf4c174032b83bb7787a1dc). To activate it, add this after creating the CMS object:

```
cms.flags.set('branch-switcher', true)
```

## Todo
- [ ] Connect to the cloud starter example to determine if screen plugin is working correctly
- [x] implement `createBranch` and `listBranches` in the plugin invocation, found in `TinaCloudProvider`
  - presumably we now want to add these endpoints to Tina Cloud, and can call out to them instead of implementing over the Github Bridge
- [ ] Re-query on branch change
- [ ] consider making the UI look slightly less prototype-y

### DIFF
--- a/packages/@tinacms/cli/src/cmds/start-server/server.ts
+++ b/packages/@tinacms/cli/src/cmds/start-server/server.ts
@@ -19,6 +19,7 @@ import { altairExpress } from 'altair-express-middleware'
 // @ts-ignore
 import bodyParser from 'body-parser'
 
+const GITHUB_ACCESS_TOKEN = process.env.GITHUB_PERSONAL_ACCESS_TOKEN
 const gqlServer = async () => {
   // This is lazily required so we can update the module
   // without having to restart the server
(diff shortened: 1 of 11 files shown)


jbevis:
@jeffsee55 

I've reopened this one and rebased latest changes from `main`. While all the check boxes haven't been deliberately addressed yet, the create and list functions have been added, and access to the branch switcher is hidden behind the feature flag work DJ implemented. 

This could in theory be merged in without it breaking any existing functionality, and then further refined once the tina cloud endpoints are in place.

jeffsee55 (approved):

PR #2120, opened 2021-10-03
Server no watch option

employee451:
In some environments (not just when `process.CI` is `true`), watching for changes to all files is not ideal/will be a problem.

This PR addresses the issue by adding a `--noWatch` option to the `server:start` command. When this option is used, file watching is disabled.


jeffsee55:
Thanks @employee451 , looks great!

jeffsee55 (approved):

PR #2121, opened 2021-10-04
Fixed issue where heading button would not work in the WYSIWYG editor

logan-anderson:
This PR does 2 things
1.
Fixes an issue where the heading button would not work in the WYSIWYG editor (using react 17)
2.
Removes the need for `react-dismissible` package


Enigmatical (approved):
Wrapping the WYSIWYG editor with the Dismissable Context (rather than the other way around) makes sense.  I feel like I encountered similar things in the Form work I did.

Also appreciate removing the package as a dependency 👍

PR #2127, opened 2021-10-06
generated client

logan-anderson:
This is only here if anyone is interested to see the graphql code generation stuff I have been working on.

Currently I have this working.

The user creates a folder called `.tina/queries/`
In that folder they can add graphql queries, fragments, ect.
Next, the CLI will generate a typed client based on those queries. 

```ts
import { getSdk } from "../.tina/__generated__/types";
const client = getSdk(requester);
const data = await client.GetSomething();
```

 
👀  Demo: https://www.loom.com/share/d67d9f86889f47af9c19ff28d8a68c46


jeffsee55:
This feels like the right direction to nail down what we want the SDK to look like.

What I think we'll run into when we want this to be prebuilt is that even if we could autogenerate some queries and then build the SDK based on them, we wouldn't be able to represent all possible requests due to the recursive nature of GraphQL.

Because of that, ultimately I think we'll need to build our own solution that probably shares a lot of logic with the plugins you're using here, so this seems like a great start.

Also, I'm curious on your thoughts for how this API might look for resolving references. Last time we talked about it we were thinking maybe something like this:
```js
client.getPostDocument({relativePath: "helloWorld.md"}, {include: {author: true}})
```
Any other thoughts around that?

logan-anderson:
@jeffsee55 

I have been thinking about this and it seems like a good API but I think that typing it could be complicated as we will be building graphql queries on the fly (client-side runtime). 
```ts
client.getPostDocument({relativePath: "helloWorld.md"}, {include: {author: true}})
```

What I was thinking is we could have a config in the `schema.ts` that allows the user to control how the queries are generated. 

Maybe something like this.

```ts
export default defineSchema({
    generator: {
        depth: 4,
         include: {
            author: true,
            posts: true,
        },
    }
})
```

Then we could generated the queries statically when the schema is built instead of doing this in runtime.

Thoughts?

packages/@tinacms/graphql/package.json
```diff
@@ -68,6 +68,7 @@
     "gray-matter": "^4.0.2",
     "js-yaml": "^3.14.0",
     "lodash": "^4.17.20",
+    "lodash.set": "^4.3.2",
```
jeffsee55, line 71: Is this needed somewhere?

packages/tinacms/src/utils/index.ts
```diff
@@ -15,6 +15,7 @@ import { Client, LocalClient } from '../client'
 import type { TinaIOConfig } from '../client'
 import * as yup from 'yup'
 
+export * from './client'
```
jeffsee55, line 18: We're not actually using any of the stuff you've added in the `util` folder are we?

packages/tinacms/src/index.ts
```diff
@@ -21,5 +21,7 @@ export * from '@tinacms/toolkit'
 export { TinaAdmin } from './admin'
 
 import { TinaCMSProvider2 } from './tina-cms'
-
+import { DocumentNode } from 'graphql'
+export type { DocumentNode }
+import { type } from 'os'
```
jeffsee55, line 26: I'm guessing we don't need this?

examples/tina-cloud-starter/.tina/__generated__/frags.json
```diff
@@ -0,0 +1,608 @@
+{
```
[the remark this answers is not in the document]
logan-anderson: It might be useful but we don't need them. I'll delete them for now.

packages/tinacms/src/utils/index.ts
logan-anderson: I will delete this import

packages/tinacms/src/index.ts
logan-anderson: No we do not

jeffsee55 (dismissed):

logan-anderson:
@jeffsee55 I updated the code to now use the generated client so an example can be seen.

logan-anderson:
Current docs are here: https://github.com/tinacms/tinacms/blob/sandbox/generated-client/packages/%40tinacms/cli/GeneratedClientDocs.md

packages/@tinacms/cli/GeneratedClientDocs.md
```diff
@@ -0,0 +1,79 @@
+# Experimental Tina Client
+
+> Note: This usage is experiment and although it is likely to maintained / supported, the API may change over time.
```
jeffsee55, line 3:
```suggestion
> Note: This usage is experimental and although it is likely to maintained / supported, the API may change over time.
```

packages/@tinacms/cli/GeneratedClientDocs.md
```diff
@@ -0,0 +1,79 @@
...
+
+## What is the generated client.
+
+This piece of work has two main functionalities,
+
+1. Generates Graphql fragments and queries that are written for you.
+2. Generates a graphql client that can be used server-side (in getStaticProps)
```
jeffsee55, line 10:
```suggestion
2. Generates a Graphql client that can be used server-side (in getStaticProps)
```

packages/@tinacms/cli/GeneratedClientDocs.md
```diff
@@ -0,0 +1,79 @@
...
+
+## Using build in Graphql Queries
```
jeffsee55, line 12:
```suggestion
## Using the built-in Graphql Queries
```

packages/@tinacms/cli/GeneratedClientDocs.md
```diff
@@ -0,0 +1,79 @@
...
+
+The client comes with several build in graphql queries to allow you to get up and running quickly. There is one caveat it **does not resolve references**. Although we plan to support an API do to this in the future, the build in generated client functions do not resolve references. To resolve reference you will have to [write your own queries](#writing-your-own-gql-queries).
```
jeffsee55, line 14:
```suggestion
The client comes with several built-in GraphQL queries to allow you to get up and running quickly. There is one caveat it **does not resolve references**. Although we plan to support an API do to this in the future, the built-in generated client functions do not resolve references. To resolve reference you will have to [write your own queries](#writing-your-own-gql-queries).
```

packages/@tinacms/cli/GeneratedClientDocs.md
```diff
@@ -0,0 +1,79 @@
...
+
+Firstly, you can import a function to get the client client from generated folder located in `.tina/__generated/types`.
```
jeffsee55, line 16:
```suggestion
Firstly, you can import a function to get the client from generated folder located in `.tina/__generated/types`.
```

packages/@tinacms/cli/GeneratedClientDocs.md
```diff
@@ -0,0 +1,79 @@
...
+
+## Writing your own GQL Queries
+
+First of all you can import the generated client
+
+```ts
+import { getTinaClient } from "../path/to/tinaFolder/.tina/__generated__/types";
+```
+
+Then you can use the typed client. The types are based on your graphql queries that you have written. The graphql Queries are written in the `.tina/queries/` folder and these will be used to generate the client.
```
jeffsee55, line 46:
```suggestion
Then you can use the typed client. The types are based on your Graphql queries that you have written. The Graphql queries are written in the `.tina/queries/` folder and these will be used to generate the client.
```

packages/@tinacms/cli/GeneratedClientDocs.md
```diff
@@ -0,0 +1,79 @@
...
+query GetAuthorDocument($path: String!) {
+  getAuthorsDocument(relativePath: $path) {
+    data {
+      ...AuthorsParts
+    }
+  }
+}
+```
+
+We are using an `AuthorsParts` Fragment. For each collection this fragment is generated and updated when the schema is updated. This means you no longer have to update your graphql query when you update your schema.ts (assuming you want to query for all the fields)
```
jeffsee55, line 75:
```suggestion
We are using an `AuthorsParts` Fragment. For each collection this fragment is generated and updated when the schema is updated. This means you no longer have to update your Graphql query when you update your schema.ts (assuming you want to query for all the fields)
```

jeffsee55 (commented):
Just some small typo fixes. We aren't super consistent with how we write "GraphQL", but I've updated your usage to at least be "Graphql", but might want to actually write it "GraphQL", which I think the more correct way.

jeffsee55 (approved):

PR #2191, opened 2021-11-02
Use graphql forms refactor

jamespohalloran:
Not sure if I'll end up undrafting this PR or not, but doing some refactoring to get a better feel on every useGraphqlForms is doing.

### DIFF
--- a/packages/tinacms/src/auth/TinaCloudProvider.tsx
+++ b/packages/tinacms/src/auth/TinaCloudProvider.tsx
@@ -19,7 +19,7 @@ import {
   MediaStore,
   //@ts-ignore why can't it find you
   BranchSwitcherPlugin,
-  Branch
+  Branch,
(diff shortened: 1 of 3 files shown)


jeffsee55:
@jamespohalloran I know you were doing this as an exercise to get familiar with things but were you thinking we should actually merge this? 

As I mentioned we've been eyeing a big refactor of this for a while and I'm worried this might make it a little harder for us, considering I'd like to use most of the work that's already done [here](https://github.com/tinacms/tinacms/pull/2069/files#diff-580eb812378c74b17309abdda8f39a9b133e433cc74d5289a80330335eb0454a)

jamespohalloran:
@jeffsee55 I'm okay closing this. I'm happy with where it landed, but it's probably too big of a step for one PR, so we can pull from it for reference if needed.

PR #2194, opened 2021-11-03
2181 - Introduces RouteMapperPlugin and FormMetaPlugin

### LINKED ISSUE #2181: [TinaAdmin] Route Mapping POC
Utilize a CMS Plugin to provide `TinaAdmin` a way to route to content on a user's site.

Enigmatical:
For a while now, `TinaAdmin` has mostly been a standalone wildcard route and has little interaction with TinaCMS inside the site.

This PR aims to provide ways for the two to interact - if only slightly.

On the `TinaAdmin` side of things, I've introduced a `RouteMapperPlugin` type that is set up inside the `cmsCallback`.  This provides the end-user with a `collection` and `document` and expects them to return the route to that document in the live site (or undefined, if there isn't one).

On the `TinaCMS` side of things, I've introduced the `FormMetaPlugin` which is an attempt to empower the user to add `meta` information to the `Form` instance.  In this case, `TinaAdmin` will place a link in the header (or where ever Scott B designs) that allows for linking back to that document in `TinaAdmin`.

This is a really barebones implementation that I expect us to iterate on.

https://www.loom.com/share/60af423a0af14a438ab84f8082aff087

Closes #2181

### DIFF
--- a/packages/@tinacms/toolkit/src/packages/react-sidebar/components/Sidebar.tsx
+++ b/packages/@tinacms/toolkit/src/packages/react-sidebar/components/Sidebar.tsx
@@ -91,6 +91,17 @@ const Sidebar = ({ sidebar }: SidebarProps) => {
             <MenuPanel visible={menuIsOpen}>
               <MenuWrapper>
                 <MenuList>
+                  {cms.flags.get('tina-admin') && (
+                    <MenuLink
+                      key="admin"
+                      value="admin"
+                      onClick={() => {
+                        window.location.href = window.location.origin + '/admin'
+                      }}
+                    >
+                      <TinaIcon /> Tina Admin
+                    </MenuLink>
+                  )}
                   {allScreens.map((view) => {
                     const Icon = view.Icon
                     return (
--- a/packages/@tinacms/toolkit/src/packages/react-sidebar/components/SidebarBody.tsx
+++ b/packages/@tinacms/toolkit/src/packages/react-sidebar/components/SidebarBody.tsx
@@ -26,6 +26,7 @@ import { FormList } from './FormList'
 import { useCMS, useSubscribable } from '../../react-core'
 import { LeftArrowIcon } from '../../icons'
 import { FormBuilder } from '../../form-builder'
+import { FormMetaPlugin } from '../../../plugins/form-meta'
 
 export const FormsView = ({
   children,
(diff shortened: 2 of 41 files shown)


packages/@tinacms/toolkit/src/packages/react-sidebar/components/SidebarBody.tsx
```diff
@@ -100,6 +103,12 @@ export const FormsView = ({
           {!isMultiform && activeForm.label && (
             <FormHeader activeForm={activeForm} />
           )}
+          {formMetas &&
+            formMetas.map((meta) => (
+              <React.Fragment key={meta.name}>
+                <meta.Component />
+              </React.Fragment>
+            ))}
```
Enigmatical, line 111: Here's where the `formMetas` are actually being inserted into the `SidebarBody`.  I originally had them placed in the `FormBuilder`, but they were appearing in _all_ forms (Global and Overlay).

packages/tinacms/src/admin/pages/CollectionListPage.tsx
```diff
@@ -95,85 +96,117 @@ const CollectionListPage = () => {
...
+                                    </span>
+                                  </td>
+                                  {livesiteRoute && (
+                                    <td className="px-6 py-4 whitespace-nowrap">
+                                      <span className="block text-xs text-gray-400 mb-1 uppercase">
+                                        Site Link
+                                      </span>
+                                      <a href={livesiteRoute}>Preview</a>
+                                    </td>
+                                  )}
```
Enigmatical, line 195: Here's where the `Preview` link is being displayed, assuming the `routeMapping` returns a `string`.

jeffsee55 (approved):

packages/@tinacms/toolkit/src/packages/react-sidebar/components/Sidebar.tsx
```diff
@@ -91,6 +91,17 @@ const Sidebar = ({ sidebar }: SidebarProps) => {
             <MenuPanel visible={menuIsOpen}>
               <MenuWrapper>
                 <MenuList>
+                  {cms.flags.get('tina-admin') && (
+                    <MenuLink
+                      key="admin"
+                      value="admin"
+                      onClick={() => {
+                        window.location.href = window.location.origin + '/admin'
```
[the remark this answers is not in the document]
spbyrne: I think the MenuLink component was generally designed to trigger forms & modals, so I guess a button made sense. I think maybe it's possible to do something like `as='a'` and use an href as normal?

PR #2201, opened 2021-11-03
Ensure the full image resolution URL it persisted

jeffsee55:
We had a great contribution from @employee451 but since we actually persist the `previewSrc` value it would make everyone's images too small. This PR makes the distinction between `previewSrc` and `src`


Enigmatical (approved):
Looks good 👍 
Got a conflict that needs to be resolved, otherwise this is fine by me.

PR #2203, opened 2021-11-04
TinaCloud by composition POC

jamespohalloran:
WIP, but wanted to get some opinions on this.
IMO, we're wrapping up a bit too much within the new `<TinaCMS />` container.

There's still some cleanup within the TinaCMS package, but the main bones of the changes are seen in the tina-cloud-starter _app.tsx. 

See: https://github.com/tinacms/tinacms/blob/7a6eac8060ed6bbf07b80d8b32a9db08aa2eda31/examples/tina-cloud-starter/pages/_app.tsx

There's more for the user to set up, but I think this approach would be easier to extend, and is closer to our SDK API, so docs would be able to be shared between TinaCMS w/ TinaCloud, & without.

A few details to note:

- `<TinaProvider />`, from the `@tinacms/toolkit` package, is now used. This attaches the cms object to context, so `cms` no longer needs to be controlled through callbacks. 
- `<TinaCloudAuthProvider>` is the main Tina Cloud wrapper. It makes sure the user is logged in.
- ` <TinaCloudQuery />` takes in a TinaCloud query & variables, and returns the data & forms as a HOC. It is then up to the user to register the returned forms. Now the user doesn't need to use formify, the can tweak/omit any forms that they want before registering them. 
- My `<TinaForms />` component is just an example of registering the forms from `TinaCloudQuery`, and also a custom dummy form.

This IMO exposes control of the underlying toolkit library, so we no longer would need two separate sets of docs. The only things specific to Tina Cloud are: `TinaCloudAuthProvider` & the `TinaCloudQuery` component.

### DIFF
--- a/examples/tina-cloud-starter/pages/_app.tsx
+++ b/examples/tina-cloud-starter/pages/_app.tsx
@@ -2,7 +2,7 @@ import "../styles.css";
 import dynamic from "next/dynamic";
 import { TinaEditProvider } from "tinacms/dist/edit-state";
 import { Layout } from "../components/layout";
-import { TinaCMS, TinaProvider } from "tinacms";
+import { TinaCMS, TinaProvider, Client, LocalClient } from "tinacms";
 import { useMemo, useEffect } from "react";
 // @ts-ignore FIXME: default export needs to be 'ComponentType<{}>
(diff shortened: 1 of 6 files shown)


jeffsee55:
Closing for now, James and I have discussed some patterns out of band and we'll likely discuss some of them as a group soon

PR #2204, opened 2021-11-04
2196 - Adds `reference` field

### LINKED ISSUE #2196: [TinaAdmin] Add `reference` Field
This seeks to add a dedicated `reference` field to our core list of Field Plugins instead of continuing to use the `select` field.

With a dedicated field, we can add additional functional to the `reference` field without overloading (or breaking) the `select` field.  Some of these new features are part of `TinaAdmin`.

Enigmatical:
Right now, `reference` field is just a copy of `select` with additional functionality if TinaAdmin is enabled.

It does an async `query { node }` to create a link to TinaAdmin under the field, updating when the value changes.

Closes #2196


packages/@tinacms/toolkit/src/packages/fields/components/Reference.tsx
```diff
@@ -0,0 +1,185 @@
...
+          {(document) => (
+            <a
+              href={`/admin/collections/${
+                document.sys.collection.name
+              }/${document.sys.breadcrumbs.join('/')}`}
+            >
+              &raquo; Edit in TinaAdmin
+            </a>
+          )}
+        </GetReference>
```
Enigmatical, line 120: Used a render-prop pattern similar to what `TinaAdmin` uses.  `document` is the response from a `query { node }` request, so we have access to all of the `sys` information for the Document.

When the `tina-admin` flag is disabled, this field behaves just like the `select` one.

spbyrne:
Screen Shot 2021-11-09 at 4 58 00 PM

Styled with inline styles to avoid styled components and tailwind (given it's not available in the sidebar yet) so there's no hover/focus state. This will be improved later with tailwind.

examples/tina-cloud-starter/pages/_app.tsx
```diff
@@ -28,6 +28,8 @@ const App = ({ Component, pageProps }) => {
               import("react-tinacms-editor").then(({ MarkdownFieldPlugin }) => {
                 cms.plugins.add(MarkdownFieldPlugin);
               });
+
+              cms.flags.set("tina-admin", true);
```
logan-anderson, line 32: We should probably put this in our Docs somewhere.  (Or in the docs you make for the admin)

logan-anderson (approved):
Looks good to me!

(side note, I ran a `yarn` from the root and it updated the pnp.js so you may want to do that and push up those changes before merging)

examples/tina-cloud-starter/pages/_app.tsx
Enigmatical: Yeah, I plan to make it a special note for the `tinacms/admin` README.md. 👍

logan-anderson (approved):

jeffsee55 (approved):

PR #2207, opened 2021-11-06
Fix missing Typescript typings for react-tinacms-inline package

wiesys:
Fixed path to Typescript typings, so that they are included in generated NPM package.


jeffsee55 (approved):
Thanks @wiesys !

PR #2210, opened 2021-11-08
Feat/basic starter

logan-anderson:
Adds a basic tinacms starter for folks to take a look at and play around with.


examples/basic/pages/posts/index.js
```diff
@@ -0,0 +1,63 @@
...
+  })
+
+  return {
+    props: {
+      ...tinaProps,
+    },
+  }
+}
+
+// export const getStaticProps = async () => {
```
logan-anderson, line 46: This should be deleted

examples/basic/pages/admin/[[...slug]].js
```diff
@@ -0,0 +1,3 @@
+import { TinaAdmin } from 'tinacms'
```
Comment (logan-anderson) [inline examples/basic/pages/admin/[[...slug]].js:1]: Is it OK to use the Admin right now?

Comment (Enigmatical) [reply examples/basic/pages/admin/[[...slug]].js:1]: I don't see why not; I assume it is working and has the right version of TinaCMS.

This is more a question for @jeffsee55 and the product team on when they want `TinaAdmin` to start being bundled with our starters.

Comment (logan-anderson) [reply examples/basic/pages/admin/[[...slug]].js:1]: Talked about it, We are keeping admin in.

jeffsee55 (approved):

PR #2221, opened 2021-11-10
Use Yarn 3

jeffsee55:
Seems to be working fine for me so far. There are 2 things I want to highlight though:

1. I had to remove this line from our bin files:
```
const require = require("esm")
```
I honestly don't know why I needed that in the first place but with it Yarn can no longer find it's dependencies

2. I had to remove the `.next` cache in the example projects. These were gitignored but due to the changes in the way zipped dependencies are now handled Next was blowing up for a missing babel/runtime dep. That'll be something each person has to do on their own when they pull down the new changes (since it's gitignored)

Full changeset https://github.com/yarnpkg/berry/blob/01586a88806a2bebd7edb28d1bee3581b1fd3762/CHANGELOG.md#300


Enigmatical (approved):
Pulled it down, deleted any `.next` folders, and ran `yarn install`.

Everything worked as expected; I mainly tested the `TinaAdmin` and was able to create and edit through both the sidebar and admin with no problems. 👍

jeffsee55:
Closing for now, will revitalize when we have time

PR #2227, opened 2021-11-11
tina cloud branch switcher

dwalkr:
This version of the branch switcher displays a dropdown list of all of the content repo's branches and allows the user to select which branch they want to send changes to.

---

To use with Tina Cloud, all you need to do is set the `branch-switcher` feature flag in your `cmsCallback`:

```tsx
<TinaCMSProvider cmsCallback={(cms) => {
    cms.flags.set('branch-switcher', true)
}}>
```

Once this flag is set, the branch switcher should automatically appear in the Global Menu.


jeffsee55 (approved):
🎉

PR #2240, opened 2021-11-16
Fixup script paths

jeffsee55:
If your local repo wasn't called `tinacms` the build/watch commands wouldn't work. This just makes the check a little less strict.

Builds on top of https://github.com/tinacms/tinacms/pull/2239


dwalkr (approved):
noticed this also has the changes from #2239, but that should be fine

PR #2241, opened 2021-11-16
Data Store 

jeffsee55:
- 9b6e50c These files were just annoyingly nested for no real reason after the move to primitive types was complete
- a8d59f6 The diffs for these tests was really hard to grok, deleting these here so the new ones don't show up as diffs
- a1458a3 and after are the more critical changes


packages/@tinacms/graphql/src/builder/index.ts
```diff
@@ -824,9 +868,140 @@ export class Builder {
       namespace: connectionNamespace,
       connectionName,
       nodeType: nodeType,
+      collections,
     })
   }
 
+  private _buildFieldFilter = async (field: TinaFieldEnriched) => {
```
jeffsee55, line 875: This is nearly identical to what we're already doing with the "mutation" params. The key difference here is that the inputs are used for building the query request, and there is an additional layer of indirection so that we can be more granular with the _type_ of request:
```graphql
query {
  getMovieList(filter: { title: { startsWith: "Star" }, archived: { eq: true } }) {
    edges {
      node {
        id
      }
    }
  }
}
```

packages/@tinacms/graphql/src/builder/index.ts
```diff
@@ -920,21 +1095,74 @@ export class Builder {
...
+    collections,
   }: {
     fieldName: string
     namespace: string[]
     connectionName: string
     nodeType: string | TypeDefinitionNode
+    collection?: Collectable
+    collections?: Collectable[]
   }) => {
+    const extra = this.experimentalData
```
jeffsee55, line 1152: The `filter` property will only appear when the `experimentalData` flag is enabled via the CLI.

packages/@tinacms/graphql/src/database/bridge/filesystem.ts
```diff
@@ -39,16 +40,16 @@ export class FilesystemBridge implements Bridge {
       return item.replace(posixRootPath, '').replace(/^\/|\/$/g, '')
     })
   }
+  public supportsBuilding() {
```
jeffsee55, line 43: Differentiates the FilesystemBridge, which _can_ build schemas from the GithubBridge, which doesn't need to because those files should be checked into source control (after being built via CLI)

packages/@tinacms/graphql/src/database/index.ts
```diff
@@ -11,65 +11,58 @@ See the License for the specific language governing permissions and
...
-        if (err instanceof GraphQLError && err.extensions?.status === 404) {
-          throw new GraphQLError(
-            `${err.toString()}.  Please confirm this location is correct and contains a '.tina' folder and a valid '.tina/schema.ts'.`
-          )
-          /** Other Errors */
-        } else {
-          throw err
-        }
-      }
+      throw new Error(`Unexpected get for config file ${filepath}`)
```
jeffsee55, line 58: `get` should only be used for _content_ data, we have system-specific functions like `getTinaSchema` instead.

packages/@tinacms/graphql/src/database/index.ts
```diff
@@ -100,6 +93,7 @@ export class Database {
       return {
         ...data,
         _collection: collection.name,
+        _keepTemplateKey: !!collection.templates,
```
jeffsee55, line 96: Indicates whether the `bridge` should keep the `_template` property when persisting. This was previously handled elsewhere but is now a higher-level concern

packages/@tinacms/graphql/src/database/index.ts
```diff
@@ -109,12 +103,71 @@ export class Database {
...
+          attribute,
+          records.filter((item) => item !== filepath)
+        )
+        return true
+      })
+      await sequential(atts2, async (attribute) => {
+        const records = (await this.store.get<string[]>(attribute)) || []
+        await this.store.put(attribute, [...records, filepath])
+        return true
+      })
```
jeffsee55, line 159: This goes through and removes the 'metadata' attributes, and re-adds the new ones.

packages/@tinacms/graphql/src/database/index.ts
```diff
@@ -160,41 +213,47 @@ export class Database {
...
         extension,
         templateInfo.type === 'union'
       )
-      await this.bridge.put(filepath, stringData)
+      return { stringifiedFile, payload }
     }
-    return true
+  }
+
+  public flush = async (filepath: string) => {
```
jeffsee55, line 225: Currently this is only used for testing purposes, it returns what we _would_ write the `bridge`, but doesn't actually write it.

packages/@tinacms/graphql/src/database/index.ts
```diff
@@ -351,3 +384,151 @@ type UnionDataLookup = {
   resolveType: 'unionData'
   typeMap: { [templateName: string]: string }
 }
+
+const _indexContent = async (tinaSchema: TinaSchema, database: Database) => {
```
jeffsee55, line 388: The indexing process just steps through each record and records a "namespaced" version of what the given record's value is, and points to an array of all the records which have that value.

For example, if I have 2 movies which both debuted in 1990, I'd the process would generate a single record that stores those references, effectively acting as a "lookup" record:
```json
"__attribute__movie#movie#released#1990": [
  "content/movies/top-gun.md",
  "content/movies/lethal-weapon.md"
]
```

packages/@tinacms/graphql/src/database/store/filesystem.ts
```diff
@@ -0,0 +1,76 @@
...
+      await fs.readFileSync(path.join(this.rootPath, filepath)).toString(),
+      path.extname(filepath),
+      (yup) => yup.object()
+    )
+  }
+
+  public supportsSeeding() {
+    return false
+  }
+  public supportsIndexing() {
```
jeffsee55, line 47: The `FilesystemStore` and `GithubStore` aren't really stores, but they're basically taking up the same role as their "bridge" counterparts, it's just easier to inject a store no matter what, and then allow it to define its capabilities. `supportsIndexing` is our way of saying that this store is sort of a fake. These will go away when we're actually ready to ship all of this.

packages/@tinacms/graphql/src/database/index.ts
```diff
@@ -0,0 +1,537 @@
...
+          attribute,
+          records.filter((item) => item !== filepath)
+        )
+        return true
+      })
+      await sequential(attributesToAdd, async (attribute) => {
+        const records = (await this.store.get<string[]>(attribute)) || []
+        await this.store.put(attribute, [...records, filepath])
+        return true
+      })
```
jeffsee55, line 168: This goes through and removes the 'metadata' attributes, and re-adds the new ones.

jeffsee55:
This readme might help navigating the tests https://github.com/tinacms/tinacms/pull/2241/commits/ab5390c904c4b868ed271f9b576b40a1152579da

packages/@tinacms/graphql/src/resolver/index.ts
```diff
@@ -511,23 +510,42 @@ export class Resolver {
...
-    )
+    let documents
+    if (args.filter) {
+      const queries = []
+      const flattenedArgs = flat(args.filter, { delimiter: '#' })
+      Object.entries(flattenedArgs).map(([key, value]) => {
+        const keys = key.split('#')
+        const realKey = keys.slice(0, keys.length - 1).join('#')
+        queries.push(
+          `__attribute__${lookup.collection}#${lookup.collection}#${realKey}#${value}`
```
Enigmatical, line 533: @jeffsee55 From our chat, this should be `${lookup.collection}#${lookup.template}` - not double `collection`.

packages/@tinacms/cli/src/cmds/compile/index.ts
```diff
@@ -61,7 +61,8 @@ export const compile = async (_ctx, _next) => {
   }
 
   // Remove old js files
-  await resetGeneratedFolder
+  await fs.remove(tinaTempPath)
+  await fs.remove(tinaConfigPath)
```
Enigmatical, line 65: Take a look at this and make sure it is calling the most recent `removeGeneratedFolder` that you made in another PR: 
https://github.com/tinacms/tinacms/pull/2239

packages/@tinacms/cli/src/cmds/start-server/index.ts
```diff
@@ -13,17 +13,26 @@ limitations under the License.
...
+  MemoryStore,
+  FilesystemStore,
+  GithubStore,
+  GithubBridge,
+  FilesystemBridge,
+  LevelStore,
+} from '@tinacms/graphql'
 import { genTypes } from '../query-gen'
-import { compile } from '../compile'
+import { compile, resetGeneratedFolder } from '../compile'
```
Enigmatical, line 27: This might also be related to the `resetGeneratedFolder` thing.

Enigmatical (changes requested):
Got a couple of bugs we found during our pairing session, but otherwise, this looks really good 👍

jeffsee55:
cc @logan-anderson I noticed a bug in the fragment generation. It only manifests with the more comprehensive tests I've written here so I wanted to just include the fix with this. Can you take a look at https://github.com/tinacms/tinacms/pull/2241/commits/e1243c5993a05cb6335503ecaff0fc7d59f9b191 and make sure it makes sense to you? Without it, collections with `templates` were generating fragments without the `fragment` keyword:
```graphql
Post {
  ... on PostPost {
    title
    body
    author {
      ... on Document {
        id
      }
    }
  }
}
```
And now:
```graphql
fragment PostParts on Post {
  ... on PostPost {
    title
    body
    author {
      ... on Document {
        id
      }
    }
  }
}
```

jeffsee55:
I noticed a few new test cases and a bug that actually would have broken for non-experimental sites.

logan-anderson:
@jeffsee55 Makes sense to me.

Enigmatical (approved):

PR #2243, opened 2021-11-16
update script for transforming deps

logan-anderson:
The script that I was using previously had some errors so it was updated

### DIFF
--- a/scripts/transformDeps.sh
+++ b/scripts/transformDeps.sh
@@ -1,4 +1,9 @@
-git checkout -b examples
+# this is ment to be run localy 
+git fetch
+git pull origin main
+git pull origin examples
+git merge origin main
+git checkout examples --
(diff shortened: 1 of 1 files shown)


scripts/transformDeps.sh
```diff
@@ -1,4 +1,9 @@
-git checkout -b examples
+# this is ment to be run localy 
+git fetch
+git pull origin main
+git pull origin examples
```
jeffsee55, line 4: Wouldn't we get a lot of conflicts here since examples will have a lot of diffs in the `package.json` files?

logan-anderson: I have not gotten any so far. But there could be

jeffsee55:
Closing as this workflow is no longer needed thanks to @logan-anderson's improvements to create-tina-app

PR #2246, opened 2021-11-17
Add ability to control plugin layout for global plugins from formify

### LINKED ISSUE #2152: `createGlobalForm` doesn't support all GlobalForm options
The `formifyCallback` supports `createGlobalForm`, but it's not possible to make it a "fullscreen" plugin:

```ts
  const createGlobalForm = (formConfig) => {
    const form = new Form(formConfig)
    cms.plugins.add(new GlobalFormPlugin(form)) // We need to expose extra options for this class
    return form
  }
  ```

jeffsee55:
Closes #2152

### DIFF
--- a/.changeset/shiny-teachers-attack.md
+++ b/.changeset/shiny-teachers-attack.md
@@ -1,0 +1,5 @@
+---
+"tinacms": patch
+---
+
+Add ability to control plugin layout for global plugins from formify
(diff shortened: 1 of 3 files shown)


Enigmatical (approved):
Well, we did agree that it was probably an easy change 👍

Issue #2245, opened 2021-11-17
[TinaAdmin] Using `/` at the end of `path` causes GitHub to 500


Enigmatical:
*Credit to @mitchmac for discovering this interesting bug.*

When using a `schema.ts` where the `collection`'s `path` has a `/` at the end:
```
export default defineSchema({
  collections: [
    {
      label: "Pages",
      name: "page",
      path: "content/pages/", <--
      fields: [
        {
          type: "string",
          label: "Title",
          name: "title",
        },
      ],
    },
  ],
});
```

causes GitHub to respond with a 500 Server Error when querying for `getCollection{}` or `getCollections{}` via `tina-cloud`.

This _does_ work locally, though.  @jeffsee55 thinks that the local filesystem is handling the extra `/` well, but GitHub isn't.

We have a few avenues for solving this:  
* We can "clean" the `path` keys when transpiling the `schema.ts`, removing any extra `/`'s (or throw a warning).  
* We can turn `collection.path` into a function that does that cleaning for us at runtime.  
* We can look at fixing the issue on the GitHub side of things.

PR #2256, opened 2021-11-19
Trim path property on collections during compilation

jeffsee55:
If you specified `content/posts/` for your `path` argument on a collection this would break when using the Github API, this trims leading and trailing slashes and validate the schema like we were doing before primitive types. 

Closes https://github.com/tinacms/tinacms/issues/2245


Enigmatical (approved):
👍 Modified a few `path` values in `tina-cloud-starter` and confirmed that the generated `_schema.json` corrected the values.

PR #2251, opened 2021-11-17
Feat/create tina app

logan-anderson:
Basic POC of a `create-tina-app` command.

This currently
1. downloads the given example
2. installs dependencies 

Most of the logic is coming from create-next-app

Source: https://github.com/vercel/next.js/tree/canary/packages/create-next-app

PR #2271, opened 2021-11-23
update danger file to reject any file with example in the path

logan-anderson:
This updates `dangerfile.ts` so that it does not check the examples sub folder added in [this PR](https://github.com/tinacms/tinacms/pull/2251).

Can test it out here: https://regex101.com/r/c3gS2L/1/


spbyrne (approved):
Nice, good idea.

dwalkr (approved):

jeffsee55 (approved):

PR #2278, opened 2021-11-24
Ensure "src" is used instead of "previewSrc" for images in MDX

jeffsee55:
We recently optimized the image src for preview thumbnails in Cloudinary. This value was also being used effectively as the src, also for MDX. This PR should update things so the correct (full) url is provided instead.

Thanks @jhuggett


Enigmatical (approved):
👍

PR #2282, opened 2021-11-26
removed unused deps from tinacms

logan-anderson:
Just some code cleanup

cc @jeffsee55


Enigmatical (approved):
👍 Related question:  What did you use to discover that these packages were not used?  I'm a little surprised by a couple of them - `final-form-arrays` and `xstate`, to name them - but if they're not used, we definitely shouldn't keep them around.

Issue #2287, opened 2021-11-29
MDX components with `list: true` fields behave unexectedly


jeffsee55:
Given this schema:
```
        {
          type: 'rich-text',
          label: 'Blog Post Body',
          name: 'body',
          isBody: true,
          templates: [
            {
              name: 'Gallery',
              label: 'Gallery',
              fields: [
                {
                  label: 'Images',
                  name: 'images',
                  type: 'object',
                  list: true,
                  fields: [
                    {
                      type: 'image',
                      name: 'src',
                      label: 'Source',
                    },
                    {
                      type: 'string',
                      name: 'width',
                      label: 'Width',
                    },
                    {
                      type: 'string',
                      name: 'height',
                      label: 'Height',
                    },
                  ],
                },
                {
                  type: 'string',
                  name: 'alignment',
                  label: 'Alignment',
                  options: ['left', 'center', 'right'],
                },
                {
                  type: 'string',
                  name: 'gap',
                  label: 'Gap',
                },
              ],
            },
          ],
        },
```
Using the gallery results in some strange behavior:
```
This is a test

How are you doing?

<Gallery images={[{}, {}, {}]}/>
```

See this branch for reproducing https://github.com/tinacms/tinacms/tree/strange-mdx-behavior

cc @jhuggett thanks for pointing this out 👍

jhuggett:
Some more notes about the different behaviours:

- saving an empty image object results in 3 empty image objects in the mdx file
- deleting one of those objects results in 6 empty image objects being saved
- saving an image object with values results in two images with the same values being saved
- and then if I refresh the page, it shows 3 images with the same values, even though only 2 are saved to the file

PR #2291, opened 2021-11-29
Fix support of `object`, `list: true` in MDX templates

jeffsee55:
When specifying `list: true` for an `object` in an MDX template there was some strange behavior, it was mapping over the template `fields` rather than the values. Tests previously weren't covering this scenario so those have also been updated.

I also noticed that empty paragraph blocks weren't getting stripped out during stringify, it may be desirable for multiple lines to represent empty paragraphs but it's hard to know when it's intentional. I think it's best to remove them for now


jeffsee55:
Closes https://github.com/tinacms/tinacms/issues/2287

examples/basic/.tina/schema.ts
```diff
@@ -15,10 +15,32 @@ export default defineSchema({
...
+    //       type: 'string',
+    //       label: 'Blog Post Body',
+    //       name: 'body',
+    //       isBody: true,
+    //       ui: {
+    //         component: 'textarea',
+    //       },
+    //     },
+    //   ],
+    // },
```
Enigmatical, line 38: If this isn't needed anymore, probably couldn't hurt to just remove it.

Enigmatical (dismissed):
Looks good 👍

logan-anderson:
Looks good to me too. Might have to resolve some conflicts to be merged

PR #2295, opened 2021-11-30
Introduce pattern for code-splittable hooks

jeffsee55:
@jamespohalloran as discussed, this is what I'm calling the "hook dependency injection" pattern. It's something @dwalkr and @logan-anderson had both expressed would be worth giving a shot and I'm really digging it. While it's not a critical capability right now since we can only fetch data from `getStaticProps`, when we move on to supporting data fetching from anywhere it'll be important to allow for this.

> Note: I've called the hook `useTina` for now with the hope that we'll come up with something better.

## Test it out
Run the `examples/basic` starter from this branch.

Here's the build to show we're not bundling tina:
```
Page                                       Size     First Load JS
┌ ● / (413 ms)                             3.19 kB        84.6 kB
├   /_app                                  0 B            81.4 kB
├ ○ /404                                   196 B          81.6 kB
├ ○ /admin/[[...slug]] (348 ms)            505 kB          586 kB
├ ● /posts (414 ms)                        2.14 kB        83.6 kB
└ ● /posts/[slug] (394 ms)                 2.13 kB        83.6 kB
    └ /posts/HelloWorld (394 ms)
+ First Load JS shared by all              81.4 kB
  ├ chunks/framework-09d69b8678b26c3f.js   48.1 kB
  ├ chunks/main-f881cbd7dc874919.js        27.9 kB
  ├ chunks/pages/_app-9cb8f0b992a4106a.js  3.64 kB
  └ chunks/webpack-7fb74210fb4d8894.js     1.84 kB
```

---
The main reason for this pattern is so we can keep data-related APIs collocated to where the data fetching actually happens. I _think_ this API achieves what we're after here.

Instead of setting up the "live" data hook (`useGraphQLForms`) from within a component at `_app.js`, we can now do it adjacent to the data-fetching logic:
```tsx
const query = `{
  getPageDocument(relativePath: "home.mdx"){
    data{
      body
    }
  }
}`

// props only needs `data` now, though you _could_ still pass the other variables 
// through. This is something I think I'd do because the SDK plays nicely with it.
  const data = useTina({ query, variables: {}, data: props.data })
  const content = data.getPageDocument.data.body
  return (
    <TinaMarkdown content={content} />
  )
}

export const getStaticProps = async () => {
  const data = await staticRequest({
    query,
    variables: {},
  })

  return {
    props: {
      data,
    },
  }
}
```

And the `_app.js` doesn't do anything related to data (it also no longer needs a render prop for it's children):
```tsx

  return (
    <>
      <TinaEditProvider
        editMode={
          <TinaCMS
            branch="main"
            clientId={NEXT_PUBLIC_TINA_CLIENT_ID}
            isLocalClient={Boolean(Number(NEXT_PUBLIC_USE_LOCAL_CLIENT))}
          >
            <Component {...pageProps} />
          </TinaCMS>
        }
      >
        <Component {...pageProps} />
      </TinaEditProvider>
    </>
  )
}

```

### Considerations

**We may still want to pass `query` and `variables` through from `getStaticProps`**
The SDK work generates the query string for you, and the ergonomics for this are great IMO. the nice thing about this is that you can do whatever you prefer

**We'll probably want to introduce another hook that does data fetching in production**
So far this pattern focuses on "hydrating" data from `getStaticProps`. In a more general sense, we should also support actually doing the data fetching, too (ie. no `data` prop is required). This would be useful in RSC and when fetching data client-side. It might look like this:

```tsx
// for client-side data fetching
export default function MyComponent() {
  const [data, isLoading] = useTinaQuery({ query, variables: {})
  if(isLoading) {
    return <Loader />
  }

  const content = data.getPageDocument.data.body
  return (
    <TinaMarkdown content={content} />
  )
}

// for use with suspense (no need to check loading state)
export default function MyComponent() {
  const data = useTinaQueryWithSuspense({ query, variables: {})

  const content = data.getPageDocument.data.body
  return (
    <TinaMarkdown content={content} />
  )
}
```
One thing we'll want to do if/when we go down this path is probably change the `TinaEditProvider` to mean something more generic, and have that be the location where you can load read-only tokens and specify the content API URL, etc so these hooks can grab that info from context. This is only necessary when we have read-only tokens.

**Might be violating the rule of hooks**
I'd kind of say it is, but I'm not sure it could possibly matter. Very relevant discussion on it the other day on [Twitter](https://twitter.com/LuisEduardoDev/status/1463986915242303493). Looks like if Dan doesn't approve, Tanner does - he used the term "hook dependency injection" 😉


jeffsee55:
Can you elaborate on what you mean here?
```tsx
const Post = (props) => {
// what is `forms` here? Why does it have props?
const data = useForms(props.forms,props.data)
 
// ...
}
```

logan-anderson:
For client side data fetching to me it would just make sense to use the same hook but make data optional. Or maybe it makes sense to split them up. Here is what it would look like if we did not split it up.

```ts
// for client-side data fetching
export default function MyComponent() {
  const  data = useTina({ query })
  const isLoading = useLoadingState()

  if(isLoading) {
    return <Loader />
  }

  const content = data.getPageDocument.data.body
  return (
    <TinaMarkdown content={content} />
  )
}
```
We could also generate some hooks (or a client) for the user the user to use on the front end. This way they would not need to remember to remember the query or have it attached to the component. 

```ts
// for client-side data fetching
export default function MyComponent() {
  const  data = useFetch{CollectionName|CollectionName}Data()
  const isLoading = useLoadingState()

  if(isLoading) {
    return <Loader />
  }

  const content = data.getPageDocument.data.body
  return (
    <TinaMarkdown content={content} />
  )
}
```

Again this would be for client side fetching only

<hr/>


> The SDK work generates the query string for you, and the ergonomics for this are great IMO. the nice thing about this is that you can do whatever you prefer

Total agree with ☝️ 

Could return the query in `getStaticProps` and then do

``` ts
const data = useTina({ query:  props.query variables: {}, data: props.data })
```


  On thing to point out is that this is a tad more work (in terms of lines of code) for the user but it is definitely more **natural** in the way a “_Typical_” data fetching works. This might be good to test in front of a few users.

dwalkr:
The implementation here definitely breaks the Rules of Hooks, and I think that poses a problem. 

The linked Twitter discussion is about something similar but subtly different - using dependency injection via props, where those dependencies happen to be Hooks. This is probably still a not-great idea, but if you promise to yourself that you will never change what hook you're referencing, it's probably OK (though that leads to questions of whether this is the only way to do this kind of dependency injection.)

The hook returned from `useTina` is one that we **know** will change, because the main selling point of this approach is that the value is hydrated with a hook that does something different when in edit mode. We are, then, expecting it to change. This is a step beyond base-level dependency injection concerns.

Fundamentally, I think the only reason this is a problem is **because we're swapping in a Hook at runtime**, and only **because the thing we're swapping in is a Hook**. This is because Hooks are weird (magically weird) and rely on a lot of assumptions about your code e.g. they can't be called conditionally, which React tries to detect and throw an error. Fooling React into conditionally calling a Hook would indicate to me that we've merely evaded their Rules of Hooks detection and not that we've found a "safe" way to create hooks dynamically.

jeffsee55:
@dwalkr I think I pretty much agree on all fronts with your concerns. I think the same thing is achievable by just lifting that state a bit higher and passing down updated values down via context.

jeffsee55:
@dwalkr lifted state up a bit, of course this is just a POC, but does this meet a better standard in your opinion?

jamespohalloran:
Can someone point out where in this change where we're breaking the rule of hooks?

jamespohalloran:
And sorry for the delay in responding to this:

> EDIT: what I meant was, can you elaborate on what you mean by this:
> 
> > Could the form registration be done outside this hook, to avoid having to use something like formify to extend things?

What I meant was, with your examples, the form registration is still done under the hood. So if the user wants to customize form registration (making a global form, doing custom behavior on submit, etc), then they need to do it in the root-level `formify` function. 

Wondered if splitting out the formRegistration/data-syncing (useContextualForms) from the content querying (useTinaQuery) might make sense.

```

  //load draft data, and forms for a piece of content
  const [draftContent, forms] = useTinaQuery({props.query, variables}, props.data)
  
  //register out forms, and keep "liveData" in sync
  const liveData = useContextualForms(forms,draftContent)
}
```

Feels maybe a bit more flexible? And `useContextualForms` _should_ stay consistent regardless if you are using another datasource outside of Tina Cloud, if we change how the SDK fetches content, etc.

I'm still a bit unsure on the level of value of this though tbh, but that's what I originally was referring to.

dwalkr:
> Can someone point out where in this change where we're breaking the rule of hooks?

The change in da58ecb58da40ecfbf59c982a050a6af03baafa8 looks like it doesn't break the rules. Previously the problem was that the `useForms2` hook's behavior was being modified at runtime (by retrieving it via context, and altering what gets returned.) This would effectively make it a conditionally-running hook.

jeffsee55:
Closing as we don't intend to ship this, but we'll use it as a reference.

PR #2302, opened 2021-11-30
2235 - Modifies `tinacms init` to add the new `TinaAdmin` page

### LINKED ISSUE #2235: CLI bootstrap the Admin files (replace existing functionality)
We should also update the tina cloud starter with this pattern

Enigmatical:
Replaces `/admin` and `/exit-admin` with the `TinaAdmin` pattern of `/pages/admin/[[..tina]].js`.

Closes #2235 

There will be a sister PR to change the documentation for `/admin` here:
https://tina.io/docs/tinacms-context/#manually-toggling-edit-mode

### DIFF
--- a/.changeset/wicked-eagles-turn.md
+++ b/.changeset/wicked-eagles-turn.md
@@ -1,0 +1,5 @@
+---
+'@tinacms/cli': patch
+---
+
+Modifies `tinacms init` for `TinaAdmin`-ready page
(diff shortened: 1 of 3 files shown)


jeffsee55 (approved):
So this will still create the `blogPost` page right? One thing we _might_ also want to do is provide a mapping function so the admin links to that page since it's kind of hard to find. Not a blocker 👍

PR #2304, opened 2021-12-01
fix: remove organization from cli init codegen

jamespohalloran:
Remove old organization param from cli codegen


Enigmatical (approved):
👍

PR #2307, opened 2021-12-01
Ensure GraphQL resolve doesn't access "bridge" documents

jeffsee55:
There were a couple of places that this was calling into the "bridge" during a GraphQL request, there's a strange mixture of when/where we ask for config files which will improve as these capabilities solidify, and the "bridge" and "store" bits of logic will be easier to pull further apart. 

cc @kldavis4


Enigmatical (commented):
Looks good to me 👍

Enigmatical (approved):

PR #2308, opened 2021-12-02
Dont memoize schema configs

jeffsee55:
@logan-anderson pointed out that schema updates weren't triggering from the CLI when a schema changed. This is due to the memoization of these into class properties and they weren't getting reinitialized. I think this should do for now, but I want to revisit this as part of a cleanup when the data store stuff settles.


logan-anderson (approved):
Working for me!

Side note: Would it be possible to enable this when not using local (on Tina Cloud or in CI)?

jeffsee55:
> Side note: Would it be possible to enable this when not using local (on Tina Cloud or in CI)?

Do you mean as an optimization?

logan-anderson:
> Do you mean as an optimization?

Yeah exactly!

jeffsee55:
Yeah probably, I think just updating that value on rebuilds will work too but things are a bit all over the place right now.

PR #2312, opened 2021-12-02
Use Github URL for the basic starter

logan-anderson:


jamespohalloran (dismissed):

.github/workflows/e2e.yml
```diff
@@ -3,6 +3,7 @@ on: [push]
 jobs:
   cypress-run:
     runs-on: ubuntu-20.04
+    container: cypress/browsers:node14.17.0-chrome88-ff89
```
Enigmatical, line 6: @logan-anderson Did this sneak in from your e2e PR?

Enigmatical (approved):
Just one question, but otherwise 👍

.github/workflows/e2e.yml
logan-anderson: Yes, @Enigmatical. Something broke (I have no idea why) but that line fixed it. Without it the tests seem to not want to run.

PR #2317, opened 2021-12-02
Pin version number and meta data during schema compilation

jeffsee55:
Pin version number and metadata from @tinacms/graphql during schema compilation. This can be used to ensure the proper version is provided when working with Tina Cloud and appropriate features (like data are switched on)

Updated schemas will now look like this:
```json
  "version": {
    "fullVersion": "0.57.2",
    "major": "0",
    "minor": "57",
    "patch": "2"
  },
  "meta": {
    "flags": [
      "experimentalData"
    ]
  },
```

@kldavis4 and I discussed this and while including the API version in the URL is great, the data layer work will mean that webhooks coming from Github also need to know which version to build the index with. The upshot from this is that it's a more fullproof way of know that if the CLI compiled with a given schema, they're guaranteed to get that same version on Tina Cloud without needing to do anything.

One thing I'm not sure about, how do we handle npm tags like `beta`, etc.? Their version numbers are like this: `0.0.0-202111212155`, and in an ideal world we at least have a "beta" tag supported on Tina Cloud 🤔

### DIFF
--- a/.changeset/sixty-kiwis-listen.md
+++ b/.changeset/sixty-kiwis-listen.md
@@ -1,0 +1,5 @@
+---
+'@tinacms/graphql': patch
+---
+
+Pin version number from @tinacms/graphql during schema compilation. This can be used to ensure the proper version is provided when working with Tina Cloud
--- a/examples/basic/.tina/__generated__/_schema.json
+++ b/examples/basic/.tina/__generated__/_schema.json
@@ -5,6 +5,7 @@
     "minor": "57",
     "patch": "2"
   },
+  "meta": {},
   "collections": [
     {
       "label": "Page Content",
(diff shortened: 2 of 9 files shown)


jeffsee55:
We don't actually have "tag" information in the `package.json`, it's just however we published it: `npm publish @tinacms/graphql --tag beta`, so that info is only on NPM. We could come up with something in the CLI, perhaps to add our own metadata to the package.json before publishing. 🤔

logan-anderson:
Was just reviewing this and noticed some strange UI behaviour.
https://www.loom.com/share/7d8c89d56f1d47e18f7f38f7dc48ab84
 I Don't think it was introduced in this PR  worth. 

cc @spbyrne

logan-anderson (approved):
Looks good to me aside from a some strange UI bugs.

jeffsee55:
Yeah definitely seems unrelated, gonna go ahead and merge

Issue #1669, opened 2020-12-14
Add an `onChange` hook to field-level configs


jeffsee55:
## Description

I want an `onChange` event at the field-level - however this isn't supported so I'm instead using `parse` - but running into an issue when trying to do something on list-type fields like `blocks`, `list`, and `group-list`, digging into the code I can see there's no `parse` supplied at this level. 

I understand that `parse` doesn't make any sense for my needs here, I'm proposing that we add an `onChange`, which is a "fire-and-forget" side effect whenever the field's value changes, whether that be a text input or a list.

### Steps to reproduce

Build any form and supply a `parse` prop to an iteratable field type:

```jsx
const formOptions = {
  fields: [
    {
      label: 'Authors List',
      name: 'rawJson.authors',
      component: 'group-list',
      description: 'Authors List',
      parse: () => {console.log("I will never be called")}
      fields: [
        {
          label: 'Name',
          name: 'name',
          component: 'text',
          parse: () => {console.log("I will be called when the text value changes")}
        },
      ],
    },
    //...
  ],
}
```

### Expected result

`parse` runs for iterable fields when the order is changed or when an item is added/removed. I consider this a bug since the Typescript types indicate this is supported.

### Actual result

`parse` never runs, `format` runs too often for my needs.

chrisdmacrae:
@jeffsee55 in my experience, `parse` runs when the data is _being sent to the field component_. `format` runs when the data is being serialized back into the form.

jeffsee55:
@chrisdmacrae Yeah that's my experience too, so parse should be the solution here.. but the issue is that it doesn't trigger for any of our list fields.

jeffsee55:
Dug into this with @chrisdmacrae on Friday, we'd expect an event from final form's `mutator` functions:

For example the block insert doesn't send an event when this happpens:

https://github.com/tinacms/tinacms/blob/master/packages/@tinacms/fields/src/plugins/BlocksFieldPlugin.tsx#L95

And it seems like it might be an issue with final form, not sure if it's __suppposed__ to send an event here

https://github.com/final-form/final-form-arrays/issues/53

jeffsee55:
I've got a workaround for now, but it's pretty messy

zenflow:
@jeffsee55 https://github.com/tinacms/tina-graphql-gateway seems to be a private repository :( Any chance of making it public?

jeffsee55:
[...]

PR #2318, opened 2021-12-03
Add event for whenever a field's value is updated

jeffsee55:
This is something we've needed for a while to improve the ergonomics of `useGraphQLForms`. The form's values don't line up directly with what a given graphql response's values are, and that split has been difficult to manage. While final form has provided a `subscribe` feature that we can use to watch form values as they change, it's been impossible to know which field actually caused the change, resetting the entire document on each 'change' event is not something we're able to do for performance reasons, especially for more complex structures like blocks and references. And fine-grained APIs like 'parse' and 'format' aren't designed for this purpose (and they don't fire for array mutations like "move", "insert", and "delete" anyway), there's a lengthy discussion with @Enigmatical [here](https://github.com/tinacms/tinacms/issues/1669) on the various issues, but I believe this work should address our main needs.

Another issue that we've had was that even if we were able to get access to the change event, we would only be able to see the _final form_ input information, not the Tina field configuration itself. This has been fixed by adding meta data to each field via the [final-form-set-field-data](https://github.com/final-form/final-form-set-field-data) plugin. With this event system, we'll be able to attach more information to a given field so we can intelligently update GraphQL values with ease. Some of the things we'll now be able to do with this event:

- Remove `path` and `mutationInfo` meta data at the root of the GraphQL response. This is used to determine of a given value change was a "reference" field, and how it should be updated when we get a new value, but this information can now be attached to the field directly and accessed from the `onChange` event.
- Only update values which are actually queried for (ie. if `title` isn't in a query, don't populate `data.title` when its value is changed)
- Add contextual information to the field input. Again, if you didn't query for `title`, it might be desirable to block that field from being edited in this context (since the change won't appear on your page).
- Easier support for [aliases](https://graphql.org/learn/queries/#aliases) in GraphQL (if you alias `title` to `myTitle: title` in GraphQL, we can now ensure that `myTitle` will be updated.
- Sync value changes to more than one location of GraphQL data. This would also require that registering 2 forms with the same id will resolve to a single instance, which isn't something we've discussed yet. Example:
```graphql
{
  getNavDocument(relativePath: "main.md") {
    data {
      featuredPost {
        ...on PostDocument {
          title
        }
      }
    }
  }
  getPostDocument(relativePath: "hello-world.md") {
    data {
      title # form changes to `title` should update `getNavDocument.featuredPost.title`
      body
    }
  }
}
```

Currently the api is designed to look like this: 

```js
cms.events.subscribe(`onChange:${form.id}`, event)
```
Where event is:
```ts
{
  type: "onChange:<formid>",
  field: FinalFormField & { data: { tinaField: TinaField } },
  value: unknown,
  mutationType: {
    type: string // one of change, insert, move, remove
    // additional meta data about the mutation
  }
}
```

To demonstrate, the `useGraphQLForms` hook is currently just logging this information to the console.


logan-anderson:
When pulling down. the code and running the `basic` starter. I got this error.
Screen Shot 2021-12-16 at 11 35 38 AM

dwalkr (approved):

jeffsee55:
@logan-anderson I'll fix that up (gonna try and make it so we don't have to deprecate that)

jeffsee55:
Closing for now, will revitalize as part of `useGraphQLForms` refactor.

PR #2319, opened 2021-12-03
refresh id_token and access_token if expired for tina cloud calls

mittonface:
This was pretty simple. The `access_token` contains the `client_id` and `iss` (the token issuer). If we have the `refresh_token` that's really all we need to make a call to get a refreshed `access_token` and `id_token`. Before any request is made to Tina Cloud we check if the token is expired, if it is we try to refresh it and swap the tokens.

I also implemented a super quick JWT parsing method because I didn't see a dependency that could parse jwts and I didn't want to add one.

Tested this out by giving my pool 5 minute token expiry and it works pretty well! Feedback welcome, JS is hard.


mittonface:
I'm not sure how to comprehend the e2e test output. But they pass locally 🤷‍♂️

logan-anderson:
@mittonface There was an issue with running e2e testing in CI. If you merge main into this it should fix it.

jeffsee55 (approved):
Thanks @mittonface !

Issue #2321, opened 2021-12-03
Print out next steps with create-tina-app after a starter's been created 


jamespohalloran:
## Summary

After creating a starter with create-tina-app, we should log out some instructions of what to do next. 

### Basic example

We currently tell them to do a `yarn dev`, but maybe we could add notes on:

- Go to localhost:3000/admin to enter edit-mode.
- Edit some content on localhost:3000 (See https://tina.io/using-tina-editor )
- Check out [our concept docs](/docs/schema/), to learn how Tina powers the starters under the hood.
- Learn how [Tina can be extended](/docs/advanced/extending-tina/) to create new field components.
- Make your site [editable with Tina on production](/docs/tina-cloud/).

### Motivation

If we show the `create-tina-app` command on our homepage, the user will have more context on how to use the starter, and where to take it next

logan-anderson:
@jamespohalloran I added this in the CLI and it is deployed now. Going to close this for now but feel free to re-open it if we should revisit it.

PR #2325, opened 2021-12-06
Fix slow init

logan-anderson:
The issue was that for some reason the `.gitignore` file was not being copied from the starter. This was causes the entire `node_modules` file to be committed. The CLI will now add a `.gitignore` file if there is not one present.


logan-anderson:
@jamespohalloran I also updated the output as suggested in https://github.com/tinacms/tinacms/issues/2321

jeffsee55 (approved):

PR #2331, opened 2021-12-07
Fix 2323 enable bold and italic in plate editor

### LINKED ISSUE #2323: MDX-Editor & TinaMarkdown not saving bold, italic, underline, striketrough correctly
## Description

When using the new mdx editor in a rich-text field I can format texts with bold, italic, underline and strikethrough and this also gets rendered correctly in the preview. Unfortunately when I save the content, the bold, italic, underline, strikethrough markdown formatting doesn't get written to the disk, despite it being present in the payload that get's sent to the TinaCMS local dev server. So on reload, the formatting is gone.

Interestingly, formatting text as headline 2 works without any problems and also gets written to disk correctly.

Schema:
image

Editor View + Onsite Rendering:
image

Console.log of payload inside formify onSubmit hook that gets sent to the TinaCMS local dev server:
image

Saved `.md` file on disk:
```markdown
---
components:
  - mdxContent: |+
      This should be bold text.

      This should be italic text.

      This should be underlined text.

      This should be strikethrough text.

      ## This is headline 2.

    _template: blogPostSection
---
```

Thus on a reload, all formatting of the text other then the headline 2 is gone.

### Steps to reproduce

1. Create a rich-text field in a document of a collection
2. Add text to the rich-text field that is formatted bold, italic, underline or strikethrough in the new mdx editor
3. Save the document from the tina sidebar
4. Reload the page in the browser

### Expected result

The `.md` file should contain respective formatting of text in bold, italic, underline or strikethrough.

### Actual result

The `.md` file doesn't contain any markdown formatting of text in bold, italic, underline or strikethrough. It's just plain text.

#### Tinacms Package Versions
```text
├── @tinacms/auth@0.50.2
├── @tinacms/cli@0.57.2
├── next-tinacms-cloudinary@3.4.0
├── tinacms@0.59.1
```

#### Tools

| Tool    | Version |
| ------- | ------- |
| browser | brave Version 1.32.113 Chromium: 96.0.4664.45 (Offizieller Build) (64-Bit)        |
| os      |    Ubuntu 20.x     |
| node    |    v16.13.0     |
| npm     |      8.1.4   |

Perni1984:
fixes #2323 

As stated I am unsure how to fix underline and strikethrough support as both don't seem to be supported in standard markdown but would need a plugin or something.

### DIFF
--- a/packages/@tinacms/graphql/src/mdx/stringify.ts
+++ b/packages/@tinacms/graphql/src/mdx/stringify.ts
@@ -439,6 +439,64 @@ ${out}
       }
 
     case 'text':
+      if (node?.bold && node?.italic) {
+        return {
+          type: 'strong',
+          children: [
+            {
+              type: 'emphasis',
+              children: [
+                {
+                  type: 'text',
+                  value: node.text || '',
+                },
+              ],
+            },
+          ],
+        }
+      }
+
+      if (node?.bold) {
+        return {
+          type: 'strong',
+          children: [
+            {
+              type: 'text',
+              value: node.text || '',
+            },
+          ],
+        }
+      }
+
+      if (node?.italic) {
+        return {
+          type: 'emphasis',
+          children: [
+            {
+              type: 'text',
+              value: node.text || '',
+            },
+          ],
+        }
+      }
+
+      // type delete is only supported via an extension
+      // see https://github.com/syntax-tree/mdast#delete
+      //
+      // if(node?.strikethrough) {
+      //   return {
+      //     type: 'delete',
+      //     children: [{
+      //       // @ts-ignore
+      //       type: 'text',
+      //       value: node.text || '',
+      //     }]
+      //   }
+      // }
+
+      // it seems that underline is not supported directly via mdast
+      // but only via github flavored markdown extension
+
       return {
         type: 'text',
         // @ts-ignore
--- a/packages/@tinacms/graphql/src/mdx/stringify.ts
+++ b/packages/@tinacms/graphql/src/mdx/stringify.ts
@@ -439,69 +439,19 @@ ${out}
       }
 
     case 'text':
-      if (node?.bold && node?.italic) {
-        return {
-          type: 'strong',
-          children: [
-            {
-              type: 'emphasis',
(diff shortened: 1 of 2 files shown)


packages/@tinacms/graphql/src/mdx/stringify.ts
```diff
@@ -439,6 +439,64 @@ ${out}
       }
 
     case 'text':
```
logan-anderson, line 441: It might be nice to refactor this with "wrapping" or "Decorator" style of code. It just might make it a bit more maintainable.

For example.
```ts
case 'text':
      let returnNode: Content = { type: 'text', value: node.text || '' }

      if (node?.bold) {
        returnNode = { type: 'strong', children: [returnNode] }
      }
      if (node?.italic) {
        returnNode = {
          type: 'emphasis',
          children: [returnNode],
        }
      }
      return returnNode
```

Should do the trick.

packages/@tinacms/graphql/src/mdx/stringify.ts
```diff
@@ -439,6 +439,64 @@ ${out}
...
+          children: [
+            {
+              type: 'text',
+              value: node.text || '',
+            },
+          ],
+        }
+      }
+
+      // type delete is only supported via an extension
```
logan-anderson, line 483: Interesting comments, not sure what to do here, thoughts @jeffsee55 ?

logan-anderson (changes requested):
Thank you so much for your contribution! I left a couple of small comments but overall really good! I hope we can fix this fully soon (support all the components ) but this is a really good start.

Thanks again 😄

packages/@tinacms/graphql/src/mdx/stringify.ts
jeffsee55: Yeah I think we should just remove strikethrough for now, it can be implemented with a custom mdx component if needed.

jeffsee55 (approved):
Thanks, @Perni1984! This works for me, I think we'll remove the strikethrough button for now from the toolbar.

jeffsee55 (approved):

PR #2334, opened 2021-12-08
Add .nvmrc file to indicate preferred node version

Perni1984:
As I ran into some node problems during development I thought it would be convenient to indicate the preferred node version in an .nvmrc file as stated here https://github.com/tinacms/tinacms/blob/main/CONTRIBUTING.md#getting-started


jeffsee55 (approved):
Yes! Thanks @Perni1984

Issue #2149, opened 2021-10-20
tinacms package readme uses dated API


jamespohalloran:
## Summary

The readme for the tinacms package uses an outdated API: https://github.com/tinacms/tinacms/tree/main/packages/tinacms
It references functions like LocalClient (which is no longer called directly), useGraphqlForms
It also has some mistakes, like tina being added as a dev dependency in the installation

This readme is what shows up for our npm package, so it may cause confusion. 
https://www.npmjs.com/package/tinacms

jeffsee55:
For now we'll update README.md files to point to Tina.io

PR #2351, opened 2021-12-13
update TinaCMS package readme

logan-anderson:
Fixes https://github.com/tinacms/tinacms/issues/2149


Enigmatical (approved):
A lot of red.  For a second, I thought we should probably replace some of this info (rather than wholesale remove it), but reading the associated issue, it is definitely better to remove and add back given just how incorrect everything is.

PR #2354, opened 2021-12-13
feat: Add UI for when in localMode

### LINKED ISSUE #2290: Visual Indicator to show when things are in local-mode.
## Summary

I question if all users know that Tina sites configured with local-mode, won't work on production. I've also made the mistake of thinking I'm in local-mode, when I'm not (and vice-versa)

### Basic example

I think something simple like this would help

Screen Shot 2021-11-29 at 12 59 07 PM

It could also link to a doc that talks more about why one should use local-mode & differences between local-mode vs prod-mode (writing to filesystem vs commiting, auth, etc)

jamespohalloran:
Add some UI for when we are in localmode.

Screen Shot 2021-12-14 at 1 16 15 PM

TODO:
- [x] Consider abstracting isLocal check. 
- [x] Create docs page detailing local-mode, vs prod-mode, and update the link.

fixes #2290

I wanted to add some tests, so you'll see there's some unrelated changes here as well:
- Added missing test scripts
- Added some missing devDependencies that were listed as peerDependencies (react-is)
- skip tests that were failing (but were being run on prod anyways). They may want to be looked at separately if they're intended to be passing


logan-anderson:
@jamespohalloran if you run `yarn` from the root of the monorepo (or merge main back into this) the tests should pass.

Enigmatical (dismissed):
This looks good to me barring any styling changes @spbyrne might suggest making 👍

jamespohalloran:
@spbyrne already removed my wonderful fuschia styles 😭 😉 , so it should be good to go on that front!

spbyrne (approved):
Looks good! I reduced the type size a bit just now, it seemed a bit big when I took another look at it.

Enigmatical (approved):
Generated schema had a version conflict that I resolved; this should be good to go 👍

PR #2355, opened 2021-12-13
chore(tinacms): fix yarn lock file

Perni1984:
should fix the yarn lock file so e2e tests go through and unblock publishing a new package on npm

image

will fix broken github action runs:
https://github.com/tinacms/tinacms/runs/4510560803?check_suite_focus=true
https://github.com/tinacms/tinacms/runs/4511212209?check_suite_focus=true

see https://stackoverflow.com/questions/67062308/getting-yn0028-the-lockfile-would-have-been-modified-by-this-install-which-is-e

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -25001,7 +25001,7 @@ fsevents@~2.1.2:
   languageName: node
   linkType: hard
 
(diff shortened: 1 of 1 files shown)


logan-anderson (approved):

logan-anderson:
thanks @Perni1984 🎉

Issue #2262, opened 2021-11-22
Tina schema should error when 2 collections use the same path


jeffsee55:
`collection.path` _and_ `collection.matches` need to be unique between one collection and the next.

```
// invalid
{
  path: 'content/posts'
},{
  path: 'content/posts'
}
// valid
{
  path: 'content/posts',
  matches: '**/*.en.md'
},{
  path: 'content/posts'
  matches: '**/*.fr.md'
}
// invalid
{
  path: 'content/posts',
  matches: '**/*.en.md'
},{
  path: 'content/posts'
}

Issue #2182, opened 2021-11-01
Don't allow kebab case for any `name` attributes in the schema


jeffsee55:
```ts
{
  name: "my-field", // not allowed
  label: "Quote",
  type: "rich-text"
}
```

jeffsee55:
Probably related. https://github.com/tinacms/tinacms/issues/1970

jeffsee55:
Closed by https://github.com/tinacms/tinacms/pull/2256

PR #2358, opened 2021-12-14
Fix/2262

logan-anderson:
Fixes https://github.com/tinacms/tinacms/issues/2262

Starting to look at the code for the CLI audit command and doing this more well defined issue first to get a sense of the code.

I also added a better error message when validating the `name` field (should close https://github.com/tinacms/tinacms/issues/2182)

cc @jeffsee55

### DIFF
--- a/.changeset/rich-mirrors-hang.md
+++ b/.changeset/rich-mirrors-hang.md
@@ -1,0 +1,5 @@
+---
+'@tinacms/graphql': patch
+---
+
+validate the schema for `path` and `matches`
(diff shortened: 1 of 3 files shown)


jeffsee55:
This looks good but it would be a good opportunity to add some tests, we have this file which you could probably copy from to see how to set things up https://github.com/tinacms/tinacms/blob/main/packages/%40tinacms/graphql/src/spec/build-schema/index.spec.ts

Enigmatical (approved):
Creative use of `Set()`; I like it 👍

PR #2362, opened 2021-12-14
2342 - [TinaAdmin] Truncates the `filename` column for the Document List View

Enigmatical:
I converted the `filename` column to use `flex` with `flex-shrink-1` on the `filename` portion to force longer filenames to truncate in the event they are too long.

Closes #2342 

Screen Shot 2021-12-14 at 5 29 39 PM

PR #2367, opened 2021-12-15
Bug/2366 style changes breaks react inline forms text input

xaviemirmon:
Adding `color: inherit;`'s back in on inline forms to fix issue #2362


jeffsee55 (approved):
Thanks @xaviemirmon - I should mention that we'll probably be deprecating the inline APIs when we get to 1.0. We haven't had success with them and will likely replace it with a more out-of-the-box solution in the new year. 

I'm curious how you're finding the API? Are you shipping the Tina bundles to your production site or are you stripping them out for users who are not in edit mode?

PR #2371, opened 2021-12-16
Ensure dates are stringified from resolver

jeffsee55:
I noticed we were sometimes getting the epoch format of a date from GraphQL. Pairing with @Enigmatical and we discovered that yaml will actually parse a string as a date if it can from frontmatter. Suprisingly GraphQL converts it to epoch when resolving, whereas `JSON.stringify` would have handled it appropriately (hence why it only fails on the `data` portion and not the `values/dataJSON` ones.


logan-anderson (approved):

packages/@tinacms/graphql/src/resolver/index.ts
```diff
@@ -623,9 +623,16 @@ export class Resolver {
     assertShape<{ [key: string]: unknown }>(rawData, (yup) => yup.object())
     const value = rawData[field.name]
     switch (field.type) {
+      case 'datetime':
+        // See you in March ;)
```
dwalkr, line 627: 😂

dwalkr (approved):

packages/@tinacms/graphql/src/resolver/index.ts
Enigmatical: Made a joke about how it seems, every three months, we have to discuss dates 😆

Issue #2320, opened 2021-12-03
Stuck with disabled document creation buttons 


scottgallant:
## Description

You can get stuck on the document creation modal where the save/reset buttons are disabled. 

[Describe the issue that you're seeing.
](https://www.loom.com/share/46634a20abdb4fac96b7676ec4f7e135) 

### Steps to reproduce

1. Click create document
2. Update name property
3. Update any other property
4. notice the disabled buttons

To enable document creation reset/save buttons, you must update name field after changing collection/template.

PR #2372, opened 2021-12-20
fixed 2320

logan-anderson:
Fixes https://github.com/tinacms/tinacms/issues/2320

@Enigmatical The issue was is that the from was being  re-created when the user selected a new "collection". This caused `initialValues`===`currentValues` this making the form ["pristine"](https://final-form.org/docs/react-final-form/types/FieldRenderProps#metapristine). When a form is pristine it can not be submitted or reset.

### DIFF
--- a/.changeset/green-lobsters-sneeze.md
+++ b/.changeset/green-lobsters-sneeze.md
@@ -1,0 +1,5 @@
+---
+'tinacms': patch
+---
+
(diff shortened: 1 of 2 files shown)


Enigmatical (approved):
The sooner we can deprecate the old `DocumentCreator` for the `TinaAdmin`, the better.

Issue #2165, opened 2021-10-26
Cannot remove image from fields


scottgallant:
## Description

After you add an image to an image field, you cannot remove it (you can only swap it for a new one).

jeffsee55:
- [ ] Add a trash icon hover to remove the image

PR #2261, opened 2021-11-19
POC for injecting tailwind

jeffsee55:
Vite automatically builds css into a JS variables as a string. This allows us to just drop that string into a `<style>` tag. This allows us to just inject styles via `<style>` tag like CSS-in-JS libs do.

PR #2373, opened 2021-12-20
Added delete button to image field

logan-anderson:
Fixes https://github.com/tinacms/tinacms/issues/2165

### DIFF
--- a/packages/@tinacms/graphql/src/resolver/index.ts
+++ b/packages/@tinacms/graphql/src/resolver/index.ts
@@ -779,6 +779,7 @@ export class Resolver {
       case 'image':
         return {
           component: 'image',
+          clearable: true,
           ...field,
           ...extraFields,
         }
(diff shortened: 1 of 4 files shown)


packages/@tinacms/graphql/src/resolver/index.ts
```diff
@@ -779,6 +779,7 @@ export class Resolver {
       case 'image':
         return {
           component: 'image',
+          clearable: true,
```
Enigmatical, line 782: Ah, simple enough. 👍

Enigmatical (approved):

packages/@tinacms/graphql/src/resolver/index.ts
logan-anderson: @Enigmatical this did present some style issues that might have to be fixed in  https://github.com/tinacms/tinacms/pull/2261

PR #2377, opened 2021-12-20
Use apiURL prop instead of other configs

jeffsee55:
We've chatted back and forth about this change a fair bit. And want to get it in front of people so we can solicit some more feedback. The general feeling is that with Tina's ability to work locally and with Tina Cloud, there's a bit of confusion on where your API requests are being made and when deploying, what pieces need to be changed to move from local to cloud. With this approach, we've made it so that it's all contained in a single property that you pass as a `<TinaCMS>` prop.

This approach replaces `branch`, `clientId`, `isLocalClient` props with single `apiURL`. When working locally, this should be `http://localhost:4001/graphql`. For Tina Cloud, use `https://content.tinajs.io/content/<my-client-id>/github/<my-branch>`.

```tsx
// _app.tsx
// ...
<TinaCMS apiURL={process.env.NEXT_PUBLIC_TINA_API_URL} {...pageProps}>
  {(livePageProps) => <Component {...livePageProps} />}
</TinaCMS>
```

DEPRECATION NOTICE: `branch`, `clientId`, `isLocalClient` props will be deprecated in the future


packages/tinacms/src/tina-cms.tsx
```diff
@@ -179,11 +200,31 @@ export const TinaCMSProvider2 = ({
...
+   *
+   * When working locally, this should be http://localhost:4001/graphql.
+   *
+   * For Tina Cloud, use https://content.tinajs.io/content/my-client-id/github/my-branch
+   */
+  apiURL: string
+  /**
+   * Point to the local version of GraphQL instead of tina.io
+   *
+   * @deprecated use apiURL instead
```
logan-anderson, line 214: Could we link to the docs in the deprecation warning?

packages/tinacms/src/tina-cms.tsx
```diff
@@ -200,6 +241,20 @@ export const TinaCMSProvider2 = ({
   if (typeof props.query === 'string') {
     props.query
   }
+
+  // branch & clientId are still supported, so don't throw if they're provided
+  if (!props.apiURL || (props.branch && props.clientId)) {
```
jamespohalloran, line 246: As is, won't this make things not work for anyone using props.branch && props.clientId instead of apiURL? Should it be:
```suggestion
  if (!props.apiURL || ( !props.clientId && !isLocalClient )) {
```

jamespohalloran (commented):
I think this is a great change, I think it will make things click much easier.

packages/tinacms/src/tina-cms.tsx
```diff
@@ -179,11 +200,31 @@ export const TinaCMSProvider2 = ({
...
+   * The URL for the GraphQL API.
+   *
+   * When working locally, this should be http://localhost:4001/graphql.
+   *
+   * For Tina Cloud, use https://content.tinajs.io/content/my-client-id/github/my-branch
+   */
+  apiURL: string
+  /**
+   * Point to the local version of GraphQL instead of tina.io
+   *
```
jamespohalloran, line 213:
```suggestion
   * https://tina.io/docs/tinacms-context/#adding-tina-to-the-sites-frontend
   *
```
as per @logan-anderson 's suggestion. Note: we'll have to update that doc along with this change being published.

PR #2387, opened 2021-12-23
fixed tailwind tests

logan-anderson:
This _should_ fix the tests 

cc @Enigmatical


logan-anderson:
they are passing! 🎉  🎉 

The issue was that  I was using `cy.wait(number)` which is  [<ins>**not** </ins> recommend](https://docs.cypress.io/api/commands/wait#Time) as different computers can take longer depending on the CPU, network, etc. It would also slow down the tests. I refactored to use callbacks instead and it seems to be working. 

I would like to make these tests even more robust in the future  but they should be _fine_  for now. 

Feel free to pull this in @Enigmatical or @jeffsee55

Enigmatical (approved):
I'm fine with these changes.  I wouldn't mind a second pass (after we merge) to clean up the commented out `wait()` just because they could be red herrings once we verify they are indeed not the correct way to do things.

Issue #2363, opened 2021-12-15
Can not create a new document.


logan-anderson:
Can not create a new document when add a new collection and "live" reload the schema.

Steps to reproduce:
1. start tina server
2. add a new collection
3. Save `schema.ts` and it will make a rebuild
4. Add a new document to that schema using `TinaAdmin`
5. See the following error

```
[
  Error: Unable to find collection for file at content/asdf/asdf1639527112498.json
      at TinaSchema.getCollectionByFullPath (/Users/logananderson-forestry-mac/dev/personal/teachcode.io/node_modules/@tinacms/graphql/dist/index.js:12561:15)
      at Database.stringifyFile (/Users/logananderson-forestry-mac/dev/personal/teachcode.io/node_modules/@tinacms/graphql/dist/index.js:23376:45)
      at processTicksAndRejections (internal/process/task_queues.js:95:5)
      at async Database.put (/Users/logananderson-forestry-mac/dev/personal/teachcode.io/node_modules/@tinacms/graphql/dist/index.js:23335:46)
      at async Resolver.createResolveDocument (/Users/logananderson-forestry-mac/dev/personal/teachcode.io/node_modules/@tinacms/graphql/dist/index.js:22405:7) {
    locations: [ [Object] ],
    path: [ 'createDocument' ]
  }
]
```

See video: https://www.loom.com/share/7a16035e43ad4ce681158c6dc31c5384

Tagging @jeffsee55 as I am guessing this is related to the new data layer stuff that was added.

PR #2398, opened 2022-01-05
rebuild database every schema change

logan-anderson:
@jeffsee55  this "Fixes" https://github.com/tinacms/tinacms/issues/2363 but I am not sure if this is the correct way to fix it.

### DIFF
--- a/.changeset/fair-pots-cry.md
+++ b/.changeset/fair-pots-cry.md
@@ -1,0 +1,5 @@
+---
+"@tinacms/cli": patch
+---
+
+Rebuild database every save while in local mode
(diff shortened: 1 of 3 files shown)


jeffsee55 (commented):
Perhaps it's easiest to make `database` a variable that has scope on the entire module:
```ts
import type { Database } from '@tinacms/graphql'
let database: Database
```
And then in the 2 places we `createDatabase`, just assign it to that variable. That way we know both the schema building process and the express server have the latest change.

logan-anderson:
> Perhaps it's easiest to make `database` a variable that has scope on the entire module:
> 
> ```ts
> import type { Database } from '@tinacms/graphql'
> let database: Database
> ```
> 
> And then in the 2 places we `createDatabase`, just assign it to that variable. That way we know both the schema building process and the express server have the latest change.

@jeffsee55 For some reason when I did this it did not work and reverted to the broken behaviour.

jeffsee55 (approved):

PR #2404, opened 2022-01-06
fixed ts types

logan-anderson:
Noticed that the types where not building on `main` so I fixed it.


Enigmatical (approved):
Simple enough change 👍 

We had some `types` issues in the `inject-tailwind` branch yesterday that mostly went under the radar.

PR #2409, opened 2022-01-07
MDX - fix "reset" and don't allow tabbing

jeffsee55:
Don't allow "tabbing" for rich-text. Tabs in markdown represent code blocks, so this isn't something we want to support.

Fixes bug where "reset" wasn't working for rich text.


Enigmatical (approved):

Enigmatical:
The `e2e` error is due to timeout which is the "classic" `e2e` error we've been experiencing.

PR #2416, opened 2022-01-11
2357 - Introduces TinaAdminApi & Consolidated Types

### LINKED ISSUE #2357: Abstract Tina admin APIs

Enigmatical:
My goal was to pull out the direct GraphQL calls into a centralized `class` and consolidate the `types` that `TinaAdmin` uses.  While simple at this point, my intent is to use the same `TinaAdminApi` in both `TinaAdmin` as well as the new `Sidebar` so that both places are using the same queries.

Closes #2357


jeffsee55 (approved):
Looks great!

PR #2424, opened 2022-01-12
2423 - Upgrades Tailwind to v3

### LINKED ISSUE #2423: Upgrade to Tailwind 3
Upgrade `tailwindcss` from `v2` to `v3` following the Upgrade Guide:
https://tailwindcss.com/docs/upgrade-guide

Enigmatical:
Released in early December, Tailwind v3 represents a pretty significant upgrade over v2:

https://tailwindcss.com/blog/tailwindcss-v3

I decided to take a stab at upgrading our `inject-tailwind` branch to v3 by following along with their upgrade guide:

https://tailwindcss.com/docs/upgrade-guide

Fortunately, we were already using the `jit` engine, so a majority of the changes revolved around renamed attributes and minor config changes.

Closes #2423


Enigmatical:
@jeffsee55 I did encounter one warning:

```
vite v2.5.0 building for production...
transforming (1) src/index.ts[vite:css] Nested @tailwind rules were detected, but are not supported.
Consider using a prefix to scope Tailwind's classes: https://tailwindcss.com/docs/configuration#prefix
Alternatively, use the important selector strategy: https://tailwindcss.com/docs/configuration#selector-strategy
1  |  .tina-tailwind {
2  |    @tailwind base;
   |     ^
3  |    @tailwind components;
4  |    @tailwind utilities;
✓ 39 modules transformed.
```

This seems to be related to `postcss-nested`.  From what I can tell, everything still _worked_, but I'm wondering if you encountered this before and possibly solved it?

jeffsee55:
Off the top of my head I don't remember that exact error, though I know the nesting there isn't something css or tailwind tools were happy with, we've been ignoring those lint errors. There may be a better approach but basically all we care about is that all of tailwind's classes get namespaced 🤷

packages/tinacms/src/styles.css
```diff
@@ -1,16 +1,16 @@
...
-
-  @layer utilities {
-    .icon-parent {
-      svg {
-        fill: currentColor;
-      }
+@layer utilities {
+  .icon-parent {
+    svg {
+      fill: currentColor;
```
Enigmatical, line 8: The `tailwindcss/nesting` plugin should be the solution for the warning related to `nesting` here:
https://tailwindcss.com/docs/using-with-preprocessors#nesting

packages/tinacms/src/styles.css
```diff
@@ -1,16 +1,16 @@
-.tina-tailwind {
-  @tailwind base;
-  @tailwind components;
-  @tailwind utilities;
+@tailwind base;
+@tailwind components;
+@tailwind utilities;
```
Enigmatical, line 3: This will be affected by the `important: '.tina-tailwind'` declaration in the config.

packages/@tinacms/scripts/src/index.ts
```diff
@@ -179,9 +180,8 @@ e.g: "forestry types:gen --help"
 
 const config = (cwd = '') => {
   return {
-    mode: 'jit',
     // prefix: 'tina-',
-    important: true,
+    important: '.tina-tailwind',
```
Enigmatical, line 184: Looking through the styles, this causes every rule to preface with `.tina-tailwind`, not unlike the nesting we were doing.

packages/@tinacms/scripts/src/index.ts
```diff
@@ -385,14 +385,7 @@ const config = (cwd = '') => {
         },
       },
     },
-    variants: {
-      backgroundColor: ['responsive', 'even', 'hover', 'focus'],
-      extend: {
-        animation: ['group-hover'],
-        padding: ['first', 'last'],
-      },
-    },
```
Enigmatical, line 394: Removed in `v3`

packages/@tinacms/scripts/src/index.ts
```diff
@@ -385,14 +385,7 @@ const config = (cwd = '') => {
...
-    purge: [path.join(cwd, 'src/**/*.{vue,js,ts,jsx,tsx,svelte}')],
+    content: [path.join(cwd, 'src/**/*.{vue,js,ts,jsx,tsx,svelte}')],
```
Enigmatical, line 388: Swapped in `v3`

logan-anderson:
Looks like some styles where added that effect the users site

The basic starter on this branch
Screen Shot 2022-01-25 at 2 19 11 PM

The basic starter on `main`
Screen Shot 2022-01-25 at 2 20 36 PM

Maybe some css resets is being applied?

cc @Enigmatical @spbyrne

spbyrne:
I'll take a look at that, @logan-anderson

spbyrne:
It appears this is coming from `@tailwind base;`, which is essentially a browser reset and not encompassed by the `important: '.tina-tailwind',` wrapping logic, likely because it has styles that apply to the html and/or body elements. When we used to wrap things the 'wrong' way it fixed the issue.

spbyrne:
Removing the important flag and going back to using the wrapper class works (with a small caveat) but gives that ugly warning, not sure how to get around that. Apparently that came with the introduction of the new JIT mode in Tailwind.

Enigmatical:
Closing this after encountering a "bug" (?) with preflight not getting wrapped in `!important: class` which causes them to affect the entire site.

> Scott B: 
Hey, spent some time digging around to figure out an issue in that Tailwind 3 PR. Basically the @tailwind base; styles are actually a browser reset and aren’t affected by their important: '.class'  wrapper, oddly (even though it includes things like button resets). We need them in the wrapper class but can’t do that without also reverting that important change & getting the big ugly yellow warning every build.

> Chris D:
Yeah... I want to avoid that big ugly yellow warning, too.

> Scott B:
Can they be suppressed with ts-ignore or is it coming from tailwind itself? I wonder if some clever configuration can get us around the issue given it’s basically them saying there’s an issue when there isn’t actually anything wrong.

Issue #2413, opened 2022-01-10
Introduce pattern for code-splittable hooks


jeffsee55:
We've done some exploration on how to make Tina feel a bit less "magical" and have landed on a decent API we'd like to introduce. The POC is [here](https://github.com/tinacms/tinacms/pull/2295) but we'll want to make sure we address a few things if we're going to make this change:

1. It should not add anything to the _production_ bundle. ie. when not in edit-mode, these hooks should essentially do nothing
2. Backwards compatibility should be considered. Right now, when you pass the `query` and `variables` through, we automatically "formify" them. We'll want to come up with a way to move away from that pattern while continuing to allow it to work.
3. What should this hook be called? Bear in mind that we will probably introduce _another_ hook which is very similar. The hook we're discussing does nothing when not in edit mode, this is because the data fetching has already occurred, but once React Server Components lands, we'll also want a hook that does data fetching _and_ formifies that data. So I think our naming should keep that in mind.

PR #2426, opened 2022-01-13
Split data from root TinaCMS provider

jamespohalloran:
Based off of the bones of [this POC](https://github.com/tinacms/tinacms/pull/2295/files).

## What is it

Giving the user the option to register forms separately from the root TinaCMS component via a new `useTina` hook.  I expect this approach will ultimately deprecate passing `query`,`variables`, & `data` into the root container.

## Example implementation

For the full example, see the basic starter in this PR: https://github.com/tinacms/tinacms/pull/2426/files#diff-42691b463aec5743dc024951e2db4ed514228f73d276082d4e6e1cdad9b1b246

```jsx
//app.jsx

  return (
    <>
      <TinaEditProvider
        showEditButton={true}
        editMode={
          <TinaCMS apiURL={apiURL}>
            <Component {...pageProps} />
          </TinaCMS>
        }
      >
        <Component {...pageProps} />
      </TinaEditProvider>
    </>
  )
}
```

```jsx
//index.jsx
// ...

const query = `{
  getPageDocument(relativePath: "home.mdx"){
    data{
      body
    }
  }
}`

  const { data } = useTina({
    query,
    variables: {},
    data: props.data,
  })

  const content = data.getPageDocument.data.body
  return (
    <Layout>
      <TinaMarkdown content={content} />
    </Layout>
  )
}

export const getStaticProps = async () => {
  const variables = {}
  const data = await staticRequest({
      query,
      variables,
    })
  return {
    props: {
      data,
    },
  }
}
```

## Reasoning behind this change.

This rework has a few short-term advantages, but also allows us to be more flexible in the future.

1) The previous approach of `<TinaCMS>` in query/variables/parameters was very rigid. All of those three variables needed to be passed from getStaticProps, and be routed through the _app.jsx layout. If one of your pages already had a "data" property, then you would need to rename that property to avoid collisions. All of your query data for a page can now completely live within a given page. You no longer need to pass your queries through the root, and have them named "query","variables","data". 
2) The root config is very simple (For basic cases, it's likely [1 prop](https://github.com/tinacms/tinacms/pull/2426/files#diff-42691b463aec5743dc024951e2db4ed514228f73d276082d4e6e1cdad9b1b246R21)). Contextual editing with `useTina` is its own concept, and might not even be required for sites that only support editing through the admin.
3) By removing the reliance on passing the `query`/`data`/`variables` props to the root layout.. our suggested implementation is less NextJS-based, and will be more universal for other implementations (E.g Remix, our basic [code-sandboxes](https://tina-gql-playground.vercel.app/), etc) 

A few things to note about this PR:
- If you aren't passing query/variables/data into `<TinaCMS>`, The `<TinaCMS>` [renders its children as a ReactNode, instead of a renderProp ](https://github.com/tinacms/tinacms/pull/2426/files#diff-42691b463aec5743dc024951e2db4ed514228f73d276082d4e6e1cdad9b1b246R22)
- Currently, if you were to have multiple useTina calls on the same page, one would override the other. If we want to change this behaviour, I think splitting up useGraphqlForms would be a precursor. 
- I tweaked useGraphqlForms so that it won't blow up if you pass in an empty query. 
- The `useTina` hook lives outside the `tinacms` bundle, and is very light. Bundle size for each page with `useTina` increases only ~-.09kB (We're importing the hook from our Tina Provider, but not TinaCMS)
Screen Shot 2022-01-14 at 12 53 02 PM

fixes: #2413

TODO: 

- [x] Allow the user to use/not-use render props
- [x] Fix issue when switching between pages
- [x] Sanity check bundle size to make sure we're not bundling tinacms in the prod bundle
- [x] Test backwards-compatibility if you are to pass query/variables into `<TinaCMS>`, using/not-using render prop etc.
- [x] Decide on naming (`useTina`?)

~~Should formify be allowed at the hook level?~~ (Probably not super important right now).


jeffsee55:
> Currently, if you were to have multiple useTina calls on the same page, one would override the other.

Is this an issue when someone uses the `<TinaCMS>` query/variables props _and_ the new hook? The `useGraphQLForms` refactor I'm working on will handle this, that being said I think we can ignore that problem for now since we don't anticipate this hook to be used anywhere besides the top of the tree until we get read-only tokens.

packages/tinacms/src/tina-cms.tsx
```diff
@@ -289,23 +249,167 @@ export const TinaCMSProvider2 = ({
...
+        return args.createForm(args.formConfig)
+      }
+    },
+  })
+
+  useDocumentCreatorPlugin(props.documentCreatorCallback)
+  return (
+    <ErrorBoundary>
+      <TinaDataProvider formifyCallback={props.formifyCallback}>
+        {isLoading && (
```
jamespohalloran, line 336: Fixed an existing issue here where we were unmounting/mounting children here when loading state changed

packages/tinacms/src/edit-state.tsx
```diff
@@ -38,6 +39,51 @@ export const TinaEditProvider = ({
...
+
+  React.useEffect(() => {
+    setRequest({ query, variables })
+  }, [JSON.stringify(variables), query])
+
+  // A bit of a hack here
+  // We need to wait 1 frame because the parent context will need to react to our
+  // new query that we've just sent when this hook was initialized.
+  // Otherwise, things get wonky when changing pages
+  useEffect(() => {
```
jamespohalloran, line 72: So I wasn't super jazzed at needing to use this state, but I couldn't think of a workaround without significant work.

When you change pages, the frontend hook on the next page will fire its new request to the Context. It's going to take a frame for the new loading/payload state to be pushed down, so we wait a frame here. 
Once we allow multiple query requests (grouped by some type of id), we won't have to do this.

jamespohalloran:
> > Currently, if you were to have multiple useTina calls on the same page, one would override the other.
> 
> Is this an issue when someone uses the `<TinaCMS>` query/variables props _and_ the new hook? ...

Yeah that functionality wouldn't be supported as is, It would likely just override the root <TinaCMS> query with the page query.

packages/tinacms/src/edit-state.tsx
```diff
@@ -38,6 +39,52 @@ export const TinaEditProvider = ({
   )
 }
 
+export function useTina<T extends object>({
```
logan-anderson, line 42: Maybe we could make this default to the type of the data provided.
Something along the lines of
 `T extends object = typeof props.data`

jamespohalloran: It should already infer its type based on the `data` property type
Screen Shot 2022-01-19 at 4 06 09 PM

Is that what you mean?

logan-anderson: Oh perfect 😄

packages/tinacms/src/edit-state.tsx
```diff
@@ -38,6 +39,52 @@ export const TinaEditProvider = ({
...
+    setRequest({ query, variables })
+  }, [JSON.stringify(variables), query])
+
+  // A bit of a hack here
+  // We need to wait 1 frame because the parent context will need to react to our
+  // new query that we've just sent when this hook was initialized.
+  // Otherwise, things get wonky when changing pages
+  useEffect(() => {
+    if (!isDummyContainer) {
+      setTimeout(() => setWaitForContextRerender(false), 0)
```
jamespohalloran, line 74: @Enigmatical Set this timeout to 0 as per your suggestion!

packages/tinacms/src/tina-cms.tsx
```diff
@@ -212,48 +180,63 @@ const parseURL = (url: string): { branch; isLocalClient; clientId } => {
...
-   * @deprecated use apiURL instead
-   */
-  clientId?: string
+type APIProviderProps =
+  | {
+      /**
+       * Content API URL
+       *
+       * @deprecated use apiURL instead
+       */
```
Enigmatical, line 189: This is a bit confusing, I'm suspecting it was just a bit of overzealous copy-paste?

packages/tinacms/src/tina-cms.tsx
```diff
@@ -212,48 +180,63 @@ const parseURL = (url: string): { branch; isLocalClient; clientId } => {
...
+      clientId?: never
+    }
+  | {
+      /**
+       * Content API URL
+       *
+       * @deprecated use apiURL instead
+       */
```
Enigmatical, line 216: Same thing here - overzealous copy-paste?

Enigmatical (approved):
Got a question about a confusing `@deprecated` notice, but otherwise, this looks good 👍

packages/tinacms/src/tina-cms.tsx
```diff
@@ -212,48 +180,63 @@ const parseURL = (url: string): { branch; isLocalClient; clientId } => {
...
-   *
-   * @deprecated use apiURL instead
-   */
-  clientId?: string
+type APIProviderProps =
+  | {
+      /**
+       * Content API URL
+       *
+       * @deprecated use apiURL instead
```
jamespohalloran, line 188:
```suggestion
```

packages/tinacms/src/tina-cms.tsx
```diff
@@ -212,48 +180,63 @@ const parseURL = (url: string): { branch; isLocalClient; clientId } => {
...
+       */
+      clientId?: never
+    }
+  | {
+      /**
+       * Content API URL
+       *
+       * @deprecated use apiURL instead
```
jamespohalloran, line 215:
```suggestion
```

packages/tinacms/src/tina-cms.tsx
jamespohalloran: whoops! yup just removed the deprecated tag here

Enigmatical (approved):
As long as the backwards capabilities are intact, I'm a fan of this approach.  It makes sense and is certainly cleaner 👍

PR #2430, opened 2022-01-14
run publish scripts in parallel

dwalkr:
I suspect that --tolerate-republish is terminating the script early when an already-published version is found, in spite of what its name suggests.

Per [the docs](https://yarnpkg.com/cli/npm/publish):

> Warn and exit when republishing an already existing version of a package

This suggests to me that by looping over packages and publishing (as we do with `yarn workspaces foreach`), we're actually exiting the loop when a republish is detected, which is exactly what we don't want to do. I'm hoping that by running the publish tasks in parallel, `--tolerate-republish` won't terminate the whole script when a republish is detected.


jeffsee55:
I don't think this would explain why things stopped working all of a sudden, but it's worth a shot.

Issue #2441, opened 2022-01-17
Able to create a document with an invalid nested object


logan-anderson:
## Description

See video: https://www.loom.com/share/35ef7513fc8d480c8389faa73cde7206

### Steps to reproduce
Use a `schema.ts` that looks like the following,

```ts

export default defineSchema({
  collections: [
    {
      label: "Some Collection",
      name: "test",
      format: "json",
      path: "content/test",
      fields: [
        {
          required: true,
          name: "objField",
          type: "object",
          fields: [{ name: "test", type: "string", required: true }],
        },
      ],
    },
// ...
})
```

Go to `/admin` (or use the document creator), and then create a new document and click save.

Noticed how the UI is valid when you have not added a valid value for the `object` component.

### Expected result

Object component should make sure that it is valid before clicking submit.

### Actual result

Can save an invalid form

NicolasRoehm:
Hi, I am currently using tinacms 2.7.5 and @tinacms/cli 1.9.5 and I am still facing the same issue.

I am able to create a page with a required object, without actually defining it. This shouldn't be possible as the object is marked as required. Do I need to create a new issue?

PR #2443, opened 2022-01-17
fixes issue {type: "object", required: true} does not get validated on the frontend.

logan-anderson:
Fixes https://github.com/tinacms/tinacms/issues/2441 

@Enigmatical Does this fix make sense? If it does I will probably do it for the other components as well. 

This will also need some styling as well before its ready to be merged.


logan-anderson:
> Does this handle recursive nesting of groups? `tinaForm.fields` would not know if the `required` `object` has a `required` `object` inside of it - in fact, I think this code would ignore the `group` field, entirely.
> 
> I would say that Tina itself does not handle "required" object fields particularly well. We've discussed figuring out a way 
to handle nested validation, but wanted to wait until after the form/field/primitive refactor, I believe.


@Enigmatical   Because of this: https://github.com/tinacms/tinacms/pull/2443/files#diff-f972affe284a8c490486748af11bb1c33b6ee11379d127906afb616927ddbdd5R273
The following is handled.
```
{
   name: "bar",
   type: "object",
   required: true,
   fields: [...]
}
```

This will tell the user that the group is required and not allow them to save an invalidate state.

The following is not handled and would probably have to solved by nested forms. 

```
{
   name: "bar",
   type: "object",
   fields: [{type: "string", name: "foo",required: true }]
}
```

In this case the user would still be able to save an invalid form.

logan-anderson:
Maybe the fix is to make it so that object are not required? 🤔

logan-anderson:
Closing in favor of https://github.com/tinacms/tinacms/pull/2461

Issue #2440, opened 2022-01-17
Select field fails to save when there is no change is state


logan-anderson:
## Description

See video: https://www.loom.com/share/9da0600841df4c358dc7cac17d8a4147

### Steps to reproduce

With a schema.ts that looks like this

```ts
export default defineSchema({
  collections: [
    {
      label: "Some Collection",
      name: "test",
      format: "json",
      path: "content/test",
      fields: [
        {
          required: true,
          name: "foo",
          type: "string",
          options: [
            { label: "One", value: "one" },
            { label: "Two", value: "two" },
          ],
        },
      ],
    },
//...
});
```

Create a new document and click save.  Although, we where allowed to save it and there was **no issues in the UI** the field foo is empty in the new document.

 `test.json`
```
{}
```

when we go to query this because foo is required we get the following error.
```
"message": "Cannot return null for non-nullable field Test.foo.",
```

### Expected result

It should save the document with this shape as indicated by the UI. 

```json
{
 "foo": "one"
}
```

and there should be no errors when querying the data

PR #2444, opened 2022-01-17
fixes 2440

logan-anderson:
This fixes https://github.com/tinacms/tinacms/issues/2440. Not sure if it is the right approach so leaving this up as a draft PR for now.


logan-anderson:
https://github.com/tinacms/tinacms/pull/2444/files#diff-f36f5a4dbc71b24bb12e037f4c9d4170fe83988c58a5454bb86b06660cdfcac5 is the only file where "changes occur" the rest is just setting up the problem.

logan-anderson:
Going to close this in favour of another way to fix it.

Issue #2147, opened 2021-10-15
Graphql float type


kauly:
## Description

Seems that we cant use float numbers, the auto generate `schema.gql` will always infer any number as integer. I cant provide a sandbox because Im using the graphql api.

### Steps to reproduce

[DEMO PROJECT](https://github.com/kauly/tina-101)

- go to `_content/pages/home.md`
- change `mapProps.longitude` and/or `mapProps.latitude` to a float value like  `-48.527` , `-27.628`
- `yarn && yarn dev`

### Expected result

A beautiful site..

### Actual result

App crashing, auto generated graphql stuff with Int type on those fields.  More context:
```
query{
  getPageDocument(relativePath: "home.md"){
   data {
    	title
    	mapProps {
        longitude
        latitude
        zoom
      }
  	}
  }
}
```
the bug
```
{
  "errors": [
    {
      "message": "Int cannot represent non-integer value: -48.527"
    }
....
}
```

#### Tinacms Package Versions

├── @tinacms/cli@0.56.0
├── react-tinacms-editor@0.52.0
├── tinacms@0.56.0

jeffsee55:
Thanks @kauly 

I'm thinking this would make sense as a subset of the `number` primitive, so:
```ts
// .tina/schema.ts

export default defineSchema({
   collections: [{
     name: "post",
     label: "Blog Posts",
     fields: [{
       name: "lat",
       label: "Latitude",
       type: "number",
       float: true,
     }]
   }]
})
```

logan-anderson:
@jeffsee55 It seems that type: "number" still maps to a graphql type of number. Can we change this to an "Float" so we do not run into this error?

jeffsee55:
Let's double check that if we use `float` by default we _don't_ necessarily get a decimal representation (eg. `1.0`). If integers become coerced into floats, I think we should use the `float: true` portion to alter the type based on that.

logan-anderson:
This is fixed on `main` and will be released when https://github.com/tinacms/tinacms/pull/2462 is merged.

PR #2446, opened 2022-01-18
Number is now a float type

logan-anderson:
Potential fix for #2147


jeffsee55:
@logan-anderson what I meant by the comment in that ticket is that only if you pass `float: true` we should change the type. I think `float` by default for all numbers may be confusing.

logan-anderson:
@jeffsee55 as far as I can tell this is completely backwards compatible and does not save 1 as 1.0 or anything like that.

see https://www.loom.com/share/b0e12cce245b4299b033d79c11770675

logan-anderson:
> I think float by default for all numbers may be confusing.

I think it might be more confusing to a user that has `{type: 'number'}`  and types `1.2` into a number input, receive no validation errors on the front end and then get an error "unable to save document".

Issue #2449, opened 2022-01-18
Implement basic metrics


logan-anderson:
- [x] CLI audit
- [x] CLI init
- [x] create-tina-app
- [x] CLI server:start
- [x] CLI server errors

Discussion can be found here: https://github.com/tinacms/tinacms/discussions/2451

PR #2450, opened 2022-01-18
Feat/metrics

logan-anderson:
Adds basic open source metrics see in https://github.com/tinacms/tinacms/issues/2449.


jeffsee55:
Closes https://github.com/tinacms/tinacms/issues/2449

logan-anderson:
closes #2449

packages/@tinacms/cli/src/cmds/audit/index.ts
```diff
@@ -23,10 +23,19 @@ import { auditCollection, auditDocuments } from './audit'
 import { logger } from '../../logger'
 import chalk from 'chalk'
 import prompts from 'prompts'
+import { Telemetry } from '@tinacms/metrics'
 
 const rootPath = process.cwd()
 
 export const audit = async (ctx: any, next: () => void, options) => {
+  const t = new Telemetry({ disabled: options.noTelemetry })
```
jeffsee55, line 31: nitpick - a better variable name would be helpful here (and for the other sections like this).

packages/@tinacms/cli/src/cmds/audit/index.ts
```diff
@@ -23,10 +23,19 @@ import { auditCollection, auditDocuments } from './audit'
...
+  await t.submitRecord({
+    event: {
+      name: 'tinacms:audit:invoke',
```
jeffsee55, line 34: Following the convention from this other ones, maybe this should be `tinacms:cli:audit:invoke`

jeffsee55 (commented):
This looks great, a couple of small comments. One thing we may also want to include (though it may be too noisy) would be [here](https://github.com/tinacms/tinacms/blob/132c630b97bcf29fb98a91784a9444a8bad4a9b9/packages/@tinacms/cli/src/cmds/start-server/server.ts#L59). If `result.errors`, we could report those.

packages/@tinacms/cli/src/cmds/audit/index.ts
logan-anderson: Changed!

packages/@tinacms/cli/src/cmds/audit/index.ts
logan-anderson: Definitely

logan-anderson:
@jeffsee55 Updated based on your comments. I also broke this piece of work into its own issue:https://github.com/tinacms/tinacms/issues/2524

jeffsee55 (approved):

PR #2455, opened 2022-01-19
fixed model provider

logan-anderson:


Enigmatical (approved):
Really good catch with this.  Hoping this is the only issue that comes from the `inject-tailwind` merge 🤞

PR #2458, opened 2022-01-20
update yarn.lock

logan-anderson:
Test where failing on `main` because the `yarn.lock` was not up to date.


spbyrne (approved):
Nice catch 👍

logan-anderson:
I think @jeffsee55 Will have to merge it because the build is not reporting

PR #2461, opened 2022-01-20
disable required on objects

logan-anderson:
Fixes: https://github.com/tinacms/tinacms/pull/2443

### DIFF
--- a/packages/@tinacms/graphql/src/resolver/index.ts
+++ b/packages/@tinacms/graphql/src/resolver/index.ts
@@ -155,6 +155,11 @@ export class Resolver {
            * Collection with no templates...
            */
           case 'object':
+            if (collectable.required) {
+              console.warn(
+                "WARNING: `{type: 'object', : required: true}` is unsupported by our User Interfaced and could result in errors"
+              )
+            }
             response[collection.name] = {
               collection,
               fields: await sequential(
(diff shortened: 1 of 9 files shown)


packages/@tinacms/graphql/src/resolver/index.ts
```diff
@@ -155,6 +155,11 @@ export class Resolver {
            * Collection with no templates...
            */
           case 'object':
+            if (collectable.required) {
+              console.warn(
+                "WARNING: `{type: 'object', : required: true}` is unsupported by our User Interfaced and could result in errors"
```
Enigmatical, line 160:
```suggestion
                "WARNING: `{type: 'object', required: true}` is unsupported by our User Interface and could result in errors"
```
Small typos here

packages/@tinacms/graphql/src/types.ts
```diff
@@ -240,6 +243,7 @@ interface InnerObjectFields<WithNamespace extends boolean> extends TinaField {
 interface InnerObjectFieldsWithNamespace<WithNamespace extends boolean>
   extends TinaField {
   type: 'object'
+  required: false
```
Enigmatical, line 246: Should this one also be `required?: false`?

Enigmatical (commented):
Just a couple of suggestions.

packages/@tinacms/graphql/src/types.ts
logan-anderson: Yes it should! good catch, thank you

Enigmatical (approved):

Issue #2481, opened 2022-01-24
Application crash when localStorage access is denied


fguitton:
Hello,

There is an issue leading to a application crash when attempting to use TinaCMS in a browser that does not provide access to `localStorage`. This can be traced to the following line: 

https://github.com/tinacms/tinacms/blob/da77ef753bb0be606c277bc6cbdcab37ec67818a/packages/%40tinacms/sharedctx/src/edit-state-ctx.tsx#L23

There are a couple of other instance on this throughout the code which might also need to have a check baked into to defend against crash.

PR #2482, opened 2022-01-24
chore: Adding defence against crash when localStorage access is denied

fguitton:
This PR looks to address issue #2481 leading to crash when `window.localStorage` access is denied.

### DIFF
--- a/packages/@tinacms/sharedctx/src/edit-state-ctx.tsx
+++ b/packages/@tinacms/sharedctx/src/edit-state-ctx.tsx
@@ -20,7 +20,9 @@ const isSSR = typeof window === 'undefined'
 
 export const isEditing = (): boolean => {
   if (!isSSR) {
-    const isEdit = window.localStorage.getItem(LOCALSTORAGEKEY)
+    // localStorage may be intentionally disabled for site visitors, 
+    // in that case Tina can't be used so just return false
+    const isEdit = window.localStorage && window.localStorage.getItem(LOCALSTORAGEKEY)
     return isEdit && isEdit === 'true'
   }
   // assume not editing if SSR
--- a/packages/@tinacms/toolkit/src/hooks/use-local-storage.ts
+++ b/packages/@tinacms/toolkit/src/hooks/use-local-storage.ts
@@ -16,8 +16,10 @@ import * as React from 'react'
 export function useLocalStorage(key, initialValue) {
   const [storedValue, setStoredValue] = React.useState(initialValue)
   React.useEffect(() => {
-    const valueFromStorage = window.localStorage.getItem(key)
-    if (valueFromStorage != null) {
+    // localStorage may be intentionally disabled for site visitors, 
+    // in that case Tina can't be used so just bail out of value storing
+    const valueFromStorage = window.localStorage && window.localStorage.getItem(key)
+    if (valueFromStorage != null && valueFromStorage != undefined) {
       setStoredValue(JSON.parse(valueFromStorage))
     }
   }, [key])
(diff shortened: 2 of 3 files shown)


jeffsee55:
Thanks @fguitton - if we aren't able to set info on `localStorage` I think a lot of other things will break so this is assumption we'll probably want to think a bit harder about. What kind of scenario are you coming across where `localStorage` isn't allowed?

fguitton:
Hello @jeffsee55, thank you for you very prompt answer ! You're perfectly right indeed. The issue we face is that as a normal visitor to a website you don't necessarily need to use TinaCMS or indeed to enter editing mode, yet the package is still checking for the `tina.isEditing` value from localStorage, and then triggers NextJS to throw an error. I hope this clarifies it.

packages/@tinacms/sharedctx/src/edit-state-ctx.tsx
```diff
@@ -20,7 +20,7 @@ const isSSR = typeof window === 'undefined'
 
 export const isEditing = (): boolean => {
   if (!isSSR) {
-    const isEdit = window.localStorage.getItem(LOCALSTORAGEKEY)
+    const isEdit = window.localStorage && window.localStorage.getItem(LOCALSTORAGEKEY)
```
jeffsee55, line 23:
```suggestion
    // localStorage may be intentionally disabled for site visitors, 
    // in that case Tina can't be used so just return false
    const isEdit = window.localStorage && window.localStorage.getItem(LOCALSTORAGEKEY)
```

jeffsee55, line 23:
```suggestion
```suggestion
    // localStorage may be intentionally disabled for site visitors, 
    // in that case Tina can't be used so just return false
    const isEdit = window.localStorage && window.localStorage.getItem(LOCALSTORAGEKEY)
```
    const isEdit = window.localStorage && window.localStorage.getItem(LOCALSTORAGEKEY)
```

Comment (jeffsee55) [reply packages/@tinacms/sharedctx/src/edit-state-ctx.tsx:23]: Oh ok, so when your editors use Tina they _will_ have access to localStorage then, correct? That seems fine, and looks like this check would do the trick 👍

Comment (fguitton) [reply packages/@tinacms/sharedctx/src/edit-state-ctx.tsx:23]: Yes, that's right ! People needing to edit via Tina will know to enable local storage API access for the site in question 😉

**Comment by jamespohalloran** [comment]
Good catch! This is probably an area that we should add some tests around.

File: packages/@tinacms/toolkit/src/hooks/use-local-storage.ts [hunk 2]
```diff
@@ -16,8 +16,11 @@ import * as React from 'react'
 export function useLocalStorage(key, initialValue) {
   const [storedValue, setStoredValue] = React.useState(initialValue)
   React.useEffect(() => {
-    const valueFromStorage = window.localStorage.getItem(key)
-    if (valueFromStorage != null) {
+    // localStorage may be intentionally disabled for site visitors, 
+    // in that case Tina can't be used so just return false
+    const isEdit = window.localStorage && window.localStorage.getItem(LOCALSTORAGEKEY)
```
logan-anderson, line 21: is this supposed to be `window.localStorage.getItem(key)`?

logan-anderson (changes requested):

packages/@tinacms/toolkit/src/hooks/use-local-storage.ts
fguitton: Oops, I ended up with an unfortunate copy-paste of the previous suggestion. This should be sorted now 🤕

logan-anderson (approved):
Looks good to me!

PR #2495, opened 2022-01-25
Fix mutation

jeffsee55:
Fixes a missing piece of the mutation string that was breaking document creation in the admin


Enigmatical (approved):
Woops 😓

Issue #2488, opened 2022-01-25
Stop publishing deprecated packages


logan-anderson:
A task from https://github.com/tinacms/tinacms/issues/2466
We must to one publish with the deprecation noticed on the README's so that it is updated in NPM
- depends on https://github.com/tinacms/tinacms/issues/2487 to be merged and released.

PR #2500, opened 2022-01-26
Feat/add deprecated folder

logan-anderson:
This will keep the deprecated code in our mono repo but we will not build, test, or publish them in our CI.

Most of the code changes is just moving files. The big change is the in the [package.json](https://github.com/tinacms/tinacms/pull/2500/files#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519)


logan-anderson:
Fixes #2488

package.json
```diff
@@ -3,28 +3,29 @@
...
     "test": "yarn workspaces foreach --no-private -vip run test",
     "test:dev": "cypress open",
     "test:e2e": "yarn cypress run",
-    "nuke": "yarn workspaces foreach --no-private -vip run g:nuke",
-    "types": "yarn workspaces foreach --topological-dev --no-private -vip run types",
+    "nuke": "yarn workspaces foreach --exclude \"deprecated-packages/*\" --no-private -vip run g:nuke",
+    "types": "yarn workspaces foreach --exclude \"deprecated-packages/*\" --topological-dev --no-private -vip run types",
     "g:echo": "echo $INIT_CWD",
     "g:nuke": "cd $INIT_CWD && yarn exec rm -rf dist && echo \"cleaned\"",
     "version": "yarn changeset version --ignore tina-cloud-starter --ignore tina-tailwind-sidebar-demo --ignore starter-basic --ignore kitchen-sink-starter",
```
jeffsee55, line 24: Does this need to exclude those packages as well?

package.json
```diff
@@ -3,28 +3,29 @@
   "private": true,
   "workspaces": [
     "packages/@tinacms/*",
+    "deprecated-packages/*",
```
jeffsee55, line 6: I wonder if we're better off just leaving this out. By including it we sort of have to opt-out of everything that might run those packages (as you've done for some of the scripts below). Seems like it will work but at the same time I'm not sure what we gain by including it.

jeffsee55 (approved):

package.json
logan-anderson: 👍  Yeah this is a much better way of doing it

package.json
logan-anderson: I am having some issues with this PR going to go ahead with https://github.com/tinacms/tinacms/pull/2500#discussion_r792929458

logan-anderson:
@jeffsee55 everything should be fixed now. had to edit a couple of old license headers

jeffsee55 (approved):

PR #2510, opened 2022-01-28
Gracefully handle useTina without a context

jamespohalloran:
When useTina is called without a parent provider, return original data


jamespohalloran:
Closing! As the issue was a red herring. This shoudl already be supported as is

PR #2522, opened 2022-02-01
restore activefieldindicator behavior

### LINKED ISSUE #2518: Active field indicator doesn't seem to be working for some fields
## Description

The Active Field Indicator doesn't appear to be highlighting in certain cases. 

### Steps to reproduce

@Perni1984 created a video of this happening within a block: https://www.loom.com/share/07ee0ea741d34bed85a413982447136c

Here's an example outside of a block (not sure if these are the same issue):

- Edit a blog post title in the tina-cloud-starter (example: http://localhost:3000/posts/anotherPost)
- The component appears to have the correct data attribute: https://github.com/tinacms/tina-cloud-starter/blob/main/components/post.tsx#L132

### Expected result

The field is highlighted.

### Actual result

The field is not highlighted.

#### Tinacms Package Versions

```text
tinacms@0.64.2
```

dwalkr:
closes #2518

### DIFF
--- a/packages/@tinacms/toolkit/src/packages/fields/plugins/GroupFieldPlugin.tsx
+++ b/packages/@tinacms/toolkit/src/packages/fields/plugins/GroupFieldPlugin.tsx
@@ -23,6 +23,7 @@ import { FieldsBuilder, useFormPortal, FormWrapper } from '../../form-builder'
 import { useCMS } from '../../react-core/use-cms'
 import { BiPencil } from 'react-icons/bi'
 import { IoMdClose } from 'react-icons/io'
+import { wrapFieldsWithMeta } from './wrapFieldWithMeta'
 
 export interface GroupFieldDefinititon extends Field {
   component: 'group'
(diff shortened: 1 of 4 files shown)


packages/@tinacms/toolkit/src/packages/fields/plugins/GroupFieldPlugin.tsx
```diff
@@ -127,6 +128,7 @@ const Header = ({ onClick, children }) => {
         onClick={onClick}
         className="group px-4 py-3 bg-white hover:bg-gray-50 shadow focus:shadow-outline focus:border-blue-500 w-full border border-gray-100 hover:border-gray-200 text-gray-500 hover:text-blue-400 focus:text-blue-500 rounded-md flex justify-between items-center gap-2"
       >
+        <h3>GROUP FIELD CARD</h3>
```
Enigmatical, line 131: This looks like some debug code, perhaps?

```suggestion
```

Enigmatical (dismissed):
I'm trying to wrack my brain on why I originally removed the `wrapFieldsWithMeta` from those fields.  My guess is that some side effect of what that HOC adds was interfering with the form's ability to prevent navigation?  Or it could have been a UI concern?

Either way, it'll come back and we'll address it then 👍

Enigmatical (approved):

spbyrne:
Screen Shot 2022-02-03 at 8 40 35 PM

Maybe I removed it to get rid of this duplicate labelling?

logan-anderson (approved):

PR #2529, opened 2022-02-01
Ensure experimental data builds during CI

jeffsee55:
cc @kldavis4 

Also - this only affected `experimentalData`


logan-anderson (changes requested):
I am not sure why but it looks like the file watch command is not running

see: https://www.loom.com/share/524369f3d36a469e90c17114f431e750

logan-anderson (approved):

Issue #2530, opened 2022-02-01
Addressing the "feeling" of Tina Chrome + Admin


Enigmatical:
For Phase 3 of #2421 , we're addressing how both sides of the new Chrome experience "feels", specifically:

* Logging into `Tina` takes you directly to the Admin's Dashboard
* Collections in the Nav link to Admin List pages
* New Documents are created via the Admin Create form and drop the user on the Admin List page
* Using a `RouteMapper` plugin:
  * If a Document has a dedicated preview, the primary link points to that
  * If a Document does not, the primary link points to the Admin Edit form

### Potential Blockers
From @mitchmac , these "Must Have" items need to be addressed:

* [ ] #2645
  * From @Enigmatical :
  > This would be achieved via the CLI so it might be best to poke @logan-anderson and see what the effort would be.
* [x] Prevent users from the loss of unsaved content. (PR: https://github.com/tinacms/tinacms/pull/2541)
  * Discussion: https://github.com/tinacms/tinacms/issues/2530#issuecomment-1028315910
* [x] A clear description and label for the field used for the file name during Content Creation (PR: https://github.com/tinacms/tinacms/pull/2536)
  * Discussion: https://github.com/tinacms/tinacms/issues/2530#issuecomment-1028400807

### For Consideration

* [x] When navigating to a Preview page from the Admin List, the Tina Sidebar is closed by default.  Should it, instead, be open?  I think this reinforces the link between Admin > Preview (the Sidebar being open shows the page's form, ready to be edited).
  * From @spbyrne:
  > Given the sidebar only renders when you’re logged in, and given the user has to get to their content via the nav, we could just default sidebar state to ‘open’ instead of ‘closed’ and it would work how we want it to. Just tested it out and it’s a one line change.
* [ ] #2644
* [ ] A discussion about possibly showing the current "active" branch in the Sidebar/Admin header: https://github.com/tinacms/tinacms/discussions/2504#discussioncomment-2059537

jeffsee55:
### Potential blockers

- [ ] We won't have a data layer before this needs to be released, but there's at least 1 major performance bottleneck that should be fixed, tracked here: https://github.com/tinacms/tinacms/issues/2523.
[...]

PR #2539, opened 2022-02-03
Single Edit Link

spbyrne:
I think we need to streamline and lean in to the 'single' editing experience offered by an integrated Admin. Given the sidebar form can go fullscreen and offer an experience essentially identical to the admin form, the admin form isn't needed beyond content creation for content types that have routing set up. That being said, it's jarring to click an 'edit' link in CMS land and load on to a page of your website with no form visible. For now, I think we should drop the admin link for pages with route mapping set up and default the Sidebar to 'open'. 

Screen Shot 2022-02-03 at 4 30 28 PM

While I think this is our best option I'd love to have others try this out locally and see how it feels.

I don't think these issues are blockers but I'll list them:
- For pages set up with multiple forms, it's possible the sidebar will pen with the wrong form being displayed.
- For users that want to default to editing in fullscreen we'd need to add more config to make that happen but it's still possible with this setup.
- As @jeffsee55 mentioned some localStorage logic would be great to maintain sidebar state & size between page loads.
- The page refresh isn't ideal.
- From @Enigmatical: When `RouteMapper` was optional it wasn't a big deal that it felt like a POC. You weren't routed through admin to get to editing so it wasn't necessary. Given Admin is now the default, this puts a lot more pressure/importance on this feature (and might bring to light the ways it is lacking) - potential DX issue.


Perni1984:
in regards to https://github.com/tinacms/tinacms/issues/2530#issuecomment-1029383312

I can confirm that we would really like to see this change happening. For us the experience "feels" better if the sidebar is open immediately when clicking the edit button in the TinaAdmin.

Enigmatical (approved):
Looks good 👍

PR #2541, opened 2022-02-03
2540 - Uses onbeforeunload to prevent navigating away from dirty form

### LINKED ISSUE #2540: Prevent navigation away from a "dirty" form

Enigmatical:
This is more of a stop-gap solution for something potentially more complex.

As we continue to blur the lines between the Sidebar Experience and the Admin Experience, a user may unknowingly seek to navigate to a Collection List page while in the middle of editing a page.  Right now, the user would navigate to the Collection List page and lose the entirety of their changes in an "dirty" form.

This solution uses `<FormSpy />` to emit a single `onbeforeunload` event when the form is "dirty" (!pristine) and removes that event when either the form is not "dirty" _or_ has been unmounted.

Navigation _within_ the app or _within_ the admin is, as of right now, not prevented.

https://www.loom.com/share/885a6105df3142af9539c40f30b3c36d

Closes #2540


packages/@tinacms/toolkit/src/packages/form-builder/FormBuilder.tsx
```diff
@@ -83,6 +83,15 @@ export const FormBuilder: FC<FormBuilderProps> = ({
     [tinaForm]
   )
 
+  /**
+   * On unmount, clear the `onbeforeunload` event (if one is lingering)
+   */
+  React.useEffect(() => {
+    return () => {
+      window.onbeforeunload = undefined
```
jamespohalloran, line 91: I wonder if it may be better to do register this like:
```window.addEventListener('beforeunload', function (e) {
```
That way, if there's other stuff happening in beforeUnload, we don't override it.

Also, Rather than unregistering/reregistering the function anytime the form state changes, should we register the form once, and check on formState within the unload function?

Enigmatical: I originally tried to use `addEventListener()`, but ran into issues with `addEventListener()` being added too many times and `removeEventListener()` not being called enough.  `window.onbeforeunload` ensured I only ever added and removed a single instance which worked well with how `<FormSpy />` works.

"check on formState within the unload function" - outside of the `<FormSpy />` component?

dwalkr: @jamespohalloran we ended up using `finalForm.subscribe` instead of `FormSpy` for code quality reasons, but ultimately they both work the same way and only trigger when the `pristine` state changes and not every time the values change. It does still add/remove the event listeners any time that state changes, but this doesn't strike me as a terribly expensive operation

dwalkr (approved):

jamespohalloran:
Sorry I forgot to circle back on this one @Enigmatical , Looks good with those changes!

PR #2553, opened 2022-02-07
fix: type error

logan-anderson:
Fixed a small type error in cloudinary package.


jeffsee55 (approved):
Any idea how this got in here all of a sudden?

PR #2554, opened 2022-02-07
Fix sidebar overlay behaviour

spbyrne:
Fixing sidebar position so it pulls from the CMS object if available.

### DIFF
--- a/.changeset/silver-buckets-wait.md
+++ b/.changeset/silver-buckets-wait.md
@@ -1,0 +1,5 @@
+---
+'@tinacms/toolkit': patch
+---
+
+Fix sidebar position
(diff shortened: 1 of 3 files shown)


spbyrne:
Looks like there's a type error. We're using a `@ts-ignore` for a similar error in another spot but I wonder if we could add sidebar to the cms type? Or was there a reason it was omitted?

Enigmatical:
> We're using a @ts-ignore for a similar error in another spot.

We do this all the time, so I would think this is fine.  As for adding `sidebar` to the `cms` type, that's worth discussing.  There's a difference between the `CMS` and `TinaCMS` types; looks like `useCMS()` returns a `CMS` type.

jeffsee55 (dismissed):

jeffsee55 (approved):

PR #2562, opened 2022-02-08
Change the default for rich-text to use the newer version

jeffsee55:


Enigmatical (approved):
Looks good; probably will do a similar thing for `tina-admin`, honestly.

Setting the flag to `false` will revert the change (rather than `true`).

And then, eventually, just retire the flag altogether. 👍

PR #2563, opened 2022-02-08
Visual Block Selector

spbyrne:
Adds an optional new visual block selector to the blocks field.

Screen Shot 2022-02-09 at 10 57 44 AM

How to provide image and category from the template schema: (both are optional; this UI can be enabled with no schema changes to start)
```
const coolEightBlockSchema: TinaTemplate = {
  name: "cooleight",
  label: "Team",
  ui: {
    category: "Page Section",
    previewSrc:
      "https://res.cloudinary.com/dghp13dmb/image/upload/v1644345838/Block%20Picker/team_zgif9y.png",
  },
  fields: [
    {
      type: "string",
      label: "String",
      name: "string",
    },
  ],
};
```

And how to enable the visual block picker:
```
...
fields: [
        {
          type: "object",
          list: true,
          name: "blocks",
          label: "Sections",
          ui: {
            visualSelector: true,
          },
...
```

Example in sidebar, some blocks without previews, no categories:
Screen Shot 2022-02-09 at 11 25 18 AM


spbyrne:
How to provide image and (optional) category from the template schema:
```
const coolEightBlockSchema: TinaTemplate = {
  name: "cooleight",
  label: "Team",
  ui: {
    category: "Page Section",
    previewSrc:
      "https://res.cloudinary.com/dghp13dmb/image/upload/v1644345838/Block%20Picker/team_zgif9y.png",
  },
  fields: [
    {
      type: "string",
      label: "String",
      name: "string",
    },
  ],
};
```

And how to enable the visual block picker:
```
...
fields: [
        {
          type: "object",
          list: true,
          name: "blocks",
          label: "Sections",
          ui: {
            visualSelector: true,
          },
...
```

spbyrne:
I'm happy with how things are working at this point. My only complaint is that CSS columns are far from perfect, but I feel like adding something like Masonry would be overkill. Basically blocks shouldn't be forced to have images of the same size, given the fact blocks come in all shapes/sizes, but this means that we can't adhere to a grid layout, we need something more like 'Pinterest'. CSS columns works pretty good, but in cases where there are fewer blocks than columns they'll often 'stick' to the first column. 

Here's what it typically looks like:
Screen Shot 2022-02-09 at 10 50 33 AM

And here's the column issue:
Screen Shot 2022-02-09 at 10 51 01 AM

So not a huge issue, but just a UI quark to be aware of.

spbyrne:
Added to the cloud starter. This is what it looks like with no categories or search. The search only appears if there's more than 6 items, but we can tweak this value.
Screen Shot 2022-02-09 at 1 46 56 PM

Enigmatical (approved):
This looks great 👍

Issue #2457, opened 2022-01-19
Use `esbuild` for parsing `.tina/schema.ts`


jeffsee55:
Right now, we pull in Typescript as a dependency from `@tinacms/cli` so that we can parse the `.tina/schema.ts`. It's cumbersome, and not very flexible, and at the end of the day we can't rely on static type-checking to ensure a valid schema. We should just use `esbuild`, which will allow us to be more flexible (ie. users can import files from wherever, not just the `.tina` folder.

PR #2565, opened 2022-02-09
Feat/use esbuild

logan-anderson:
closes https://github.com/tinacms/tinacms/issues/2457

This PR allows the user to
- use other extensions (js, ts, jsx, tsx) (Needs testing)
- import from outside the .tina folder (needs testing on external project but is working inside the mono)

Everything is working but it needs a bit more work in the way of testing, and still needs some update to the CLI. It will be a draft PR for now.


experimental-examples/tina-cloud-starter/.tina/schema.ts
```diff
@@ -706,6 +550,10 @@ export default defineSchema({
       name: "pages",
       path: "content/pages",
       fields: [
+        {
```
logan-anderson, line 553: This can be removed

experimental-examples/tina-cloud-starter/package.json
```diff
@@ -3,10 +3,10 @@
   "version": "0.2.7",
   "private": true,
   "scripts": {
-    "dev": "yarn tinacms server:start -c \"next dev\"",
-    "dev-data": "yarn tinacms server:start -c \"next dev\" --experimentalData",
-    "build": "yarn tinacms server:start -c \"next build\"",
-    "start": "yarn tinacms server:start -c \"next start\""
+    "dev": "yarn tinacms server:start -c \"next dev\" --noTelemetry",
```
logan-anderson, line 6: We should add this to all of our starters

packages/@tinacms/cli/src/cmds/compile/index.ts
```diff
@@ -81,32 +80,46 @@ export const compile = async (_ctx, _next) => {
...
-        const newPath = file
-          .replace(posixProjectDir, posixTempDir)
-          .replace('.ts', '.js')
-        await fs.outputFile(newPath, newContent)
-        return true
-      })
+  const packageJSON = JSON.parse(
+    fs.readFileSync(packageJSONFilePath).toString() || '{}'
   )
+  const deps = packageJSON?.dependencies || []
```
logan-anderson, line 87: Is this the correct way to go about the external modules @jeffsee55 ?

yarn.lock
```diff
@@ -4962,6 +4962,8 @@ __metadata:
     "@types/progress": ^2.0.3
     "@types/prompts": ^2.0.13
     "@types/yup": ^0.29.11
+    "@yarnpkg/esbuild-plugin-pnp": ^2.0.1-rc.3
```
logan-anderson, line 4965: It turns out that I did not need this, it was working without but there might be some edge cases I am not hitting where it is needed.

packages/@tinacms/cli/src/cmds/compile/index.ts
jeffsee55: It is, but I don't think want to exclude them, it works fine for now - but if/when the `defineSchema` moves of to `tinacms` we actually need to "bundle" this so our bundled output includes everything we need (specifically the `defineSchema` function). See comment for `esbuild-plugin-pnp`. At the very least we'll want to be sure to include `tinacms`, if it's problematic to do more than that then we can cross that bridge when we get there.

yarn.lock
jeffsee55: We'll probably need to use this when we move `defineSchema` to `tinacms`, because at that point the function will be coming from another module and esbuild will need to know where to find it. The tricky part will be only including this plugin when the environment is PNP, which hopefully is as easy as checking for one of these  https://yarnpkg.com/advanced/pnpapi#runtime-constants.

Find to leave it/out for now.

packages/@tinacms/cli/src/cmds/compile/index.ts
```diff
@@ -81,32 +80,46 @@ export const compile = async (_ctx, _next) => {
...
+  const peerDeps = packageJSON?.peerDependencies || []
+  const devDeps = packageJSON?.devDependencies || []
+  const external = Object.keys({ ...deps, ...peerDeps, ...devDeps })
+
+  // Get file
+  const inputPathTS = path.join(projectDir, 'schema.ts')
+  const inputPathJS = path.join(projectDir, 'schema.js')
+  const inputPathTSX = path.join(projectDir, 'schema.tsx')
+  const inputPathJSX = path.join(projectDir, 'schema.jsx')
```
jeffsee55, line 96: This will be handy, one thing we may want to do in the near future is when `js` or `jsx` is provided, the generated SDK should be compiled to `js`.

logan-anderson: Yeah we would also have to compile the SDK to a separate file too. (it lives in type right now)

packages/@tinacms/cli/src/cmds/compile/index.ts
logan-anderson: Makes sense

logan-anderson:
Should be able to use this PR with `yarn add --dev @tinacms/cli@use-es-build`

logan-anderson:
Seeing some issues when using with external sites. This is ready for review but needs more testing before it is merged

logan-anderson:
@tinacms/maintainers 
Update:

I think this is ready to be merged one we get some eyes on it other than myself. to make this easy I added to the [tina-cloud-starter](https://github.com/tinacms/tina-cloud-starter/tree/testing/use-esbuild) so you can pull down that code and checkout the `testing/use-esbuild` branch.

spbyrne (approved):
Not a code review (I don't understand the build system) but a 'tested and this is working for me' review 👍

packages/@tinacms/cli/src/cmds/baseCmds.ts
```diff
@@ -50,7 +50,7 @@ const noSDKCodegenOption = {
 const cleanOption = {
   name: '--clean',
   description:
-    'Submit gql mutation to all files to git rid of any data that is not defined in the `schema.ts`',
+    'Submit gql mutation to all files to git rid of any data that is not defined in the `schema.{ts,tsx,js}`',
```
dwalkr, line 53:
```suggestion
    'Updates all content files to remove any data not explicitly permitted by the current schema definition',
```

packages/@tinacms/cli/src/cmds/compile/index.ts
```diff
@@ -41,74 +44,107 @@ export const resetGeneratedFolder = async () => {
...
   logger.info(logText('Compiling...'))
-  // FIXME: This assume it is a schema.ts file
-  if (
-    !fs.existsSync(tinaPath) ||
-    !fs.existsSync(path.join(tinaPath, 'schema.ts'))
-  ) {
+  let schemaExists = true
+  try {
+    getSchemaPath({ projectDir: tinaPath })
+  } catch (_) {
```
dwalkr, line 50:
```suggestion
  } catch {
```
this syntax works if you don't need to do anything with the thrown error

packages/@tinacms/cli/src/cmds/compile/index.ts
```diff
@@ -41,74 +44,107 @@ export const resetGeneratedFolder = async () => {
...
-  )
-  await fs.remove(tinaTempPath)
+  try {
+    const schemaFunc = require(path.join(tinaTempPath, 'schema.js'))
+    const schemaObject: TinaCloudSchema = schemaFunc.default
+    await fs.outputFile(
+      path.join(tinaConfigPath, 'schema.json'),
+      JSON.stringify(schemaObject, null, 2)
+    )
+    await fs.remove(tinaTempPath)
```
dwalkr, line 90: related issue here as above - the temp file won't get cleaned up if an error is thrown. you can consider using a `finally` block, though I'm not completely clear on whether `finally` will execute when you re-throw from `catch`

logan-anderson: Yeah I _dont think_ it would. I will do some refactoring to account for this.

logan-anderson: @dwalkr I updated the code to account for this

packages/@tinacms/cli/src/cmds/compile/index.ts
```diff
@@ -41,74 +44,107 @@ export const resetGeneratedFolder = async () => {
...
     await fs.writeFile(file, defaultSchema)
   }
 
-  // Turn the TS files into JS files so they can be exacted
-  await transpile(tinaPath, tinaTempPath)
+  // Turns the schema into JS files so they can be run
+  try {
+    await transpile(tinaPath, tinaTempPath)
+  } catch (e) {
+    throw new BuildSchemaError(e)
```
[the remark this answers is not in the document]
logan-anderson: I updated the code to account for this. Good catch @dwalkr

packages/@tinacms/cli/src/cmds/compile/index.ts
```diff
@@ -39,76 +42,119 @@ export const resetGeneratedFolder = async () => {
...
+    }
+  })
+  try {
+    const schemaFunc = require(path.join(tinaTempPath, 'schema.js'))
+    const schemaObject: TinaCloudSchema = schemaFunc.default
+    await fs.outputFile(
+      path.join(tinaConfigPath, 'schema.json'),
+      JSON.stringify(schemaObject, null, 2)
+    )
+    await fs.remove(tinaTempPath)
```
dwalkr, line 65: this line will still get skipped if anything throws from this `try` block, and we probably want to make sure it runs regardless

packages/@tinacms/cli/src/cmds/start-server/errors/index.ts
```diff
@@ -0,0 +1,51 @@
...
+  constructor(message) {
+    super(message)
+    this.name = 'ExecuteSchemaError'
+  }
+}
+
+export const handleServerErrors = (e: Error) => {
+  if (e instanceof BuildSchemaError) {
+    logger.error(`${dangerText(
+      'ERROR: your schema was not successfully build: see https://tina.io/docs/errors/esbuild-error/ for more details'
```
dwalkr, line 34:
```suggestion
      'ERROR: your schema was not successfully built: see https://tina.io/docs/errors/esbuild-error/ for more details'
```

packages/@tinacms/cli/src/cmds/compile/index.ts
logan-anderson: Aw yes, I overlooked what the cleanup function was doing. Should be working as expected now.

PR #2567, opened 2022-02-09
Add experimental formify (part 1 of useGraphQLForms refactor)

jeffsee55:


logan-anderson:
@jeffsee55 What can be done to unblock this work?

jeffsee55:
@logan-anderson this can be reviewed now, it's a subset of changes I expect to get through end of week

jamespohalloran:
Are there any user-facing changes with this @jeffsee55? Or is it all an under the hood refactor?

jeffsee55:
@jamespohalloran it's all under the hood, and this one is only activated with a feature flag

packages/tinacms/src/client/index.ts
```diff
@@ -179,12 +182,63 @@ mutation addPendingDocumentMutation(
...
+   * ```graphql
+   * {
+   *   getPostDocument(relativePath: "") {
+   *     data {
+   *       title
+   *     }
+   *   }
+   * }
+   */
+  getOptimizedQuery = async (documentNode: DocumentNode) => {
```
logan-anderson, line 211: Is there any reason why we don't call `optimizeDocuments` here? is it because we want to keep graphql dependencies on the backend?

logan-anderson:
Also, another quick question. What is the best way we can review this work?

packages/tinacms/src/client/index.ts
jeffsee55: The library is Node-only, to we need to get it from the server

jeffsee55:
> Also, another quick question. What is the best way we can review this work?

Probably just seeing if anything breaks for you, set `cms.flags.set('use-unstable-formify', true)`

experimental-examples/unit-test-example/pages/_app.js
```diff
@@ -0,0 +1,35 @@
+import dynamic from 'next/dynamic'
+import { TinaEditProvider } from 'tinacms/dist/edit-state'
+
+// @ts-ignore FIXME: default export needs to be 'ComponentType<{}>
+const TinaCMS = dynamic(() => import('tinacms'), { ssr: false })
+
+const NEXT_PUBLIC_TINA_CLIENT_ID = process.env.NEXT_PUBLIC_TINA_CLIENT_ID
+const NEXT_PUBLIC_USE_LOCAL_CLIENT =
+  process.env.NEXT_PUBLIC_USE_LOCAL_CLIENT || true
```
dwalkr, line 9:
```suggestion
  typeof process.env.NEXT_PUBLIC_USE_LOCAL_CLIENT !== 'undefined' ? process.env.NEXT_PUBLIC_USE_LOCAL_CLIENT : true 
```

Enigmatical:
* Switched to `use-graphql-refactor` branch
* Ran `yarn install` and `yarn watch`
* Set `cms.flags.set('use-unstable-formify', true)` in the `cmsCallback` for `experimental-examples/tina-cloud-starter`

Immediately ran into an unstable error:
```
TypeError: props.data is null
Source

pages/[filename].tsx (8:9) @ HomePage

   6 |   props: AsyncReturnType<typeof getStaticProps>["props"]
   7 | ) {
>  8 |   return <Blocks {...props.data.getPagesDocument.data} />;
     |         ^
   9 | }
  10 | 
  11 | export const getStaticProps = async ({ params }) => {
  ```
  
  * Setting `use-unstable-formify` to `false` causes the page to load correctly again.

jeffsee55:
Thanks @Enigmatical this should be fixed. I guess this confirms that the feature flag works properly at least 😆

packages/tinacms/src/client/index.ts
logan-anderson: Makes sense!

logan-anderson (approved):
Looks good to me! was able to pull it down and run with no issues.

PR #2570, opened 2022-02-10
Upgrade to Tailwind 3

spbyrne:
This is essentially the same as what @Enigmatical previously did, but disables the built-in preflight and provides a built-in preflight that's slightly customized. There's a number of benefits to Tailwind 3 that I'd like to take advantage of.

Other fixes:
- Provides a value for gray-150
- Sets a base font size for the Tina wrapper

Notes:
- The preflight is just copied into both places; can we import from a single preflight.css file?
- Given all the styles are the same, if we can import the preflight from a single location, can we also have a 'tine-tailwind.css' file that's used in both locations? The only addition in `@toolkit` are the plate styles.


jeffsee55:
> The preflight is just copied into both places; can we import from a single preflight.css file?

Since `tinacms` can't exist without `@tinacms/toolkit` we can just leave it in `@tinacms/toolkit`

Enigmatical:
> > The preflight is just copied into both places; can we import from a single preflight.css file?
> 
> Since `tinacms` can't exist without `@tinacms/toolkit` we can just leave it in `@tinacms/toolkit`

That's a good point.  I was thinking we would place it in a shared package (like how `edit-state` works), but this is a different situation than that.

Enigmatical (approved):
I've poked around both the Contextual Editor and the CMS and I see no significant changes.

Given that this both upgrades our tailwind version to current (v3) _and_ lets us take control over the preflight (which was causing us issues before), this sounds like an upgrade with no (noticeable) drawbacks. 👍

PR #2586, opened 2022-02-15
Feat/add client side hook

### LINKED ISSUE #2560: Add hook for client-side data fetching
With the addition of read-only tokens support coming soon, we'd like to expose a hook that allows us to do data fetching _and_ "hydrating" that data with contextual editing. We currently have `useTina` which accepts `query`, `variables`, and `data` - but now we'll need one which would not accept a `data` property, and would instead do the actual data fetching itself.

This hook needs to know the API url and the read-only token, I don't want to get into implementation details here too much, but we've talked about exposing a lightweight HTTP client which makes this setup easy, something that just takes `apiURL` and `token` or something similar. Perhaps that could be passed in to this hook:
```ts
const client = createClient({apiURL: "", token: ""})
const { data, isLoading } = useTinaQuery({query, variables, client})
// or more simply
const { data, isLoading } = useTinaQuery({query, variables, apiURL: "", token: ""})

```

---

Down the road we'll want this to support React Server Components, it's too early to say what that should look like but I expect we'll be able to use this same hook in that context, though I'm not sure.

Note: This work is important for Read-Only Tokens

logan-anderson:
This is ready for review but I will have to change the name before it get merged (thats why it is a draft PR)

Adds a client side hook for data fetching. This hook will load the data using a read only token. When in edit mode it will load the from data and return the live props.

Also, I need a better name then `useClientSideDataTina` any suggestions?
Here are some suggestions
- `useTinaClientQuery`
- `useClientQuery`
- `useTinaQuery`
- `useQuery`
- `useFetchData`
- `useTinaFetch`

closes #2560


logan-anderson:
Thinking about this a little bit I am starting to think we might consider having a `client` that can be used on the front end and backend. This would make it easier to do the same logic on the frontend and the backend. We could use the client in both places to ensure unified data fetching. Thoughts?

It _might_ look something like this for this hook.
```ts
import { createClient } from 'tinacms/dist/edit-state'
export const client = createClient({apiURL: "", token: ""})
```

```ts
import { client } from '../path/to/client'
//...
const { data, isLoading } = useTinaQuery({query, variables, client})
```

We could even use the same client in our provider

```ts
// pages/_app.js
import dynamic from 'next/dynamic'
import { TinaEditProvider } from 'tinacms/dist/edit-state'
import client from '../path/to/client'
const TinaCMS = dynamic(() => import('tinacms'), { ssr: false })

  return (
    <>
      <TinaEditProvider
        editMode={
          <TinaCMS client={client}>
            <Component {...pageProps} />
          </TinaCMS>
        }
      >
        <Component {...pageProps} />
      </TinaEditProvider>
    </>
  )
}

export default App
```



cc @jamespohalloran @jeffsee55 

PS
this is what sanity does: https://www.sanity.io/guides/build-your-first-blog-using-react#09943e6d2b68

jamespohalloran:
One thing that I'm a bit stuck on is that this wraps up `useTina` into a separate API, and it may make it harder to communicate, compared to having a consistent `useTina` hook, that always does the same thing.

Just spit-balling, but what if.. useTina were to look something like:
```js
const data = useTina(async (client) => {
	return client.request({query,variables})
, props.data}
```

- The first param is an async function, that returns the form/data that is used in edit-mode.
- The second is the prod data that is passed through in prod-mode

Maybe the second prop could optionally be a Promise, so for a client-side only app, it would look like:
```js
const data = useTina(async (client) => {
	const tinaResponse = await client.request({query,variables})
	return tinaResponse
, client.request({query,variables}).then(({data}) => data)
})
```

jeffsee55:
@logan-anderson I agree, just exporting `client` is all you need and is a pattern I've seen a lot from other libraries, 👍 to just passing that to `TinaCMS`.

logan-anderson:
@jamespohalloran I am confused on why we need to let the user provide how the fetching is done? I think they should provide the client and then we can do the data fetching for them. This way they don't have to worry about how the data is fetched.

Can you elaborate on why the this is a user concern?

jamespohalloran:
> @jamespohalloran I am confused on why we need to let the user provide how the fetching is done? I think they should provide the client and then we can do the data fetching for them. This way they don't have to worry about how the data is fetched.
> 
> Can you elaborate on why the this is a user concern?

So right now, in the onboarding we have 3 core concepts:
- Content modeling: Make a schema.ts
- Data fetching: Load your production content from the Tina API
- Contextual Editing: Filter your content through useTina, and make your pages live-update with the form data.

The `useClientSideDataTina` kind of ties together the Data fetching and Contextual Editing concepts (but only when you're using client-side data-fetching). I think in the past we made a mistake by introducing too many abstractions on top of our `useForm` hook (E.g `useGitHubForm`, `useJsonForm`), and what ended up happening was that it was hard to tell what was the core API, and what was happening under the hood.

My proposed solution might overcomplicate things, but what I was going for was a consistent solution of our core concepts (specifically `useTina`), regardless of whether client-side data-fetching or doing it on the backend.

jeffsee55:
> The useClientSideDataTina kind of ties together the Data fetching and Contextual Editing concepts (but only when you're using client-side data-fetching)

I think `useClientSideDataTina` (or my preference, `useTinaQuery`) - should be the default way we talk about data fetching. It'll be easier to say this is how you do it for _any_ React framework, and then branch into framework-specific scenarios when the user is ready for a bit more complexity (Remix and Next have the same constraints here).

Also, I know it's a bit late, but `useTina` is probably not an appropriate name for that hook.

logan-anderson:
Just had a chat with @jamespohalloran and we decided on this API for now with the idea of being open minded to having a unified client in the future. We decided on this.

```ts
const { data, isLoading } = useTina({
    query,
    variables,
    data: async ()=>{
        const res = await fetch(
      "https://content.tinajs.io/content/<ClientId>/github/<Branch>",
      {
        method: "POST",
        body: JSON.stringify({ query, variables }),
        headers: {
          "X-API-KEY": "<ReadOnlyToken>",
          "Content-Type": "application/json",
        },
      }
    )
  const data = await res.json()
  return data
    }
   })
```

This makes it much easier to understand as its the same `useTina` hook being used in both cases (static and dynamic content).

And when we have a better data fetching solution we can have this be the client side data fetching API.
```ts
const { data, isLoading } = useTina({
    query,
    variables,
    data: async ()=>{
        const data = await client.request({query, variables})
        return data 
    }
   })
```


for context here is all the notes I have produced.

--- 

# Data fetching in tina

## Right now we have

```tsx
const branch = 'main'
// When working locally, hit our local filesystem.
// On a Vercel deployment, hit the Tina Cloud API
const apiURL =
  process.env.NODE_ENV == 'development'
    ? 'http://localhost:4001/graphql'
    : `https://content.tinajs.io/content/${process.env.NEXT_PUBLIC_TINA_CLIENT_ID}/github/${branch}`

  return (
    <>
      <TinaEditProvider
        showEditButton={true}
        editMode={
          <TinaCMS
            // API url includes branch info, and client id
            apiURL={apiURL}
          >
            <Component {...pageProps} />
          </TinaCMS>
        }
      >
        <Component {...pageProps} />
      </TinaEditProvider>
    </>
  )
}

export default App
```

### Static fetching with the local client

```ts
const client = new LocalClient()
const data = client.request({query,variables})
```
## Front end data fetching in edit mode
```ts
const data = useTina({query, variables, data })
```



## What we need is a way to fetch data with read only tokens (Front end and backend)

### idea 1 just use fetch

```ts
import { useState, useEffect } from "react";

// This query can be any query
const query = `
query ContentQuery($relativePath: String!) {
  get<CollectionName>Document(relativePath: $relativePath) {
    data {
      body
      title
    }
  }
}
`; 

// Variables used in the GraphQL query; 
const variables = {
  relativePath: "voteForPedro.md",
};

const  BlogPostPage = ()=> {
  const [data, setData] = useState(null);
  const [isLoading, setLoading] = useState(false);

  useEffect(() => {
    setLoading(true);
    fetch(
      "https://content.tinajs.io/content/<ClientId>/github/<Branch>",
      {
        method: "POST",
        body: JSON.stringify({ query, variables }),
        headers: {
          "X-API-KEY": "<ReadOnlyToken>",
          "Content-Type": "application/json",
        },
      }
    )
      .then((res) => res.json())
      .then((data) => {
        console.log({ data });
        setData(data?.data?.getPostsDocument?.data);
        setLoading(false);
      })
      .catch((e) => {
        console.error(e);
      });
  }, []);

  if (isLoading) return <p>Loading...</p>;


  if (!data) return <p>No data</p>;

  return (
    <div>
      <h1>{data.title}</h1>
      <p>{data.body}</p>
    </div>
  );
}
export default BlogPostPage;
```

Now lets say you want to fetch data on the backend too

```ts
data = await fetch(
      "https://content.tinajs.io/content/<ClientId>/github/<Branch>",
      {
        method: "POST",
        body: JSON.stringify({ query, variables }),
        headers: {
          "X-API-KEY": "<ReadOnlyToken>",
          "Content-Type": "application/json",
        },
      }
    )
```

Problems 
- duplicating information from the frontend and backend
  - This might be able to be solved by a unified client (used on both backend and frontend)
- When you go into edit mode you will not be able to edit the content because the `useTina` hook was not being called
  - This is solved by adding a data fetching hook




## planned solution 

Add a hook that does the dataFetching for the user so that they will be able to edit the data. 

instead of the above we can do
```ts
import { useState, useEffect } from "react";

// This query can be any query
const query = `
query ContentQuery($relativePath: String!) {
  get<CollectionName>Document(relativePath: $relativePath) {
    data {
      body
      title
    }
  }
}
`; 

// Variables used in the GraphQL query; 
const variables = {
  relativePath: "voteForPedro.md",
};

const  BlogPostPage = ()=> {
   const { data, isLoading } = useClientSideDataTina({
    query,
    variables,
    token: '<ReadOnlyToken>',
    url: 'https://content.tinajs.io/content/<ClientID>/github/main',
  })

  if (isLoading) return <p>Loading...</p>;


  if (!data) return <p>No data</p>;

  return (
    <div>
      <h1>{data.title}</h1>
      <p>{data.body}</p>
    </div>
  );
}
export default BlogPostPage;
```

Now we still have the problem of duplicating the clientID, token and branch across every request.



## Idea add a unified client (this is the solution that A LOT of OSS libraries use)

`lib/client.ts`
```ts
import createClient from 'tinacms/dist/client'
export const client = createClient({
    // URL contains client ID and the branch
    url:" ",
    // Token is not needed when using the local graphql server
    token: "...",
})
```

Now we could do

```ts
import { client } from '../path/to/client'
//...
const { data, isLoading } = useTinaQuery({query, variables, client})
```

We could even use the same client in our provider

```ts
// pages/_app.js
import dynamic from 'next/dynamic'
import { TinaEditProvider } from 'tinacms/dist/edit-state'
import client from '../path/to/client'
const TinaCMS = dynamic(() => import('tinacms'), { ssr: false })

  return (
    <>
      <TinaEditProvider
        editMode={
          <TinaCMS client={client}>
            <Component {...pageProps} />
          </TinaCMS>
        }
      >
        <Component {...pageProps} />
      </TinaEditProvider>
    </>
  )
}

export default App
```

now server side data fetching can use that client as well.

(in get staticProps)
```ts 
const data = client.request({query, vars,})
```
or maybe in the future
```ts 
const data = client.posts()
```


### One thing to consider

For a lot of workflows we will not know what branch we are requesting the data on until runtime so we will NEED to support this.

```ts 
const data = client.request({query, vars, branch})
```

Also, have a feeling that one may want some request to go to the local server (build time) and then content API (runtime) so we might have to support (needed for preview mode or ISR or `fallback: blocking`).

See https://tinacms-site-next-git-feat-read-only-tokens-docs-tinacms.vercel.app/docs/graphql/read-only-tokens/#nextjs-fallback-blocking

```ts 
const data = client.request({query, vars, branch, url})
```

But now we have the same initial problem


### total tangent maybe we have

```ts
const client = createClient({
    localURL: ...,
    url: ...,
    token: ...,
})
```

```ts
// Static content
const data = client.request({query, variables, local: true })
// Dynamic content with read only tokens
const data = client.request({query, variables, local: false })
```

jeffsee55:
2 thoughts:

1. It's not clear why we'd make the user do the `fetch` boilerplate code instead of just do the `createClient` pattern.  

2. I think this essentially amounts to changing the API of `useTina`, right now - we don't recommend blocking rendering `data` if `isLoading` is `true`, under this new approach, we're now saying `isLoading` actually means the `data` is not going to be there, though it _might_ be if you gave it a `data` value rather than a `data` function. For example, this code would break https://tina.io/docs/tinacms-context/#adding-contextual-editing-to-a-page.

I'm cool with us using 1 hook, but my feeling is that this just moves the problem of "we have 2 hooks" to "we have 2 ways of providing data for the hook".

jeffsee55:
> isLoading will return false in production mode if the user provides an object (same way it works now).

My point is that we're changing what `isLoading` means, right now it means "there's data, but it's not 'live' yet", and going forward it will mean "we don't have any data yet", and it's conditional based on what you give the `data` key, which I find confusing. 

Also, one small thing, we should probably provide `query` and `variables` as part of the callback:

```js
const { data, isLoading } = useTina({
    query,
    variables,
    // this way we don't get into any inconsistent behaviors
    data: async (query, variables)=>{
        const res = await fetch(
      "https://content.tinajs.io/content/<ClientId>/github/<Branch>",
      {
        method: "POST",
        body: JSON.stringify({ query, variables }),
        headers: {
          "X-API-KEY": "<ReadOnlyToken>",
          "Content-Type": "application/json",
        },
      }
    )
  const data = await res.json()
  return data
    }
   })
```

logan-anderson:
@jeffsee55 just as a POC I wanted to make sure it was possible so I coded the proposed solution here: https://github.com/tinacms/tinacms/pull/2598

logan-anderson:
closing in favour of https://github.com/tinacms/tinacms/issues/2628

PR #2594, opened 2022-02-16
fix publish failing

logan-anderson:
The publish was failing because `tina-tailwind-sidebar-demo` was still in the package.json scrips when it did not exist.


jamespohalloran (approved):

spbyrne (approved):
Good catch 👍

Issue #2523, opened 2022-02-01
`getCollections` includes all _documents_ automatically


jeffsee55:
While the query for `getCollections` offers the ability to query through a collection to its documents, this should only run the documents look ups when requested.

This should not trigger the documents lookup
```graphql
{
  getCollections {
    name
  }
}
```
But this obviously should.
```graphql
{
  getCollections {
    name
    documents {
      totalCount
    }
  }
}
```

PR #2595, opened 2022-02-16
do not resolve all documents in `getCollection` if it is not needed

logan-anderson:
fixes https://github.com/tinacms/tinacms/issues/2523

I am not 100% sure if I am doing this the _correct_ way. Will defer to @jeffsee55 for confirmation.

### DIFF
--- a/packages/@tinacms/graphql/src/resolve.ts
+++ b/packages/@tinacms/graphql/src/resolve.ts
@@ -162,8 +162,20 @@ export const resolve = async ({
             return value
           }
           if (info.fieldName === 'getCollections') {
+            const getCollectionNode = info.fieldNodes.find(
+              (x) => x.name.value === 'getCollections'
+            )
+            const hasDocuments = getCollectionNode.selectionSet.selections.find(
+              (x) => {
+                // @ts-ignore
+                return x?.name?.value === 'documents'
+              }
+            )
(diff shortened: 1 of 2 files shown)


packages/@tinacms/graphql/src/resolve.ts
```diff
@@ -162,11 +162,23 @@ export const resolve = async ({
...
             return tinaSchema.getCollections().map((collection) => {
-              return resolver.resolveCollection(collection.name)
+              return resolver.resolveCollection(
+                collection.name,
+                Boolean(hasDocuments)
+              )
             })
           }
-          return resolver.resolveCollection(args.collection)
+          return resolver.resolveCollection(args.collection, true)
```
jeffsee55, line 181: Wouldn't we want to do the same here?

jeffsee55 (changes requested):
This is looking great! Just one small question

packages/@tinacms/graphql/src/resolve.ts
logan-anderson: @jeffsee55 I was not sure. Left it as true so that I would not break anything. What query would cause this code to run?

logan-anderson: Just updated to account for `getCollection` is there any other collection field?

jeffsee55 (approved):

Issue #1452, opened 2020-09-10
The Media interface should represent both files and directories


ncphillips:
# Summary

Allow `Media` to be files or directories

### Basic example

Add a `type` property to the interface:

```ts
interface Media {
  type: 'file' | 'dir'
  // ...
}
```

### Motivation

It is common for media to be organized inside of nested directories. In order to support this the media objects must be able to represent either files or directories.

PR #2605, opened 2022-02-23
fix: #1452: update indexDocument to handle adding new docs

kldavis4:
# what

With the experimentalData flag enabled, adding a new document in the admin interface was failing with this error:

```
{
  "errors": [
    {
      "message": "Unable to find '<filename>' in store for project <client_id>: '<owner>/<repo>', Ref: 'main'",
      "locations": [
        {
          "line": 3,
          "column": 9
        }
      ],
      "path": [
        "createDocument"
      ],
      "extensions": {
        "status": 404
      }
    }
  ],
  "data": null
}
```
I traced this issue to Database.indexDocument where was attempting to fetch an existing document from the store and not checking whether the document already existed. This PR adds an exception check to prevent this error from bubbling up. 

Ideally we'd probably be better off explicitly checking if a file exists (rather than checking if the get fails) _or_ if we know a request involves a new document, pass that info to this function and use it to avoid making the get.

### DIFF
--- a/.changeset/cyan-singers-smile.md
+++ b/.changeset/cyan-singers-smile.md
@@ -1,0 +1,5 @@
+---
+"@tinacms/graphql": patch
+---
+
+fix: #1452: update indexDocument to handle adding new docs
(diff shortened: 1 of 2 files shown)


jeffsee55 (dismissed):
Looks good!

> Ideally we'd probably be better off explicitly checking if a file exists (rather than checking if the get fails) or if we know a request involves a new document, pass that info to this function and use it to avoid making the get.

Might make sense to have a "safe" get function from the store

jeffsee55 (approved):

PR #2607, opened 2022-02-23
use defineSchema on the frontend

logan-anderson:
This is the first step in the "Extending Tina work" this moves `defineSchema` and it's types to the `tinacms` package.

This is more to show that is it possible we will benefit from this work when we take on some of the extending tina work.

### DIFF
--- a/examples/basic/.tina/schema.ts
+++ b/examples/basic/.tina/schema.ts
@@ -11,7 +11,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-import { defineSchema } from '@tinacms/cli'
+import { defineSchema } from 'tinacms'
 
 export default defineSchema({
   collections: [
--- a/.changeset/breezy-melons-carry.md
+++ b/.changeset/breezy-melons-carry.md
@@ -1,0 +1,5 @@
+---
+'tinacms': patch
+---
+
+Adds define schema to the `tinacms` package (instead of `@tinacms/cli`. This is done in preparation of the extending tina work)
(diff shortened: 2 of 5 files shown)


examples/basic/.tina/schema.ts
```diff
@@ -11,7 +11,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-import { defineSchema } from '@tinacms/cli'
+import { defineSchema } from 'tinacms'
```
jeffsee55, line 14: Does this actually work in the monorepo? I would expect we'd need the `pnpPlugin` for esbuild to know where to find `tinacms`?

logan-anderson: It was working for me.

jeffsee55: Strike that, looks like it's been fixed in esbuild https://github.com/evanw/esbuild/blob/master/CHANGELOG.md#0131. Awesome! 👍

jeffsee55: My only hesitation here is that issues _like_ the PNP one will arise at this point, even though that specific one has been fixed. We'll want to test this with pnpm and npm obviously. I guess at the end of the day if this doesn't work we can tell users to just `'@tinacms/cli'` while we make any necessary fixes. Will these compile errors be reported to our telemetrics system?

jeffsee55 (approved):

examples/basic/.tina/schema.ts
logan-anderson: Yeah I share that concern, I think this will be fine for _most_ use cases. We can leave it in the CLI for these cases.

> Will these compile errors be reported to our telemetrics system?

I think so, let me check. If they are not part of it they should be.

PR #2613, opened 2022-02-24
2612 - Sets the tina-admin flag to true by default

### LINKED ISSUE #2612: Make `tina-admin` flag default to `true`

Enigmatical:
A user can still disable `tina-admin` by setting the flag to `false`, but any lookups related to the flag will assume it is `true` if it isn't provided.

Closes #2612

### DIFF
--- a/packages/@tinacms/toolkit/src/packages/fields/components/Reference.tsx
+++ b/packages/@tinacms/toolkit/src/packages/fields/components/Reference.tsx
@@ -92,7 +92,7 @@ export const Reference: React.FC<ReferenceProps> = ({
   options,
 }) => {
   const cms = useCMS()
-  const hasTinaAdmin = cms.flags.get('tina-admin') || true
+  const hasTinaAdmin = cms.flags.get('tina-admin') === false ? false : true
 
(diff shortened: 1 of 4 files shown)


packages/@tinacms/toolkit/src/packages/fields/components/Reference.tsx
```diff
@@ -92,7 +92,7 @@ export const Reference: React.FC<ReferenceProps> = ({
   options,
 }) => {
   const cms = useCMS()
-  const hasTinaAdmin = cms.flags.get('tina-admin')
+  const hasTinaAdmin = cms.flags.get('tina-admin') || true
```
jeffsee55, line 95: If the user has `cms.flags.get('tina-admin')` set to `false` this would evaluate to `true`

jeffsee55 (approved):

Enigmatical:
Momentarily putting this into a **blocked** status.

After discussions with @gitsstewart , this PR will act as the final "nail" for the Tina Admin + Chrome initiative, to be merged once we're confident in the updated starters _and_ adequate documentation is in place.

PR #2618, opened 2022-02-25
initial work on for pagination/filtering/sorting in leveldb store

kldavis4:
# Overview

This pull request adds support for pagination, filtering and sorting with the LevelDB store implementation (and removes the MemoryStore implementation). An in-depth walk through of the changes are below (slightly outdated as I noticed a few issues that needed to be corrected during the recording). Lastly, there are a set of test queries that can be executed locally to validate functionality.

# Video Walkthroughs
- [Resolver](https://youtu.be/vSQk4wEpP54) 42m
- [Database & Indexing](https://youtu.be/CE8N94_OpUM) 20m
- [Level store query](https://youtu.be/XbEug1KO6V0) 19m

# Questions / Open Issues
- Should we keep the tina-cloud-starter changes in this PR? ie Zodiac collection linked to author and published/rating fields on the blog posts. It is useful for some of the queries below.
- should we exclude indexing of body? in dynamodb there is a limited key length, so we will need to atleast truncate long bodies
- it is possible to provide full text search now, we just don't have a substring operator. Do we want to add that?
- note: we talked about excluding items from an index if a field value is not present - this work hasn't been done yet in this PR
- we don't currently index "embedded" (non-list) objects but theoretically we can. Would this be useful? We can still filter on these without indexing, but we can't sort by an embedded object field

# Sample Queries
## Setup
`cd experimental-data/tina-cloud-starter && yarn dev-data`

### Query
```
        query GetPostsList {
          getPostsList {
            edges {
              node {
                data {
                  title
                  date
                  published
                  author {
                      ...on AuthorsDocument {
                          id
                      }
                  }
                }
              }
            }
          }
        }
```
### Variables
 ```
{}
```

### Query
```
        query GetPostsList(
            $first: Float,
            $after: String, 
            $last: Float, 
            $before: String, 
            $sort: String) {
          getPostsList(first: $first, after: $after, last: $last, before: $before, sort: $sort) {
            edges {
              cursor
              node {
                data {
                  title
                  date
                  published
                }
              }
            }
            pageInfo {
                hasNextPage
                endCursor
            }
          }
        }
```
### Variables
 ```
{
    "sort": "date"
}
```

### Query
```
        query GetPostsList(
            $first: Float,
            $after: String, 
            $last: Float, 
            $before: String, 
            $sort: String) {
          getPostsList(first: $first, after: $after, last: $last, before: $before, sort: $sort) {
            edges {
              cursor
              node {
                data {
                  title
                  date
                  published
                }
              }
            }
            pageInfo {
                hasNextPage
                endCursor
            }
          }
        }
```
### Variables
 ```
{
    "sort": "title"
}
```

### Query
```
        query GetPostsList(
            $first: Float,
            $after: String, 
            $last: Float, 
            $before: String, 
            $sort: String,
            $filter: PostsFilter!) {
          getPostsList(first: $first, after: $after, last: $last, before: $before, sort: $sort, filter: $filter) {
            edges {
              cursor
              node {
                data {
                  title
                  date
                  published
                }
              }
            }
            pageInfo {
                hasNextPage
                endCursor
            }
          }
        }
```
### Variables
 ```
{
    "filter": {
        "title": {
            "in": ["Salmon and Moonshine", "Vote For Pedro"]
        }
    },
    "sort": "date"
}
```

### Query
```
        query GetPostsList(
            $first: Float,
            $after: String, 
            $last: Float, 
            $before: String, 
            $sort: String,
            $filter: PostsFilter!) {
          getPostsList(first: $first, after: $after, last: $last, before: $before, sort: $sort, filter: $filter) {
            edges {
              cursor
              node {
                data {
                  title
                  rating
                  date
                  published
                }
              }
            }
            pageInfo {
                hasNextPage
                endCursor
            }
          }
        }
```
### Variables
 ```
{
    "filter": {
        "rating": {
            "in": [4, 5]
        }
    },
    "sort": "date"
}
```
### Query
```
        query GetPostsList(
            $filter: PostsFilter!, 
            $first: Float,
            $after: String, 
            $last: Float, 
            $before: String, 
            $sort: String) {
          getPostsList(filter: $filter, first: $first, after: $after, last: $last, before: $before, sort: $sort) {
            edges {
              cursor
              node {
                data {
                  title
                  date
                  published
                }
              }
            }
            pageInfo {
                hasNextPage
                endCursor
            }
          }
        }
```
### Variables
 ```
{
    "filter": {
        "date": {
            "after": "2021-03-12T07:00:00.000Z"
        }
    },
    "sort": "date"
}
```
### Query
```
        query GetPostsList(
            $first: Float,
            $after: String, 
            $last: Float, 
            $before: String, 
            $sort: String, $filter: PostsFilter!) {
          getPostsList(first: $first, after: $after, last: $last, before: $before, sort: $sort, filter: $filter) {
            edges {
              cursor
              node {
                data {
                  title
                  date
                  published
                }
              }
            }
            pageInfo {
                hasNextPage
                endCursor
            }
          }
        }
```
### Variables
 ```
{
    "filter": {
        "date": {
            "after": "2021-03-12T07:00:00.000Z",
            "before": "2021-07-04T07:00:00.000Z"
        },
        "published": {
            "eq": true
        }
    },
    "sort": "published-by-date"
}
```
### Query
```
        query GetPostsList(
            $filter: PostsFilter!,
            $first: Float,
            $after: String, 
            $last: Float, 
            $before: String, 
            $sort: String) {
          getPostsList(filter: $filter, first: $first, after: $after, last: $last, before: $before, sort: $sort) {
            edges {
              node {
                data {
                  title
                  date
                  published
                  author {
                      ...on AuthorsDocument {
                          id
                      }
                  }
                }
              }
            }
          }
        }
```
### Variables
 ```
{
    "filter": {
        "author": {
            "authors": {
                "name": {
                    "in": ["Napolean", "Pedro"]
                }
            }
        }
    }
}
```
### Query
```
        query GetAuthorsList(
            $filter: AuthorsFilter!,
            $first: Float,
            $after: String, 
            $last: Float, 
            $before: String, 
            $sort: String) {
          getAuthorsList(filter: $filter, first: $first, after: $after, last: $last, before: $before, sort: $sort) {
            edges {
              node {
                data {
                  name
                }
              }
            }
          }
        }
```
### Variables
 ```
{
    "filter": {
        "name": {
            "in": ["Napolean", "Pedro"]
        }
    }
}
```
### Query
```
        query GetPostsList(
            $filter: PostsFilter!,
            $first: Float,
            $after: String, 
            $last: Float, 
            $before: String, 
            $sort: String) {
          getPostsList(filter: $filter, first: $first, after: $after, last: $last, before: $before, sort: $sort) {
            edges {
              cursor
              node {
                data {
                  title
                  date
                  published
                  author {
                      ...on AuthorsDocument {
                          id
                      }
                  }
                }
              }
            }
          }
        }
```
### Variables
 ```
{
    "filter": {
        "author": {
            "authors": {
                "name": {
                    "in": ["Pedro","Napolean"]
                }
            }
        }
    },
    "sort": "author"
}
```
### Query
```
        query GetPostsList(
            $filter: PostsFilter!, 
            $first: Float,
            $after: String, 
            $last: Float, 
            $before: String, 
            $sort: String) {
          getPostsList(filter: $filter, first: $first, after: $after, last: $last, before: $before, sort: $sort) {
            edges {
              cursor
              node {
                data {
                  title
                  date
                }
              }
            }
            pageInfo {
                hasNextPage
                endCursor
            }
          }
        }
```
### Variables
 ```
{
    "filter": {
        "date": {
            "after": "2021-03-12T07:00:00.000Z"
        }
    },
    "sort": "date",
    "first": 
}
```
### Query
```
        query GetPostsList(
            $filter: PostsFilter!,
            $first: Float,
            $after: String, 
            $last: Float, 
            $before: String, 
            $sort: String) {
          getPostsList(filter: $filter, first: $first, after: $after, last: $last, before: $before, sort: $sort) {
            edges {
              node {
                data {
                  title
                  date
                  published
                  author {
                      ...on AuthorsDocument {
                          data {
                            name
                          }
                      }
                  }
                }
              }
            }
          }
        }
```
### Variables
 ```
{
    "filter": {
        "author": {
            "authors": {
                "name": {
                  "startsWith": "Nap"
                }
            }
        }
    },
  "sort": "author"
}
```
### Query
```
query getPages($first: Float, $filter: PagesFilter!){
  getPagesList(first: $first, filter: $filter) {
    edges {
       node {
        id
        data {
          blocks {
            __typename
          }
          __typename
        }
      }
    }
  }
}
```
### Variables
 ```
{
  "first": 1,
  "filter": {
    "blocks": {
      "hero": {
        "color": {
          "eq": "default"
        }
      }
    }
  }
}
```
### Query
```
query getPages($first: Float, $filter: PagesFilter!){
  getPagesList(first: $first, filter: $filter) {
    edges {
       node {
        id
        data {
          blocks {
            __typename
          }
          __typename
        }
      }
    }
  }
}
```
### Variables
 ```
{
  "first": 1,
  "filter": {
    "blocks": {
      "features": {
        "items": {
          "icon": {
            "color": {
              "eq": "red"
            }
          }
        }
      }
    }
  }
}
```
### Query
```
        query GetPostsList(
            $filter: PostsFilter!,
            $first: Float,
            $after: String, 
            $last: Float, 
            $before: String, 
            $sort: String) {
          getPostsList(filter: $filter, first: $first, after: $after, last: $last, before: $before, sort: $sort) {
            edges {
              node {
                data {
                  title
                  date
                  published
                  author {
                      ...on AuthorsDocument {
                          id
                      }
                  }
                }
              }
            }
          }
        }
```
### Variables
 ```
{
    "filter": {
        "author": {
            "authors": {
              "details": {
                "zodiac": {
                	"zodiacs": {
                  	"name": {
                    	"eq": "Sagittarius"
                  	}
                	}
              	}
              }
            }
        }
    }
}
```
### Query
```
        query GetAuthorsList(
            $filter: AuthorsFilter!,
            $first: Float,
            $after: String, 
            $last: Float, 
            $before: String, 
            $sort: String) {
          getAuthorsList(filter: $filter, first: $first, after: $after, last: $last, before: $before, sort: $sort) {
            edges {
              node {
                data {
                  name
                  details {
                    zodiac {
                      __typename
                    }
                  }
                }
              }
            }
          }
        }
```
### Variables
 ```
{
    "filter": {
      "details": {
        "zodiac": {
          "zodiacs": {
            "name": {
              "eq": "Aquarius"
            }
          }
        }
      }
    }
}
```


jeffsee55:
This is awesome, I didn't run into _any_ issues while testing locally. I just want to recap my understanding of how this is all tied together and ask a few general questions:

**All filterable fields are stored in their own LevelDB store (`db.index_date` etc)?**

- I really like this solution, it's clean and straightforward. My only question is will there performance overhead to initializing each DB, but if there ever is we can partition a single LevelDB with [one of these strategies](https://github.com/Level/awesome#sublevels).
- Also, there's no need for the metadata in the "main" DB (eg. `__attribute__...`) anymore right?
- Is each index DB is roughly equivalent to a GSI in Dynamo?

**This doesn't remove stale index values yet, right?**

The strategy I'd been implemented dealt with this from within the `database.put` function, we talked about how that responsibility should probably be at the `store` level instead, so this makes perfect sense to me. Seems like [this logic](https://github.com/tinacms/tinacms/blob/d0d5e04a577e7578ad46a08a5a8b70bdaf817798/packages/@tinacms/graphql/src/database/index.ts#L125-L129) can all go away

**The multi-column index are implicitly an AND clause, do you have any thoughts on supporting OR clauses?**

**Along those same lines, do you have any thoughts about filtering "IN" an array of items?**

**Instead of the `index` key, can we namespace the indexes?**

I'm not sure this is _better_, but want to clarify that it'd be doable:
```graphql
{
  getPostsList(filter: { 
    publishByDate: {
      published: {
        eq: true
      },
      date: {
      	after: "2021-03-12T07:00:00.000Z"
      }      
    }
  }) {...}

}
```

Doing it this way auto-complete works and we can infer the index from that key. And for single-column filters we could just use the field's name:

```graphql
{
  getPostsList(filter: { 
    title: {
      startsWith: "Just"      
    }
  }) {...}
}
```

That would require a "graphql-friendly" name so instead of "published-by-date" it'd have to be "publishedByDate".

---

## Other thoughts

We have a hard limit on GSI with Dynamo, and if that means we can only have 20 filters, and from where we are right now that seems ok, but then I also think about the fact that a reference will be an implicit filter:

"Get all posts where the author is "pedro.md"

So would that inflate the number of indexes depending on how many relationships our schema has? Or would those all be considered the same GSI?

kldavis4:
> This is awesome, I didn't run into _any_ issues while testing locally. I just want to recap my understanding of how this is all tied together and ask a few general questions:
> 
> **All filterable fields are stored in their own LevelDB store (`db.index_date` etc)?**
> 
> * I really like this solution, it's clean and straightforward. My only question is will there performance overhead to initializing each DB, but if there ever is we can partition a single LevelDB with [one of these strategies](https://github.com/Level/awesome#sublevels).

I'll look into this - I was using a different library to simulate secondary indexes but had issues and discovered it was super old, so I pivoted to separate files but I think if one of these other libs allows partitioning, a single file should be fine.

> * Also, there's no need for the metadata in the "main" DB (eg. `__attribute__...`) anymore right?

I _think_ so.. that's on my list to clean-up before this is ready to merge. I am wondering if something like this might still be useful for the FK relationships

> * Is each index DB is roughly equivalent to a GSI in Dynamo?

yes, that's the way it is modeled now

> 
> **This doesn't remove stale index values yet, right?**

Correct, that's a todo item

> 
> The strategy that had been implemented dealt with this from within the `database.put` function, we talked about how that responsibility should probably be at the `store` level instead, so this makes perfect sense to me. Seems like [this logic](https://github.com/tinacms/tinacms/blob/d0d5e04a577e7578ad46a08a5a8b70bdaf817798/packages/@tinacms/graphql/src/database/index.ts#L125-L129) can all go away

👍🏻 
> 
> **The multi-column index are implicitly an AND clause, do you have any thoughts on supporting OR clauses?**

We are limited by the dynamodb expression only allowing a single comparison (eq, gt, lt, startsWith, etc), so I think without some really clever indexing, OR probably won't be possible in dynamo without post-processing at the application level.

> 
> **Along those same lines, do you have any thoughts about filtering "IN" an array of items?**

I think if you mean something like <field> in [val1, val2, val3], that would not be possible for the same reason - as it can be translated into an OR (<field> == val1 OR <field> == val2 ...)

> 
> **Instead of the `index` key, can we namespace the indexes?**
> 
> I'm not sure this is _better_, but want to clarify that it'd be doable:
> 
> ```graphql
> {
>   getPostsList(filter: { 
>     publishByDate: {
>       published: {
>         eq: true
>       },
>       date: {
>       	after: "2021-03-12T07:00:00.000Z"
>       }      
>     }
>   }) {...}
> 
> }
> ```

I like this - I am not sure how exactly it would be done, but it seems like it would be a better experience

> 
> Doing it this way auto-complete works and we can infer the index from that key. And for single-column filters we could just use the field's name:
> 
> ```graphql
> {
>   getPostsList(filter: { 
>     title: {
>       startsWith: "Just"      
>     }
>   }) {...}
> }
> ```
> 
> That would require a "graphql-friendly" name so instead of "published-by-date" it'd have to be "publishedByDate".
> 
> ## Other thoughts
> We have a hard limit on GSI with Dynamo, and if that means we can only have 20 filters, and from where we are right now that seems ok, but then I also think about the fact that a reference will be an implicit filter:
> 
> "Get all posts where the author is "pedro.md"
> 
> So would that inflate the number of indexes depending on how many relationships our schema has? Or would those all be considered the same GSI?

Yeah, I've been thinking about this and I'm thinking it might be a case for keeping the attributes (writing them to the primary index as additional entries), at least for references. It may make sense to do the same for the other fields as well, and then just reserve the user defined indexes for multi-column values.  I'm going to be looking into this today and will follow-up with you.

jeffsee55:
> I'll look into this - I was using a different library to simulate secondary indexes but had issues and discovered it was super old, so I pivoted to separate files but I think if one of these other libs allows partitioning, a single file should be fine.

I would expect that we could also just prefix within the same store, too right?

---

> We are limited by the dynamodb expression only allowing a single comparison (eq, gt, lt, startsWith, etc), so I think without some really clever indexing, OR probably won't be possible in dynamo without post-processing at the application level.

> I think if you mean something like in [val1, val2, val3], that would not be possible for the same reason - as it can be translated into an OR ( == val1 OR == val2 ...)

So in your mind if we want to support these things in "post-processing" that means, "get the data in the first pass, get another data set in the second pass" and look at them together. At this point, are we doing sorting and pagination on our own (ie. in memory)?

I have been really curious about how we're going to tackle this kind of problem. Due to the nested nature of GraphQL, this "post-processing" (or round-trips to the DB as I've been referring to them) is somewhat unavoidable (and is one of the reasons SQL works better with GraphQL). It's possibly avoidable in these 2 scenarios because we can just say those features aren't supported, but the fact that you can do `getPostList.edges.node.data.author.name` (stepping through _each_ post and getting the associated author's name) poses the same kind of problem. On the flip side, one of our core tenets so far has been that SQL will be _hard_ to support in a way that fully normalizes the data (since schemas are dynamic and differ from one repo to the next).

I think there's a whole topic here to discuss, and the direction it goes can really vary wildly. A few thoughts I want to put out for how we might want to think about this

- This is a problem specific to any non-SQL strategy that I know of. So even if we moved to Elastic for this we'd still need "round-trips". SQL would handle this, but at what cost to our infrastructure (managing migrations, how many tables?)
- I'd be willing to bet we _can_ afford to to pay that "round-trip" penalty at some level of degraded performance. The point I want to stress is that this system is for _content management_, which by it's nature is high read-volume (for queries that don't often change) and relatively low write-volume. So a good caching mechanism can make up for a whole host of N+1 problems. If this something we need to address, I'd like to at least see what turning on DAX (and putting our API in a VPC) does for us here.

---

> I like this - I am not sure how exactly it would be done, but it seems like it would be a better experience

This shouldn't be too hard, at least worth playing around with.

---

> Yeah, I've been thinking about this and I'm thinking it might be a case for keeping the attributes (writing them to the primary index as additional entries), at least for references. It may make sense to do the same for the other fields as well, and then just reserve the user defined indexes for multi-column values. I'm going to be looking into this today and will follow-up with you.

My expectation for foreign keys has been that it's identical to our needs for filters, so maybe in your pattern we'd have a `db.index_reference` where the keys are: `{ key: 'posts:content/authors/pedro.md', value: [ 'posts/document_1' ] }`. There would be some limitations there, at first glance we wouldn't be able to say what field the reference is on, so in this example, `post` references `pedro.md`, but we don't know which field (ie. `post.data.author`) so maybe there's a better pattern.

---

package.json
```diff
@@ -36,13 +36,15 @@
     "@typescript-eslint/eslint-plugin": "^2.3.0",
     "@typescript-eslint/parser": "^2.3.0",
     "@yarnpkg/pnpify": "^2.0.0-rc.20",
-    "cypress": "9.2.0",
```
logan-anderson, line 39: Why is cypress removed as a dependency?

jamespohalloran:
Hey! So I was wondering if having to define indexes might be a hurdle for the average user (and potential for issues if they forget about an index). 

I wonder if things might be more foolproof if:
- Suggest a best practice, where users put their queries in a `queries` directory
- We generate a set of indexes based on their defined queries
- Devs could also optionally provide their own set of indexes, and use queries outside of the queries directory (at their own risk of poor performance). 

## What this might look like:

Let's say I want to show my "featured" posts on my homepage

Inside `.tina/queries.gql`, I write the following query:

```gql
query featuredPostsQuery($featured: Boolean!, $date: Datetime!) {
  getPosts(filter: { featured:$featured}, sort: { date: $date }) {
    id
    data {
      ...PostsParts
    }
  }
}
```

This now generates the following in my .tina/__generated__/indexes.ts

```ts
export const indexes = {
  post: [{ date, featured }],
};
```

Now in my page, I can use the above query:

```ts
import {  featuredPostsQuery } from '../.tina/_generated_/types'

export const getStaticProps = async ({ params }) => {
   
    // Down the road, maybe the user just uses a prisma-like client for fetching, instead of importing the query
    const featuredPosts = (await staticRequest({ 
          query: , featuredPosts
    }));
    // ...
})
```

In the above case, I never have to worry about indexing. 

Also, we could provide a separate .tina/indexes.ts (outside of the autogenerated directory), where the advanced users are free to add any explicit indexes that they choose. That way, they can go back to inlining their queries in getStaticProps, they can mess around in altair with custom queries (that may or may not be indexed appropriately), etc. 

Thoughts @kldavis4 @jeffsee55 ?

jeffsee55:
> In the above case, I never have to worry about indexing.

My general feeling is that this concern is premature. We're not sure this will be a problem, and we're even less sure about all the various ways we may be able to solve it. From a technical perspective there are a couple of issues with this:

1. GraphQL variables can be an entire filter:

```graphql
query MyQuery($filter: PostsFilter) {
  getPostsList(filter: $filter) {
    edges {
      cursor
      node {
        data {
          title
        }
      }
    }
  }
}
```
The variables you'd need to provide here at runtime:
```json
{
  "filter": {
    "title": {
      "eq": "Hello"
    }
  }
}
```

2. Allowing filters to be "flat" like this (ie no `index` key or no namespacing) means we can't really tell what index you want, which order should this be in? Because the order of the index matters, would we assume that if `featured` came first, that's the order you want? What if another one of your queries accidentally reversed the order.

jamespohalloran:
> 1. GraphQL variables can be an entire filter:
> 
> ```graphql
> query MyQuery($filter: PostsFilter) {
>   getPostsList(filter: $filter) {
>     edges {
>       cursor
>       node {
>         data {
>           title
>         }
>       }
>     }
>   }
> }
> ```
> 
> The variables you'd need to provide here at runtime:
> 
> ```json
> {
>   "filter": {
>     "title": {
>       "eq": "Hello"
>     }
>   }
> }
> ```
> 

Right, so in your example, you're passing in dynamic filter shapes, however you still need to know what these possible shapes of your filters can be at build-time. If that were me, I'd be scared that I'd miss defining an index, and later change my filter which would break things. However, if you REALLY want to do what you're proposing, you can with my second option, of using the explicit `.tina/indexes.ts`

> Allowing filters to be "flat" like this (ie no index key or no namespacing) means we can't really tell what index you want, which order should this be in? Because the order of the index matters, would we assume that if featured came first, that's the order you want? What if another one of your queries accidentally reversed the order.

I don't exactly know the answer to this, (and I have no idea if my `indexes.ts` data structure makes sense, it's really just a fuzzy example) but if we're also expecting users to take this into consideration while manually defining indexes, then I think that's potentially a pretty significant DX hurdle.

jeffsee55:
> If that were me, I'd be scared that I'd miss defining an index

I'm not sure what the issue is here, the GraphQL API will tell you that you've passed an invalid filter if it's not defined. And if you've got intellisense working (or your in a client like Altair) you'd see errors as you type it.

> we're also expecting users to take this into consideration while manually defining indexes

My feeling is that there's a much more explicit step here, we can give Typescript hints and document it. Whether this is actually harder seems pretty subjective at this stage.

package.json
kldavis4: @logan-anderson for some reason having this in the root package.json was causing the my ide to infer that mocha was the test runner for the @tinacms/graphql and @tinacms/datalayer projects.

I'll put it back before this is taken out of draft mode and I have time to figure out why its screwing up the test runner

kldavis4:
> > I'll look into this - I was using a different library to simulate secondary indexes but had issues and discovered it was super old, so I pivoted to separate files but I think if one of these other libs allows partitioning, a single file should be fine.
> 
> I would expect that we could also just prefix within the same store, too right?

That's correct. It's a little tricky because you need to think about the defaults for gt/lt when they aren't specified, but its still pretty simply

kldavis4:
@jeffsee55 

Just pushed up the following changes:
- Allow filtering on unindexed fields (There is logic which uses the index for filtering if possible, otherwise it fetches the document and performs application layer filtering on the full doc)
- Change 'index' parameter to 'sort'. This is optional - when not specified, items will be sorted according to whatever the natural order is (I think filesystem order). If specified, results are sorted according to the associated index. If sort is specified and the fields being filtered are on the index, the query will execute faster because it doesn't need the full doc for filtering.
- Got rid of the separate leveldb files and now have a single partitioned database
- Support for IN operator and filtering on Referenced objects

I think the main thing that is missing now is nested object filtering which I'm working on now

Here are some sample queries (experimental-examples/tina-cloud-starter + `yarn dev-data`):

**Get posts by author:**
```
        query GetPostsList(
            $filter: PostsFilter!,
            $first: Float,
            $after: String, 
            $last: Float, 
            $before: String, 
            $sort: String) {
          getPostsList(filter: $filter, first: $first, after: $after, last: $last, before: $before, sort: $sort) {
            edges {
              node {
                data {
                  title
                  date
                  published
                  author {
                      ...on AuthorsDocument {
                          id
                      }
                  }
                }
              }
            }
          }
        }

{
    "filter": {
        "author": {
            "authors": {
                "name": {
                    "in": ["Napolean", "Pedro"]
                }
            }
        }
    }
}
```
**published by date:**
```
        query GetPostsList(
            $first: Float,
            $after: String, 
            $last: Float, 
            $before: String, 
            $sort: String, $filter: PostsFilter!) {
          getPostsList(first: $first, after: $after, last: $last, before: $before, sort: $sort, filter: $filter) {
            edges {
              cursor
              node {
                data {
                  title
                  date
                  published
                }
              }
            }
            pageInfo {
                hasNextPage
                endCursor
            }
          }
        }
{
    "filter": {
        "date": {
            "after": "2021-03-12T07:00:00.000Z",
            "before": "2022-08-03T07:00:00.000Z"
        },
        "published": {
            "eq": false
        }
    },
    "sort": "published-by-date"
}
```
(Note this is still keeping sort and filter separate)

**filter on rating sort by date:**
```
        query GetPostsList(
            $first: Float,
            $after: String, 
            $last: Float, 
            $before: String, 
            $sort: String,
            $filter: PostsFilter!) {
          getPostsList(first: $first, after: $after, last: $last, before: $before, sort: $sort, filter: $filter) {
            edges {
              cursor
              node {
                data {
                  title
                  rating
                  date
                  published
                }
              }
            }
            pageInfo {
                hasNextPage
                endCursor
            }
          }
        }
{
    "filter": {
        "rating": {
            "eq": 4
        }
    },
    "sort": "date"
}
```
**filter on title using IN operator**
```
        query GetPostsList(
            $first: Float,
            $after: String, 
            $last: Float, 
            $before: String, 
            $sort: String,
            $filter: PostsFilter!) {
          getPostsList(first: $first, after: $after, last: $last, before: $before, sort: $sort, filter: $filter) {
            edges {
              cursor
              node {
                data {
                  title
                  date
                  published
                }
              }
            }
            pageInfo {
                hasNextPage
                endCursor
            }
          }
        }
{
    "filter": {
        "title": {
            "in": ["Salmon and Moonshine"]
        }
    },
    "sort": "title"
}
```

packages/@tinacms/graphql/src/database/index.ts
```diff
@@ -106,29 +124,41 @@ export class Database {
...
+      collectionIndexDefinitions = indexDefinitions?.[collection.name]
+    }
     if (this.store.supportsSeeding()) {
       await this.bridge.put(filepath, stringifiedFile)
     }
-    await this.store.put(filepath, payload, keepTemplateKey)
+    await this.store.put(filepath, payload, { keepTemplateKey, collection: collection.name, indexDefinitions: collectionIndexDefinitions })
   }
 
   public put = async (filepath: string, data: { [key: string]: unknown }) => {
```
kldavis4, line 142: @jeffsee55 @logan-anderson just calling your attention to the changes I am making here to the put and addPendingDocument functions to lookup the collection and indexDefinitions so they can be passed to the store.put function. Not sure if this is the best way to do this.

jeffsee55: That makes sense for now, hopefully we can remove `addPendingDocument` soon 👍

logan-anderson (approved):

Issue #2542, opened 2022-02-04
Admin catch all route isn't supported on Next.js static exports


jeffsee55:
If you use Next.js [static export functionality](https://nextjs.org/docs/advanced-features/static-html-export), catch all routes _without a `getStaticPaths` function_ will return a 404. This is because a fully static Next.js SSG needs to know all of the paths ahead of time. Since it's not possible to know all routes for the admin at build time, this breaks the usage of `/admin/[[...tina]].tsx`.

What we actually want it be a URL that is wildcarded from Next.js's routing, but seems like that's not supported. Perhaps in these cases we'd need to use a hash router.

In the short term, users who are using the `/admin` for login purposes only can just name their page `/pages/admin/index.jsx` or `/pages/admin.jsx`, but obviously for using the full admin that won't help much.

Note: 
This is only on Netlify
Timebox research to 1 day

**Expected Output:**
* Research/poc suggested path
* If it proves out, a PR will be created otherwise an alternative path will be proposed

**Pull Requests**
* Admin: https://github.com/tinacms/tinacms/pull/2623
* CLI: https://github.com/tinacms/tinacms/pull/2622
* Docs: https://github.com/tinacms/tinacms.org/pull/1248

cc @Enigmatical

gitsstewart:
Having this problem solved helps with several areas of concern beyond just what is stated here.

Enigmatical:
All parts of this issue have been addressed and resolved 👍

PR #2622, opened 2022-02-28
update admin path in CLI

logan-anderson:
Helping @Enigmatical complete the work of supporting static export by updating the admin path in the CLI

I am going to deploy this PR and test it with npx before merging.

Part of the fix for #2542

PR #2623, opened 2022-02-28
2542 - Uses HashRouter for Admin

Enigmatical:
Identified by Jeff, TinaAdmin unfortunately does not work for Next.js static exports due to the use of the wildcard route (`[[...tina]].tsx`).

Fortunately, by using the `HashRouter` we can completely remove the need for the wildcard route and get things working again.

Related to #2542


logan-anderson (approved):

jamespohalloran:
Does this section also need to change? @logan-anderson @Enigmatical 
https://github.com/tinacms/tinacms/blob/feature/2542-changed-to-using-hash-router-for-admin/packages/%40tinacms/cli/src/cmds/init/index.ts#L176-L179

logan-anderson:
> Does this section also need to change? @logan-anderson @Enigmatical https://github.com/tinacms/tinacms/blob/feature/2542-changed-to-using-hash-router-for-admin/packages/%40tinacms/cli/src/cmds/init/index.ts#L176-L179

@jamespohalloran There is a separate PR for the CLI https://github.com/tinacms/tinacms/pull/2622

Issue #2519, opened 2022-01-31
Link to FAQ for common "Content API unavailable" errors


jamespohalloran:
As a developer I want to have some error messaging for common "Content API unavailable" errors so that I can have better information to help me debug the issue. 

**Expected Results:**
- Confirm the below list of potential issues
- Create FAQ for this category of "Content API not available" errors
- The Error is visually displayed for all of the defined error-cases. The error body should link to the FAQ.

**Notes:**

Currently, we have several common usage errors that can lead to an abiguous error such as:

![](https://cdn.discordapp.com/attachments/835172431836479508/937675650360279080/unknown.png)

### Common causes of the GraphQL request failing

- Invalid branch (or undefined branch)
- .tina directory (with generated subdirectory) not pushed to git
- api url is misformatted
- Trying to access localhost graphql in prod (or locally when it's not running). 
- Document doesn't exist

### Where should we display this error?

- On the admin, when `getCollections` fails
- In the admin, in an individual document
- When contextual editing, and on a document

Note: For V1, from the clientside, we should just show this generic error that links to the FAQ
(In the future, we can probably should specific errors for certain specific cases. 
E.g, For the case where the schema isn't pushed to a given branch, the backend could check if there's a schema missing, and give an error with an error code)

jeffsee55:
@jamespohalloran I like it, should we start to think about a list of error codes to help with this? For example, yarn does a nice job of this https://yarnpkg.com/advanced/error-codes

gitsstewart:
Outstanding question:
1. Does adding the link to the new FAQ page belong in the client side (which is only one place in the code) or in the graphQL api everywhere that "Content API Unavailable" is thrown?

jamespohalloran:
I added some notes in the description. For v1, I believe we can handle this all from the clientside

mittonface:
@tinacms/maintainers I'm having a slightly tricky time actually knowing where to find where to the different errors. Not quite sure what I'd search to find the different locations. Any hints?

mittonface:
I wasn't overthinking it. [Changing the error boundary](https://github.com/tinacms/tinacms/commit/43d9f3bf7e451a1d7368a9e514eb8c0f6273729c) is easy enough but only covers a couple of the errors. 

Other errors that you might get just get sent out as a `cms.alert`, which are not really able to handle links. i.e. https://github.com/tinacms/tinacms/blob/main/packages/tinacms/src/admin/components/GetCollections.tsx#L32-L35 
[...]

PR #2629, opened 2022-03-02
2600 - Handle errors better in the CMS

### LINKED ISSUE #2600: Handle errors better in Admin
> As a developer, I would like to capture and share error messages encountered when using Tina's Admin.

We're currently not handling error responses from GraphQL very elegantly in the Admin side of Tina.

At the very least, I would like to leverage the `cms.alerts` and present any `errors` responses to avoid situations like the one @jhuggett recently encountered during dogfooding - where the file was created, but failed when using TinaCloud + DynamoDB Layer:

```
{
  "errors": [
    {
      "message": "Unable to find 'content/post/example.md' in store for project 84e0aba1-502f-49d5-bba4-8e0e652f0cdc: 'jhuggett/barebones-starter-db-test-2', Ref: 'main'",
      "locations": [
        {
          "line": 3,
          "column": 9
        }
      ],
      "path": [
        "createDocument"
      ],
      "extensions": {
        "status": 404
      }
    }
  ],
  "data": null
}
```

Possible candidates for error handling:
- [x] `createDocument`
- [x] `updateDocument`
- [x] `getCollection`
- [x] `getCollections`
- [x] `getDocument`
- [x] `getDocumentFields`

Enigmatical:
Right now, the CMS was built for "happy path" and does not particularly hold up well when it encounters an error from the GraphQL server.

While an enterprising developer can poke through the network log and find the issue, I would like to surface _something_ to the user if at least to get the ball rolling on what might've gone wrong.

Closes #2600

As an aside, I see this as the first iteration of handling errors.

Once the dust settles on the CMS, I would like to pair up with @spbyrne and create something even more useful.  I began to envision an `<ErrorPage />` that would universally appear for any `error` with handy links to Documentation and Community Resources for help but decided to pull back as it felt like scope creep.

The purpose of this PR is to _surface_ the errors outside of the debugging tools.  
Later, we can format and present them better.

Screenshot 2022-03-02 at 15-43-15 Tina


jeffsee55:
Seems like a good fix for now, and yes a more dedicated error page in the future feels like a necessary step. I also wonder how often admin issues would be related to configuration problems that we're surfacing on this ticket https://github.com/tinacms/tinacms/issues/2519#issuecomment-1057328527. I think there's a good chance thats  going to be the common issues we see in admin, and maybe we should leverage the same or similar messaging that links to the FAQ

logan-anderson (approved):

PR #2634, opened 2022-03-03
update language on "unregistered field" placeholder

dwalkr:
The original link now 404s. Additionally, I don't think linking to the PR is necessary any longer.


Enigmatical (approved):
I like it 👍 much clearer.

jeffsee55 (approved):

Issue #2633, opened 2022-03-03
CLI Init: Override dev,start,build scripts instead of creating new scripts.


jamespohalloran:
As a dev I want to have my existing `dev`, `start`, & `build` commands overrided when I use @tinacms/cli so that I don't need to manually update them later.

**Expected Results:**

When a dev runs `@tinacms/cli init`, we should take their existing `dev`,`build`, & `start` scripts, and build overrided scripts from their values.

E.g:
```diff
 -   "dev": "next dev PORT=3006",
 -   "build": "next build",
 -   "start": "next start"
 +   "tina-dev": "yarn tinacms server:start -c \"next dev PORT=3006\"",
 +   "tina-build": "yarn tinacms server:start -c \"next build\"",
 +   "tina-start": "yarn tinacms server:start -c \"next start\""
``` 

**Notes:**

## Background

Currently, when you use `@tinacms/cli init`, it creates 3 new scripts:

```
    "tina-dev": "yarn tinacms server:start -c \"next dev\"",
    "tina-build": "yarn tinacms server:start -c \"next build\"",
    "tina-start": "yarn tinacms server:start -c \"next start\""
``` 

We originally opted to create new scripts because it was safer rather than overriding the user's existing dev,build,start scripts, (incase they already were doing something custom.)

## Motivation

Our current approach causes a few problems:
- Devs are very used to using specific commands to run their site, and will often forget to run `yarn tina-dev` instead of `yarn dev`.
- When deploying to Vercel, the dev needs to manually ovverride the default build command.

PR #2639, opened 2022-03-04
feat: override existing scripts with cli init

jamespohalloran:
fixes [#2633](https://github.com/tinacms/tinacms/issues/2633)

Overrides existing script names instead of introducing new `tina-*` scripts

Docs here: https://github.com/tinacms/tinacms.org/pull/1251

### DIFF
--- a/packages/@tinacms/cli/src/utils/script-helpers.test.ts
+++ b/packages/@tinacms/cli/src/utils/script-helpers.test.ts
@@ -1,0 +1,62 @@
+/**
+
+Copyright 2021 Forestry.io Holdings, Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+
+*/
+
+import { generateGqlScript, extendNextScripts } from './script-helpers'
+
+describe('generateGqlScript', () => {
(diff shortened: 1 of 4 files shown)


packages/@tinacms/cli/src/utils/script-helpers.test.ts
```diff
@@ -0,0 +1,62 @@
...
+
+*/
+
+import { generateGqlScript, extendNextScripts } from './script-helpers'
+
+describe('generateGqlScript', () => {
+  it('wraps original script correctly', () => {
+    const newScript = generateGqlScript('yarn dev -p 3000')
+
+    expect(newScript).toEqual('yarn tinacms server:start -c "yarn dev -p 3000"')
```
logan-anderson, line 25: We probably do not want to have "yarn in here" only tinacms server:start -c "what was in there before".

The  `yarn` is not needed

logan-anderson (changes requested):

packages/@tinacms/cli/src/utils/script-helpers.test.ts
jamespohalloran: Ah good catch, just removed it

packages/@tinacms/cli/src/utils/script-helpers.test.ts
```diff
@@ -0,0 +1,62 @@
...
+
+*/
+
+import { generateGqlScript, extendNextScripts } from './script-helpers'
+
+describe('generateGqlScript', () => {
+  it('wraps original script correctly', () => {
+    const newScript = generateGqlScript('next dev -p 3000')
+
+    expect(newScript).toEqual('tinacms server:start -c "next dev -p 3000"')
```
logan-anderson, line 25: We also might beed to add a `\` char in front of `"`. I _think_ that this will evaluate in JSON to

`"dev": "'tinacms server:start -c "next dev -p 3000""`


I will do a PR deploy and test 


EDIT: nvm it works as expected!

logan-anderson (approved):

Issue #2636, opened 2022-03-04
Missing new line (<br />) in MDX


zeropaper:
## Description

In Markdown (and MDX), it should be possible to make HTML `<BR />` by adding multiple space characters (` `) before a new line (`\n`) but it doesn't seem possible to do so with Tina.

### Steps to reproduce

In a Markdown / MDX file, add something like

```md
first line  
second line
```

### Expected result

The resulting AST should be something like:
```js
{
  "type": "root",
  "children": [
    // ...

    {
      "type": "p",
      "children": [
        {
          "type": "text",
          "text": "first line"
        },
        {
          "type": "break",
        },
        {
          "type": "text",
          "text": "second line"
        },
      ]
    },
   // ...
```

The rendered HTML should contain: `<br />`

### Actual result

It's impossible to have a line break from Markdown / MDX content.

#### Tinacms Package Versions

```text
├── @tinacms/auth@0.50.2
├── @tinacms/cli@0.60.5
├── next-tinacms-cloudinary@3.5.8
├── react-tinacms-editor@0.53.5
├── tinacms@0.66.4
```

#### Tools

| Tool    | Version |
| ------- | ------- |
| browser |   N/R    |
| os      |  linux |
| node    |     v16.6.2    |
| npm     |    8.4.0     |

zeropaper:
In this case, the markdown is correct, however, the editor is not properly handling.
![image](https://user-images.githubusercontent.com/65971/156756449-f65ddf43-0eef-4c0a-9170-78f1b3caa231.png)

In that case, the markdown was "edited" with the editor (adding a few "spaces" and "SHIFT + RETURN" - as I imagine the most common), saved and its result in the markdown file is... "strange"
![image](https://user-images.githubusercontent.com/65971/156756926-3f05c86f-23d0-4b6a-8df5-c86b1b912296.png)

PR #2652, opened 2022-03-05
WIP: properly handle markdown and editor line breaks

zeropaper:
Related to #2636 2636

### DIFF
--- a/packages/@tinacms/graphql/src/mdx/parse.ts
+++ b/packages/@tinacms/graphql/src/mdx/parse.ts
@@ -26,7 +26,7 @@ import type { RichTypeInner } from '../types'
 import { isNull } from 'lodash'
 
 export const parseMDX = (value: string, field: RichTypeInner) => {
-  const tree = unified().use(markdown).use(mdx).parse(value)
+  const tree = unified().use(markdown, { commonmark: true }).use(mdx).parse(value)
   return parseMDXInner(tree, field)
 }
 /**
@@ -364,6 +364,7 @@ export interface NodeTypes {
   delete_mark: string
   inline_code_mark: string
   thematic_break: string
+  hard_break: string
 }
 
 export type SlateNodeType =
@@ -435,8 +436,11 @@ export type SlateNodeType =
       caption: string
     }
   | {
-      type: 'thematic_break'
-    }
+    type: 'thematic_break'
+  }
+  | {
+    type: 'hard_break'
+  }
 //   block_quote: 'block_quote',
 //   code_block: 'code_block',
 //   link: 'link',
@@ -511,6 +515,7 @@ export const plateElements = {
   ELEMENT_H5: 'h5',
   ELEMENT_H6: 'h6',
   ELEMENT_HR: 'hr',
+  ELEMENT_BR: 'br',
   ELEMENT_ALIGN_CENTER: 'align_center',
   ELEMENT_ALIGN_JUSTIFY: 'align_justify',
   ELEMENT_ALIGN_LEFT: 'align_left',
@@ -561,6 +566,7 @@ export const defaultNodeTypes: NodeTypes = {
   delete_mark: plateElements.MARK_STRIKETHROUGH,
   inline_code_mark: plateElements.ELEMENT_CODE_LINE,
   thematic_break: plateElements.ELEMENT_HR,
+  hard_break: plateElements.ELEMENT_BR,
   image: plateElements.ELEMENT_IMAGE,
 }
 
@@ -672,6 +678,12 @@ export default function remarkToSlate(node: MdxAstNode) {
     case 'thematicBreak':
       return {
         type: types.thematic_break,
+        children: [{ type: 'text', text: '' }],
+      }
+    case 'break':
+    // case 'hardBreak':
+      return {
+        type: types.hard_break,
         children: [{ type: 'text', text: '' }],
       }
     case 'text':
--- a/packages/@tinacms/toolkit/src/packages/fields/plugins/MdxFieldPlugin/updated/plate/plugins/core/autoformat/autoformatBlocks.ts
+++ b/packages/@tinacms/toolkit/src/packages/fields/plugins/MdxFieldPlugin/updated/plate/plugins/core/autoformat/autoformatBlocks.ts
@@ -21,6 +21,7 @@ import {
   ELEMENT_CODE_BLOCK,
 } from '@udecode/plate-code-block'
 import { ELEMENT_HR } from '@udecode/plate-horizontal-rule'
+import { ELEMENT_BR } from '@udecode/plate-break'
 import { ELEMENT_BLOCKQUOTE } from '@udecode/plate-block-quote'
 import {
   ELEMENT_H1,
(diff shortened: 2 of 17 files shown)


logan-anderson:
Hello @zeropaper, thank you for the contribution. I see this is still a draft PR is there any questions you have? or is there any support you need on our end?

Thanks again.

zeropaper:
Hi @logan-anderson  :)

Thanks for asking!
The issue seems somewhat... complex.

At some point(s), between the parsing of the markdownfile, the rendering in the RTE, the parsing / conversion of the RTE value to markdown and the generation of the markdown to be written in files... something goes bonkers...

I wrote some reliable Cypress tests (that will correctly fail) and should better illustrate what I mean.
If you want to give it a shot, check that branch out and run Cypress....
(https://github.com/tinacms/tinacms/pull/2652/commits/671c2f2d4b027b9c2badc4687169217ed198ab02 contains some instructions on how to run the E2E tests - as I couldn't find any - it also has some useful Cypress "commands" and "tasks" for future test development).

So... Yeah, any help (hints, feedback, ...) would be welcome. :)

packages/@tinacms/graphql/src/mdx/parse.ts
```diff
@@ -26,7 +26,7 @@ import type { RichTypeInner } from '../types'
 import { isNull } from 'lodash'
 
 export const parseMDX = (value: string, field: RichTypeInner) => {
-  const tree = unified().use(markdown).use(mdx).parse(value)
+  const tree = unified().use(markdown, { commonmark: true }).use(mdx).parse(value)
```
logan-anderson, line 29: I was getting a type error here in the editor. I am not 100% sure if this is doing what you want.

logan-anderson, line 29: I was getting a type error here in the editor. I am not 100% sure if this is doing what you want.

packages/@tinacms/toolkit/src/packages/fields/plugins/MdxFieldPlugin/updated/plate/plugins/core/autoformat/autoformatBlocks.ts
```diff
@@ -21,6 +21,7 @@ import {
   ELEMENT_CODE_BLOCK,
 } from '@udecode/plate-code-block'
 import { ELEMENT_HR } from '@udecode/plate-horizontal-rule'
+import { ELEMENT_BR } from '@udecode/plate-break'
```
logan-anderson, line 24: I was also getting a type error here as well. Using `"br"` directly seemed to have better luck for me (although might not be totally correct).

logan-anderson, line 24: I was also getting a type error here as well. Using `"br"` directly seemed to have better luck for me (although might not be totally correct).

packages/@tinacms/toolkit/src/packages/fields/plugins/MdxFieldPlugin/updated/plate/plugins/core/autoformat/autoformatBlocks.ts
```diff
@@ -109,4 +110,17 @@ export const autoformatBlocks: AutoformatRule[] = [
       })
     },
   },
+  // that probably should not be here...
+  {
+    mode: 'inline',
```
logan-anderson, line 115: I was also getting a type error here. What promoted you to use `mode: 'inline'`

packages/@tinacms/toolkit/src/packages/fields/plugins/MdxFieldPlugin/updated/plate/plugins/ui/components.tsx
```diff
@@ -18,6 +18,7 @@ limitations under the License.
 
 import React from 'react'
 import { ELEMENT_HR } from '@udecode/plate-horizontal-rule'
+import { ELEMENT_BR } from '@udecode/plate-break'
```
logan-anderson, line 21: I was getting a type error here and could not find any docs about using this package like this. I had better luck using the `"br"` string directly (although this might not be totally correct either )

logan-anderson (commented):
It feels like you are really close to the solution, as far as I can tell almost everything is correct and I really like the re-work of the tests! That is much more clean than what we are doing right now. That would be worth merging alone if we can't figure out this line-break issue.

packages/@tinacms/toolkit/src/packages/fields/plugins/MdxFieldPlugin/updated/plate/plugins/core/autoformat/autoformatBlocks.ts
zeropaper: just poking around TBH... semantically speaking `inline` here seems more adequate since it is not like a paragraph (and 
contains something / has children).  
But then... It's probably not the place it should be.

jeffsee55:
@zeropaper just catching up on this, this actually isn't supported from the plugin library we're using on top of slate ([see comment](https://github.com/udecode/plate/discussions/501)). While this would be fine to add as our own slate logic, it explains some of the missing elements from `@udecode/plate-break` we're seeing. 

> In Markdown (and MDX), it should be possible to make HTML <BR /> by adding multiple space characters ( ) before a new line (\n) but it doesn't seem possible to do so with Tina.

One question I have about what you'd expect the behavior to be, inserting spaces isn't something an editor would really know about, and right now that logic will insert a `.`, much like most rich-text editors do. For that reason I'm inclined to want to leave this feature out, curious what that would prevent you from accomplishing?

Of course if you hit `shift+enter` that will create a newline `\n`, and this works as expected from my testing, though our `<TinaMarkdown>` will need the `<p>` to be overridden:

```js
<TinaMarkdown components={{
  p: (props) => <p style={{whiteSpace: "break-spaces"}} {...props} />
}} content={data._body} />
```

zeropaper:
Hi @jeffsee55,

I get why you would accept that as a "feature" rather than a bug (it's complicated to fix because it lies in a dependency and many other users won't notice a thing).

Unfortunately, I don't care much (in that particular case) about UX and the CSS workaround is not going to float in my case (I need the markdown to be valid / standard / compatible with other "editors" and "renderers").. 

Right now, I don't have the strength or patience to look at the RTE library that Tina uses... but it may happen.

If in the meantime you want the stuff that I wrote for the Cypress tests, feel free to cherry-pick it :)

jeffsee55:
@zeropaper you brought up some good points, after discussing internally it feels like the `<br>` tag should be the default soft-break behavior. I've re-opened a PR [here](https://github.com/tinacms/tinacms/pull/2724) and added your work as a commit (not sure if it matters, but only was able to use your email I found in the git log as the author).

Notably, I had to skip [one of your tests](https://github.com/tinacms/tinacms/pull/2724/files#diff-205dbaa814e14382091e8f7f69adca13b80cfb7e95e4e46549ce932b481c2446R90-R103), the parser automatically removes empty blocks before sending it through the API. Let me know if you think that's an error we need to address for some reason.

zeropaper:
@jeffsee55 sounds good to me :)

sorry I couldn't get fixed myself... but happy if was of any help.

I'm closing that PR then, cheers!

PR #2658, opened 2022-03-07
2657 - Checks isAuthenticated() before making requests to GraphQL

### LINKED ISSUE #2657: GetCollection Error appearing during login with TinaCloud
Thanks to @scottgallant for discovering this:

https://www.loom.com/share/6faa25b0c93849d1a0a31c429aa0553f

There is a brief period between the `cms` existing and the user is authenticated where an errant `GetCollections` call is being made (probably to render the Global Nav).  This results in an `error` that could be misleading to a user who might assume their configuration is incorrect (it isn't).

Locally, `cms` and `isAuthenticated()` both exist at nearly the same time.  For `TinaCloud` (and production, in general), we need to check for both `cms` and `cms.api.tina.isAuthenticated()` before making any request.

Enigmatical:
This was discovered during some dogfooding that the Global Nav attempts to retrieve the Collections list (via `GetCollections()`) _before_ the user is authenticated resulting in a misleading error message.

I've pulled `isAuthenticated()` into `TinaAdminApi` where it is simply calling `cms.api.tina.isAuthenticated()`.  I've wrapped the hooks in a check against that.

I also _fixed_ a bug with the loading animation not appearing (one I introduced, whoops).

Closes #2657

### DIFF
--- a/.changeset/sixty-teachers-lie.md
+++ b/.changeset/sixty-teachers-lie.md
@@ -1,0 +1,6 @@
+---
+'@tinacms/toolkit': patch
+'tinacms': patch
+---
+
+Checks isAuthenticated() before making requests to the GraphQL client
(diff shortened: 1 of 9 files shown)


kldavis4 (approved):
👍🏻

logan-anderson (approved):
LGTM

PR #2663, opened 2022-03-08
Allow dot (".") in file names

### LINKED ISSUE #2653: Allow dot in file names in Tina Admin
## Description

The way we do localization is by having the locale appended to the filename (some-article.en.md). But in Tina admin it's now not allowed to include have a dot (.) in the filename when creating a new document

### Steps to reproduce

Open Tina Admin try adding a (.) to a file name.

### Expected result

Allow (.) in filenames

### Actual result

Not allow to use dots.

logan-anderson:
fixes #2653


Enigmatical (commented):
I'm okay with this assuming @jeffsee55 is.  The regex I was using on the front-end for `relativePath` was based on the same regex being used in the backend.  I suspect we'd need to make a change there, too?

jeffsee55:
Unfortunately we don't have any test coverage of this on the backend but as long as `path.extname` works fine this seems ok https://github.com/tinacms/tinacms/blob/9a2fc83e195dfcb33f49ba0caa9017041b2eebdd/packages/@tinacms/graphql/src/database/index.ts#L186

logan-anderson:
> Unfortunately we don't have any test coverage of this on the backend but as long as `path.extname` works fine this seems ok
> 
> https://github.com/tinacms/tinacms/blob/9a2fc83e195dfcb33f49ba0caa9017041b2eebdd/packages/@tinacms/graphql/src/database/index.ts#L186

@jeffsee55 this is the GQL resoponce. Seems like it is working
```
"sys": {
              "filename": "test.en",
              "basename": "test.en.md",
              "relativePath": "test.en.md",
              "extension": ".md"
            }
```

jeffsee55 (approved):

PR #2666, opened 2022-03-09
Fix/z index issue

logan-anderson:
Fixes a bug where if you had an object (list false) and then another object as a child (list true) you would not be able to click on any of the templates.

See https://www.loom.com/share/06404c49de004177b5d19a3c2e0ab45f


Enigmatical (approved):
I like it 👍

PR #2678, opened 2022-03-10
Batch of console.error fixes

jeffsee55:
We've been getting a few messages in the console about these, most are benign but the one that does error in dev is the `error` message being an object.

The biggest change here is to `react-beautiful-dnd` which has gone from 11 to 13, though I haven't run into any issues, and from reading through them it doesn't look like we're relying on things that have broken https://github.com/atlassian/react-beautiful-dnd/releases.

### DIFF
--- a/packages/@tinacms/toolkit/src/packages/form-builder/fields-builder.tsx
+++ b/packages/@tinacms/toolkit/src/packages/form-builder/fields-builder.tsx
@@ -44,7 +44,12 @@ export function FieldsBuilder({ form, fields }: FieldsBuilderProps) {
     // @ts-ignore FIXME twind
     <FieldsGroup>
       {fields.map((field: Field) => (
-        <InnerField field={field} form={form} fieldPlugins={fieldPlugins} />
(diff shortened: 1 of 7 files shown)


packages/@tinacms/toolkit/src/packages/form-builder/fields-builder.tsx
```diff
@@ -44,7 +44,12 @@ export function FieldsBuilder({ form, fields }: FieldsBuilderProps) {
     // @ts-ignore FIXME twind
     <FieldsGroup>
       {fields.map((field: Field) => (
-        <InnerField field={field} form={form} fieldPlugins={fieldPlugins} />
+        <InnerField
+          key={field.name}
```
Enigmatical, line 48: Glad to see this; I had the same change in my own PR (to clean up the `console`) 👍

Enigmatical (approved):
LGTM and I'm all for anything that silences those `react-beautiful-dnd` error messages 🎉

PR #2693, opened 2022-03-15
Feat/move schema to frontend

logan-anderson:
~~99% of the work is done here. Just testing a few things before I open it up.~~

Here is what this work allows the user to do: https://www.loom.com/share/1af3099087e54a4395b462d9d4d53b37


logan-anderson:
@Enigmatical i had to add another option to the drop down so we do no re introduce #2440

jamespohalloran:
This stuff looks awesome! 

We briefly talked about this during today's meeting, but just wondering more about the promoted pattern for exporting both the schema and config together. 
If we want that to be the approach, does `defineConfig` need to change to include the schema properties?

E.g: 
```
// schema.ts
export default defineConfig({
  collections: [ /**/],
  apiURL,
  mediaStore: async () => {
        // ...
  }
})
```

packages/@tinacms/cli/package.json
```diff
@@ -60,6 +60,7 @@
     "@tinacms/datalayer": "workspace:*",
     "@tinacms/graphql": "workspace:*",
     "@tinacms/metrics": "workspace:*",
+    "@tinacms/schema-tools": "workspace:*",
```
jeffsee55, line 63: Is this dependency necessary?

packages/@tinacms/schema-tools/src/schema/TinaSchema.ts
```diff
@@ -0,0 +1,273 @@
...
+      (template) => template.name === templateName
+    )
+    if (!globalTemplate) {
+      throw new Error(
+        `Expected to find global template of name ${templateName}`
+      )
+    }
+    return globalTemplate
+  }
+  public getCollectionByFullPath = async (filepath: string) => {
```
jeffsee55, line 112: Not a blocker, but this doesn't need to be async anymore.

packages/@tinacms/schema-tools/package.json
```diff
@@ -0,0 +1,43 @@
...
+  },
+  "devDependencies": {
+    "@tinacms/scripts": "workspace:*",
+    "@types/yup": "^0.29.10",
+    "jest": "^27.0.6",
+    "react": "16.14.0",
+    "typescript": "^4.3.5"
+  },
+  "peerDependencies": {
+    "react": ">=16.14.0",
```
jeffsee55, line 26: I wouldn't think we need this as a peer dep or devDep, I don't see any usage of React in here and if/when we use this package inside `@tinacms/graphql` it won't meet this requirement.

jeffsee55 (commented):
Looks great! The only thing we'll want to make sure we do is circle back and replace the schema that's in the GraphQL package https://github.com/tinacms/tinacms/blob/4d2215b36d967698e75fc0aed9c4aad0eaf67b99/packages/@tinacms/graphql/src/schema/index.ts#L1

packages/@tinacms/schema-tools/src/schema/resolveField.ts
```diff
@@ -0,0 +1,183 @@
...
+          templates[lastItem(template.namespace)] = {
+            // @ts-ignore FIXME `Templateable` should have name and label properties
+            label: template.label || templateName,
+            key: templateName,
+            inline: template.inline,
+            name: templateName,
+            // fields: await sequential(
+            //   template.fields,
+            //   async (field) => await resolveField(field, schema)
+            // ),
```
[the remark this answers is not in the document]
logan-anderson: Yup!

packages/@tinacms/schema-tools/package.json
logan-anderson: It is only needed for the `FC` type

packages/tinacms/src/utils/index.ts
```diff
@@ -22,20 +23,24 @@ export interface CreateClientProps {
...
+  schema?: TinaCloudSchema<false>
 }
 export const createClient = ({
   clientId,
   isLocalClient = true,
   branch,
   tinaioConfig,
+  schema,
 }: CreateClientProps) => {
+  console.log({ schema })
```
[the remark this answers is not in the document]
logan-anderson: Good catch

packages/@tinacms/cli/package.json
logan-anderson: Removed

logan-anderson:
> This stuff looks awesome!
> 
> We briefly talked about this during today's meeting, but just wondering more about the promoted pattern for exporting both the schema and config together. If we want that to be the approach, does `defineConfig` need to change to include the schema properties?
> 
> E.g:
> 
> ```
> // schema.ts
> export default defineConfig({
>   collections: [ /**/],
>   apiURL,
>   mediaStore: async () => {
>         // ...
>   }
> })
> ```
@jamespohalloran The schema is still needed to be a default export (its still used on the backend) but this can be done (and works)
```ts
const schema = defineSchema({...})

export const config = defineConfig({
   schema: schema,
   apiURL: apiRIL,
   mediaStore: //...
})

export default schema 
```

packages/@tinacms/schema-tools/package.json
jeffsee55: Ok, when we fixup our types we can probably keep this over in `tinacms` and have the `ui` portion of these types be inserted via generic.

jeffsee55 (approved):

PR #2700, opened 2022-03-16
Uses async function to get reference options

Enigmatical:
As part of the "Extending Tina" effort, `reference` field is not responsible for retrieving its own `options` (rather than expecting them from `field.options`).

I also took the opportunity to simplify and split apart `Reference` into multiple components:  `Reference`, `ReferenceSelect`, and `ReferenceLink`.

Closes https://github.com/tinacms/product-roadmap/issues/25


logan-anderson:
@Enigmatical would it be easier to make this a PR into `feat/move-schema-to-frontend` that way we can test them together and merge them together or if we want to merge this into `main` **first** that would be ok too!

packages/@tinacms/toolkit/src/packages/fields/components/Reference/ReferenceSelect.tsx
```diff
@@ -0,0 +1,142 @@
...
+      <select
+        id={input.name}
+        value={input.value}
+        onChange={input.onChange}
+        className={selectFieldClasses}
+        {...input}
+      >
+        {optionSets.length > 0 &&
+          optionSets.map(({ collection, edges }: OptionSet) => (
+            <optgroup key={`${collection}-group`} label={collection}>
```
jeffsee55, line 128: Slick 👍 !

jeffsee55 (approved):

PR #2718, opened 2022-03-21
2707 - Allows Documents to be created with subfolders in the filename

### LINKED ISSUE #2707: Allow Documents to be created within subfolders via the CMS
This feature is currently blocked by a bug with the GraphQL Client and `breadcrumbs`:
https://github.com/tinacms/tinacms/issues/2708

## Summary

As a Tina User, I want to be able to create documents nested within subfolders under my collection's `path` (for example, `posts` is my Blog Post collection, but I would like to have my recipe `posts` stored in `posts/recipes/new-favorite-recipe.mdx`) so that I have better control over how my content is organized.

### Expected Results

* Within the CMS, on the Create Document page, modify the `filename` (`relativePath`) field to allow for `/` characters.
  * Modify the `regex` to allow for `/` characters and update the invalid message to include `/` character.
  * For `Example`, include more examples: `My_Document`, `subfolder/My_Document`, and `My_Document.en`
* Within the CMS, on the Document List page, include the subfolder(s) as part of the `filename`.
  * Will need left-side overflow ellipsis so that `filename` is always the most visible part.
* Within the CMS, on the Document List page, ensure all routing accounts for the subfolders + filename (currently does not).
* Will also want to make sure subfolder(s) + filename works with `RouteMapper` (it should, but just test and confirm).
Screen Shot 2022-03-21 at 2 06 33 PM

### Notes

Initially, I purposefully restricted the `filename` to match the allowed pattern for `relativePath`, but lately, we've been relaxing those criteria and allowing the user more control over the `filename` (i.e. allow for `.` so that localization can be included).  I feel this is another step in that direction, but the CMS was not built for handling subfolder routes.

This is a feature request from Discord users.

Enigmatical:
Brought up in a Discord discussion, we currently do not support Documents nested within subfolders within the CMS.  The documents can be queried, but creating and updating via the CMS fails.

This PR allows for a `/` to be used in the `filename` field and accounts for nested documents by using the document's `breadcrumbs` (which also needed to be fixed).

Screen Shot 2022-03-21 at 2 06 33 PM

Closes #2707 

Note:  This is not a _complete_ solution for Documents and subfolders.  Ideally, we would want to update the Global Nav to have a more "tree"-like UI where subfolders under each collection would be clickable to filter down the Document List.

For now, this merely adds support for creating and updating nested documents via the CMS.


packages/tinacms/src/admin/index.tsx
```diff
@@ -78,7 +78,7 @@ export const TinaAdmin = () => {
                         element={<CollectionCreatePage />}
                       />
                       <Route
-                        path="collections/:collectionName/:filename"
+                        path="collections/:collectionName/*"
```
Enigmatical, line 81: In previous version of `react-router`, you could create wildcard routes like this:
```
path="collections/:collectionName/:filename+"
```
where everything at the end of the match would be associated with `filename`.

In `react-router` v6, that does not appear to work anymore.  Instead, you can use `*` to capture "everything else" into an object:
```
const {collectionName, ...rest} = useParams()
```

packages/tinacms/src/admin/pages/CollectionListPage.tsx
```diff
@@ -149,9 +149,13 @@ const CollectionListPage = () => {
...
                             <tbody className="divide-y divide-gray-150">
                               {documents.map((document) => {
+                                const subfolders = document.node.sys.breadcrumbs
+                                  .slice(0, -1)
+                                  .join('/')
+
                                 return (
                                   <tr
-                                    key={`document-${document.node.sys.filename}`}
+                                    key={`document-${document.node.sys.relativePath}`}
```
Enigmatical, line 158: More unique in situations where you have two Documents of the same name, but in different folders.

jeffsee55 (approved):

PR #2719, opened 2022-03-21
Formify fixes

jeffsee55:
Most changes are due to moving the tests. The relevant commits are:

- [Add parentTypename to field builders](https://github.com/tinacms/tinacms/commit/612c3194e2aafe25f1a538647c095c40c576d18f)
- [Cleanup formify logic, rely on parentTypename to disambiguate field e…](https://github.com/tinacms/tinacms/commit/9e9da051f2b521304c6c8e259473293b1af69ab9)

Most notably this adds a new field on the fields builder (working both client-side and via GraphQL until we get rid of the backend form generator logic). The field is called `parentTypename` and it tells us what we need to know to complete the picture on the event-driven system. One of the pitfalls to this approach was that we didn't know 100% for sure what kind of change the event belongs to. This presented a problem when 2 blocks at the same depth shared a common field _that was a list field_. So if we had a `blocks` field which was polymorphic and one of the templates had `items` which was a list of features, and another one had `items` which was a list of slideshow images, the _insert_ event on `blocks.0.items` would push to both sections of the payload. So `parentTypename` is used to disambiguate the 2 ([see here](https://github.com/tinacms/tinacms/blob/9e9da051f2b521304c6c8e259473293b1af69ab9/packages/tinacms/src/hooks/formify/util.ts#L401-L417)). This use-case is captured in [this test case](https://github.com/tinacms/tinacms/blob/f02bf75b7635a29f397790b14e9badaeae09f74f/packages/tinacms/src/hooks/formify/spec/18-matching-field-signatures/index.spec.ts)


logan-anderson (approved):
I am going to put a ✅  on this assuming we are testing it with the @beta tag tomorrow.

Also, does allow the new schema work to be used with the new `formify` stuff. I don't think I added that  in my original PR

jeffsee55:
@logan-anderson yes it does add that [here](https://github.com/tinacms/tinacms/pull/2719/files#diff-5e6a39a698c838d3f3d59267f8338cf63e5e14d4676ff1569d6521759fec95f2R212)

jeffsee55:
Merging this, deep-dive walkthrough coming soon!

PR #2721, opened 2022-03-21
2590 - Improves no TinaEditProvider error message

### LINKED ISSUE #2590: Improve "Error: No `TinaEditProvider` found" error message
As a dev (user) I want clearer instructions on what to do next when the Tina wrapper is missing so that I can quickly experience Tina.

Expected Results:

- Our error message should note the [new wrapper](https://github.com/tinacms/tinacms/pull/2549) (User doesn't need to add TinaEditProvider anymore, they need '../.tina/components/TinaProvider')
- We should link to our setup doc: https://tina.io/docs/introduction/tina-init/#adding-tina  

## Description

If you run `tina init`, and navigate to the admin (without adding the TinaCMS container to the _app layout), you currently get this page:

Screen Shot 2022-02-15 at 12 58 02 PM

I could see this being a common error for new users. We should give the user a bit more guidance here, and link to the docs (either a docs page specifically for this error, or the general onboarding docs).

Also note, with [this change](https://github.com/tinacms/tinacms/pull/2549), the user doesn't need to include the TinaEditProvider directly, it will be the TinaProvider from their .tina directory

#### Tools

```
├── @tinacms/cli@0.60.0
└── tinacms@0.66.1 invalid: "^0.65.3" 
```

Enigmatical:
A small, but (potentially) significant change to the error message presented when TinaCMS is not found in the _app layout.

This updates the language to reference the new `TinaProvider` pattern and a link to the `Adding Tina` part of the setup docs:

```
Unable to find `TinaProvider`; did you forget to add the TinaCMS container to your _app layout?  See our setup docs: https://tina.io/docs/introduction/tina-init/#adding-tina
```

Closes #2590

### DIFF
--- a/packages/@tinacms/sharedctx/src/edit-state-ctx.tsx
+++ b/packages/@tinacms/sharedctx/src/edit-state-ctx.tsx
@@ -20,9 +20,10 @@ const isSSR = typeof window === 'undefined'
 
 export const isEditing = (): boolean => {
   if (!isSSR) {
-    // localStorage may be intentionally disabled for site visitors, 
+    // localStorage may be intentionally disabled for site visitors,
     // in that case Tina can't be used so just return false
-    const isEdit = window.localStorage && window.localStorage.getItem(LOCALSTORAGEKEY)
+    const isEdit =
+      window.localStorage && window.localStorage.getItem(LOCALSTORAGEKEY)
     return isEdit && isEdit === 'true'
   }
   // assume not editing if SSR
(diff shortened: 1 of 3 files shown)


packages/@tinacms/sharedctx/src/edit-state-ctx.tsx
```diff
@@ -79,7 +80,9 @@ export const EditProvider: React.FC = ({ children }) => {
 export const useEditState = () => {
   const { edit, setEdit } = useContext(EditContext)
   if (!setEdit) {
-    throw new Error('No `TinaEditProvider` found')
+    throw new Error(
+      'Unable to find `TinaProvider`; did you forget to add the TinaCMS container to your _app layout?  See our setup docs: https://tina.io/docs/introduction/tina-init/#adding-tina'
```
jeffsee55, line 84: Only nitpick is that `_app` is a Next.js-specific thing, we may forget to amend this later on.

logan-anderson (dismissed):

packages/@tinacms/sharedctx/src/edit-state-ctx.tsx
Enigmatical: Talked with @jeffsee55 about this and we agreed to replace `_app` with `app root` to remove the Next.js-specific language 👍

jeffsee55 (approved):

PR #2722, opened 2022-03-21
Check for appropriate versions of react and react-dom before initializing with tinacms init

jeffsee55:
- [ ] Works on Yarn v2
- [ ] Works on Yarn v1


mittonface (dismissed):
It worked for me!
```
➜  nextjs-blog git:(main) ✗ npx @tinacms/cli@0.0.0-202222413501 init

  Error: Unable to initialize Tina due to outdated dependencies, try upgrading the following packages:
      "react@>=16.14.0"
      "react-dom@>=16.14.0"

  Then re-rerun "@tinacms/cli init"
  ```
  
My only question would be, is there a place we could more globally define the React major/minor version so that when we bump up the minimum supported version we don't need to remember to update this?

jeffsee55:
> is there a place we could more globally define the React major/minor version

Good question, I added a test thats... kinda awkward but might be good enough. We're really saying that the check here should satisfy the peer dependency that we have on the `tinacms` package. So whenever we bump that the test should fail.

logan-anderson (approved):

PR #2724, opened 2022-03-22
Add support for br

jeffsee55:
Closes https://github.com/tinacms/tinacms/issues/2636

When inserting a soft-break (`shift+enter`), this will now result in a `<br>` tag being inserted. Note that this will save the markdown with a backslash to indicate line break (instead of multiple empty spaces):

```markdown
#### Address:
123 Abc St\
Charlottetown, PEI
```

I wasn't able to easily pull back on some of the other stuff in the PR, but credit here is to [zeropaper](https://github.com/zeropaper) from this PR https://github.com/tinacms/tinacms/pull/2652. Thanks!


spbyrne:
Works for me. Tested with a file that uses the double spacing format and it rendered correctly in the editor & page, and when I save the document with new changes it saves existing double-space breaks with `\` instead. New line breaks with shift + enter work great.

spbyrne (approved):
Seems like a good solution, confirmed it's working as expected 👍

PR #2728, opened 2022-03-24
Feat/prettier

logan-anderson:
Wondering if we should unify our code formate  to accommodate for those using different editors or not using the format on save?

If we pull this in we should also merge [this PR](https://github.com/tinacms/tinacms/pull/2730) to unify all the formats


Enigmatical:
FWIW, I legit thought we _were_ using `.prettier`.  It has been such a staple in all of my development repos that I only just now realized we don't have a `.prettier.config` in `tinacms`.

jeffsee55 (approved):

PR #2738, opened 2022-03-25
Prevent reset error on formify

jeffsee55:
Resolving fields wasn't getting the right even shape when doing a form reset. It was also not running a clean restart when the query changed, only when using the new `useTina` hook.


Enigmatical (changes requested):
Outside of the lingering `log`, this looks good 👍

PR #2748, opened 2022-03-29
update Database.indexContentByPaths to not require collection parameter

kldavis4:
In the initial datalayer work for querying, the indexContentByPaths method was modified to take a Collection. This isn't available to all callers, so this change removes the parameter and uses the documentPath to determine the collection.

### DIFF
--- a/packages/@tinacms/graphql/src/database/index.ts
+++ b/packages/@tinacms/graphql/src/database/index.ts
@@ -472,10 +472,6 @@ export class Database {
         (collection) => documentPath.startsWith(collection.path)
       )
 
-      if (!collection) {
-        throw new Error(`Unable to find collection for ${documentPath}`)
-      }
-
       if (!pathsByCollection[collection.name]) {
         pathsByCollection[collection.name] = []
       }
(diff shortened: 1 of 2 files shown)


logan-anderson (approved):

packages/@tinacms/graphql/src/database/index.ts
```diff
@@ -463,13 +463,31 @@ export class Database {
...
-  ) => {
-    await _indexContent(this, documentPaths, collection)
+  public indexContentByPaths = async (documentPaths: string[]) => {
+    const pathsByCollection: Record<string,string[]> = {}
+    const nonCollectionPaths: string[] = []
+    const collections: Record<string,CollectionFieldsWithNamespace<true> | CollectionTemplatesWithNamespace<true>> = {}
+    const tinaSchema = await this.getSchema()
+    for (const documentPath of documentPaths) {
+      const collection = tinaSchema.schema.collections.find(
+        (collection) => documentPath.startsWith(collection.path)
```
jeffsee55, line 473: This is probably a change for the better, if/when we want to actually support the `matches` property in our collection config it'll be easier to do so here.

jeffsee55 (approved):

PR #2751, opened 2022-03-30
Formify check for blueprint

jeffsee55:
The SDK generates a query for references like:

```graphql
{
  getPostDocument(relativePath: $relativePath) {
    title
    author {
      ...on Document {
        id
      }
    }
  }
}
```
Previously this was expecting that `Document` would always be the actual type (eg. `AuthorDocument`) rather than the `Document` interface. If it _is_ the document interface, there's no need to formify those fields (they can't have `title` or anything like that in them).


Enigmatical (approved):
Looks good 👍

PR #2762, opened 2022-03-31
fix: #2740: escape index field separator in input strings

### LINKED ISSUE #2740: Data layer indexing fails to escape separator characters
## Description

The level store implementation use a colon for separating the parts of the index and doesn't escape colons in the data, causing fields to be incorrectly indexed

### Steps to reproduce

Put colon(s) in title field (with data layer enabled) in a document and query that field.

### Expected result

Query should include that doc's results

### Actual result

Query does not include that doc's results

kldavis4:
Update the datalayer movie tests in graphql and updates the data layer store code to fix these issues:
- properly escape the index field separate ('#') when it is present in a field and when used in a query
- pad numeric values for proper sorting. (NOTE: this hardcodes numeric values to 4 places. I'm currently thinking we can expose this as an optional field property in the schema but for now this should be sufficient for most use-cases)
- fix a bug in how fields were being ordered
- removes unused default property in the TinaIndex type

closes #2740 

# testing

You can test this by running `yarn dev-data` in the experimental examples tina-cloud-starter after adding a '#' character to the title of one of the blog posts and then executing a query like:

```
        query GetPostsList(
            $first: Float,
            $after: String, 
            $last: Float, 
            $before: String, 
            $sort: String,
            $filter: PostsFilter!) {
          getPostsList(first: $first, after: $after, last: $last, before: $before, sort: $sort, filter: $filter) {
            edges {
              cursor
              node {
                data {
                  title
                  date
                  published
                }
              }
            }
            pageInfo {
                hasNextPage
                endCursor
            }
          }
        }
```

```
{
    "filter": {
        "title": {
            "eq": "Just# Another Blog Post"
        }
    },
    "sort": "title"
}
```

### DIFF
--- a/.changeset/dry-mirrors-smile.md
+++ b/.changeset/dry-mirrors-smile.md
@@ -1,0 +1,5 @@
+---
+'@tinacms/datalayer': patch
+---
+
+Escape index field separator in input strings
--- a/packages/@tinacms/datalayer/src/database/store/index.test.ts
+++ b/packages/@tinacms/datalayer/src/database/store/index.test.ts
@@ -21,6 +21,7 @@ import {
   makeFilterChain,
   makeFilterSuffixes,
   makeKeyForField,
+  INDEX_KEY_FIELD_SEPARATOR,
 } from '.'
 
 describe('datalayer store helper functions', () => {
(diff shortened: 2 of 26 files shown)


Enigmatical (dismissed):
Looks good to me 👍

kldavis4:
@Enigmatical this is ready to review again - added a bunch of integration tests and fixed a few bugs that I found after adding the tests

Enigmatical (approved):
LGTM! 👍

PR #2767, opened 2022-04-01
Style Rich Text Input Lists

spbyrne:
New list styles:
Screen Shot 2022-04-01 at 11 37 04 AM

Previous, broken list styles:
Screen Shot 2022-04-01 at 11 37 44 AM

Changes:
- Adds `li:before` reset to try and catch common user list styling.
- Adds left padding to lists.
- Resets padding on list elements.
- Explicitly sets list styling and positioning for both unordered and ordered lists.

### DIFF
--- a/.changeset/poor-elephants-sleep.md
+++ b/.changeset/poor-elephants-sleep.md
@@ -1,0 +1,5 @@
+---
+'@tinacms/toolkit': patch
+---
+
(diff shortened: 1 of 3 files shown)


perkinsjr (approved):
LGTM!

perkinsjr (approved):

PR #2771, opened 2022-04-04
Feat/validation

logan-anderson:
Adds validation and error messages to the schema.

A couple of things to note
- Adds [zod](https://github.com/colinhacks/zod) as the validation library 
- Adds very basic and very verbose error handling and  messages, (would love feedback on these)
- Adds tests for the validation


jeffsee55 (commented):
This looks awesome, the fact that it handles recursion so well makes it worth the effort to switch from Yup. I suspect we'll want to slowly replace Yup in other areas.

A couple of thoughts:

#### Error messages

As you mentioned - the error message is super verbose and seems to point to a lot of messages that aren't relevant. I would consider it a blocker until we can clean it up somewhat

It was also a bit misleading, I tried misspelling `value` on an options item through a huge error, and I kept seeing:

`"Expected string, received object"`

And then for the parent:
`"type must be one of string, boolean, number, datetime, image, object, reference, rich-text"`

I was able to override some messages by providing a `message` prop on some of the fields and that seemed to work nicely. Might just be a matter of finding the right balance.

#### Using this for external types

Our types are pretty messy, would be great to leverage this, but I did see in the docs that when you use recursion you get unexpected results. I tried it briefly using [infer](https://github.com/colinhacks/zod#type-inference) but broke in confusing ways. Not much worse than what we have today, but I'm hoping we can do something better.

logan-anderson:
> Error messages
> ... 

Agreed! going to parse these better and update the PR. Pulling back into "In progress" for now

logan-anderson:
>Using this for external types
> Our types are pretty messy, would be great to leverage this, but I did see in the docs that when you use recursion you get unexpected results. I tried it briefly using [infer](https://github.com/colinhacks/zod#type-inference) but broke in confusing ways. Not much worse than what we have today, but I'm hoping we can do something better.

I agree one thing to keep in mind is that there are things that are in the types that might not be validated by the schema. The UI property contains functions and other stuff that we may not want to validate.

jeffsee55:
> there are things that are in the types that might not be validated by the schema

Yeah that's enough of an extra hurdle to where it's probably not worth it.

jeffsee55:
That's much more legible IMO. One thing (that I don't think should be a blocker) is that multiple messages pile up from what I'm assuming I think the last one is the only that makes sense.

I changed `value` to `vlue` in one of the options, and the last message was the one that made the most sense, but the parent object was also considered invalid

```ts
 options: [
  { label: "Default", vlue: "default" },
  { label: "Tint", value: "tint" },
  { label: "Primary", value: "primary" },
 ],
```

```
ERROR: your schema was not successfully validated: see https://tina.io/docs/schema/ for instructions on how to setup a schema
  Error Message Below
  TinaSchemaValidationError: Error Invalid input at path collections.4.fields.0.templates.1.fields.1.options.0, Error Expected string, received object at path collections.4.fields.0.templates.1.fields.1.options.0, Error Required at path collections.4.fields.0.templates.1.fields.1.options.0.value
```

Since `value` is required: `"Error Required at path collections.4.fields.0.templates.1.fields.1.options.0.value"` makes sense, but the others don't.

If this is somewhat simple to solve maybe it's worth it, but having this as it is now would be way better than what we have now.

packages/@tinacms/schema-tools/src/validate/fields.ts
```diff
@@ -0,0 +1,225 @@
...
+] as const
+
+const typeTypeError = `type must be one of ${TypeName.join(', ')}`
+const typeRequiredError = `type is required and must be one of ${TypeName.join(
+  ', '
+)}`
+
+const nameProp = z.string({
+  required_error: 'Name must be provided',
+  invalid_type_error: 'Name must be a sting',
```
jeffsee55, line 36:
```suggestion
  required_error: 'name must be provided',
  invalid_type_error: 'name must be a sting',
```

packages/@tinacms/schema-tools/src/util/parseZodErrors.ts
```diff
@@ -0,0 +1,51 @@
...
+    if (issue.code === 'invalid_union') {
+      issue.unionErrors.map((unionError) => {
+        moreInfo.push(parseZodError({ zodError: unionError }))
+      })
+      // moreInfo.push(issue.unionErrors.map((x) => x.flatten()))
+    }
+    const errorMessage = `Error ${issue?.message} at path ${issue.path.join(
+      '.'
+    )}`
+    const errorMessages = [errorMessage, ...moreInfo]
```
jeffsee55, line 28: Is `moreInfo` necessary? It seems like this adds some of the messages that aren't as relevant as the sort of main issue.

packages/@tinacms/schema-tools/src/validate/fields.ts
```diff
@@ -0,0 +1,225 @@
...
+})
+
+// ==========
+// Non Scaler fields
+// ==========
+const ReferenceField = FieldWithList.extend({
+  type: z.literal('reference' as const, {
+    invalid_type_error: typeTypeError,
+    required_error: typeRequiredError,
+  }),
```
jeffsee55, line 106: Out of scope of this ticket but - `type: 'reference'` should validate if the `collections` prop is actually referencing collections. This was something we could do in Yup, and seems a bit clunky in Zod, but maybe can be done with something like [this](https://github.com/colinhacks/zod/issues/1016). 

When `collections` is invalid, our GraphQL API errors during SDK generation with obscure messages like: `GraphQLDocumentError: Fragment cannot be spread here as objects of type "PostsAuthorDocument" can never be of type "Document"`

jeffsee55 (commented):
That error message seems better. I'm still not sure why we get so many dumped out, maybe it's due to the `moreInfo` part I commented on.

packages/@tinacms/schema-tools/src/util/parseZodErrors.ts
logan-anderson: It is necessary it also adds valid error message as well it some cases.

jeffsee55: Ok, I don't want to keep pushing this back but the error messages here are pretty tough to decipher. Maybe if each error is on it's own line that would help.

packages/@tinacms/schema-tools/src/validate/fields.ts
logan-anderson: We could definitely handle this in a separate ticket.

packages/@tinacms/schema-tools/src/util/parseZodErrors.ts
logan-anderson: Sounds good. PR is updated

jeffsee55 (dismissed):

jeffsee55 (approved):

PR #2788, opened 2022-04-07
fix: reduce _graphql.json file size by eliminating whitespace

### LINKED ISSUE #2787: Large `.tina/__generated__/_graphql.json` prevents data layer indexing
## Description

If _graphql.json is over 1MB you get this error from the github api when retrieving:
```
This API returns blobs up to 1 MB in size. The requested blob is too large to fetch via the API, but you can use the Git Data API to request blobs up to 100 MB in size.: {"resource":"Blob","field":"data","code":"too_large"}
```

Potential solution is to remove whitespace from the file

kldavis4:
# what

This change remove the whitespace when writing _graphql.json to help reduce the file size below the 1MB github api limit

Fix #2787

### DIFF
--- a/.changeset/cyan-bulldogs-judge.md
+++ b/.changeset/cyan-bulldogs-judge.md
@@ -2,4 +2,4 @@
 '@tinacms/graphql': patch
 ---
 
-Modify putConfig to write \_graphql.json without whitespace to reduce file size
+Modify database to write config json files without whitespace to reduce file sizes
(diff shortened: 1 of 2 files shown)


kldavis4:
@jeffsee55 wondering if same should be done for _lookup.json and _schema.json?

jeffsee55:
@kldavis4 yes none of those really need to be legible.

jeffsee55 (approved):

Issue #2785, opened 2022-04-06
Unregistered MDX template error should link to the docs


jamespohalloran:
## Description

The error message when you don't register an MDX component should link to the docs: https://tina.io/docs/editing/mdx

<img width="998" alt="Screen Shot 2022-04-06 at 9 31 43 AM" src="https://user-images.githubusercontent.com/3323181/161975101-bede0663-d2ed-4ed5-a29d-76debfb8509d.png">

### Steps to reproduce

From the docs starter, add an html component.
E.g: `<hr />`

### Expected result

Error message should link to https://tina.io/docs/editing/mdx/
This should happen when developing locally, and when on prod (with the fancy error modal)

#### Tinacms Package Versions

├── @tinacms/auth@0.50.2
├── @tinacms/cli@0.60.8
├── next-tinacms-cloudinary@3.4.3
├── tinacms@0.66.8

PR #2796, opened 2022-04-11
fix: #2785: link to MDX docs when unregistered component is used

kldavis4:
# what
Update the error message displayed when an unregistered MDX component is used to include link to documentation


jeffsee55 (approved):
🚀

PR #2799, opened 2022-04-12
feat: --schemaFileType cli option and NO_TELEMETRY env var

kldavis4:
# what
- Update tinacms CLI to support schemaFileType option (default 'ts') to allow user to specify the schema file type
- Update telemetry module to optionally check NO_TELEMETRY environment variable for disabling telemetry


logan-anderson (approved):
LGTM

PR #2815, opened 2022-04-15
fix: 2810: Database.put fails on windows

kldavis4:
# what

Modify the schema generation in cli init to use the correct path separator depending on the platform and do some general refactoring of the database code which determines the collection based on the path.


jeffsee55 (approved):

nodgear:
This just happened to me on @latest.
slightly different erroring but same result and same fix (changing `/` to `\` manually)

```
TinaFetchError: Error in PUT for content\services\Quicada.json
Error: Unable to find collection for file at content\services\Quicada.json
```

Issue #2753, opened 2022-03-30
Add instructions in the CLI of how to enter the CMS


jamespohalloran:
## Summary

As a dev running a TIna site with the CLI, I want clearer instructions of how to enter the CMS  so that if I only skimmed the docs, I can still be successful.

### Expected Results
- When process.env.CI is not set, we should add something like this to the console "Enter the CMS at [your-development-url]/admin"
- Would be nice to style this output a bit more so this URL, and the Altair URL stick out more. 

### Notes

Here's our current output: 

<img width="482" alt="Screen Shot 2022-03-30 at 9 38 07 AM" src="https://user-images.githubusercontent.com/3323181/160836035-c48248dc-37ad-4028-8eb0-11837ec2a4a5.png">

Hard to get analytics on this, but I would bet that some users run Tina with the CLI, and never actually make it to the CMS.

_Also, I know that technically a dev can customize the "admin" path, and that technically they could run the graphQL API without running the site, but IMO, this covers 99% of cases._

PR #2819, opened 2022-04-19
Make dev mode CLI a bit more friendly.

mittonface:
For #2753 

* I've added a `--verbose` option to the server:start command that cuts out a bit of the noise so that the altair URL is better highlighted.
image

The ticket also wants us to put in message that points the user to `/admin` so they can make it to their CMS. However we're not really doing anything besides running `next dev` in the starters to get their site up and running. So: 

@jamespohalloran 
**Do I just tack on something to the `server:start` command that tells the user to go to `/admin`?** 

a) They can technically run `server:start` separately
b) Technically the `server:start`command has nothing to do with their Next site actually running.

The ticket seems to imply you that this is maybe what you want, but just want to double check.

### DIFF
--- a/packages/@tinacms/cli/src/cmds/compile/index.ts
+++ b/packages/@tinacms/cli/src/cmds/compile/index.ts
@@ -21,7 +21,6 @@ import type { Loader } from 'esbuild'
 import type { TinaCloudSchema } from '@tinacms/graphql'
 import { dangerText, logText } from '../../utils/theme'
 import { defaultSchema } from './defaultSchema'
-import fs from 'fs-extra'
 import { getSchemaPath } from '../../lib'
 import { logger } from '../../logger'
 


jamespohalloran:
> **Do I just tack on something to the `server:start` command that tells the user to go to `/admin`?**
> 
> a) They can technically run `server:start` separately b) Technically the `server:start`command has nothing to do with their Next site actually running.

Yeah, I think that's fine for now! All our examples/starters have them running together, so if they're running them independently, they are probably rolling with an advanced use-case and won't get too tripped up by the console.log

kldavis4:
need a documentation update for the verbose flag?: https://github.com/tinacms/tinacms/blob/main/packages/%40tinacms/cli/README.md

packages/@tinacms/cli/README.md
```diff
@@ -24,6 +24,7 @@ Usage: @tinacms/cli command [options]
 Options:
   -V, --version             output the version number
   -h, --help                display help for command
+  -, --verbose              increase verbosity of console output
```
kldavis4, line 27:
```suggestion
  -v, --verbose              increase verbosity of console output   
```

packages/@tinacms/cli/src/cmds/compile/index.ts
```diff
@@ -50,9 +52,13 @@ const cleanup = async ({ tinaTempPath }: { tinaTempPath: string }) => {
 export const compile = async (
   _ctx,
   _next,
-  options = { schemaFileType: 'ts' }
+  options: { schemaFileType?: string; verbose?: boolean }
 ) => {
-  logger.info(logText('Compiling...'))
+  if (!options.schemaFileType) options = { ...options, schemaFileType: 'ts' }
+
+  if (options.verbose) {
```
logan-anderson, line 59: @mittonface We could use the logger for this (it does not really matter to me).

It would look something like this.
```ts
if(options.verbose) {
  // set the logger level to verbose
  logger.setLevel('verbose')
}
// instead of logger.into
logger.verbose(logText('Compiling...'))
```

This is more sudo-code (I am not 100% sure of the logger interface)

jeffsee55 (approved):

PR #2820, opened 2022-04-19
2781 - Removes backend calls for `getCollections` and `getDocumentFields` from CMS

### LINKED ISSUE #2781: Simplify GraphQL API: Remove reliance on backend for form generation
**Is a Blocker For:** 
- [ ] https://github.com/tinacms/product-roadmap/issues/91

**Summary:**
Now that we have access to the Tina schema on the frontend we can start to remove the logic for generating forms on the backend. There are a few things we do today that rely on this, and we should move the responsibility to the `TinaAdmin` or `TinaSchema` modules:

- [x] Listing collections in the CMS
- [x] Getting fields for a given collection form
- [x] `mutationInfo` provides the mutation string, as well as some info on how to transform the form payload before sending it to the server.
  - `transformDocumentIntoMutationRequestPayload` has some shaky logic in it because it has no knowledge of the schema. Now that we have it client-side, we should be able to remove it's reliance on `mutationInfo`, and instead just look at the schema
  - This has been spun into a separate card: https://github.com/tinacms/tinacms/issues/2822 to address technical debt
- [x] ~~Another natural improvement to make here is to ensure `formifyCallback` is called for admin form generation too.~~
- [ ] Ensure when schema is not provided we throw an error

**Notes:**
This will be a breaking change as users will now have to provide the `schema` as a `TinaCMS` prop

cc @Enigmatical I know we've discussed this, please chime in with any other thoughts.

Enigmatical:
Now that we have `schema` available on the frontend (thanks to "Extending Tina"), the CMS can retrieve a lot of its structural information from that rather than making calls to the backend.

This means that the Global Nav no longer needs to call `getCollections` and the CMS can use the `schema` to build a `form` rather than `getDocumentFields`.

The only calls the CMS still makes are those related to `document` retrieval (which makes sense):
- When getting a list of a `collection`'s `documents` (via a `getCollection` call)
- When getting a specific `document` (via a `getDocument` call)

Closes #2781


packages/@tinacms/graphql/src/resolver/index.ts
```diff
@@ -747,7 +747,6 @@ export class Resolver {
     const value = rawData[field.name]
     switch (field.type) {
       case 'datetime':
-        // See you in March ;)
```
jeffsee55, line 750: 😆

jeffsee55 (approved):

jeffsee55:
I can't make sense of why the cypress tests are failing https://github.com/tinacms/tinacms/runs/6086121295?check_suite_focus=true. Sort of seems unrelated to this work and was previously failing so I'm going to go ahead and merge so I can build on top of these changes.

cc @logan-anderson any suggestions on the failing cypress tests?

PR #2826, opened 2022-04-20
Improvement of select dropdown. Make arrow icon clickable

Brunni:
Status Quo: The arrow icon on top of the HTML5 select element cannot be clicked by the user. There is also no hover effect.
The arrow icon is created using an react-icon svg which is moved on top of the dropdown.

Solution:
https://tailwindcss.com/docs/pointer-events


jeffsee55 (dismissed):
Thanks @Brunni!

jeffsee55 (approved):

PR #2839, opened 2022-04-24
Replaced tinacms dependency with @tinacms/toolkit

Brunni:
With this pull-request I want to be able to use the `react-tinacms-editor` without dependency to the "big" tina package, as we are just using parts of the @tinacms/toolkit project.

Main changes:
* Removed devDependency and peerDependency on "tina"
* Added devDependency and peerDependency on "@tinacms/toolkit"
* Fixed imports.

### DIFF
--- a/.changeset/beige-toes-stare.md
+++ b/.changeset/beige-toes-stare.md
@@ -1,0 +1,5 @@
+---
+'react-tinacms-editor': minor
+---
+
+Replaced dependendy from react-tinacms-editor to tinacms with dependency to @tinacms/toolkit
--- a/packages/react-tinacms-editor/package.json
+++ b/packages/react-tinacms-editor/package.json
@@ -22,6 +22,7 @@
     "@storybook/react": "^5.3.19",
     "@testing-library/react": "^12.0.0",
     "@tinacms/scripts": "workspace:*",
+    "@tinacms/toolkit": "workspace:*",
     "@types/codemirror": "^0.0.71",
     "@types/color-string": "^1.5.0",
     "@types/jest": "^27.0.1",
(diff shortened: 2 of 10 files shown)


logan-anderson:
Thanks for the contribution @Brunni. Just a couple of quick questions.

1. What is your use case for this? I think this is only useful when someone is installing  not `tinacms` into there project ( which seems rare )

2. This package can be lazy loaded and only loaded when an editor is editing the page (it will never load in production) for this reason larger bundle size should not be a huge deal. (That being said we are always open to making it smaller if editors notice long loading times for JS assets)
   -  See [this doc](https://tina.io/docs/editing/markdown/#registering-the-field-plugins) for how to lazy load it.

3. I think (although I could be wrong) that most of the size of `tinacms` comes from `@tinacms/toolkit` since `@tinacms/toolkit` is a direct dependency of  `tinacms`. It might be good to take a look at the sizes of both of these packages to see how much saving there is.

Again, thanks for the contribution and looking forward to hearing back.

Brunni:
Hi Logan,

were are using only tinacms/toolkit in a special setup, where we need users/admins to be able change the content (text, image) of some views in our application. As we don't use any GraphQL for Tina or TinaCloud, we have implemented a solution, where we only use the tina sidebar (form, fields) and a small REST interface for saving data.

We use lazy loading for tina.

I don't see any reason, why the react-tinacms-editor needs to depend on the whole tina package.

schierla:
As far as I can see, the direct imports from `tinacms` are actually defined in `@tinacms/toolkit` and just reexported in `tinacms`.

jeffsee55:
@schierla sorry for the delay, but I don't think we can say for certain that this change is something we want to support going forward. The mixture of modules between `tinacms` and `@tinacms/toolkit` is still a bit fuzzy (we previously had over 20 packages and have been working to combine them) and we'll likely have to do a lot of shuffling in the near future to handle these types of requests. We hope to soon publish es versions of these packages properly which should address those concerns.

I'm going to close this for now, but please feel free to continue to comment here.

Issue #2817, opened 2022-04-17
Upgrading past v0.66.9 breaks collections with multiple templates (and causes other small bugs)


coreyaus:
## Description

Hey legends - I have a project using `tinacms@0.66.9` and I see the following series of issues appear when upgrading to version `0.66.10` or version `0.67.0`:

1. Collections break if they implement multiple different templates (i.e. passing an array of `templates` rather than an array of `fields`). The error message in the Tina Admin view when trying to create or edit a document in that collection is:
    > Failed to assertShape - _template is a required field
    - You can also see that in the screenshot below
    - That persists even when settings `cms.flags.set('use-unstable-formify', false)`

2. The issue described here: #2816

3. The `TinaCollection`/`TinaCloudCollection` types seem to break
    - That is, a TypeScript error appears in `schema.ts` when running code like the following:
        ```typescript
        import pagesCollection from './schema/collections/pages'
        
        export default defineSchema({
            collections: [pagesCollection],
        })
        ```
    - The TS error appears on the `collections: [pagesCollection]` line. It claims that the types expected for that `collections` property are not compatible with the `TinaCollection` or `TinaCloudCollection` types used in the `pagesCollection` file, which are:
        ```typescript
        import type { TinaCollection, TinaCloudCollection } from '@tinacms/cli'
        const pagesCollection: TinaCollection | TinaCloudCollection = {
            label: 'Pages',
            name: 'pages',
            path: 'content/pages',
            fields: [...],
        }
        export default pagesCollection
        ```
    - Please let me know if there is something wrong with those types I'm using - they work fine when running `tinacms@0.66.9`

4. The Tina hamburger menu to open the list of collections, media manager, etc. stops working, unless we set `reactStrictMode: false` in our `next.config.js` file

Please let me know if you have any questions about those points or need more info. Thanks for all your efforts building such a sensational tool! 🙌 

### Steps to reproduce

I have a private repository that currently works on `v0.66.9` and all these issues are visible when upgrading. I'd be happy to share access to that repository with the relevant people from the Tina core team. Alternatively, I'm happy to run tests myself with a patch version of tinacms if you have an idea for these fixes and want me to review whether an updated/edge version resolves the issues above 👍 

### Expected result vs actual result

I expect to be able to upgrade without the errors listed above but currently we need to run tinacms 2 versions behind the latest in order to avoid those issues

#### Tinacms Package Versions

[...]

PR #2842, opened 2022-04-26
Fix issues with finding the template for multitemplate collections

jeffsee55:
Fixes part 1 of https://github.com/tinacms/tinacms/issues/2817

> Collections break if they implement multiple different templates (i.e. passing an array of templates rather than an array of fields). [...]


Enigmatical (approved):
👍

PR #2844, opened 2022-04-27
Fix/use rm

### LINKED ISSUE #2835: Deprecation Warning `fs.rmdir` -> `fs.rm`
The following deprecation warning is appearing when running `dev` on >14.4.0:

```
(node:425930) [DEP0147] DeprecationWarning: In future versions of Node.js, fs.rmdir(path, { recursive: true }) will be removed. Use fs.rm(path, { recursive: true }) instead
(Use `node --trace-deprecation ...` to show where the warning was created)
```

https://nodejs.org/api/deprecations.html#DEP0147

logan-anderson:
fixes #2835

### DIFF
--- a/.changeset/three-feet-warn.md
+++ b/.changeset/three-feet-warn.md
@@ -1,0 +1,6 @@
+---
+'@tinacms/cli': patch
+'@tinacms/graphql': patch
+---
+
+Use rm instead of rmdir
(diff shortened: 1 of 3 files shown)


Enigmatical:
Hey!  Given that this is using `fs-extra`, should we be using `.remove()` instead?  Maybe `rm` is an alias to `remove`?
https://github.com/jprichardson/node-fs-extra/blob/master/docs/remove.md

jeffsee55:
Is this different from https://github.com/tinacms/tinacms/pull/2836?

logan-anderson:
Closing in favour of #2836

PR #2853, opened 2022-04-29
Fix issue where pages that didnt use `useTina` would get a loading spinner

jeffsee55:
When a page doesn't use the `useTina` hook, no query ever comes through so our `status` remained `idle`. Just return early in that case


Enigmatical (approved):
Makes sense.

PR #2856, opened 2022-04-29
Style list page overflow menu

spbyrne:
This adjusts the list page overflow menu styling given the existing styling was designed for another context. This also removes the `showEmbed` prop given it was specific to the other context (MDX toolbar) and reverts to rendering the button as a button instead of a `<span>` (Maybe that was necessary for the toolbar?). 

Screen Shot 2022-04-29 at 12 02 32 PM

### DIFF
--- a/.changeset/fifty-houses-confess.md
+++ b/.changeset/fifty-houses-confess.md
@@ -1,0 +1,6 @@
+---
+'@tinacms/toolkit': patch
+'tinacms': patch
+---
+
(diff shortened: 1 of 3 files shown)


jeffsee55:
Yes the span is needed for the toolbar because it has to make sure we don't steal focus from the editor so it's a span and uses onMouseDown events instead.

jeffsee55 (dismissed):

Enigmatical (approved):

PR #2863, opened 2022-04-30
Remove old MDX editor component

jeffsee55:
Should remove the "it looks like there are several instances of styled-components" warning in the console.


Enigmatical (approved):
This is mostly just deleting files and some other fixes to MDX that we've already merged.  Looks good 👍

PR #2870, opened 2022-05-03
Popover panel

spbyrne:
Adds a popover panel to the collection list page overflow menu based on @jeffsee55's fix #2861 for the MDX toolbar issue.


jamespohalloran (approved):
Just tested this out on a site, and it's working for me!

PR #2879, opened 2022-05-06
Persist sidebar state across reload

spbyrne:
This saves and loads sidebar state to local storage. Both the open/closed/fullscreen state and sidebar width are persisted. 

I've added inline comments for clarity but here's what it's doing:

- On initialization, set sidebar open state and width to local values if available
- If the default sidebar state changes, update current state if it wasn't set from local state
- Update the local value of the sidebar state any time it updates, if the CMS is loaded 
- Update the local value of the sidebar width any time the user drags to resize it


kldavis4 (approved):
tested and seems to be working as described

PR #2887, opened 2022-05-10
Feat/use is title

### LINKED ISSUE #2723: Ability to designate a field as the title (of a document)
## Summary

As a user I want to see a more readable title than the filename in my collection view so that I can easily find the content that I want

### Depends On:
Data Layer no longer experimental OR a flag indicating opt-in to data layer (so we can see easily whether data layer is enabled)

### Expected Results

- In the schema, I should be able to set a "isTitle" property on a field definition, which marks it as the "title" field.
- If "isTitle" is true, it is also required
- When looking at the collection list view in the CMS, the "title" should be used as the main label (with the filename as a "secondary" field to differentiate between same titles that are not unique)
- Ideally, if there are more than one defined "isTitle" fields, the schema build will fail with an error.
- "isTitle" field must comply with the following constraints:
      - Must be a non-list field of type string with required: true.
      - Must be a top-level field (ie. not inside an object)
- Update documentation for this and for the addition of the data layer opt-in flag
- Tests pass

### Notes
If "isTitle" is false (on all fields of a collection), filename can be used as the title

### LINKED ISSUE #2886: Sort does not work on `collection` query
## Description

When running a query like this.

```gql
query($collection: String!, $includeDocuments: Boolean!, $sort: String) {
  postsConnection(sort: "title") {
    edges {
      node {
        title
      }
    }
  }
  collection(collection: $collection) {
    name
    label
    format
    templates
    documents(sort: $sort) @include(if: $includeDocuments) {
      totalCount
      edges {
        node {
          ... on Document {
            __typename
            ... on Posts {
              title
            }
          }
        }
      }
    }
  }
}

```
With variables
```json
{
  "collection": "posts",
  "sort": "title",
  "includeDocuments": true
}
```

I would expect that the documents would be returned in order of the `title` field.  (ie `postsConnection` and `collection.docuemnts` should be in the same order )

### Actual result
They are not ordered by title but ordered by filename

```json
{
  "data": {
    "postsConnection": {
      "edges": [
        {
          "node": {
            "title": "Just Another Blog Post"
          }
        },
        {
          "node": {
            "title": "Vote For Pedro"
          }
        },
        {
          "node": {
            "title": "Z"
          }
        }
      ]
    },
    "collection": {
      "name": "posts",
      "label": "Blog Posts",
      "format": "mdx",
      "templates": null,
      "documents": {
        "totalCount": 3,
        "edges": [
          {
            "node": {
              "__typename": "Posts",
              "title": "Just Another Blog Post"
            }
          },
          {
            "node": {
              "__typename": "Posts",
              "title": "Z"
            }
          },
          {
            "node": {
              "__typename": "Posts",
              "title": "Vote For Pedro"
            }
          }
        ]
      }
    }
```

### Environment

monorepo

logan-anderson:
fixes #2886
fixes #2723

TODO: @spbyrne  is going to make it look good!


experimental-examples/tina-cloud-starter/content/posts/asdf.mdx
```diff
@@ -0,0 +1,5 @@
+---
```
kldavis4, line 1: did you mean to keep this?

packages/@tinacms/graphql/src/resolver/index.ts
```diff
@@ -64,12 +67,15 @@ export class Resolver {
     // if (res.type === "union") {
     //   extraFields["templates"] = res.templates;
     // }
-    let documents = {}
-    if (hasDocuments) {
-      documents = await this.getDocumentsForCollection(collectionName)
-    }
+    // let documents = {}
```
kldavis4, line 70: I'm confused by this - I thought the commented out code here was the fix for the issue you mentioned

Enigmatical (approved):
Barring @kldavis4 's questions, I'm good with the changes to the CMS.  Surprised at how quickly it came together, nice work 👍

experimental-examples/tina-cloud-starter/content/posts/asdf.mdx
logan-anderson: No I should delete this. Thanks

logan-anderson:
Here is a video explaining the code. It's a bit complicated but I hope this helps.
https://www.loom.com/share/1eb3c70d062d4ce7b8c41e31108377a9


cc @kldavis4

kldavis4 (approved):
@logan-anderson thanks for the video - that makes sense :)

PR #2889, opened 2022-05-10
Adds a paginate package to @tinacms

Enigmatical:
This is an iteration on Kelly's approach to handling cursor-based pagination with patterns like `page/1`, `page/2`, etc.

It requires the use of a filesystem cache and the experimentalData flag enabled.

### DIFF
--- a/packages/@tinacms/paginate/package.json
+++ b/packages/@tinacms/paginate/package.json
@@ -17,8 +17,9 @@
   },
   "devDependencies": {
     "@tinacms/scripts": "workspace:*",
-    "typescript": "^4.3.5",
-    "jest": "^27.0.6"
+    "@types/node": "^17.0.31",
+    "jest": "^27.0.6",
+    "typescript": "^4.3.5"
   },
   "scripts": {
     "build": "echo \"Run `yarn build` from the root of the repository instead\"",
--- a/packages/@tinacms/paginate/package.json
+++ b/packages/@tinacms/paginate/package.json
@@ -1,0 +1,36 @@
+{
+  "name": "@tinacms/paginate",
+  "main": "dist/index.js",
+  "typings": "dist/index.d.ts",
+  "files": [
+    "dist",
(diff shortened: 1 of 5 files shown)


packages/@tinacms/paginate/package.json
```diff
@@ -0,0 +1,36 @@
...
+  "buildConfig": {
+    "entryPoints": [
+      {
+        "name": "src/index.ts",
+        "target": "node"
+      }
+    ]
+  },
+  "devDependencies": {
+    "@tinacms/scripts": "workspace:*",
```
logan-anderson, line 19: I think jest needs to be a devDep (I _think_ that is why the test is failing)

Enigmatical: Ah, thanks!
I didn't think I needed it cause I don't have any tests, but that makes sense.

jamespohalloran:
I wonder if it makes sense to call the package something like `offset-paginate`.
Might also be good to have a readme where we specify the limitations / use-cases.
E.g. This will ONLY work statically (it uses a filesystem cache under the hood)

Enigmatical:
> I wonder if it makes sense to call the package something like `offset-paginate`. Might also be good to have a readme where we specify the limitations / use-cases. E.g. This will ONLY work statically (it uses a filesystem cache under the hood)

I'm not sure if we _should_ call it `offset-paginate` or maybe something even more specific?  `static-offset-paginate`?  If our spike determines that we do offer an actual way to handle offset pagination, I wouldn't want to cause confusion with this.

Also, definitely needs to be all manner of stuff written around it - hence the idea of a mini-guide (and a README).  

This is just to see if there's anything of value here :)

Enigmatical:
We've decided to take a step back from offering a pagination solution for now until we better understand what the community is looking for.  There's a good chance this work will turn into a "Mini Guide" on tina.io rather than a package we would want to maintain and continue to support.

PR #2893, opened 2022-05-11
Adds `useRelativeMedia` to local GraphQL

Enigmatical:
Part of the Repo-based Media push, this adds an extra `config` to `resolve()` that contains (for now) just `useRelativeMedia: true`.

Locally, this value will always be `true`.

In the future, a companion PR to `tina-cloud` will include a slightly different `config`:
```
{ useRelativeMedia: false, clientId: app.client_id }
```
which will be used to determine the location of the file on `assets.tinajs.io`. (https://github.com/tinacms/product-roadmap/issues/161)

> Note: This can merge without the sister PR.

Closes https://github.com/tinacms/product-roadmap/issues/50

### DIFF
--- a/packages/@tinacms/graphql/src/resolver/index.ts
+++ b/packages/@tinacms/graphql/src/resolver/index.ts
@@ -29,11 +29,14 @@ import type {
   TinaSchema,
 } from '@tinacms/schema-tools'
 
+import type { GraphQLConfig } from '../types'
+
 import { TinaError } from './error'
 import { FilterCondition, makeFilterChain } from '@tinacms/datalayer'
 import { collectConditionsForField, resolveReferences } from './filter-utils'
 
 interface ResolverConfig {
+  config: GraphQLConfig
   database: Database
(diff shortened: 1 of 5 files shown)


packages/@tinacms/graphql/src/resolver/index.ts
```diff
@@ -706,9 +712,21 @@ export class Resolver {
...
         accumulator[field.name] = value
         break
+      case 'image':
+        if (this.config) {
+          if (this.config.useRelativeMedia === true) {
+            accumulator[field.name] = value
+          } else {
+            accumulator[
+              field.name
+            ] = `https://assets.tinajs.io/${this.config.clientId}/${value}`
```
Enigmatical, line 724: @kldavis4 Is this `assets.tina.io` now?  Sounded like that was an easy change to make.

kldavis4: it _should_ be - just waiting for the prod deploy to complete

PR #2902, opened 2022-05-12
update sdk so it works with unified client

### LINKED ISSUE #2778: SDK works with new "unified" client
## Summary

SDK continues to work with new client: (See [here](https://github.com/tinacms/internal-product-discussions/discussions/116#discussioncomment-2440057) for more details )

### Expected Results
- SDK does not import from `tinacms`
- SDK generates a class that is passed to create client.

So before one would have 
```ts
import { createClient } from 'tinacms/dist/client' 
export const client = createClient({same options as before}
```
becomes
```ts
import { createClient } from 'tinacms/dist/client' 
import { ExperimentalSDK } from './__generated__/types'

export const client = createClient({same options as before, sdk: ExperimentalSDK }) 
```

this expands the client interface to allow queries to be run at the top level.
```ts
client.getPost({...})
// or
client.myCustomQuery({...})
```

### Notes
Depends on https://github.com/tinacms/internal-product-discussions/discussions/116
If this winds up being more difficult than expected, another option is to switch the order by passing in the ExperimentalClient vs. passing in the ExperimentalSDK

logan-anderson:
fixes #2778

### DIFF
--- a/.changeset/gentle-games-heal.md
+++ b/.changeset/gentle-games-heal.md
@@ -1,0 +1,6 @@
+---
+'@tinacms/cli': patch
+'tinacms': patch
+---
+
+Adds support for using the generated client on the frontend
--- a/experimental-examples/tina-cloud-starter/.tina/__generated__/types.ts
+++ b/experimental-examples/tina-cloud-starter/.tina/__generated__/types.ts
@@ -1216,31 +1216,43 @@ export type Requester<C= {}> = <R, V>(doc: DocumentNode, vars?: V, options?: C) 
   export type Sdk = ReturnType<typeof getSdk>;
 
 // TinaSDK generated code
-import { createClient } from 'tinacms/dist/client'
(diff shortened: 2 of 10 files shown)


logan-anderson:
Can be tested in mono repo or by doing

`yarn add tinacms@new-sdk` and `yarn add @tinacms/cli@new-sdk`

Enigmatical (approved):
The Typescript stuff is a bit above my head, so I'll take your word on it, but the code itself looks solid.  Placing the `sdk` one level off `client` is going to pay off in the long run.

It keeps the `client` namespace focused entirely on generic requests while the `sdk` is focused on query-based requests.

And, as I mentioned to you in Slack, it also quarantines the `sdk` namespace so we avoid collisions.  Not that I think it would happen, but if somebody names a personal query `request`, that could potentially collide with our `request` and clobber it.

Looks good 👍 Looking forward to updating my dogfooding site with it.

PR #2912, opened 2022-05-17
Feat/local media routes

logan-anderson:
Adds media routes to the local server


packages/@tinacms/cli/src/server/models/media.ts
```diff
@@ -0,0 +1,89 @@
...
+interface FileRes {
+  src: string
+  fileName: string
+  size: number
+  isFile: boolean
+}
+interface ListMediaRes {
+  directories: string[]
+  files: File[]
+  curser?: string
```
kldavis4, line 36:
```suggestion
  cursor?: string
```

kldavis4 (dismissed):

PR #2913, opened 2022-05-18
Replace field meta styled components with tailwind

### LINKED ISSUE #2891: Tina Admin Panel: Not all text wraps
## Description

I have noticed when creating a custom component that has paragraph text in it and validation error messaging, the text does not wrap. However things like `description` do wrap.  Maybe this isn't a bug, if not feel free to change the label.

Here is an example of what I mean
image

 I noticed that there is an explicit `no-wrap` set on the validation message span so maybe it is intentional?.

### Steps to reproduce

Create a field with a validation method and trigger it to be invalid with a long error message.  It won't wrap.

Alternatively I recall creating a custom component that had a paragraph of long text and that also didn't wrap.  I could see this not being a bug because components are unknown and arbitrary, but i figured i'd mention it anyway

### Expected result

I would expect at least for validation errors, since they are strings, that they should wrap like descriptions. 

### Actual result

What actually happened?

#### Tinacms Package Versions

```text

```

#### Tools

| Tool    | Version |
| ------- | ------- |
| browser |         |
| os      |         |
| node    |         |
| npm     |         |

spbyrne:
Fixes #2891 (error messages not wrapping) and replaces the remaining field meta styled components with tailwind components.

### DIFF
--- a/.changeset/chatty-olives-wink.md
+++ b/.changeset/chatty-olives-wink.md
@@ -1,0 +1,5 @@
+---
+'@tinacms/toolkit': patch
+---
+
+Replace field meta components, fix text wrapping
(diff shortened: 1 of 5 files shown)


Enigmatical (approved):
Looks good 👍

PR #2919, opened 2022-05-18
Adds `waitForDB` command to Tina CLI

Enigmatical:
For use when polling for DB indexing, `waitForDB` is introduced into the `tinacms server:start` and `tinacms server:waitForDB` commands.

This leverages the unified client (`.tina/client.ts`) to ping a specific endpoint on `tina-cloud` that returns a status of indexing - `inprogress | completed | failed | unknown`.

I also refactored `getPath` into `getSchemaPath` and `getClientPath` that both leverage `getPath`.  I also split `compile` into `compileSchema` and `compileClient` and made `transpile` more generic so it could be used by both.

Closes https://github.com/tinacms/product-roadmap/issues/155


Enigmatical:
Additional work needed:

* Add `--dev` option to `tinacms` for setting `NODE_ENV` to `development` vs `production`
* Pull in `dotenv` to process `env` before compiling the client (`compileClient()`)
* Confirm all of this works in CI

packages/@tinacms/cli/src/cmds/waitForDB/index.ts
```diff
@@ -0,0 +1,109 @@
...
+      } else if (status === STATUS_INPROGRESS) {
+        if (options.verbose) {
+          logger.info(logText(`${statusMessage}, trying again in 5 seconds`))
+        }
+        setTimeout(pollForStatus, POLLING_INTERVAL)
+
+        // Index Failed
+      } else if (status === STATUS_FAILED) {
+        logger.error(
+          dangerText(`${statusMessage}, encountered error: ${error}`)
```
logan-anderson, line 91: We should probably throw an error here (With an message) instead of doing a logger.error. Failing silently causes a very strange and cryptic  error message on vercel.

logan-anderson:
Looks good. I tested in Tina-cloud and it seemed to be working as expected. 

This did expose an issue with tina-cloud and we should consider not merging this until [this ticket](https://github.com/tinacms/tina-cloud/issues/1551) is resolved.

kldavis4 (dismissed):

Issue #2925, opened 2022-05-23
Branching doesn't work correctly when using contextual editing


perkinsjr:
## Description

When using branch switching, it successfully will switch branches and use the data from the branch when using the Admin CMS, however, when you go to use contextual editing it will only use main. 

https://user-images.githubusercontent.com/45409975/169816226-8f6a0638-0639-484e-a852-54f1cd8db1aa.mp4

### Steps to reproduce

1. Create a new Branch 
2. Try and use the branch when using Contextual editing, in the network and commits you will see it is using main only. 
3. Create a branch using contextual editing and it will show in the UI as the branch but the commits and data will be retrieved only from the `main` branch.

### Expected result

The branch selected should be used to retrieve data in contextual editing.

### Actual result

`main` is being used regardless. 

#### Tinacms Package Versions

tinacms@0.68.5

PR #2926, opened 2022-05-23
fix: #2925: update current branch from local storage

kldavis4:
Addresses #2925. It appears that for contextual editing, the client was being configured with the default branch instead of the branch in local storage. This PR updates the logic in the tina cloud provider to update the branch when the value is read for local storage

### DIFF
--- a/packages/tinacms/src/auth/TinaCloudProvider.tsx
+++ b/packages/tinacms/src/auth/TinaCloudProvider.tsx
@@ -157,7 +157,9 @@ export const TinaCloudProvider = (
     [props.cms]
   )
   if (!cms.api.tina) {
-    cms.registerApi('tina', createClient(props))
+    cms.registerApi('tina', createClient({ ...props, branch: currentBranch }))
+  } else {
+    cms.api.tina.setBranch(currentBranch)
   }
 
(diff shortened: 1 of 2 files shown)


packages/tinacms/src/auth/TinaCloudProvider.tsx
```diff
@@ -157,7 +157,9 @@ export const TinaCloudProvider = (
     [props.cms]
   )
   if (!cms.api.tina) {
-    cms.registerApi('tina', createClient(props))
+    cms.registerApi('tina', createClient({ ...props, branch: currentBranch }))
+  } else {
+    cms.api.tina.setBranch(currentBranch)
```
Enigmatical, line 162: This makes sense to me.  Would explain why `branch` wasn't working if we weren't instantiating the `client` with it.

I wonder if the CMS working properly is more a race condition than intentional coding?

kldavis4: yeah, I was wondering about a race condition as well. I'll make this ready for review (James P has already tested and confirmed it fixed the issue for him)

Enigmatical (approved):

PR #2939, opened 2022-05-30
feat: #51: default TinaMediaStore for repo-based media

kldavis4:
# What

Implement default Tina Media Store that works in both local dev mode and against Tina Cloud's assets api

# Testing

There is a Tina Cloud PR that needs to be merged before this will work against the prod environment. The changes have been deployed to staging. See loom on associated ticket for configuration details.

resolves ENG-174


packages/@tinacms/toolkit/src/packages/core/media-store.default.ts
```diff
@@ -42,3 +54,229 @@ export class DummyMediaStore implements MediaStore {
...
+      }
+    }
+  }
+
+  async isAuthenticated() {
+    this.setup()
+    return await this.api.isAuthenticated()
+  }
+
+  accept = 'text/*,  application/*, image/*'
```
Enigmatical, line 93: Is this accurate?  I seem to recall we were limiting the filetypes to just image formats.

packages/@tinacms/toolkit/src/packages/core/media-store.default.ts
```diff
@@ -42,3 +54,235 @@ export class DummyMediaStore implements MediaStore {
...
+          item.directory && item.directory !== '/'
+            ? `${item.directory}/${item.file.name}`
+            : item.file.name
+        }`
+        const res = await this.api.fetchWithToken(
+          `${this.url}/upload_url/${path}`,
+          { method: 'GET' }
+        )
+
+        const { signedUrl } = await res.json()
```
jeffsee55, line 114: So glad to see we could do something like this, very clean 👍

packages/@tinacms/toolkit/src/packages/core/media.ts
```diff
@@ -272,3 +272,22 @@ export class MediaListError extends Error {
     this.docsLink = config.docsLink
   }
 }
+
+export const E_UNAUTHORIZED = new MediaListError({
```
jeffsee55, line 276: These error messages are kind of a mixture between Tina Cloud and Cloudinary messages (the docs links will obviously get updated when we have them)

jeffsee55 (commented):
This is awesome, testing locally works great for me.

logan-anderson:
Just want to make sure we have the same understanding. 

When we initially scoped this, I was under the impression that this was only for preview images and that it would always use the localURL when saved to the content.

Right now this does not seem to be the case.

When add a new image using the media manager (in production editing environment) the saved file looks like this.

```
"imageURL": "https://assets.tina.io/05477644-3e42-44dd-87ea-ab81992db347/Screen Shot 2022-06-01 at 1.05.42 PM.png"
```

and I would expect that it would use that URL when in edit mode but when it saves the content it would look like this.

```
"imageURL": "/Screen Shot 2022-06-01 at 1.05.42 PM.png"
```

This way I am not tied to using Tina's URL and it would not effect me if our DAM wen't down. 

I know we talked about always using the DAM in the future but I _don't think_ this was supposed to default implementation

Let me know if this makes sense.


cc @kldavis4 @Enigmatical @jeffsee55

Enigmatical:
It does and I believe you are correct, @logan-anderson.

The switching to the DAM link is made on the backend (that's the bit of work I did in `tina-cloud` with the `useRelativeMedia` flag).

We should _always_ be saving the non-DAM link in the document.  This way, if I am working locally (and do not have access to the DAM), my files will still work (pulling from the filesystem).

In fact, having the full DAM link in the document could lead to a larger problem where the `value` (the full DAM link) is appended to the DAM url:
```
https://assets.tina.io/05477644-3e42-44dd-87ea-ab81992db347/https://assets.tina.io/05477644-3e42-44dd-87ea-ab81992db347/Screen Shot 2022-06-01 at 1.05.42 PM.png
```

kldavis4:
@logan-anderson @Enigmatical sounds like a bug ? can someone recommend a fix?

Enigmatical:
> @logan-anderson @Enigmatical sounds like a bug ? can someone recommend a fix?

Let's huddle and talk about it on Monday, but I'm thinking a simple regex to remove the:
```https://<assets-url>/<client-id>```
when on `TinaCloud` might suffice.

That should leave us with only the `media` path, store that.  When that `value` is then looped through the backend, it will get the `assets.tina.io` part added back in the preview.

Enigmatical (approved):

PR #2941, opened 2022-05-31
Remove Form Header Label

spbyrne:
When there's a single form, the label doesn't provide useful context, so I've removed it.

Screen Shot 2022-05-31 at 10 37 41 AM


jeffsee55:
When I click on an item from the admin and get taken to the contextual editing page, it feels like it'd make sense for me to still see the title to reinforce that I'm in the right place.

spbyrne:
Yeah I'll take a look at how things render in admin and think about this a bit.

PR #2943, opened 2022-06-01
Feat/better error message in gql

### LINKED ISSUE #2880: Difficult to debug invalid MD for a file in a collection
## Description

When there is an error in a collection's file, it's difficult to debug which file has the issue

https://www.loom.com/share/30bd09e574aa489e8264a824f0299767

### Steps to reproduce

Add "invalid" content to a`.md` file in a collection. E.g:

```md
{{ WarningCallout text="" }}
```

Make sure the body is set up with the rich-text field

```
{
        type: "rich-text",
        name: "body",
        label: "body",
        isBody: true,
}
```

Try and navigate to that collection in the CMS (or run `yarn run tinacms audit`)

### Expected result

- The file(s) that is causing the issue should be included in the error message
- If possible, improve the error message as well
- Add error to console
- Add error to cms alerts

### Actual result

Ambiguous error is logged, without knowing which file is causing the issue.

#### Tinacms Package Versions

├── @tinacms/auth@0.50.2
├── @tinacms/cli@0.60.15
├── @tinacms/graphql@0.60.0
├── next-tinacms-cloudinary@3.5.12
├── react-tinacms-editor@0.53.9
├── tinacms@0.68.3

logan-anderson:
fixes #2880


packages/@tinacms/graphql/src/database/index.ts
```diff
@@ -448,26 +453,36 @@ export class Database {
...
+      return {
+        edges: await sequential(edges, async (edge) => {
+          try {
+            const node = await hydrator(edge.path)
+            return {
+              node,
+              cursor: btoa(edge.cursor),
+            }
+          } catch (error) {
+            console.error(`error in fetching ${edge.path}`)
```
jeffsee55, line 472: Rather than logging this out here, would it be better to just throw a new Error with the more contextual message? This way you don't get the bubbling up of console.log statements:

```
error in fetching content/posts/anotherPost.mdx
error in querying posts
```
And instead could have one error that actually makes sense like
```
AuditError: Error while reading "content/posts/anotherPost.mdx"
  Could not parse expression with acorn: Unexpected token
    ... rest of stack trace
```
Also it might be nice to only show the stack trace when a `--verbose` flag is passed

packages/@tinacms/graphql/src/mdx/parse.ts
```diff
@@ -329,12 +330,8 @@ const parseField = (attribute, field: TinaField, props) => {
            *  </>}
            * />
            */
-          try {
```
logan-anderson, line 332: @jeffsee55  is it ok to remove this try catch?

packages/@tinacms/graphql/src/database/index.ts
logan-anderson: @jeffsee55 Agreed, fixed.

.changeset/wild-mice-sit.md
```diff
@@ -0,0 +1,5 @@
+---
+'@tinacms/graphql': patch
+---
+
+Added more context to error messages to help to user debug issues`
```
kldavis4, line 5:
```suggestion
Added more context to error messages to help to user debug issues
```

packages/@tinacms/graphql/src/database/index.ts
```diff
@@ -211,27 +212,32 @@ export class Database {
...
       }
-      await this.store.put(normalizePath(filepath), payload, {
-        keepTemplateKey,
-        collection: collection,
-        indexDefinitions: collectionIndexDefinitions,
-      })
+      return true
+    } catch (error) {
+      console.error(`Error in PUT for ${filepath}`)
+      throw error
```
jeffsee55, line 239: Is `console.log` the primary way of communicating an error happened? I think ideally the error message speaks for itself, so in this scenario wed' throw a new error with a more clear message than the one we're catching, and then not also log the error.

packages/@tinacms/graphql/src/database/index.ts
```diff
@@ -696,12 +716,17 @@ const _indexContent = async (
...
+      const dataString = await database.bridge.get(normalizePath(filepath))
+      const data = parseFile(dataString, path.extname(filepath), (yup) =>
+        yup.object({})
+      )
+      if (database.store.supportsSeeding()) {
+        await database.store.seed(normalizePath(filepath), data, seedOptions)
+      }
+    } catch (error) {
+      console.error(`Unable to seed ${filepath}`)
+      throw error
```
jeffsee55, line 729: Same as the comment above, it seems like the error we throw should be more user-friendly, and we wouldn't need the console log.

packages/@tinacms/graphql/src/resolve.ts
```diff
@@ -77,279 +81,303 @@ export const resolve = async ({
...
+
+          if (verboseValue) {
+            console.error('More error connext below')
+            console.error(e.message)
+            console.error(e)
+          }
+        })
+        // }
+
+        // if (!verbose) {
```
jeffsee55, line 374: Is this code finished?

packages/@tinacms/graphql/src/database/index.ts
logan-anderson: Fixed

packages/@tinacms/graphql/src/database/index.ts
logan-anderson: For sure! Fixed.

packages/@tinacms/graphql/src/resolve.ts
```diff
@@ -81,279 +85,291 @@ export const resolve = async ({
...
     })
 
     if (res.errors) {
       if (!silenceErrors) {
-        console.error(res.errors)
+        res.errors.map((e) => {
+          console.error(e.toString())
+
+          if (verboseValue) {
+            console.error('More error connext below')
```
jeffsee55, line 368:
```suggestion
            console.error('More error context below')
```

jeffsee55 (approved):

PR #2945, opened 2022-06-02
Always use query when data layer is enabled.

### LINKED ISSUE #2777: Data layer: queries which use glob are not limited
## Description

Any queries which do not specify a `sort` or a `filter` arg use store.glob(), which won't limit the results. For a sufficiently large collection, this request may take a long time to complete and may timeout.

(Always using  ~`store.query()` (which has a default limit) might be the best solution)

Additionally, pageInfo is not populated, so it is not possible to implement pagination without using sort or filter args

Note:
No expected change to the resolver, only changes in the data store supporting the data layer

logan-anderson:
fixes #2777

### DIFF
--- a/.changeset/quick-teachers-grow.md
+++ b/.changeset/quick-teachers-grow.md
@@ -2,4 +2,4 @@
 '@tinacms/graphql': patch
 ---
 
-Always use the glob function when the dataLayer in enabled
+Always use the query function when the dataLayer in enabled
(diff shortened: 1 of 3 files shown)


kldavis4:
@logan-anderson is this 'always use glob' or 'always use query'?

logan-anderson:
> @logan-anderson is this 'always use glob' or 'always use query'?

@kldavis4 You are correct. Updated

kldavis4 (approved):

PR #2971, opened 2022-06-13
Added new config key to schema

logan-anderson:
Adds a new config key to the schema that is used for media config.

It is also validated and passed to `_schema.json`


kldavis4 (approved):
lgtm

Enigmatical (approved):

PR #2975, opened 2022-06-13
2974 - Uses new config for resolving media URLs

Enigmatical:
Uses the new `schema.ts` key: `config: { media: { tina: { ... } } }` for resolving relative to cloud media URLs.

This required passing the `tinaSchema.schema` into every function that was already responsible for resolving (including `parseMDX` and `stringifyMDX`.

I also added some `config` validation and clean up functions for `syncFolder`.

I removed `--tinaCloudMediaStore` from the `cli`, too.

I also added new tests!


packages/@tinacms/graphql/src/mdx/parse.ts
```diff
@@ -23,16 +23,18 @@ import { TinaField } from '..'
 import type { Content } from 'mdast'
 import { visit } from 'unist-util-visit'
 import type { GraphQLConfig, RichTypeInner } from '../types'
+import type { TinaCloudSchemaEnriched } from '@tinacms/schema-tools'
 import { isNull } from 'lodash'
 import { resolveMediaRelativeToCloud } from '../resolver/media-utils'
 
 export const parseMDX = (
```
logan-anderson, line 30: More of a design comment,

Would it make sense to refactor this into a class, that way `graphQLconfig` and `schema` would not need to be passed around everywhere.

packages/@tinacms/graphql/src/resolver/media-utils.ts
```diff
@@ -23,19 +24,28 @@ import { GraphQLConfig } from '../types'
...
+  config: GraphQLConfig = { useRelativeMedia: true },
+  schema: TinaCloudSchemaEnriched
 ) => {
   if (config && value) {
     if (config.useRelativeMedia === true) {
       return value
-    } else {
+    }
+
+    if (isValidTinaSchema(schema) === true) {
```
logan-anderson, line 35: This name was a little confusing at first. Maybe it could be renamed to something that describes what it is doing a bit better?
Maybe
- hasSchemaMediaConfig ?
- isValidSchemMediaConig ?

packages/@tinacms/graphql/src/resolver/media-utils.ts
```diff
@@ -23,19 +24,28 @@ import { GraphQLConfig } from '../types'
...
       const assetsURL = `https://${config.assetsHost}/${config.clientId}`
       if (typeof value === 'string' && value.includes(assetsURL)) {
-        return value.replace(assetsURL, '')
-      } else {
-        return value
+        const cleanSyncFolder = cleanUpSlashes(
```
logan-anderson, line 38: I also made a very similar  function. We should probable use the same one once everything is merged.

logan-anderson:
@Enigmatical What is the best way to test this? (outside regression testing) With out the Tina Cloud media store and my PR it seems like it might be hard to test (This is OK just wanted to ask if you know if anything we can do to test )

packages/@tinacms/graphql/src/resolver/media-utils.ts
Enigmatical: Agreed.  Changed to `hasTinaMediaConfig` 👍

logan-anderson: awesome. When that is cleaned up we can merge it.

logan-anderson (approved):

Issue #2976, opened 2022-06-14
Media does not restart when schema is re-generated on save.


logan-anderson:
Media does not restart when schema is re-generated on save.

Steps to re-produce

1. have a schema with a media config
2. update the media config and save your schema file
3. See the schema is re-generated but  since the [GQL server is not restarted](https://github.com/tinacms/tinacms/blob/main/packages/%40tinacms/cli/src/cmds/start-server/index.ts#L240) the media config is not updated

To fix this we will have to restart the media server on save.

PR #2977, opened 2022-06-14
Feat/config local media endpoints

### LINKED ISSUE #2973: Repo-Based Media: Support `config` in `localhost` endpoints
With this PR (https://github.com/tinacms/tinacms/pull/2971), we introduced `schema.config.media.tina` to determine public and upload paths.  We'll need to make sure our local endpoints recognize and use them.

logan-anderson:
fixes #2973

~~But exposes [a small bug](https://github.com/tinacms/tinacms/issues/2976). If we are not ok with merging this with this bug I can fix it in this PR (Just did not want to scope creep and make this PR take a long time)~~

Update: I fixed that bug as well.


packages/@tinacms/cli/src/cmds/start-server/index.ts
```diff
@@ -192,6 +192,8 @@ export async function startServer(
   let isReady = false
 
   const start = async () => {
+    // we do not want to start the server while the schema is building
+    await lock.promise
```
kldavis4, line 196: is there a risk that the lock will be taken right after this? should this hold the lock until it starts?

kldavis4 (commented):
minor nit, but calling the module `removeStartingAndTrailingSlash` makes that a single function module. Maybe throw it in utils/index.ts?

packages/@tinacms/cli/src/cmds/start-server/index.ts
logan-anderson: I don't think there is a risk of this since the schema compiling always starts before the server.

logan-anderson:
> minor nit, but calling the module `removeStartingAndTrailingSlash` makes that a single function module. Maybe throw it in utils/index.ts?

Good point. Fixed.

Enigmatical (approved):
LGTM 👍

PR #2978, opened 2022-06-14
Fix Login Styles

spbyrne:
The css preflight was only being included in `@tinacms/toolkit` and not `tinacms` given it was assumed that would always be wrapped in the toolkit's styles, but this isn't the case of at least the login screen. I just copied it over, but it's possible the `tinacms` package could import the preflight from the toolkit.

Before: 
Screen Shot 2022-06-14 at 12 19 57 PM

After:
Screen Shot 2022-06-14 at 12 16 52 PM

### DIFF
--- a/.changeset/red-lemons-mate.md
+++ b/.changeset/red-lemons-mate.md
@@ -1,0 +1,6 @@
+---
+'@tinacms/toolkit': patch
+'tinacms': patch
+---
+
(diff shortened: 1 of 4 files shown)


logan-anderson:
@spbyrne I think you forgot to add a changeset. 

They can be added with

`yarn changeset` in the monorepo.

PR #2984, opened 2022-06-14
2974 - Uses DummyMediaStore if config.media.tina is not available

Enigmatical:
Before, we were defaulting to `TinaMediaStore` regardless if `--tinaCloudMediaStore` existed or not.

Now that the flag is gone (and replaced by `config.media.tina`), we only want to use the `TinaMediaStore` if `tina` is provided.

Otherwise, use `DummyMediaStore`.


logan-anderson (approved):
LGTM, as long as we make sure to update the dummy media store content when we have the correct docs

PR #2995, opened 2022-06-17
media manage uses relieve dir paths

logan-anderson:
It turns out out media manage expects the paths of dirs to be relative to the parent. The CLI was returning local media folders with full paths which was causing issues.


perkinsjr (approved):

Enigmatical (approved):
LGTM!

PR #2999, opened 2022-06-17
Enables pagination for local media `list` endpoint

Enigmatical:
Basically uses `offset` and `limit` by treating the `offset` value like a `cursor`.

Tested with both 26 and 6 items and appears to be working correctly for both.

### DIFF
--- a/packages/@tinacms/cli/src/server/models/media.ts
+++ b/packages/@tinacms/cli/src/server/models/media.ts
@@ -17,6 +17,8 @@ import { parseMediaFolder } from '../../utils/'
 
 interface MediaArgs {
   searchPath: string
+  cursor?: string
+  limit?: string
 }
 
 interface File {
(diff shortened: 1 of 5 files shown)


packages/@tinacms/toolkit/src/packages/fields/plugins/BlocksFieldPlugin/BlockSelectorBig.tsx
```diff
@@ -167,6 +167,7 @@ export const BlockSelectorBig = ({
                         <CardColumns className="pt-3">
                           {uncategorized.map(([name, template]) => (
                             <BlockCard
+                              key={`${template}-${name}`}
```
Enigmatical, line 170: Noticed I was getting a nasty `missing key` React Warning when using the `tina-cloud-starter`; this should fix that.

packages/@tinacms/cli/src/server/models/media.ts
```diff
@@ -92,11 +94,21 @@ export class MediaModel {
...
 
-      const files = await Promise.all(filesProm)
+      const offset = Number(args.cursor) || 0
+      const limit = Number(args.limit) || 20
+
+      const allItems = await Promise.all(filesProm)
+      const files = allItems.filter((x) => x.isFile)
+      const directories = allItems.filter((x) => !x.isFile).map((x) => x.src)
+
+      const limitFiles = files.slice(offset, offset + limit)
```
kldavis4, line 104: @Enigmatical in the tina-cloud implementation, for paging purposes, there is no distinction between files and directories. Here it looks like if you specify a limit of 20, and have say 25 directories, you will end up with more than 20 items in the list

packages/@tinacms/toolkit/src/packages/fields/plugins/BlocksFieldPlugin/BlockSelectorBig.tsx
logan-anderson: 👍

logan-anderson (commented):
Looks good to me as soon as kelly's comment is addressed.

packages/@tinacms/cli/src/server/models/media.ts
Enigmatical: Ah, good catch.  I didn't realize directories counted against the "page" limit.  Easy to solve.

kldavis4 (approved):
lgtm

PR #3005, opened 2022-06-21
fix: relativePath calculation

kldavis4:
Fixes an issue when pathToTina is specified as rootPath


jeffsee55 (approved):
@kldavis4 looks good just needs a changeset so it will publish

kldavis4:
> @kldavis4 looks good just needs a changeset so it will publish

@jeffsee55 does it need a changeset if the other isomorphic stuff hasn't been published yet?

PR #3011, opened 2022-06-23
Fix rich text focus styles

spbyrne:
Given the rich text input styles are provided by a wrapper div, I've changed `focus` for `focus-within` so the styles will work properly. 

Screen Shot 2022-06-23 at 2 52 46 PM


jeffsee55 (approved):
Nice!

PR #3018, opened 2022-06-27
Move to turbo

jeffsee55:
This PR does a few things, the actual meaningful changes are best seen by looking at the commites rather than the whole PR because the first commit removes the `.yarn` folder so it's bascially impossible to view on Github. 

- Removes `yarn` in favor of `pnpm`
- Publishes packages as es modules. @logan-anderson I expect this will overlap with the work for standalone, vite didn't work for me until I added this. 
- Removes the `react-tinacms-editor` package, we haven't touched it over a year - if we really need to fix something with it we can pull it up from an older git tag
- Skips some of the tests in `@tinacms/graphql`. Jest and ESM don't mix well, Yarn pnp was doing some magic to make it work but I honestly don't even understand how it was working. The reason the issue only comes up here is because of the [unified won't publish CJS versions of their packages](https://github.com/unifiedjs/unified/issues/121). Considering a move to more modern testing lib like [uvu](https://github.com/lukeed/uvu) for these tests.
- Pins `typescript` to `4.3.5`. This was a constraint of being stuck on Yarn 2, we can move up but there were some type errors so should do that incrementally.


.github/workflows/e2e-dev.yml
```diff
@@ -12,15 +12,19 @@ jobs:
...
       - uses: actions/checkout@v2
+      - uses: pnpm/action-setup@v2.0.1
+        with:
+          version: 7.4.0
+          run_install: true
       - name: Cypress run
         uses: cypress-io/github-action@v2
         with:
-          install-command: yarn install
+          install: false
```
jeffsee55, line 22: cc @logan-anderson just FYI the cypress image doesn't know about pnpm, so we do the install ourselves.

packages/@tinacms/graphql/tsconfig.json
```diff
@@ -2,8 +2,7 @@
   "extends": "../../../tsconfig.json",
   "compilerOptions": {
     "outDir": "dist",
-    "rootDir": "src"
   },
-  "exclude": ["node_modules", "dist", "src/**/*.test.tsx", "src/**/*.test.ts", "src/**/*.spec.ts", "src/**/*.spec.tsx"],
+  "exclude": ["node_modules", "dist", "src/**/*.test.tsx", "src/**/*.test.ts", "src/**/*.spec.ts", "src/**/*.spec.tsx","src/**/*.skip.ts"],
```
jeffsee55: FYI I renamed the files from `.spec` to `.skip` so jest wouldn't pick them up

packages/@tinacms/scripts/src/index.ts
```diff
@@ -123,6 +124,35 @@ export const run = async (args: { watch?: boolean; dir?: string }) => {
   }
 }
 
+const watch = () => {
```
jeffsee55: Same watch logic as before, but easier with `pnpm` thanks to it's `list` command.

logan-anderson:
> Removes the react-tinacms-editor package, we haven't touched it over a year - if we really need to fix something with it we can pull it up from an older git tag

Could we also move it into deprecated packages folder?

logan-anderson (dismissed):
Looks good to me. It might be good for another dev to run thought it as well.

cc @kldavis4 @spbyrne

experimental-examples/tina-cloud-starter/pages/[filename].tsx
```diff
@@ -33,6 +33,7 @@ export const getStaticProps = async ({ params }) => {
 
 export const getStaticPaths = async () => {
   const pagesListData = await client.queries.pagesConnection();
+  console.log("gsp", pagesListData.data);
```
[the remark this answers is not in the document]
Comment (jeffsee55) [reply experimental-examples/tina-cloud-starter/pages/[filename].tsx]: This was fixed in this commit 0ed5c3db51eb761a7e01823cff246f49929a880f

jeffsee55:
> also noticed the token/team env vars for turborepo and curious about how turbo is being used in the project. Is it replacing any existing build functionality?

It sets up a cache that allows to reuse old build artifacts if nothing has changed from the previous run ([docs](https://turborepo.org/docs/ci/github-actions)). You can see the output for a cache hit here https://github.com/tinacms/tinacms/runs/7122366235?check_suite_focus=true. One thing I hadn't considered is that this may be nullifying the value of running these builds on multiple environments. If it runs the build on Node 12 ubuntu for example, will it just reuse those artifacts on the others 🤔 . I'll go ahead and merge this but I've asked about that in the turborepo Discord

packages/@tinacms/datalayer/src/database/store/level.ts
```diff
@@ -50,6 +50,7 @@ export class LevelStore implements Store {
     this.rootPath = rootPath || ''
     this.useMemory = useMemory
     if (useMemory) {
+      // @ts-ignore FIXME: something from the upgrade is now causing `errors` to be missing
```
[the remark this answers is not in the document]
jeffsee55: I think this was fixed when I pinned Typescript to 4.3.5

kldavis4 (approved):

jeffsee55:
> Could we also move it into deprecated packages folder?

I think we should remove that folder now, we haven't touched in ages either and none of the code would build properly anymore anyway.

Issue #3004, opened 2022-06-21
Repo-Based Media:  Introduce a button to sync media with Tina Cloud


Enigmatical:
Today, `tina-cloud` only syncs _new_ media added after the `config` is in place.  

It **does not** sync existing files (which can create a confusing situation when using the `MediaManager`).

Instead, we would like to introduce a "Sync My Media" button to the `MediaManager` that is only available in Production that would let the user trigger a sync on their `mediaRoot`.

This would compose of changes to the `MediaManager`, the `MediaStore`, and backend changes to `tina-cloud`.

NOTES
- The client should specify the branch
- will not delete anything from the cloud folder

PR #3019, opened 2022-06-28
Feat/add media sync button

logan-anderson:
Frontend part of #3004.
FIXES ENG-371

Will be a draft until @spbyrne  has time to make it look good.


packages/@tinacms/toolkit/src/components/media/media-manager.tsx
```diff
@@ -197,6 +264,53 @@ export function MediaPicker({
...
+              const totalDone = Object.values(status.status).filter(
+                Boolean
+              )?.length
+              const total = Object.keys(status.status)?.length
+              setLoadingText(`${totalDone}/${total} Media items loaded`)
+              return status
+            },
+            3000,
+            // Will time out after 10 seconds
+            10000
```
kldavis4, line 289: might want this to be higher than 10 seconds. we can test & tweak later, but when I was testing w/ a large amount of big-ish files, it definitely took more than 10 seconds.

kldavis4 (dismissed):
code looks good - just commented on the timeout

packages/@tinacms/toolkit/src/components/media/media-manager.tsx
logan-anderson: Sounds good!

kldavis4 (approved):

PR #3022, opened 2022-06-29
Feat/new media config

logan-anderson:
This PR
- Adds new media config
- Adds zod checks for the new config
- Fixes a bug with to many restarts of the `schema.ts` file


kldavis4 (commented):
this looks good to me. The only issue I noticed is that if you leave the mediaStore in the config and enable tina media store in the schema config, the mediaStore will take precedence. It seems like it might be better (if possible) to have the tina store take precedence.

PR #3031, opened 2022-07-05
better errors in the client

logan-anderson:
New error message in the client that links to the FAQ and also throws an error message when a non 200 status code is returned.

Screen Shot 2022-07-05 at 10 04 39 AM

### DIFF
--- a/.changeset/quiet-moose-return.md
+++ b/.changeset/quiet-moose-return.md
@@ -1,0 +1,5 @@
+---
+'tinacms': patch
+---
+
+Client now throws an error when a non 200 status code is returned.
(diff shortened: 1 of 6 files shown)


jamespohalloran:
Sweet! Do you know if this also has the same behavior for the `client.queries.collection` requests?

logan-anderson:
> Sweet! Do you know if this also has the same behavior for the `client.queries.collection` requests?

Right now its a warning because we do a `try...catch` and swallow the error so document creation works. Should we remove it? (This would be a small and technically breaking change) but would probably be a better behaviour.

logan-anderson:
> > > Sweet! Do you know if this also has the same behavior for the `client.queries.collection` requests?
> > 
> > 
> > Right now its a warning because we do a `try...catch` and swallow the error so document creation works. Should we remove it? (This would be a small and technically breaking change) but would probably be a better behaviour.
> 
> In our docs right now, most of our docs have the data-fetching call already wrapped in a try/catch to handle that (although looks like some of the starters don't have it.) I'd say it's worth the small breaking change though, as when we soon move to API-first, that try/catch won't be required at all.

Makes sense to me! updated PR

PR #3051, opened 2022-07-25
Feat/use ts compile

logan-anderson:
Generate a `types.js` and a `types.ts` so that the client will work with JS and TS sites.

I know the "correct way" would be to generate a types.d.ts and a types.js but that compilation took around 10 seconds.


packages/@tinacms/cli/src/cmds/start-server/index.ts
```diff
@@ -201,7 +202,7 @@ export async function startServer(
           process.exit(0)
         }
       })
-      .on('all', async () => {
```
logan-anderson, line 204: this can be undone

logan-anderson:
closing in favour of https://github.com/tinacms/tinacms/pull/3052

PR #3052, opened 2022-07-25
FIX: Allow new client to work with js and ts next.js projects

logan-anderson:
If a `schema.js` is provided  then 
- `types.js`
- `types.d.s`
- `client.js`
is generated. 

If a `schema.{ts,tsx}` is provided then
- `types.ts`
- `client.ts`
is generated.


packages/@tinacms/cli/src/cmds/compile/index.ts
```diff
@@ -36,15 +38,16 @@ export const resetGeneratedFolder = async ({
     console.log(e)
   }
   await fs.mkdir(tinaGeneratedPath)
+  const etx = usingTs ? 'ts' : 'js'
```
kldavis4, line 41: should this be ext?

.changeset/pretty-ties-lick.md
```diff
@@ -0,0 +1,5 @@
+---
+'@tinacms/cli': patch
+---
+
+Genearte types and client based on the file type of the schema
```
kldavis4, line 5:
```suggestion
Generate types and client based on the file type of the schema
```

packages/@tinacms/cli/src/buildTina/index.ts
```diff
@@ -167,6 +167,7 @@ export const build = async ({
   skipIndex,
 }: BuildOptions) => {
   const rootPath = ctx.rootPath as string
+  const tinaGeneratedPath = path.join(rootPath, '.tina', '__generated__')
```
kldavis4, line 170: should this be down below the rootPath null check?

packages/@tinacms/cli/src/cmds/compile/index.ts
logan-anderson: yup, fixed.

packages/@tinacms/cli/src/buildTina/index.ts
logan-anderson: Yes, fixed.

jamespohalloran (approved):
Tested this out in the barebones starter and it's working nicely!

PR #3059, opened 2022-07-27
Fix Audit cmd

logan-anderson:


jamespohalloran:
Other than that one rogue log noted above, I made a PR deploy and tested this locally, seems to be working!

logan-anderson:
> Other than that one rogue log noted above, I made a PR deploy and tested this locally, seems to be working!

@jamespohalloran I just removed it. Thanks.

jamespohalloran (approved):

PR #3066, opened 2022-07-28
Throw an error message when name contains spaces

### LINKED ISSUE #3065: TinaCMS hangs if there is a space in a field name
## How to reproduce

Add a space (`" "`) to a field name

EX
```
 fields: [
        { name: 'sub text', label: 'Sub text', type: 'string' },
//...
```

See output of `yarn dev` hangs

```
yarn run v1.22.17
$ tinacms dev -c "next dev
```

## Solution

Check for a space in the name when validation the schema and throw an error with a useful error message.

logan-anderson:
fixes #3065

### DIFF
--- a/.changeset/slow-swans-brake.md
+++ b/.changeset/slow-swans-brake.md
@@ -1,0 +1,5 @@
+---
+'@tinacms/schema-tools': patch
+---
+
+Throw an error message when name contains spaces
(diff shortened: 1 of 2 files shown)


jeffsee55 (approved):
Looks good, but it still sounds like we're swallowing errors from `@tinacms/graphql` somewhere. Instead of hanging I would expect to see a low-level error message

PR #3071, opened 2022-08-03
Sort by any field in cms

### LINKED ISSUE #2774: Sort posts by date in CMS
## Summary

As a content editor, I want to edit the **most recent post** so that I can fix a typo that just went out. 

The problem is that I don't know the filename, and can't easily find this post. I can't sort by date, nor can I search the filename (related to [this issue](https://github.com/tinacms/tinacms/issues/2723))

Screen Shot 2022-04-05 at 11 33 56 AM

### Expected Results

* See post title and meta information in list (date last edited, author, etc)
* Sort by fields

logan-anderson:
This adds the ability to sort in the CMS

closes #2774

### DIFF
--- a/packages/@tinacms/graphql/src/database/index.ts
+++ b/packages/@tinacms/graphql/src/database/index.ts
@@ -483,7 +483,10 @@ export class Database {
             cursor: btoa(edge.cursor),
           }
         } catch (error) {
-          if (error instanceof Error) {
+          if (
+            error instanceof Error &&
+            !edge.path.includes('.tina/__generated__/_graphql.json')
+          ) {
             throw new TinaQueryError({
               originalError: error,
               file: edge.path,
--- a/packages/tinacms/src/admin/pages/CollectionListPage.tsx
+++ b/packages/tinacms/src/admin/pages/CollectionListPage.tsx
@@ -40,6 +40,7 @@ import { TinaAdminApi } from '../api'
 import { useState } from 'react'
 import { CursorPaginator } from '@tinacms/toolkit'
 import { useEffect } from 'react'
+import type { TinaCloudCollection } from '@tinacms/schema-tools'
 
 const TemplateMenu = ({ templates }: { templates: Template[] }) => {
   return (
@@ -126,6 +127,7 @@ const CollectionListPage = () => {
   })
   const [endCursor, setEndCursor] = useState('')
   const [prevCursors, setPrevCursors] = useState([])
+  const [sortKey, setSortKey] = useState('')
   const loc = useLocation()
   useEffect(() => {
     // reset state when the route is changed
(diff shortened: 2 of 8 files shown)


jamespohalloran:
Sweet!! This one's going to be so helpful.

Small thing @spbyrne (and don't think it needs to hold this PR up) but the loading state flash is a bit jarring:
https://www.loom.com/share/72947a36c6d44874a04f899250113369

packages/@tinacms/graphql/src/database/index.ts
```diff
@@ -483,7 +483,10 @@ export class Database {
             cursor: btoa(edge.cursor),
           }
         } catch (error) {
-          if (error instanceof Error) {
+          if (
+            error instanceof Error &&
+            !edge.path.includes('.tina/__generated__/_graphql.json')
```
kldavis4, line 488: @logan-anderson is this change related to the sorting feature?

packages/tinacms/src/admin/pages/CollectionListPage.tsx
```diff
@@ -174,11 +208,70 @@ const CollectionListPage = () => {
...
+                                    label: 'No Selection',
+                                    value: JSON.stringify({
+                                      order: 'asc',
+                                      name: '',
+                                    }),
+                                  },
+                                  ...fields
+                                    .map((x) => [
+                                      {
+                                        label: x.label + ' (Acscending)',
```
kldavis4, line 239:
```suggestion
                                        label: x.label + ' (Ascending)',
```

kldavis4:
Instead of "No Selection" should it say Default maybe (which is the filename sort)?

jamespohalloran:
> Instead of "No Selection" should it say Default maybe (which is the filename sort)?

I may be wrong on this, but I thought someone said recently the default sort was the isTitle field (if it's set)

kldavis4:
> > Instead of "No Selection" should it say Default maybe (which is the filename sort)?
> 
> I may be wrong on this, but I thought someone said recently the default sort was the isTitle field (if it's set)

yeah, I think if isTitle is set then that field is the default, otherwise it is the filename. Either way, a "default" sort is applied - "No selection" seems to be saying that there is no sort.

packages/@tinacms/graphql/src/database/index.ts
logan-anderson: Sorta. I was getting a strange error message when an invalid sort key was given. This changes to just the raw error message

logan-anderson:
> Sweet!! This one's going to be so helpful.
> 
> Small thing @spbyrne (and don't think it needs to hold this PR up) but the loading state flash is a bit jarring: https://www.loom.com/share/72947a36c6d44874a04f899250113369

@jamespohalloran Do you have a suggestion of how to fix this? In production the loading will take a bit longer and the loading message will be very helpful and feel better.

logan-anderson:
> > > Instead of "No Selection" should it say Default maybe (which is the filename sort)?
> > 
> > 
> > I may be wrong on this, but I thought someone said recently the default sort was the isTitle field (if it's set)
> 
> yeah, I think if isTitle is set then that field is the default, otherwise it is the filename. Either way, a "default" sort is applied - "No selection" seems to be saying that there is no sort.

@jamespohalloran @kldavis4 

The default is the 'IsTitle" field if there is no isTitle field then it will be the file name. 

I can change is to "Default" instead of the "No selection"

kldavis4 (approved):

jamespohalloran:
> 

@logan-anderson Here's how Linear does it: https://www.loom.com/share/021159a491b744ca8f6a9fb0cd608cdc
It just looks a bit smoother without the global loading state flicker.

PR #3073, opened 2022-08-03
Feat/always reset

logan-anderson:
Always reset the generated folder even when using CI.

This is to fix an issue where the a "dummy" client.ts file was not being created and that would cause the the schema to not be able to compile.

### DIFF
--- a/.changeset/spotty-ads-help.md
+++ b/.changeset/spotty-ads-help.md
@@ -1,0 +1,5 @@
+---
+'@tinacms/cli': patch
+---
+
+Always reset the generated folder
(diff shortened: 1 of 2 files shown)


kldavis4:
@logan-anderson can you provide some info on the reason for this change? Is it related to the local mode issue that @jamespohalloran was having?

logan-anderson:
> @logan-anderson can you provide some info on the reason for this change? Is it related to the local mode issue that @jamespohalloran was having?

Yes it is related. I updated the comment

kldavis4 (approved):

PR #3103, opened 2022-08-17
Treat blockquote and code blocks inside list items as invalid

jeffsee55:
Update markdown logic to treat blockquote and code blocks inside list items as invalid since our rich-text editor can't support it.

PR #3104, opened 2022-08-17
Improved MDX & Audit error handling

jamespohalloran:
Currently, the document list blows up if there's an unexpected error in the body of any individual document's rich-text. 
This will handle those unexpected errors that we aren't specifically handling better, so we'll display an error instead in the rich-text field.

Screen Shot 2022-08-17 at 3 29 35 PM

Screen Shot 2022-08-17 at 3 26 24 PM

To test, Add this invalid rich-text body to a markdown file:
https://github.com/tinacms/tina.io/blob/master/content/docs/contributing/releasing.md

---

Update: the audit functionality also needed to be updated with this change. Since we were letting requests with rich-text issues move the content api, we needed to update to CLI audit to still capture these issues.

Other than some general refactoring, the user-facing changes to audit:
- Fixed a bug where we were only auditing first 10 docs per collection
- Now that collection lists can be long, I'm only printing the actual warnings/errors
- We're now more likely to log ALL the document's issues, and not just break & stop on the first issue.
- We display the count of errors/warnings

Screen Shot 2022-08-23 at 3 44 42 PM

fixes ENG-497
fixes ENG-489

### DIFF
--- a/packages/@tinacms/cli/src/cmds/audit/audit.ts
+++ b/packages/@tinacms/cli/src/cmds/audit/audit.ts
@@ -73,13 +73,8 @@ const interateCollectionDocuments = async (
   } while (hasNextPage)
 }
 
-const auditDocument = async (
-  node,
-  args: AuditArgs
-): Promise<{ warnings: string[]; errors: string[] }> => {
-  let errors: string[] = []
+const validateRichText = (node) => {
   let warnings: string[] = []
-
   Object.keys(node._values)
     .map((fieldName) => node._values[fieldName])
     .filter((field) => {
@@ -95,6 +90,10 @@ const auditDocument = async (
       })
     })
 
+  return warnings
+}
+
+const validateMutations = async (node, args: AuditArgs) => {
   const topLevelDefaults = {}
 
   // TODO: account for when collection is a string
@@ -136,8 +135,23 @@ const auditDocument = async (
     silenceErrors: true,
     verbose: true,
   })
-  if (mutationRes.errors) {
-    mutationRes.errors.forEach((err) => {
+
+  return mutationRes.errors
+}
+
+const auditDocument = async (
+  node,
+  args: AuditArgs
+): Promise<{ warnings: string[]; errors: string[] }> => {
+  let errors: string[] = []
+  let warnings: string[] = []
+
+  warnings = [...warnings, ...validateRichText(node)]
+
+  const mutationErrors = await validateMutations(node, args)
+
+  if (mutationErrors) {
+    mutationErrors.forEach((err) => {
       errors.push(err.message)
     })
   }
--- a/packages/@tinacms/cli/src/cmds/audit/audit.ts
+++ b/packages/@tinacms/cli/src/cmds/audit/audit.ts
@@ -15,46 +15,11 @@ import { resolve } from '@tinacms/graphql'
 import type { Database, TinaCloudCollection } from '@tinacms/graphql'
 import p from 'path'
 import { logger } from '../../logger'
-import { assertShape } from '@tinacms/graphql'
-import chalk from 'chalk'
+import { AuditIssue } from './issue'
+import { validateRichText } from './validations/validateRichText'
+import { validateMutations } from './validations/validateMutations'
 
-export interface AuditIssue {
-  level: 'error' | 'warning'
-  print: () => void
-}
-
-abstract class BaseAuditIssue implements AuditIssue {
-  message: string
-  level: 'error' | 'warning'
-  constructor(message: string, level: 'error' | 'warning') {
-    this.level = level
-    this.message = message
-  }
-
-  print() {
-    throw new Error('Not Implemented')
-  }
-}
-
-class AuditError extends BaseAuditIssue {
-  constructor(message) {
-    super(message, 'error')
-  }
-  public print() {
-    logger.error(chalk.red(this.message))
-  }
-}
-
-class AuditWarning extends BaseAuditIssue {
-  constructor(message) {
-    super(message, 'warning')
-  }
-  public print() {
-    logger.warn(chalk.yellowBright(this.message))
-  }
-}
-
-type AuditArgs = {
+export type AuditArgs = {
   collection: TinaCloudCollection
   database: Database
   rootPath: string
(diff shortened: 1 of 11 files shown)


jeffsee55:
Works for me, I was handling them a bit more granularly in this PR https://github.com/tinacms/tinacms/pull/3103 but seems easy enough to just catch all errors. The tradeoff here is that for the errors we don't expect we won't have positional information, which is handly in the raw editor but seems reasonable.

jamespohalloran:
> Works for me, I was handling them a bit more granularly in this PR #3103 but seems easy enough to just catch all errors. The tradeoff here is that for the errors we don't expect we won't have positional information, which is handly in the raw editor but seems reasonable.

Seems like an easy tradeoff to me! We'll go from the CMS completely blowing up, to a small piece of it showing an error. 

One thing I notice though, is that the CLI audit doesn't show an error on the files with invalid rich-text:

Screen Shot 2022-08-18 at 3 54 33 PM

Previously, the audit would blow up on those files, but at least you knew there was an error (Although still not ideal because you'd have to debug 1 file at a time)

Screen Shot 2022-08-18 at 3 57 31 PM

Going to turn this PR to a draft until that's resolved

jeffsee55:
> One thing I notice though, is that the CLI audit doesn't show an error on the files with invalid rich-text:

Are you saying you still expect audit to show the errors but not the GraphQL API? We're now catching the errors and audit just uses the GraphQL API under the hood. We'll have to add some logic for a special use-case.

jamespohalloran:
> Are you saying you still expect audit to show the errors but not the GraphQL API? 

I'd expect the GraphQL API to 200, but return an error with the response. Mainly though, I just think from the editor perspective, it shouldn't blow up the full document list when there's an error in one content file. 
What do you think?

In the screenshot, it's also showing an error for one file in the blog, but not others. Seems inconsistent when things are flagged in the audit.

jeffsee55:
> but return an error with the response

It's tricky because `audit` is meant to catch errors that will stop you from using the CMS, so now that invalid markdown nodes aren't going to break your site (in some cases) it's almost like audit needs another log level: `error` and `warning`, rather than just `error`.

jamespohalloran:
> It's tricky because audit is meant to catch errors that will stop you from using the CMS, so now that invalid markdown nodes aren't going to break your site (in some cases) it's almost like audit needs another log level: error and warning, rather than just error.

Would something like this hold us over?

https://github.com/tinacms/tinacms/pull/3104/files#diff-8f8fb3295b7d89cb3ca29b496bcd17d284345964f41ba3f8df78b676e4a3c945R124-R141
(I'd still like to clean it up a bit and add some tests)

I could see different log levels being useful, but overall I think that just being able to work your way out of when a document is busted is my main goal here. This allows the editors to keep working when document(s) have issues, and it shows all potentially problematic documents in the audit. 

Screen Shot 2022-08-22 at 4 36 11 PM

packages/@tinacms/cli/src/cmds/audit/audit.ts
```diff
@@ -121,6 +121,24 @@ export const auditDocuments = async (args: AuditArgs) => {
...
+      .forEach((fieldName) => {
+        const errorMessages = docResult.data.document._values[
+          fieldName
+        ].children
+          .filter((f) => f.type == 'invalid_markdown')
+          .map((f) => f.message)
+
+        errorMessages.forEach((errorMessage) => {
+          error = true
+          logger.error(chalk.red(errorMessage))
```
jeffsee55, line 138: perhaps this should be a `warn`. [Here's](https://github.com/tinacms/tinacms/blob/392d8870281ee4d081d39cdb37ad4235011b86a2/packages/%40tinacms/cli/src/cmds/audit/audit.ts#L62-L71) an example of that.

jeffsee55 (commented):
Sounds good to me, minor nitpick but whatever you think makes sense.

packages/@tinacms/cli/src/cmds/audit/audit.ts
```diff
@@ -11,232 +11,70 @@ See the License for the specific language governing permissions and
...
           }
-        }
-      }
-      `
-  const result = await resolve({
-    database,
-    query,
-    variables: {},
-  })
-  const format = result.data.collection.format
```
jamespohalloran, line 54: This chunk of code stopped actually doing anything, since documents not matching the collections format are no longer included in the response

packages/@tinacms/cli/src/cmds/audit/validations/validateMutations.ts
```diff
@@ -0,0 +1,122 @@
+import { AuditError, AuditIssue } from '../issue'
+import { assertShape } from '@tinacms/graphql'
+import { AuditArgs } from '../auditArgs'
+
+export const validateMutations = async (
```
jamespohalloran, line 5: Mostly unchanged, but pulled this section in its own file to keep the higher level "audit" clean

packages/@tinacms/cli/src/cmds/audit/audit.ts
```diff
@@ -11,232 +11,70 @@ See the License for the specific language governing permissions and
...
-          format
-          documents {
-            edges {
-              node {
-                ...on Document {
-                  _sys {
-                    extension
-                    path
-                  }
+const interateCollectionDocuments = async (
```
jamespohalloran, line 19: We were only auditing the first page of documents (10), so I had to introduce some logic to iterate through each page

packages/@tinacms/cli/src/cmds/audit/validations/validateRichText.ts
```diff
@@ -0,0 +1,34 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+import { AuditIssue, AuditWarning } from '../issue'
+
+export const validateRichText = (node) => {
+  let issues: AuditIssue[] = []
+  Object.keys(node._values)
```
jeffsee55, line 18: I don't think this will work for errors in nested `rich-text` fields.

jamespohalloran: Updated to search through nested fields

packages/@tinacms/cli/src/cmds/audit/validations/validateRichText.ts
```diff
@@ -0,0 +1,45 @@
...
+*/
+
+import { AuditIssue, AuditWarning } from '../issue'
+
+const validateValues = (values, addWarning: (message: string) => void) => {
+  let fields = Object.keys(values)
+    .map((fieldName) => values[fieldName])
+    .filter((field) => typeof field === 'object')
+
+  fields.forEach((field) => {
```
jeffsee55, line 21: This isn't the best way to iterate through recursively because for `object` fields with `templates` this won't work. Whenever we need to iterate, we always use the schema to do it. I'm not sure if this is the best approach or if it'd just be easier to throw a special error during `audit` rather than catching things in the `invalid_markdown` node.

jamespohalloran: > This isn't the best way to iterate through recursively because for object fields with templates this won't work

This is just searching through a document's `_values` (not actually searching on a "fields" object). Wouldn't a response setup with templates vs fields still output the same response object?

I've kind of just been going off of what the backend API response actually gives us, but if it could provide a root-level list of errors that would be easier to handle in the CLI.

jeffsee55: Yeah if it works that's fine. It's definitely feeling a bit of a weird API to step through rather than just through a special kind of error and catching it higher up. Is this also handling `rich-text` inside an array of values?

jamespohalloran: Hmm, okay well I see a few paths forward (let me know if you can think of others).
1) We can remove the rich-text check from the audit for now. 
2) We can take the time now to revamp how errors are handled on the backend, so a document-list request can return a root-level errors list, while still providing a valid response. That way we can cleanly handle these in the CLI, and the document list. I'm handle to poke around and see if I can figure this out, but you may be able to navigate the GraphQL code quicker if you have bandwidth.

jeffsee55: I definitely think the whole query blowing up on an error in one of the documents should be fixed up. I can tackle that this week. And if we did that, perhaps we can just allow the `invalid_markdown` error to bubble up only during `audit`. That would allow audit to work and just show the error on a document level I think.

jamespohalloran: Would it be possible to return partial data with an errors array for all requests? (It seems that when an error is thrown from the resolver, the full response comes back null) 

I could also see this type of behaviour being useful to fetch/show our document list view, but show "alert" icons next to the documents with issues.

jamespohalloran: @jeffsee55 I logged an internal ticket to capture how this PR is blocked: https://linear.app/tina/issue/ENG-506/surface-errors-in-audit-that-arent-breaking-in-the-cms

PR #3114, opened 2022-08-19
Iframe with vite

### LINKED ISSUE #2673: Tina Style Isolation
Right now, Tina's style reset is designed to normalize elements between browsers _(given it's pulled from Tailwind's preflight)_ and doesn't guard us from global user styles. Our wrapper class prevents our styles from interfering with the user's website, but not the reverse. In general, our UI is exposed and can be broken in a number of ways.

Here's an example from @jhuggett:
Screen Shot 2022-03-10 at 9 59 32 AM

There are only two approaches I can think of to definitively resolve these types of issues:
1. Use an iFrame. _(I believe we've ruled this out in the past)_
2. Use a combination of [`all: unset;`](https://developer.mozilla.org/en-US/docs/Web/CSS/all) and `px` for all units (we're nearly there with this, but our rich text editor and some other elements are still subjected to root font size).
    - This seems to be the approach to have a blank slate to work with if you're putting a component somewhere with preexisting styles, but it will require a lot of work.

jeffsee55:
Steps for using iframe:

- Switch `useTina` hook to come from `tinacms/dist/react`
- Remove Tina code in `_app.js`
- Remove `/pages/admin` if you have set that up (as it will no longer work)
- Export `defineConfig` _as the default export_ from `.tina/config.ts` (this is really awkward right now, see [the starter PR](https://github.com/tinacms/tina-cloud-starter/pull/318) for an example)
- When viewing a document in the admin, if it returns a value for route mapper you'll be taken to the preview version of it
- You'll need to visit the route including the `index.html` file, so `/admin/index.html`. If you want things to work with just `/admin` you'll need to update your `next.config.js` to handle it. This will be a framework-specific step so I hope we can avoid automating this and just point users to the index.html.

```ts
module.exports = {
  async rewrites() {
    return [
      {
        source: "/admin",
        destination: "/admin/index.html",
      },
    ];
  },
};

```

fixes #2673


kldavis4 (approved):
there is a fair amount of this that I don't have a lot of understanding of in this PR (lib stuff, schema-tools changes, etc) but it appears to be working as expected in the starter branch. The only thing I wanted to ask about was the content changes in the monorepo starters, and if any of those changes need to be reverted.

jeffsee55:
> the only thing I wanted to ask about was the content changes in the monorepo starters

Good point, I'll undo those. It's going to be kind of a pain until we make this the happy path. Maybe I'll make another example that's dedicated to this pattern.

> there is a fair amount of this that I don't have a lot of understanding of in this PR

I did this move now because this part is subtly different from what we have in our `formify` logic now, we have things like unmounting and registering forms happening at different times. And then I moved a few bits of logic to the `schema-tools` because we've been wanting to do that, but it's just moving them the formify logic at this point.

examples/basic-iframe/README.md
```diff
@@ -0,0 +1,24 @@
+This is a [Tina CMS](https://tina.io/) project.
+## Getting Started
+
+First, run the development server:
+
+```bash
```
jamespohalloran, line 6: Wasn't able to get this example working locally:
https://www.loom.com/share/de31d36ea66b4313b27be4c0f61ab005

jamespohalloran: Scratch that, I'm dumb and this is an example without contextual editing

logan-anderson:
This might not be a bug but a bit of a strange behaviour (Maybe we are aware of it. ). 

1. Make a content change.
2. Navigate away
3. Content change gets wiped without the user knowing.

See video: https://www.loom.com/share/0aa05d475b1641259837e7df3fb67f77


We might want to consider having a popup for when the user navigates away? or maybe this will be solved better when we do more stuff with multiple forms?

jeffsee55:
Yeah we might still be able to do this by moving the unmount logic to the new `useTina` hook, but it feels like this might all be part of a larger conversation about form state. I didn't want to implement new logic for this but keeping the old behavior isn't really doable.

logan-anderson:
Looks like there is an issue where it is hardcoded to `tina`, instead of looking at the value of the output folder?

See https://www.loom.com/share/9b7e13d005a54d21a0d959c4ba73e135

See 
- https://github.com/tinacms/tinacms/blob/d50a3acd1bb6a0da3e6a2833090c7ae7154b71e6/packages/%40tinacms/toolkit/src/packages/fields/components/Reference/ReferenceLink.tsx#L97  
- https://github.com/tinacms/tinacms/blob/d50a3acd1bb6a0da3e6a2833090c7ae7154b71e6/packages/%40tinacms/toolkit/src/packages/react-sidebar/components/Sidebar.tsx#L521

examples/basic-iframe/.tina/__generated__/out.jsx
```diff
@@ -0,0 +1,457 @@
+// .tina/schema.js
```
logan-anderson, line 1: @jeffsee55 Should this be gitignored?

logan-anderson (dismissed):

PR #3115, opened 2022-08-21
feat: allow paths to start with numeric characters

danielknell:
Allows numeric characters at the start of filenames to facilitate existing practices which include storing blog posts with the date prefixed or date based folder hierarchies. (e.g. Jekyll style `2022-08-21-blog-title-slug.md` files)

also updated a regex that appeared to be malformed and did not match the error description by allowing `,` characters.


jeffsee55 (approved):
Thanks @danielknell !

Issue #3078, opened 2022-08-07
Nested inline marks dropped from rich text output


mAAdhaTTah:
## Description

If I have a sentence that contains nested bold & italics, that text is dropped completely from the output.

### Steps to reproduce

1. Create a new collection with a rich-text field.
2. Write a sentence that works similar to one like this: This is **bold with _nested_ italics**.
3. Query the results for output.

### Expected result

The word "nested" should be a text field with `bold` & `italic` properties.

### Actual result

Word is dropped completely. Part of the schema is output as follows:

```json
{
  "bold": true
  "text": "bold with  italics."
  "type": "text"
  "undefined": true
  "value": " italics."
}
```

Markdown appears like this:

```md
This is a **bold with *****nested***** italics.**
```

I've also had this drop the word `nested` completely, although not 100% sure how to reproduce that.

#### Tinacms Package Versions

```text
├─┬ @tinacms/cli@0.60.25
│ ├─┬ @tinacms/datalayer@0.2.0
│ ├─┬ @tinacms/graphql@0.61.0
│ ├─┬ @tinacms/metrics@0.0.3
│ ├─┬ @tinacms/schema-tools@0.0.6
├─┬ react-tinacms-editor@0.53.26
│ ├─┬ @tinacms/toolkit@0.56.34
│ │ ├── @tinacms/sharedctx@0.1.1
├─┬ tinacms@0.68.13 invalid
│ ├── @tinacms/schema-tools@0.0.8 extraneous
│ ├── @tinacms/sharedctx@0.1.2 extraneous
│ ├── @tinacms/toolkit@0.56.35 extraneous
```

#### Tools

| Tool    | Version |
| ------- | ------- |
| browser |  Chrome       |
| os      |     OSX    |
| node    |       14.17.6  |
| npm     |     6.14.15    |

mAAdhaTTah:
Seems related to the fix in #2964 but that never landed.

mAAdhaTTah:
This appears to be resolved now.

PR #3116, opened 2022-08-21
Fix stringification of code element nested in link

mAAdhaTTah:
Add missing call to `linkifyTextNode` when handling `inlineCode` elements.
Also add a test case for #3078.

### DIFF
--- a/packages/@tinacms/mdx/src/stringify/marks.ts
+++ b/packages/@tinacms/mdx/src/stringify/marks.ts
@@ -77,7 +77,7 @@ const replaceLinksWithTextNodes = (content: Plate.InlineElement[]) => {
                 type: 'link',
                 url: item.url,
                 title: item.title,
-                children: [text({ text: a.value })],
+                children: [a],
               }
             },
           })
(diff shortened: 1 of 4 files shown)


packages/@tinacms/mdx/src/stringify/marks.ts
```diff
@@ -77,7 +77,7 @@ const replaceLinksWithTextNodes = (content: Plate.InlineElement[]) => {
                 type: 'link',
                 url: item.url,
                 title: item.title,
-                children: [text({ text: a.value })],
+                children: [a],
```
mAAdhaTTah, line 80: Nothing failed when I removed this. Its inclusion was stripping the `code: true` property required to properly linkify the `inlineCode` element.

jeffsee55: Thanks @mAAdhaTTah, `inlineCode` and `text` nodes happen to have the [same shape](https://github.com/syntax-tree/mdast#inlinecode), and they're the only types that should get through to this point, so your fix makes sense. 👍

jeffsee55 (approved):
Awesome! Thanks for the updated test case

PR #3117, opened 2022-08-21
Fix list page breaking for templates on collection

### LINKED ISSUE #3087: types.ts break when using "templates" in a collection
## Description

When I switch a collection to use "templates" instead of "fields", "types.ts" seems to get cleared out. 
[See the example in the starter](https://github.com/tinacms/tina-cloud-starter/compare/templates-test?expand=1), specifically this change in [types.ts](https://github.com/tinacms/tina-cloud-starter/blob/17d419ea01048032b4ca4bdfed91c40b910b4f0c/.tina/__generated__/types.ts#L10)

Console gives:

```
TypeError: Failed to fetch
    at <anonymous>:1:876
    at __webpack_require__.hmrM (http://localhost:3000/_next/static/chunks/webpack.js?ts=1660070703891:1280:20)
error - unhandledRejection: Error: aborted
    at connResetException (node:internal/errors:691:14)
    at abortIncoming (node:_http_server:596:17)
    at socketOnClose (node:_http_server:590:3)
    at Socket.emit (node:events:402:35)
    at TCP.<anonymous> (node:net:687:12) {
  code: 'ECONNRESET'
}
```

And the browser gives:

Screen Shot 2022-08-09 at 3 49 16 PM

### Steps to reproduce

[See the example in the starter](https://github.com/tinacms/tina-cloud-starter/compare/templates-test?expand=1)
I replaced a collection's fields with a few dummy templates

#### Tools

Node 16
├── @tinacms/cli@0.61.0
├── tinacms@0.68.14

mAAdhaTTah:
A collection doesn't have to have fields, so optional chain off the `fields` property.
This does mean collections with `templates` aren't sortable.

Fixes #3087.

### DIFF
--- a/.changeset/quick-seas-eat.md
+++ b/.changeset/quick-seas-eat.md
@@ -1,0 +1,5 @@
+---
+'tinacms': patch
+---
+
+Fix list page breaking for templates on collection
--- a/packages/tinacms/src/admin/components/GetCollection.tsx
+++ b/packages/tinacms/src/admin/components/GetCollection.tsx
@@ -40,7 +40,7 @@ export const useGetCollection = (
       if (await api.isAuthenticated()) {
         const { name, order } = JSON.parse(sortKey || '{}')
         const validSortKey = collectionExtra.fields
-          .map((x) => x.name)
+          ?.map((x) => x.name)
           .includes(name)
           ? name
           : undefined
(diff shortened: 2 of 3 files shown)


mAAdhaTTah:
I suspect there's a deeper solution here for sorting on fields that are common across all templates? Or maybe treating templates w/o those fields as if those fields were null? But this was the simplest solution to unbreak the collection list & edit pages.

jeffsee55 (approved):

PR #3125, opened 2022-08-24
Alert Improvements

### LINKED ISSUE #3101: Improve CMS Error UI
## Summary

Our error UI isn't super usable, because it's limited in size and disappears when clicked
Screen Shot 2022-08-17 at 10 50 38 AM

We also can't have any links in the error messages right now

### Expected Results

I would expect either a modal that persists until clicked, or to have the error persisted in the CMS view. 
I'll defer to @spbyrne on that.

spbyrne:
- Fixes alerts so only the close button dismisses them (or the timeout) and text wraps, allowing for longer alert messages.
- Converts error alerts to a blocking modal that doesn't have a timeout.
- Error modal supports providing a component instead of a string so that formatting and links can be provided, but it will render out strings inside a paragraph so that it's compatible with existing errors.

Fixes #3101

Example:

Code: 

```
cms.alerts.error(() => {
  return (
    <>
      <p className="text-base mb-3">
        This is an example error with a link. You can click the link
        without dismissing it.
      </p>
      <a
        className="flex items-center transition-all duration-150 ease-out text-blue-600 hover:text-blue-400 hover:underline no-underline"
        href="#"
      >
        Example Link{' '}
        <MdArrowForward className="w-4 h-auto ml-1.5 opacity-80" />
      </a>
    </>
  )
})
```

How it's rendered:

Screen Shot 2022-08-24 at 2 44 14 PM

Right now a lot of code is needed to render a pretty trivial error, given you have to provide your own styling. There's a few ways I could possibly improve the DX but I'm not sure what approach to take.

closes ENG-488

### DIFF
--- a/packages/@tinacms/toolkit/src/packages/react-alerts/Alerts.tsx
+++ b/packages/@tinacms/toolkit/src/packages/react-alerts/Alerts.tsx
@@ -22,6 +22,14 @@ import {
   CloseIcon,
 } from '../icons'
 import { useSubscribable } from '../react-core'
+import {
+  Modal,
+  ModalActions,
+  ModalBody,
+  ModalHeader,
+  PopupModal,
+} from '../react-modals'
+import { Button } from '../styles'
 
 export interface AlertsProps {
   alerts: AlertsCollection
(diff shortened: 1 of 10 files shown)


packages/@tinacms/toolkit/src/packages/react-alerts/Alerts.tsx
```diff
@@ -35,27 +43,72 @@ export function Alerts({ alerts }: AlertsProps) {
...
+                <CloseAlert
+                  onClick={() => {
+                    alerts.dismiss(alert)
+                  }}
+                />
+              </Alert>
+            )
+          })}
+      </AlertContainer>
+      {alerts.all
```
jeffsee55, line 68: I suspect that one of the reasons the previous UI was good was because it handled multiple messages by just stacking them. So for this, what happens when there are 2? And along with that, what about when 1 is an error and another is not? 

For what it's worth, that might be a task for another day,  just want to make sure it's not buggy, if we need to only show a single error message at a time that's probably fine.

jeffsee55 (commented):
> There's a few ways I could possibly improve the DX but I'm not sure what approach to take.

I think it's fine to be open-ended because we have a lot of errors messages from the server where there is html that needs to be called with `dangerouslySetInnerHTML`, so not something we'd want to build into the low-level component anyway.

packages/@tinacms/toolkit/src/packages/react-alerts/Alerts.tsx
spbyrne: I think the error modals will stack, with the most recent on top. The alerts display overtop the modal backdrop. I haven't tested with multiple errors, though, so I'm not 100% sure.

spbyrne: Confirmed modals stack & other alerts are rendered on top.

Screen Shot 2022-08-29 at 12 10 20 PM

jeffsee55:
Merging this, Vercel was my fault I was trying to set up a preview site so this is good to go.

PR #3136, opened 2022-08-30
feat: ENG-510: sync status widget & log viewer

kldavis4:
# what

Implement a sync status widget and log viewer


spbyrne:
Here's what I've come up with for widget placement & table styling. 

Widget is located next to the overflow menu where 'log out' is found:
Screen Shot 2022-09-01 at 12 48 56 PM

Fullscreen modal with scrolling table:
Screen Shot 2022-09-01 at 12 50 16 PM

One change I made to get around a bug I was seeing was to move the `useCMS()` call to inside `SyncStatus` instead of passing the CMS object, but I think this means it loads later than if it were passed in from the sidebar. When I had it passing in for whatever reason it wasn't making it past the loading state. With it set up the way I have it the mobile nav sync widget only loads when the nav is opened, leading to a delay.

kldavis4:
@jeffsee55 @jamespohalloran this is ready for review - you can install `tinacms@0.0.0-20220901185350` in a tina cloud project if you want to try it out

jeffsee55 (approved):

Issue #3132, opened 2022-08-27
Top-level templates not working in 0.69.2


cereallarceny:
## Description

I'm trying to create a top-level list of templates with which to base my only colleciton on. The purpose is to have a pages collection and have a homepage, about, and contact page, each with different schemas (and thus different templates). I'm basing this entirely on [this page in your documentation](https://tina.io/docs/reference/templates/). Yet, for some reason, no variation of this seems to actually work (json, md, or mdx). I get a very cryptic error saying that basically cannot get the collection at all, despite one document existing and the audit passing.

### Steps to reproduce

**Schema definition:**
```js
const schema = defineSchema({
  config: {
    token: process.env.TINA_TOKEN,
    clientId: process.env.TINA_CLIENT_ID,
    branch:
      process.env.NEXT_PUBLIC_TINA_BRANCH ||
      process.env.NEXT_PUBLIC_VERCEL_GIT_COMMIT_REF ||
      process.env.HEAD ||
      'main',
  },
  collections: [
    {
      label: 'Pages',
      name: 'page',
      path: 'content/pages',
      format: 'json',
      templates: [
        {
          name: 'homepage',
          label: 'Homepage',
          fields: [
            {
              type: 'string',
              name: 'title',
              label: 'Title',
            },
          ],
        }
      ]
    },
  ]
});
```

`content/pages/homepage.json`:
```json
{
  "title": "page title",
  "_template": "homepage"
}
```

### Expected result

Renders a collection of pages at `/admin` with one result: `homepage`. This page should be of `template` "homepage" and contain a filled `title` field with a value of "page title".

### Actual result

<img width="1728" alt="Screen Shot 2022-08-27 at 1 59 57 PM" src="https://user-images.githubusercontent.com/1297930/187044445-3d7a76e4-f574-44bf-a155-f98cff0c54c9.png">

#### Tinacms Package Versions

```text
├── @tinacms/cli@0.61.6
├── tinacms@0.69.2
```

#### Tools

| Tool    | Version |
| ------- | ------- |
| browser | Brave 1.42.97 |
| os      | MacOS 12.3 |
| node    | 16.13.1 |
| npm     | 8.12.2 |

cereallarceny:
Solved by https://github.com/tinacms/tinacms/pull/3137, close at will @jeffsee55 .

PR #3137, opened 2022-08-30
Fix admin list error on collections with templates

jeffsee55:
Fixes an issue where collections which used  would error on the admin list page due to the recent addition of filters. Filters will only work for collections with  at this time

Closes https://github.com/tinacms/tinacms/issues/3132


cereallarceny:
LGTM! ❤️

kldavis4 (approved):

PR #3142, opened 2022-08-30
Fix audit issues

jeffsee55:
Fix audit issues where listing the collection itself blows up before audit has a chance to run, don't run standalone build during audit command

- [x] Fix issue where audit only audits the first 10 items
- [x] Fix issue where document-list blows up before the audit actually runs when an item is invalid
- [x] During an audit error, the error message should not say `Please run "tinacms audit" or add the --verbose option for more info`
- [x] Invalid markdown should throw an error during audit, but not during runtime requests
- [x] Verbosity of error output makes it hard to decipher, add support for `--verbose` argument but otherwise keep error message concise.

NOTE: there are a lot of spec changes from a related fix to the formify logic (which was preventing some errors from populating on the frontend)


packages/tinacms/src/hooks/formify/formify.ts
```diff
@@ -380,6 +383,8 @@ export const formify = async ({
     }
 
     if (util.isFormifiableDocument(type)) {
+      // console.log('formifyInlineFragmentDocument', inlineFragmentNode)
+      // console.trace()
```
kldavis4, line 387: delete?

kldavis4 (dismissed):
lgtm (assuming the formify fixes were intended to be in this PR)

packages/@tinacms/cli/src/cmds/audit/index.ts
```diff
@@ -68,20 +68,40 @@ export const audit = async (ctx: any, next: () => void, options) => {
...
+      // on by default because when we index the item we
+      // must do a check to see of the format matches,
+      // so it never actually exists in the collection
+      const returnWarning = await auditCollection({
+        collection,
+        database,
+        rootPath,
+        useDefaultValues: options.useDefaultValues,
+        documents: docs.edges,
+      })
```
jamespohalloran, line 90:
```suggestion
```
Can confirm it's not doing anything

packages/@tinacms/cli/src/cmds/audit/audit.ts
```diff
@@ -22,94 +22,39 @@ type AuditArgs = {
...
-  })
+  const { collection, database, useDefaultValues, documents } = args
   let error = false
-  const documents: any[] = result.data.collection.documents.edges
   for (let i = 0; i < documents.length; i++) {
     const node = documents[i].node
-    const fullPath = p.join(rootPath, node._sys.path)
-    logger.info(`Checking document: ${fullPath}`)
+    const relativePath = node.path.replace(`${collection.path}/`, '')
+    logger.info(`Checking document: ${node.path}`)
```
jamespohalloran, line 55:
```suggestion
```
Now that we don't exit after a failure, I think it's more readable if we only output the issues

jamespohalloran (changes requested):

jamespohalloran:
For using `-1` for fetching all pages, will this cause issues for large sites? (I'm not sure if on the backend if we're enforcing any sort of max on the backend, but curious what would happen for sites with thousands of posts)

packages/@tinacms/cli/src/cmds/audit/index.ts
```diff
@@ -68,20 +68,40 @@ export const audit = async (ctx: any, next: () => void, options) => {
...
-      useDefaultValues: options.useDefaultValues,
-    })
-    warning = warning || returnWarning
-    error = error || returnError
+    try {
+      // Not a huge fan of querying the database from outside of GraphQL
+      // but this allows us to return the list of items unhydrated, so errors
+      // in a single document don't cause the entire list query to fail
+      const docs = await database.query(
+        { collection: collection.name, first: -1, filterChain: [] },
```
jamespohalloran, line 76: For using -1 for fetching all pages, will this cause issues for large sites? (I'm not sure if on the backend if we're enforcing any sort of max on the backend, but curious what would happen for sites with thousands of posts)

kldavis4: Does audit hit the tina cloud content api?

jeffsee55: No it doesn't this is probably pretty safe to run on large projects. It only returns the id for each node (and then loops through them 1 by 1).

packages/@tinacms/cli/src/cmds/audit/audit.ts
```diff
@@ -22,94 +22,39 @@ type AuditArgs = {
...
-    variables: {},
-  })
+  const { collection, database, useDefaultValues, documents } = args
   let error = false
-  const documents: any[] = result.data.collection.documents.edges
   for (let i = 0; i < documents.length; i++) {
     const node = documents[i].node
-    const fullPath = p.join(rootPath, node._sys.path)
-    logger.info(`Checking document: ${fullPath}`)
+    const relativePath = node.path.replace(`${collection.path}/`, '')
```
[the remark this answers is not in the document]
jeffsee55: Not a fan of this code, but we _do_ trim that value when we process it so I think it's ok.

packages/@tinacms/cli/src/cmds/audit/audit.ts
jeffsee55: yeah, much cleaner.

jamespohalloran (approved):

PR #3143, opened 2022-09-01
Refactor-CLI build

jamespohalloran:
Refactoring some of the code around the .tina directory build. There's lots of different ways to call this function with all of the variations of params.. so I split the function up a bit. Now we don't have to pass quite so many flags (like `skipIndex`) and we instead just conditionally index afterwards.

Properties are added/pulled to/from context throughout layers of nested functions, so it's difficult to track how different options are used. 

Main remaining todo's (maybe best served for another PR):
- We pass around a lot of different "schema" objects, and it's hard to know which schema gets used where. 
- It's a bit weird that the .tina files and the content all needs to get indexed together. Some of the logic in the `database` class depends on the schema files being indexed. 

### Behaviour changes:
- Fix so that we don't build the admin, client, or types on `tinacms audit`

### Testing
I tested tinacms build, audit, & dev on our barebones starter (js schema) and cloud starter (ts schema)


packages/@tinacms/graphql/src/index.ts
```diff
@@ -32,26 +31,23 @@ export type DummyType = unknown
 export const buildSchema = async (
```
jamespohalloran, line 31: Does this need to live in the GraphQL package? I'm assuming it's only used by the CLI

packages/@tinacms/cli/src/buildTina/index.ts
```diff
@@ -59,15 +52,15 @@ export const buildSetupCmdBuild = async (
...
+  const { bridge, database } = await buildSetup({
     ...opts,
     rootPath,
     useMemoryStore: true,
   })
   // attach to context
   ctx.bridge = bridge
   ctx.database = database
-  ctx.store = store
+  ctx.builder = new Builder(database)
```
jamespohalloran, line 63: TODO: Don't find a less generic name

packages/@tinacms/graphql/src/database/index.ts
```diff
@@ -338,6 +338,7 @@ export class Database {
     const graphqlPath = path.join(GENERATED_FOLDER, `_graphql.json`)
     return this.store.get(normalizePath(graphqlPath))
   }
+  //TODO - is there a reason why the database fetches some config with "bridge.get", and some with "store.get"?
```
jamespohalloran, line 341: cc @jeffsee55 , 

> is there a reason why the database fetches some config with "bridge.get", and some with "store.get"? Wondering if it could make sense to split some of these functions into multiple classes.

jeffsee55: There's a lot of nuance here because we needed to continue to support multiple stores/bridges at the same time. In general I expect these differences to go away when we remove the logic for non-indexable stores (ie. `Filesystem`). I imagine that will need to be a large-ish refactor and it'd be best to leave this alone for now.

logan-anderson (approved):

logan-anderson:
@jamespohalloran This looks good to me.

> Main remaining todo's (maybe best served for another PR):

> We pass around a lot of different "schema" objects, and it's hard to know which schema gets used where.
> It's a bit weird that the .tina files and the content all needs to get indexed together. Some of the logic in the database class > depends on the schema files being indexed.

This is probably best as an another PR and might change a bit with the Iframe stuff.

jeffsee55 (approved):

PR #3149, opened 2022-09-06
Add router and global properties to collections config

jeffsee55:
Add `router` property on collections. This replaces the need for using the RouteMapper plugin.

```ts
...
  name: 'post',
  path: 'posts',
  ui: {
    router: ({ document }) => {
      // eg. post items can be previewed at posts/hello-world
      return `/posts/${document._sys.filename}`;
    },
  },
...
```

Add `global` property on collections. This replaces the need for `formifyCallback` in most cases

```ts
...
  name: 'post',
  path: 'posts',
  ui: {
    global: true
  },
...
```


experimental-examples/tina-cloud-starter/.tina/schema.ts
```diff
@@ -251,6 +255,7 @@ const schema = defineSchema({
       label: "Global",
       name: "global",
       path: "content/global",
+      ui: true,
```
logan-anderson, line 258: @jeffsee55 Getting a type error here. Is this supposed to be 

```
ui: {
    global: true
},
``` 
?

jeffsee55: Oh woops, yeah fixed.

logan-anderson (changes requested):
Looking good. The router seems to work as expected, just found one issue with the global form. 

It seems to work in contextual editing.


Screen Shot 2022-09-06 at 2 05 24 PM

But does not seem to work in the admin.

Screen Shot 2022-09-06 at 2 04 58 PM

Maybe it needs to be wired up in the Admin UI?

logan-anderson:
Another small issue;

Seems like the other types for UI are being wiped out:
Screen Shot 2022-09-06 at 2 01 10 PM

jeffsee55:
> Maybe it needs to be wired up in the Admin UI?

Does `formifyCallback` currently run for stuff in the admin? I don't see where that would be happening?

jeffsee55:
> Seems like the other types for UI are being wiped out:

Is `global` provided in your screenshot on line 60 (and isn't visible because of the intellisense box).

This is the first use of the `ui` property at the collections level isn't it?

logan-anderson:
> > Seems like the other types for UI are being wiped out:
> 
> Is `global` provided in your screenshot on line 60 (and isn't visible because of the intellisense box).
> 
> This is the first use of the `ui` property at the collections level isn't it?
> 
> Here's what I'm seeing
> 
> Screen Shot 2022-09-06 at 11 14 17 AM

disregard my comment. I was thinking of the ui types for fields (validate, component, etc) not showing up. They don't show up here which is correct.

logan-anderson:
> > Maybe it needs to be wired up in the Admin UI?
> 
> Does `formifyCallback` currently run for stuff in the admin? I don't see where that would be happening?

No, I don't think it is being called. It might added. The other properties work like validate, component, etc, so it would be nice if this one worked in admin as well.

jeffsee55:
@logan-anderson if `formifyCallback` doesn't currently operate at the admin I'd like to continue that behavior and not support it there. The idea is that a form in the "global" sidebar is contextual, which wouldn't make sense in the admin.

logan-anderson:
> @logan-anderson if `formifyCallback` doesn't currently operate at the admin I'd like to continue that behavior and not support it there. The idea is that a form in the "global" sidebar is contextual, which wouldn't make sense in the admin.

Yup! makes sense to me.

logan-anderson (approved):

PR #3151, opened 2022-09-06
Fix build error handling

jamespohalloran:
Surface errors from type generation

Fixes #ENG-524

### DIFF
--- a/packages/@tinacms/cli/src/codegen/index.ts
+++ b/packages/@tinacms/cli/src/codegen/index.ts
@@ -34,73 +34,49 @@ export const generateTypes = async (
   }
 ) => {
   if (options.verbose) logger.info('Generating types...')
-  try {
-    let docs = []
-    let fragDocs = []
 
-    // Load GQL queries from user queries folder
-    try {
-      if (!options.noSDK) {
-        docs = await loadDocuments(queryPathGlob, {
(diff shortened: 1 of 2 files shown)


packages/@tinacms/cli/src/codegen/index.ts
```diff
@@ -34,73 +34,49 @@ export const generateTypes = async (
...
-      if (!options.noSDK) {
-        docs = await loadDocuments(queryPathGlob, {
-          loaders: [new GraphQLFileLoader()],
-        })
-      }
-    } catch (e) {
-      let showErrorMessage = true
-      const message: string = e.message || ''
-      if (
-        message.includes(
```
jamespohalloran, line 52: I'm not sure if this check is because this error in particular is just verbose, but I figured it would be better than not logging anything.

jamespohalloran: Oh, I see the cypress tests are failing with that error:
> Unable to find any GraphQL type definitions for the following pointers:

Not sure if that test should be updated, or if there's context I'm missing

jamespohalloran: Update: Added that check back, as we need it incase the user doesn't have a queries folder

PR #3153, opened 2022-09-07
Fix/add router to init

logan-anderson:
- update default schema to use router prop
- update router to have correct type (before it was forced to return a string)

### DIFF
--- a/packages/@tinacms/cli/src/cmds/compile/defaultSchema.ts
+++ b/packages/@tinacms/cli/src/cmds/compile/defaultSchema.ts
@@ -37,7 +37,7 @@ const schema = defineSchema({
       ui: {
         router: ({ document }) => {
           // This can be used to add contextual editing to your site. See https://tina.io/docs/tinacms-context/#accessing-contextual-editing-from-the-cms for more information.
-          return undefined
+          return \`/demo/blog/\${document._sys.filename}\`
         },
       },
       fields: [
--- a/packages/@tinacms/cli/src/cmds/compile/defaultSchema.ts
+++ b/packages/@tinacms/cli/src/cmds/compile/defaultSchema.ts
@@ -34,6 +34,12 @@ const schema = defineSchema({
       name: 'post',
       path: 'content/posts',
       format: 'mdx',
+      ui: {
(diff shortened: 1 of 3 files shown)


packages/@tinacms/schema-tools/src/types/SchemaTypes.ts
```diff
@@ -98,7 +98,7 @@ interface BaseCollection {
     router?: (args: {
       document: Document
       collection: TinaCloudCollection<true>
-    }) => string
+    }) => string | undefined
```
jeffsee55, line 101: I had thought this could be more strict since it's no longer a global config but you're probably right, doesn't need to enforce this.

packages/@tinacms/cli/src/cmds/compile/defaultSchema.ts
```diff
@@ -34,6 +34,12 @@ const schema = defineSchema({
       name: 'post',
       path: 'content/posts',
       format: 'mdx',
+      ui: {
+        router: ({ document }) => {
+          // This can be used to add contextual editing to your site. See https://tina.io/docs/tinacms-context/#accessing-contextual-editing-from-the-cms for more information.
+          return undefined
```
jeffsee55, line 40: Wouldn't we want to send them to the demo page?

logan-anderson: Yeah that would make sense! I will update

packages/@tinacms/schema-tools/src/types/SchemaTypes.ts
jamespohalloran: So undefined just brings the user to the admin instead of a contextual page right? I think your use-case makes sense @logan-anderson. We might just want to add a note to our reference docs if it's not already there.

logan-anderson: @jamespohalloran Yeah exactly. It should be in the docs PR https://github.com/tinacms/tina.io/pull/1394

jamespohalloran (approved):

PR #3154, opened 2022-09-07
Event Log UI

spbyrne:
This revises the sync event log UI based on a discussion we had at office hours. 

When there's no error, the event log is available through the 'Tina' menu:
Screen Shot 2022-09-07 at 3 02 42 PM

When there's a sync error, an icon is displayed next to the overflow menu to indicate something's wrong:
Screen Shot 2022-09-07 at 3 02 04 PM

The menu item icon is changed to guide the user to the event log:
Screen Shot 2022-09-07 at 3 02 17 PM


packages/@tinacms/toolkit/src/packages/react-sidebar/components/Nav.tsx
```diff
@@ -26,7 +26,8 @@ import { FormModal } from '../../react-forms'
 import { useEditState } from '@tinacms/sharedctx'
 import type { ScreenPlugin } from '../../react-screens'
 import { LoadingDots } from '../../form-builder'
-import { SyncStatus } from './SyncStatus'
+import { SyncStatus, SyncErrorWidget } from './SyncStatus'
+import { useCMS } from '../../react-core'
```
jamespohalloran, line 30: Our monorepo setup seems to be handling this (I see our other imports do the same thing). But to confirm, is importing like this from another sibling package okay? @logan-anderson @jeffsee55

jeffsee55: @jamespohalloran yes because these aren't really packages, they're just folders from when we moved all of this to be part of `@tinacms/toolkit`.

jeffsee55 (approved):

PR #3161, opened 2022-09-08
do not require schema when config is provided

logan-anderson:
No longer require a schema file when a config file is present.

### How to test.
- In the tina-cloud-starter move the code from the schema.ts file into the config.ts 
-  delete the schema.ts file
- everything should work the same


packages/@tinacms/cli/src/cmds/compile/index.ts
```diff
@@ -313,15 +297,49 @@ export const compileSchema = async (options: {
...
+    allowedTypes: ['js', 'jsx', 'tsx', 'ts'],
+  })
+  const configExists = fileExists({
+    projectDir: tinaPath,
+    filename: 'config',
+    allowedTypes: ['js', 'jsx', 'tsx', 'ts'],
+  })
+
+  // If there is not Schema and no config
+  if (!schemaExists && !configExists) {
```
jeffsee55, line 319: For what it's worth I'm not sure there's a ton of value in creating the schema at this stage. I know this command is run for various tasks, but I think the main path it'd be preferred to just throw an error saying you need to run `tinacms init`.

logan-anderson: Agreed. Pushed up a fix for that.

jeffsee55 (approved):

PR #3165, opened 2022-09-09
More context to schema validators

logan-anderson:
add better error messages for duplicate values


packages/@tinacms/schema-tools/src/util/hasDuplicates.ts
```diff
@@ -11,10 +11,31 @@ See the License for the specific language governing permissions and
...
 }
+
+/**
+ *
+ * @param array
+ * @returns False if the array is undefined or has no duplicates.
+ */
+export function findDuplicates<T = any>(
+  array: T[] | undefined
+): false | string {
```
jamespohalloran, line 29: Should this return undefined instead of false if no duplicated are found?

jamespohalloran (changes requested):

packages/@tinacms/schema-tools/src/util/hasDuplicates.ts
logan-anderson: It can. I am just doing a IF check on them so it would not matter.

packages/@tinacms/schema-tools/src/util/hasDuplicates.ts
```diff
@@ -11,10 +11,31 @@ See the License for the specific language governing permissions and
...
   }
 }
+
+/**
+ *
+ * @param array
+ * @returns False if the array is undefined or has no duplicates.
+ */
+export function findDuplicates<T = any>(
+  array: T[] | undefined
```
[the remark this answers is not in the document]
logan-anderson: I updated it.

packages/@tinacms/schema-tools/src/util/hasDuplicates.ts
logan-anderson: Updated

jamespohalloran (approved):
Looks good!!

Issue #2766, opened 2022-04-01
No longer warning user that there are unsaved changes when navigating away 


scottgallant:
Tina isn't warning the user about unsaved changes when navigating away from the page
![](http://g.recordit.co/rLY10xQNj8.gif)

PR #3174, opened 2022-09-12
Fix not warning unsaved issue

Phoenix-Alpha:


Phoenix-Alpha:
Fix #2766

jamespohalloran:
Hey! So this looks like the type of thing we're looking for, but the ticket was moreso for the forms of existing documents. This looks like the code that used to prevent navigating away from a dirty document: https://github.com/tinacms/tinacms/blob/main/packages/%40tinacms/toolkit/src/packages/form-builder/FormBuilder.tsx#L90-L93

However at some point, it seems it stopped working

Phoenix-Alpha:
@jamespohalloran , `onBeforeUnload` is called when the document is unloaded and navigated to another document.
And the issue is `onBeforeUnload` is not called when route navigation in React SPA.

Phoenix-Alpha:
To prevent react route navigation, we used to use `usePrompt` hook of `react-router-dom` but it was removed from `react-router-dom` v6 so I defined custom `usePrompt` hook

jeffsee55:
Isn't this supposed to work on edit forms (both contextual and in the admin)?

In general, the iframe stuff uses the same code as current contextual forms. But this won't work regardless for that because there is no window location change.

jamespohalloran:
> Isn't this supposed to work on edit forms (both contextual and in the admin)?

That was the original intent yeah. I was wondering if this could live inside the useTina hook (so the location change would be inside the iframe), but I guess if I'm understanding correctly, that wouldn't work because it wouldn't have the react-router scope?

jeffsee55:
> I was wondering if this could live inside the useTina hook (so the location change would be inside the iframe)

It might be doable but yeah it'll be totally different logic because the `useTina` hook might represent multiple forms.

jamespohalloran:
Just converted this one to a draft. My fault for pulling into the "ready to be worked on issues", but I think it might just need some more discussion based on the thread.

Phoenix-Alpha:
@all-contributions please add @Phoenix-Alpha for its contribution to doc, code and plugin.

jeffsee55:
Closing as I think we'll need a pretty different approach for this, feel free to reopen if necessary

PR #3177, opened 2022-09-12
Fix rich text dropdown overlap issue

### LINKED ISSUE #2648: Rich Text Dropdown menu Z-index issue
## Description
The rich text field dropdown menu is behind the toolbar. This probably isn't noticeable if you have one really long rich text field, but  it is noticeable if you have a second rich text field like in the "two columns" block of our new homepage. 

Screen Shot 2022-03-04 at 5 21 23 PM

Phoenix-Alpha:
close #2648

### DIFF
--- a/.changeset/calm-bees-buy.md
+++ b/.changeset/calm-bees-buy.md
@@ -1,0 +1,5 @@
+---
+'@tinacms/toolkit': patch
+---
+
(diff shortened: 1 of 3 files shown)


logan-anderson (approved):
Looks good to me! 😄

PR #3183, opened 2022-09-13
replace loading message and hide while loading

### LINKED ISSUE #2637: When loading data, we should replace the Tina didn't find any queries.
When loading data, we place an overlay over the site with "Tina is loading data, please wait" but in the side bar we have "Tina didn't find any queries" 

Screenshot_2022-03-04_at_12 10 15

From a UX perspective this is confusing as you might believe that your site is in fact not working correctly. When we are loading the data, we should provide better feedback, so the user knows that we are attempting to load the data. 

Then if it fails, we can return the usual errors and replace the text in the sidebar. If successful return the forms.

Phoenix-Alpha:
close #2637
closes ENG-51


logan-anderson (changes requested):
This is _almost_ there. While loading the message "Looks like there's nothing to edit on this page." is confusing (and false). When the data is loading we should display a loading message in the sidebar (or nothing at all would be fine). If no forms are in use then we can display 
"Looks like there's nothing to edit on this page.". If there is forms then we can display them. 

https://www.loom.com/share/605e5d8b235d4d3cbf045d1403a5bdc4

Phoenix-Alpha:
@logan-anderson , I hide the sidebar message while data is loading. Plz review again

logan-anderson (changes requested):
Looks like some tests are failing. Once they are passing again, this looks good to me.

Phoenix-Alpha:
@logan-anderson , I fixed the testing failing. Plz review again.

logan-anderson (approved):

logan-anderson (approved):

PR #3187, opened 2022-09-14
update the development setup instructions

zeropaper:
The `pnpm install` step was missing in the instructions

- Add note about NVM
- Add a step about installing yarn globally

The code doesn't change, so I doubt there's a need for a changeset.
Correct me if I'm wrong :)


logan-anderson (approved):
Thanks @zeropaper !

PR #3190, opened 2022-09-14
change hardcoded content api url

### LINKED ISSUE #3085: Configurable port for "yarn dev"
## Summary

As a dev I want to be able to configure the "yarn dev" port so that I can run more than one instance of Tina at a time

### Expected Results

```
yarn dev --port 4002
```

### Notes

The CLI has a configurable port, but it's hardcoded to 4001 in tinacms.

Phoenix-Alpha:
close #3085
closes ENG-467


packages/@tinacms/cli/src/cmds/query-gen/genTypes.ts
```diff
@@ -48,8 +48,10 @@ export async function genClient(
     )
   }
 
+  const portIndex = process.argv.findIndex((a) => a === '--port') + 1
+  const PORT = (portIndex ? process.argv[portIndex] : null) || 4001
```
logan-anderson, line 52: The arguments are already being parsed with commander so we should not be doing our own parsing logic. 

It would be best to use the port option available at `options?.port`. That can [be accessed in this function](https://github.com/tinacms/tinacms/blob/feat/configurable-port-for-yarn-dev/packages/@tinacms/cli/src/buildTina/index.ts#L73) and will have to be passed to `genClient`

logan-anderson (changes requested):
Once parsing port logic is removed and we use the port from the options, this looks good to me.

packages/@tinacms/cli/src/cmds/query-gen/genTypes.ts
Phoenix-Alpha: @logan-anderson , I passed the port to `genClient`. Plz review again

logan-anderson (changes requested):
## **UPDATE:**

Sorry this is happening because the client was not configured properly on the basic starter. This is handled.

I will push up a change the fixes the basic starter

@Phoenix-Alpha things are looking good for the client ~~but it is not working when in edit mode~~

~~But when I go into edit mode using the `--port` option it breaks. You will need to update the url for the edit mode "internal" client as well.~~


~~See this video for more detail~~
 https://www.loom.com/share/e6d353b459764beb900d57c2fc7a5de5

packages/@tinacms/cli/src/cmds/query-gen/genTypes.ts
logan-anderson: Found another issue.  See https://github.com/tinacms/tinacms/pull/3190#pullrequestreview-1109631857

logan-anderson (approved):

logan-anderson (approved):

PR #3192, opened 2022-09-14
Add static init

jeffsee55:


examples/hugo-quickstart/resources/_gen/assets/css/ananke/css/main.css_bb5467e0521bbea6b1e66429f6ec028e.json
```diff
@@ -0,0 +1 @@
+{"Target":"ananke/css/main.min.css","MediaType":"text/css","Data":{}}
```
logan-anderson, line 1: @jeffsee55  can these `resources/_gen/**`  files be git ignored? or are they needed?

packages/@tinacms/cli/src/buildTina/index.ts
```diff
@@ -98,6 +98,9 @@ export const buildSetupCmdAudit = async (
     ? new FilesystemBridge(rootPath)
     : new AuditFileSystemBridge(rootPath)
 
+  if (!fs.pathExistsSync(path.join(rootPath, '.tina', '__generated__'))) {
```
logan-anderson, line 101: For some reason (not sure why) we have ran into windows issues with similar code.
I have had better success with the [`ensureDir`](https://github.com/jprichardson/node-fs-extra/blob/HEAD/docs/ensureDir.md) function from fs-extra, might be worth switching to that.

packages/@tinacms/cli/src/cmds/init/static.ts
```diff
@@ -0,0 +1,269 @@
...
+
+  await addConfigFile(publicFolder, usingTypescript)
+
+  await addContentFile()
+
+  logNextSteps(packageManager)
+}
+const baseDir = process.cwd()
+
+const choosePackageManager = async () => {
```
logan-anderson, line 64: Might be out of scope but it might be nice to eventually add pnpm to the normal init as well.

logan-anderson (commented):
Couple of small things but looks good overall.

examples/hugo-quickstart/resources/_gen/assets/css/ananke/css/main.css_bb5467e0521bbea6b1e66429f6ec028e.json
jeffsee55: I'm not actually sure, if it gets annoying we can always add the `gitignore`

packages/@tinacms/cli/src/cmds/init/static.ts
jeffsee55: I'm hoping we can just start moving pieces of the normal init into this one as standalone becomes the happy path. For example, the main difference between the two now (aside from `defineStaticConfig`) is the stuff for NextJS, which feels like it could be simple to add

logan-anderson (approved):

PR #3193, opened 2022-09-14
Pre build checks

logan-anderson:
- Makes cli checks the error message clearer 
- Adds a "Spinner" to the messages to make it cleaner
- Does a GQL request to check the clientID, token,  branch are all valid.


packages/@tinacms/cli/src/cmds/statusChecks/waitForIndexing.ts
```diff
@@ -34,27 +37,23 @@ class IndexFailedError extends Error {
...
-    logger.info(logText('Waiting for DB...'))
-  }
-
   if (!ctx.client) {
-    if (options.verbose) {
-      logger.info(logText('client is unavailable, skipping...'))
-    }
-    return next()
+    throw new Error(
+      'No Tina Cloud find. For more information on how to setup the tina cloud see https://tina.io/docs/features/data-fetching/#making-requests-with-the-tina-client'
```
jeffsee55, line 42:
```suggestion
      'No Tina Cloud found. For more information on how to setup the tina cloud see https://tina.io/docs/features/data-fetching/#making-requests-with-the-tina-client'
```

Though I'm not sure this error makes a lot of sense in this context. Is this happening when we don't pass in a `client`?

packages/@tinacms/cli/src/cmds/statusChecks/waitForIndexing.ts
```diff
@@ -95,18 +94,23 @@ export const waitForDB = async (ctx, next, options: { verbose?: boolean }) => {
...
-          `Attempting to index DB responded with status 'failed', ${error}`
+          `Attempting to index but responded with status 'failed', ${error}`
         )
 
         // Index Unknown
       } else {
-        if (options.verbose) {
-          logger.info(logText(`${statusMessage}`))
-        }
+        // TODO: Should we throw an error here?
```
jeffsee55, line 102: Would there be a reason not to throw here?

examples/basic/.tina/schema.js
```diff
@@ -14,6 +14,11 @@ limitations under the License.
 import { defineSchema, defineConfig } from 'tinacms'
 
 const schema = defineSchema({
+  config: {
+    branch: 'main',
+    clientId: 'ce702ac4-743f-434a-bb50-48b84ec75b04',
+    token: '07b20ae78b8bf163447f17f142fccaf6a9ac50d4',
```
jeffsee55, line 20: Probably don't want these hanging around

packages/@tinacms/cli/src/cmds/statusChecks/checkClientInformation.ts
```diff
@@ -0,0 +1,47 @@
...
+    })
+    bar.tick({
+      prog: '✅',
+    })
+  } catch (e) {
+    bar.tick({
+      prog: '❌',
+    })
+    console.warn(
+      'Error when checking client information. Please check you have the correct client ID, URL and read only token configured. For more information see https://tina.io/docs/tina-cloud/connecting-site/'
```
jeffsee55, line 41: I think it'd be helpful to list the values we have here. From my experience I usually forget the token altogether or get the branch wrong.

packages/@tinacms/cli/src/cmds/statusChecks/waitForIndexing.ts
logan-anderson: Yeah, I think this might happen if the user does not provide a client ID, token and branch.

packages/@tinacms/cli/src/cmds/statusChecks/waitForIndexing.ts
logan-anderson: I am not sure. @kldavis4 When would the indexing return "unknown" ?

kldavis4: unknown because means there are no events associated with the branch

packages/@tinacms/cli/src/cmds/statusChecks/checkClientInformation.ts
logan-anderson: Agreed.

examples/basic/.tina/schema.js
logan-anderson: Nope 😆

packages/@tinacms/cli/src/cmds/statusChecks/waitForIndexing.ts
logan-anderson: Do you think it would make sense to throw an error if this happens?

examples/basic/.tina/schema.js
logan-anderson: Removed them.

packages/@tinacms/cli/src/cmds/statusChecks/waitForIndexing.ts
kldavis4: It means that the branch is not indexed and indexing hasn't started, so yeah, probably an error would make sense

logan-anderson: Makes sense, updated

jamespohalloran (approved):

PR #3199, opened 2022-09-15
Adds video support for media manager

wem5637:
This work is based on https://github.com/tinacms/tinacms/pull/1581.

This PR attempts to add video uploading to the media manager and filters the content depending on filter type, be it image or video.

Screen Shot 2022-09-15 at 2 23 25 PM
Screen Shot 2022-09-15 at 2 11 37 PM
Screen Shot 2022-09-15 at 2 11 44 PM


packages/@tinacms/toolkit/src/components/media/get-mime.js
```diff
@@ -0,0 +1,174 @@
...
+  const split_arr = filename.split('.')
+  const suffix = split_arr[split_arr.length - 1]
+  if (VIDEO_FILETYPES.includes(suffix)) {
+    return 'video'
+  } else if (IMAGE_FILETYPES.includes(suffix)) {
+    return 'image'
+  } else {
+    return 'file'
+  }
+}
```
wem5637, line 174: this was a last resort at determining the file type. was hoping to save type on file `persist` but couldn't figure out how to make the property stick. this just looks at the filename instead and in `media-item.tsx` uses it to determine icon, and then also media-manager.tsx uses it to filter the list of media to only show types relevant to the current filter.

packages/@tinacms/toolkit/src/components/media/media-manager.tsx
```diff
@@ -341,7 +345,19 @@ export function MediaPicker({
...
+  function filterMedia(item: Media) {
+    if (!filter) return true
+    const mime = getMime(item.filename)
+    if (filter === 'video') {
+      return mime === 'video'
+    } else if (filter === 'image') {
+      return mime === 'image'
+    } else {
+      return false
+    }
```
jamespohalloran, line 357:
```suggestion
return mime === filter
```
^ Could we simplify this down to that conditional?

jamespohalloran (commented):
Thanks for this great PR! It might take the team some more time to discuss and make sure it aligns with our roadmap before we merge. I left a few comments, but let me discuss with the team and figure out next steps before worrying about any of the change-requests.

jamespohalloran:
Hey @wem5637,
First of all, thank you for your contribution to TinaCMS. The team took a look, and while we'd love video support in the future, there are still some details that we'll need to spend some more time considering. Having video as a core field type may be a bit problematic, because by default, the content is stored in the git repo. Storing large files in the repo often doesn't scale well, so we may want to limit videos to external media providers.
The team is currently busy on a pretty focused roadmap as we near our 1.0 release. I'll close this PR for now, but we hope to circle back to video support in this future, and we will continue to take this PR into consideration!

johannesmutter:
Hey,
I agree that git is not ideal storing large video files.

But in my use case I need to store **MP4** files as a more storage efficient replacement for **GIFs**.
Besides the ability to upload small looping videos, I would like to use the Media Manager for storing **PDFs** and **SVGs**. Both not possible currently.

I don't really need a video preview or PDF preview in the media manager, just displaying the file name + extension is sufficient. 

Could you reconsider adding the option for adding a generic "file" type? 
Or what's a quick and dirty way to make this possible myself?

PR #3209, opened 2022-09-20
Remove esm package & upgrade to Vite 3

jeffsee55:
We were using the [esm](https://github.com/standard-things/esm) package, which helped when we need to convert CJS to ES modules in Yarn v2 contexts. But since we've moved to PNPM this is no longer necessary. 

ESM was used in all of our bin's, so `@tinacms/scripts` (is working already), `@tinacms/cli` and `create-tina-app` all need to be tested externally from the monorepo.

Removing the ESM package allows us to move to Vite 3 for `@tinacms/app`, there was a new node feature used in the vite react plugin that the ESM package didn't support.


jeffsee55:
Closing for now as we don't need Vite 3 yet. This requires all users to be on Node 16

PR #3210, opened 2022-09-20
port in use: CLI error handling

jamespohalloran:
Fixes a few issues with the CLI error handling:
- Fixes issue where subprocess would still run if port was in use.
- Don't exist with success when initial build fails


logan-anderson:
@jamespohalloran Can you add a `changeset`? Can be done with `pnpm changeset`.

logan-anderson (dismissed):

logan-anderson (approved):

PR #3216, opened 2022-09-21
Refactor to use schema tools

logan-anderson:
Removes the schema from `@tinacms/graphql` and uses `@tinacms/schema-tools` everywhere.


packages/@tinacms/graphql/src/schema/createSchema.ts
```diff
@@ -0,0 +1,44 @@
...
+
+export const createSchema = async ({
+  schema,
+  flags = [],
+}: {
+  schema: TinaCloudSchemaBase
+  flags?: string[]
+}) => {
+  // TODO: fix types
+  // @ts-ignore
```
jamespohalloran, line 27:
```suggestion
```

Not 100%, but I don't think you need this anymore

jamespohalloran (changes requested):
Nice 👍 Just left one small comment

packages/@tinacms/graphql/src/schema/createSchema.ts
logan-anderson: @jamespohalloran I updated and removed some other spots that I could remove that comment.

PR #3218, opened 2022-09-21
Enable dev server for faster dev builds in standalone

jeffsee55:
The way our vite build step has worked so far is we run the production build of the app source files in `@tinacms/app/appFiles` along with an aliased reference to the user's config, allowing them to include custom components and things of that sort. This has worked nicely, but the builds take a long time, and there was some trickery to getting things to work, we were building the app from the context of the `@tinacms/app` location - so for installs it would be something like `<my-roo>/node_modules/@tinacms/app`, which is no issue for pnpm but with yarn resulted in the vite dev server not being able to find external modules for it's cjs->esm transform in dev mode. We've also up to this point required that users install `react`, `react-dom`, and `styled-components` even when we're in control of the artifacts. 

This PR adds a pre-build step to treat the standalone app files as a single artifact, including the dependencies we use (we bring our own version of react). It also now treats the vite root as the `cwd` (default on most vite apps) which allows the vite dev server to properly transform cjs->esm for dev mode. This unlocks a much faster dev experience and should lead to hot-reloading capabilities in the near future.

Things this doesn't do:
- When you change your config file, you'll need to refresh your browser for now to see the latest changes
- It doesn't compile tailwind classes from the user's config file (this would be handy when the user wants to provide their own component)

I think these are both small-ish tasks but want to get this out before moving on to them.


packages/@tinacms/app/src/index.ts
```diff
@@ -112,23 +118,44 @@ vite.svg`
...
+    }
+    server = await createServer(config)
     await server.listen()
     await server.printUrls()
+  } else {
+    /**
+     * This is kind of awkward, we're putting files in the specified
+     * output folder because we want to run the vite build
+     * from the context of the user's site, so dependencies are
+     * discovered properly. So his drops in the scaffolding
```
kldavis4, line 147:
```suggestion
     * discovered properly. So this drops in the scaffolding
```

logan-anderson (approved):

PR #3225, opened 2022-09-23
Change behaviour of defaults

logan-anderson:
- deprecate: `defaultValue`
- add `defaultItem` to the collection (as a function or an object)
```ts
      defaultItem: () => {
        const m = new Date()
        return {
          title: 'New Page',
          test: 'This is a default value of the test field',
          filename: `new-page-${
            m.getUTCFullYear() +
            '-' +
            (m.getUTCMonth() + 1) +
            '-' +
            m.getUTCDate()
          }`,
        }
      },
```
- Allow `datetime` field to be undefined or empty

One strange behaviour (outlined in video below) that I could try and fix but I think might be better addressed as a separate ticket. (We might have to refactor this form)

https://www.loom.com/share/9ee5aadbeb0d4891a3a62e65432f51bf


jamespohalloran:
I wonder if it makes sense to restrict setting a default on "filename". Filename needs to be unique, so I don't know if the default would ever be helpful. 
(Filename behaviour will likely be changing soon anyways: https://linear.app/tina/issue/ENG-520/provide-filename-customization-api)

logan-anderson:
> I wonder if it makes sense to restrict setting a default on "filename". Filename needs to be unique, so I don't know if the default would ever be helpful.

In my example I used function to set the default as the current date (I think this is used in hugo as well) this might be one way of being unique. 

We could restrict it but I have a feeling that the FileName function may be a bit tricky to implement. So at least if we keep it we can allow users to set a default filename for the meantime.

jeffsee55:
Regarding this:

> One strange behaviour (outlined in video below) that I could try and fix but I think might be better addressed as a separate ticket. (We might have to refactor this form)

I would say the current behavior is incorrect. It shouldn't be disabled based on whether or not the form has been touched, only on whether the form is valid.

That is _different_ than the related behavior we have of showing a form changes indicator, the value _should_ be based on if the form has been touched.

jeffsee55 (approved):

PR #3231, opened 2022-09-23
Revert "Fix rich text dropdown overlap issue"

jeffsee55:
Reverts tinacms/tinacms#3177

This seems to be causing some issues. The label is under the fields. 

Screen Shot 2022-09-23 at 1 42 43 PM

I'd like to know a bit more context about this PR and why it's necessary to reverse the fields and handle the order with `flex-col-reverse`

cc @Phoenix-Alpha

### DIFF
--- a/.changeset/slow-clocks-play.md
+++ b/.changeset/slow-clocks-play.md
@@ -1,0 +1,5 @@
+---
+"@tinacms/toolkit": patch
+---
+
+Revert "Fix rich text dropdown overlap issue" which was causing field labels to live underneath their inputs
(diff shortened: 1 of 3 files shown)


logan-anderson (approved):

Phoenix-Alpha:
@jeffsee55 

I think reverting is not a good idea.

Rich text toolbars have same z-index so later siblings(in the below) overlap dropdown of previous siblings.
It is a bit hard to fix this overlap issue by changing z-index or position css because dropdown is child of toolbar.
So I reversed the order and display it in reverse using flex-col-reverse.

jeffsee55:
After discussing on Slack with @Phoenix-Alpha. We'll revert this and then rework the original issue.

PR #3235, opened 2022-09-27
Correctly update filename

logan-anderson:
Fixed an issue where the filename would not be update on an array of values.


kldavis4:
@logan-anderson when would the filename be an array?

jeffsee55 (approved):

PR #3242, opened 2022-09-28
Fix/slugigy function breaks templates

logan-anderson:
Fix issue where slugify function would break templates


packages/tinacms/src/admin/pages/CollectionCreatePage.tsx
```diff
@@ -15,6 +15,7 @@ import { Form, FormBuilder, FormStatus } from '@tinacms/toolkit'
 import { Link, useNavigate, useParams } from 'react-router-dom'
 import React, { useMemo, useState } from 'react'
 import { TinaSchema, resolveForm } from '@tinacms/schema-tools'
+import type { GlobalTemplate } from '@tinacms/schema-tools'
```
jeffsee55, line 18: I'm not sure it matters much but we'll be getting rid of GlobalTemplate, I think `CollectionTemplates` is what we want here (it has a generic so `CollectionTemplates<false>` I think)

jeffsee55 (dismissed):
Small comment, but looks good

packages/tinacms/src/admin/pages/CollectionCreatePage.tsx
logan-anderson: I just updated the types again. I don't need to use `GlobalTemplates`  but we need to use a  different template type here. We need a different one for `collections[].templates.ui` because the one at the "top level" is different from the one at the lower levels.

jeffsee55 (approved):

PR #3247, opened 2022-10-03
Fix/update object UI types

logan-anderson:
Adds `defaultItem` to object field types (already supported it was just not in the types) 

fixes: #3248


jamespohalloran (approved):
lgtm!

PR #3256, opened 2022-10-05
fix: fs.rm "callback is not a function"

jamespohalloran:
Fix for "callback is not a function" when running tinacms dev

### DIFF
--- a/packages/@tinacms/graphql/src/index.ts
+++ b/packages/@tinacms/graphql/src/index.ts
@@ -48,7 +48,7 @@ export const buildSchema = async (
   const config = await fs
     .readFileSync(path.join(tempConfig, 'schema.json'))
     .toString()
-  fs.rmSync(tempConfig, { recursive: true, force: true })
+  await fs.remove(tempConfig)
 
   // only build the files, do not index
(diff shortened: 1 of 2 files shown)


packages/@tinacms/graphql/src/index.ts
```diff
@@ -48,7 +48,7 @@ export const buildSchema = async (
   const config = await fs
     .readFileSync(path.join(tempConfig, 'schema.json'))
     .toString()
-  await fs.rm(tempConfig, { recursive: true })
+  fs.rmSync(tempConfig, { recursive: true, force: true })
```
jeffsee55, line 51: Should this be awaited?
```suggestion
  await fs.rmSync(tempConfig, { recursive: true, force: true })
```

jamespohalloran: It looks like it's just a synchronous function that returns void:

> Synchronously removes files and directories (modeled on the standard POSIX rm utility).

logan-anderson (approved):

logan-anderson (commented):
@jamespohalloran I just made an update. Wanna double check that it didn't beak anything?

PR #3260, opened 2022-10-06
Feat/add vue example

logan-anderson:
Adds an example of using Tinacms with vue.


jeffsee55:
Looks like this is related https://github.com/vitejs/vite/issues/3617

PR #3264, opened 2022-10-07
Remove esm upgrade vite 2

jeffsee55:
This can be tested with the `remove-esm-upgrade-vite-2` tag:

```
yarn add @tinacms/cli@remove-esm-upgrade-vite-2
```
Or
```
npx create-tina-app@remove-esm-upgrade-vite-2 
```

We were using the [esm](https://github.com/standard-things/esm) package, which helped when we need to convert CJS to ES modules in Yarn v2 contexts. But since we've moved to PNPM this is no longer necessary. 

ESM was used in all of our bin's, so `@tinacms/scripts` (is working already), `@tinacms/cli` and `create-tina-app` all need to be tested externally from the monorepo.

Removing the ESM package allows us to move to Vite 3 for `@tinacms/app`, there was a new node feature used in the vite react plugin that the ESM package didn't support.


jamespohalloran (approved):

jeffsee55:
Closing in favor of https://github.com/tinacms/tinacms/pull/3285

PR #3267, opened 2022-10-11
Feat/update cli for new iframe

logan-anderson:
Updates @tinacms/cli and create-tina-app to use the new "I frame" way of doing things

- Uses static init by default
    - Now asks for to choose a framework  
- create-tina-app has a hugo example

Still a WIP


jeffsee55 (approved):
Awesome, this lays the groundwork for remix/vue and many more 👍

packages/tinacms/src/react.tsx
```diff
@@ -37,3 +37,15 @@ export function useTina<T extends object>(props: {
   }, [])
   return { data } as any
 }
+
+export function useEditState(): { edit: boolean } {
+  const [edit, setEdit] = React.useState(false)
+  if (typeof window !== 'undefined') {
+    window.addEventListener('message', (event) => {
+      if (event.data?.type === 'tina:updateData') {
```
jeffsee55, line 45: This will only trigger when the user is using a `useTina` hook elsewhere on the page. but if they're on a page without the `useTina` hook they could still be in an edit state (ie. in the iframe). 

For some context, some of our users are using the `useEditState` hook to modify the image URL (when in edit mode, keep the assets.tina.io portion of the URL, when not in edit mode, use the relative path.

I think the fix here can be to add very similar logic to the current postMessage stuff, but to put it in the `preview.tsx` file and yeah, probably using a separate `type`. (I think adding the `tina:updateData` type is good and should stay.

logan-anderson: @jeffsee55 I updated the PR to resolve this issue.

packages/@tinacms/app/appFiles/src/preview.tsx
```diff
@@ -97,6 +97,12 @@ const QueryMachine = (props: {
 
   React.useEffect(() => {
     if (props.iframeRef.current) {
+      window.addEventListener('message', (event: MessageEvent<PostMessage>) => {
+        if (event?.data?.type === 'isEditMode') {
+          console.log('sending edit mode')
+          send({ type: 'EDIT_MODE' })
```
jeffsee55, line 103: Instead of sending this to the machine, why not respond back immediately?

```suggestion
          props.iframe?.contentWindow?.postMessage({
            type: 'tina:editMode',
          })
```

jeffsee55 (commented):
Since this turns on static by default for tina init, do we want to wait to merge this?

packages/@tinacms/app/appFiles/src/preview.tsx
logan-anderson: That makes more sense to me!

logan-anderson: Fixed

jamespohalloran (approved):

PR #3269, opened 2022-10-12
fix(draggable): update key

SenseiCain:
An attempt to fix an issue preventing users from drag/dropping blocks in the editor view


SenseiCain:
blah, created from a forked repo

PR #3284, opened 2022-10-17
Fix drag'n'drop impage upload not working issue

### LINKED ISSUE #2692: Drag'n'drop Image Uploader Doesn't Work
https://user-images.githubusercontent.com/5075484/158445184-10b5ae67-1454-402b-b804-119d7cdd8334.mp4

When using drag'n'drop to upload an image it doesn't complete successfully. In the setup above the image is making it to Cloudinary and shows up in the media manager, but the field isn't updated. This is inside an image component in a rich-text field.

I also find the 'click to add image' UI looks a lot like the 'drag'n'drop to upload' UI, which can be confusing:

Screen Shot 2022-03-15 at 3 20 29 PM

Screen Shot 2022-03-15 at 3 20 18 PM

### LINKED ISSUE #2851: next-tinacms-cloudinary plugin drag-n-drop should select the image by default
## Description

When I drag an image into an image field using the `next-tinacms-cloudinary` plugin, it will upload successfully to cloudinary but the image field itself won't reflect the change.  

### Steps to reproduce

Drag an image into an image field for a project using `next-tinacms-cloudinary` plugin and see the field itself does not update with the image.

### Root Cause

I have dug into the source trying to improve the cloudinary plugin for my own purposes and believe i have detected the issue.

`parse` expects the media object to have a `src` property https://github.com/tinacms/tinacms/blob/main/packages/next-tinacms-cloudinary/src/cloudinary-media-store.ts#L112

However, during the drag n drop upload process, the `src` is not set, only `previewSrc` is https://github.com/tinacms/tinacms/blob/main/packages/next-tinacms-cloudinary/src/cloudinary-media-store.ts#L68.

I do not have enough context on all the uses of this plugin but I believe renaming `previewSrc` to `src` resolves this issue, or at least also adding `src` to the object fixes it.

Phoenix-Alpha:
Closes #2692
Fix ENG-63
fixes #2851

### DIFF
--- a/.changeset/lovely-icons-listen.md
+++ b/.changeset/lovely-icons-listen.md
@@ -1,0 +1,7 @@
+---
+'next-tinacms-cloudinary': patch
+'next-tinacms-dos': patch
+'next-tinacms-s3': patch
+---
+
+Fix dran'n'drop uploaded image not shown issue
--- a/packages/next-tinacms-cloudinary/src/cloudinary-media-store.ts
+++ b/packages/next-tinacms-cloudinary/src/cloudinary-media-store.ts
@@ -109,7 +109,7 @@ export class CloudinaryMediaStore implements MediaStore {
   }
 
   parse = (img) => {
-    return img.src
+    return img.previewSrc
   }
 
   private buildQuery(options: MediaListOptions) {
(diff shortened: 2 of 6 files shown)


jamespohalloran (commented):
Tested this out with the dos media store and looks like drag & drop is working well! I'm not super familiar with the media store details though, mind giving this a second look, @logan-anderson ?

logan-anderson:
Looks good to me. Testing cloudinary and it is working.

PR #3285, opened 2022-10-17
Iframe punchlist

jeffsee55:


packages/@tinacms/toolkit/src/packages/react-sidebar/components/SidebarBody.tsx
```diff
@@ -44,7 +45,20 @@ export const FormsView = ({
       : true
   const formPlugins = cms.plugins.getType<Form>('form')
   const { setFormIsPristine } = React.useContext(SidebarContext)
-  const { formsRegistering } = React.useContext(EditContext)
+  const { formsRegistering, setFormsRegistering } =
+    React.useContext(EditContext)
+
+  React.useMemo(
```
logan-anderson, line 51: Should this be a `useEffect` ?

logan-anderson:
Another small thing (could be addressed in another PR)
On page refresh it says something like "tina could not find any forms" and then loading state and then the forms. It might be nice if it started in a loading state.

https://www.loom.com/share/3a6f061fa3a6468ca853d98473f0bd90

jeffsee55:
> On page refresh it says something like "tina could not find any forms"

Yeah, that could be a separate PR, but basically the first message is the default, so we don't know if a form is loading until we get a message from the iframe. Maybe we can delay the initial message if we care about that.

PR #3287, opened 2022-10-18
Raw-mode

jamespohalloran:
WIP:
With this approach, you don't need to supply `json` or `raw` in the query. TinaMarkdown takes in the root node, so it's mostly backwards compatible unless you're doing something custom with your rich-text nodes outside of TinaMarkdown.

Some limitations:
- Currently, the "raw" content isn't live updated. 
     - Unknown: If we want the raw content to be editable through a tab in the rich-text editor, will this approach work?
- Since you don't need to specify `raw` or `json` in the query, the response may be larger than you need. 

Not 100% if this is the best approach yet or not, so mostly just prototyping


jamespohalloran:
Chatted with the team & closing this PR. Including the raw document value under the root `_sys` property might be a better solution than auto-including it under the rich-text field.

PR #3296, opened 2022-10-20
show a deprecation message for old schema file usage

Phoenix-Alpha:
closes ENG-637

### DIFF
--- a/packages/@tinacms/cli/src/cmds/compile/index.ts
+++ b/packages/@tinacms/cli/src/cmds/compile/index.ts
@@ -314,7 +314,7 @@ export const compileSchema = async (options: {
   let schema: any
 
   // only do this if there is a schema file
-  if (schemaExists) {
+  if (schemaExists && !configExists) {
     console.warn(
       `schema.{ts,tsx,js,jsx} will soon be deprecated, in favor of the new config.{ts,tsx,js,jsx}\nSee here for migration steps, see here: https://tina.io/blog/upgrading-to-iframe`
     )
--- a/packages/@tinacms/cli/src/cmds/compile/index.ts
+++ b/packages/@tinacms/cli/src/cmds/compile/index.ts
@@ -226,18 +226,6 @@ export const compileFile = async (
     )
   }
 
-  let schemaExists = true
-  try {
-    getPath({
-      projectDir: tinaPath,
-      filename: fileName,
-      allowedTypes: ['js', 'jsx', 'tsx', 'ts'],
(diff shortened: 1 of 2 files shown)


packages/@tinacms/cli/src/cmds/compile/index.ts
```diff
@@ -226,18 +226,6 @@ export const compileFile = async (
...
-  try {
-    getPath({
-      projectDir: tinaPath,
-      filename: fileName,
-      allowedTypes: ['js', 'jsx', 'tsx', 'ts'],
-      errorMessage: `Must provide a ${fileName}.{js,jsx,tsx,ts}`,
-    })
-  } catch {
-    // getSchemaPath will throw an error if it is not found
-    schemaExists = false
```
jamespohalloran, line 239: Looks like this needed to be wrapped in a try/catch before. Do you know if that's still the case? @Phoenix-Alpha

jamespohalloran (changes requested):

packages/@tinacms/cli/src/cmds/compile/index.ts
Phoenix-Alpha: Actually, `schemaExisits` is not used anymore after this assignment and this try catch block does nothing, I think.
Also, similar file existing check is done in the next try catch block.

jamespohalloran: Are you sure? It doesn't look like they're being wrapped in a try/catch 
https://github.com/tinacms/tinacms/blob/660dc62fbbdb25647d899002b9b6106682190b24/packages/%40tinacms/cli/src/cmds/compile/index.ts#L296-L305

Phoenix-Alpha: For this, I am not sure when it throws an exception as it will return false if a file does not exist.
However, I will wrap with a try/catch block if you want. Please let me know your opinion.

jamespohalloran: Ok cool, yeah just returning true/false seems like the ideal behaviour, so that's fine by me as-is in that case!

jamespohalloran (approved):

logan-anderson (changes requested):
@Phoenix-Alpha Its _technically_  still possible to have a valid setup while having a schema.ts file. Would it be possible to only display this message when the user does not have a config.{ts,tsx,js}

logan-anderson (approved):

PR #3300, opened 2022-10-21
update error message

logan-anderson:
Update couple of error messages.

### DIFF
--- a/packages/@tinacms/toolkit/src/components/media/media-manager.tsx
+++ b/packages/@tinacms/toolkit/src/components/media/media-manager.tsx
@@ -390,7 +390,7 @@ export function MediaPicker({
           <input {...getInputProps()} />
 
           {listState === 'loaded' && list.items.length === 0 && (
-            <EmptyMediaList />
+            <EmptyMediaList hasTinaMedia={hasTinaMedia} />
           )}
 
           {list.items.map((item: Media) => (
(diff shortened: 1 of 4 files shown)


packages/@tinacms/toolkit/src/components/media/media-manager.tsx
```diff
@@ -498,8 +500,7 @@ const SyncModal = ({ close, syncFunc, folder, branch }) => {
         <ModalHeader close={close}>Sync Media</ModalHeader>
         <ModalBody padded={true}>
           <p>
-            {`This will copy all media from the \`${folder}\` folder on branch \`${branch}\` in your git repository to Tina Cloud. Are
-            you sure you would like to perform this action?`}
+            {`This will copy media assets from the \`${folder}\` folder on branch \`${branch}\` in your git repository to Tina Cloud's asset service for better performance.`}
```
jamespohalloran, line 503:
```suggestion
            {`This will copy media assets from the \`${folder}\` folder on branch \`${branch}\` in your git repository to Tina Cloud's asset service.`}
```
Nitpick, but "for better performance", makes it seem like it's an optional thing

jamespohalloran (changes requested):

packages/@tinacms/toolkit/src/components/media/media-manager.tsx
logan-anderson: Updated the message

jamespohalloran (approved):

PR #3306, opened 2022-10-25
Collections info btn

jamespohalloran:
Add an info link button to the collections label in the sidebar
Screen Shot 2022-10-25 at 10 27 33 AM


logan-anderson:
Looks good. to me!  Maybe we should only show this in "Local Mode" since it's only useful to devs?

logan-anderson (dismissed):

logan-anderson (approved):

Issue #3309, opened 2022-10-26
Z-Pos of editor dropdown menus is rendered incorrectly


nodgear:
## Description

As title suggests, this issue happens on the rich-text editor of tina.
Clicking the options button (tree dots) opens a dropdown menu, which has the z-pos lower than the content blocks.
![firefox_6UzVqTm6Kc](https://user-images.githubusercontent.com/22383863/197908910-db5fbe37-5e4a-42d8-b3b6-47112afb2816.png)

### Steps to reproduce
Click on it while having another non-text content below 

### Expected result

The dropdown menu should have a higher value (preferably some absurdly high number)

### Actual result
The dropdown is rendered below the content, making it near impossible to click on some options

### Environment
Development: Arch Linux & Windows 11+ NodeJS 16.x LTS running NextJS
Production: Vercel running NextJS + Tina cloud 

#### Tinacms Package Versions

```
"@tinacms/cli": "^0.61.24"
"tinacms": "^0.69.18"
```

#### Tools

| Tool    | Version |
| ------- | ------- |
| browser |  Firefox 106.0, Edge 106.0, Chrome: 107.0     |
| os      |    Windows 11 22H2, Linux, <TinaCloud???>     |
| node    |     16.x LTS    |
| yarn     |     1.22    |

PR #3317, opened 2022-10-26
Fix dot menu popover displayed under blocks

breizh24:
As reported on the following issue: #3309 When two blocks having the "DotMenu" are below each other, if you try to open the menu of the first block it will be shown below the second block.

The issue was solved by moving the classes that are defining the position of the element from the `div` that is setting its appearance to the `Popover.Panel` element.

Before: 
image

After:
image


jeffsee55:
Awesome! Thanks @breizh24 👍

jeffsee55 (approved):

PR #3318, opened 2022-10-26
Don't restart vite server on every schema change

jeffsee55:


logan-anderson:
@jeffsee55 When I save a file and the GQL server restarts I get a white screen and can not view the admin.

https://www.loom.com/share/9aa26c35560b44eaa06026a65d3c6aed

jeffsee55:
@logan-anderson I think this also needs to not clear out the `app/prebuild` folders in that reset logic. I'll update this PR on top of main now that that's merged and remove those two from the cleanup logic

logan-anderson (approved):

logan-anderson (changes requested):
Getting this Error when starting the server @jeffsee55 
Screen Shot 2022-10-28 at 7 32 09 AM

logan-anderson (approved):

Issue #3324, opened 2022-10-26
Toolbar "Embed" button wrong background when opened and hovered


breizh24:
## Description

The toolbar "embed" button shows a wrong background when opened and hovered.
Under these conditions the background turns blue and hides the text, since the text has the same color.

### Steps to reproduce

Click on the "Embed" button on the toolbar and move the mouse inside and outside of the button.

This is a GIF click on it please.
![toolbar-embed-button-bug](https://user-images.githubusercontent.com/38522984/198158818-56dc3b87-f2e3-4594-a073-ed4d7fd78312.gif)

### Expected result

What should happen?

The background color should not change, or the text should become white

### Actual result

What actually happened?

The background is blue and hides the text

#### Tools

| Tool    | Version |
| ------- | ------- |
| browser | Google Chrome |
| os      |  Windows 11       |
| node    |    v17.1.0     |
| npm     |   8.1.2      |

breizh24:
I can also open a PR to fix this bug if you tell me what should be the correct behavior of the button. Thank you!

PR #3325, opened 2022-10-27
Fix wrong background when hovering "Embed" button inside the toolbar

breizh24:
As described in the issue opened earlier #3324 : The toolbar "embed" button shows a wrong background when opened and hovered.
Under these conditions the background turns blue and hides the text, since the text has the same color.

The issue was solved by removing the `hover:bg-blue-500` class from the always applied classes of the button, since it has the `bg-blue-500` class when the open state is `false`. 

Let me know if this behavior is the correct one, if not, I will fix it by changing the text color.

Before: 

This is a GIF click on it please.
toolbar-embed-button-bug

After:

This is a GIF click on it please.
background-embed-button-fix

### DIFF
--- a/.changeset/strange-items-cheat.md
+++ b/.changeset/strange-items-cheat.md
@@ -1,0 +1,5 @@
+---
+'@tinacms/toolkit': patch
+---
+
+fix wrong background when hovering the "Embed" button inside the toolbar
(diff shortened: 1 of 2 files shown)


jeffsee55 (approved):
You're on a roll! Thanks again

PR #3337, opened 2022-10-28
Feat/add client build

logan-anderson:
Adds adds a new command, `tinacms build:client`, that builds the only the client and types file. This use useful for CI type checks and other situations where you need the client and types file but not the admin UI.

### DIFF
--- a/packages/@tinacms/cli/src/buildTina/index.ts
+++ b/packages/@tinacms/cli/src/buildTina/index.ts
@@ -185,6 +185,10 @@ export const buildCmdBuild = async (
     usingTs: ctx.usingTs,
     port: options.port,
   })
+  // @ts-ignore
+  if (ctx?.skipBuild) {
+    return next()
+  }
   await buildAdmin({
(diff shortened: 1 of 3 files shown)


packages/@tinacms/cli/src/buildTina/index.ts
```diff
@@ -185,6 +185,10 @@ export const buildCmdBuild = async (
     usingTs: ctx.usingTs,
     port: options.port,
   })
+  // @ts-ignore
+  if (ctx?.skipBuild) {
```
jamespohalloran, line 189: Could this be maybe called "skipAdmin" or something more specific? My first thought was that this would skip the client build

logan-anderson: Updated

jamespohalloran (approved):

jeffsee55:
Closing as we'll probably add a flag to turn off the admin UI build instead.

PR #3348, opened 2022-10-31
Display an error when node version < 14.18

Phoenix-Alpha:
close ENG-671

### DIFF
--- a/packages/@tinacms/cli/bin/tinacms
+++ b/packages/@tinacms/cli/bin/tinacms
@@ -1,3 +1,11 @@
 #!/usr/bin/env node
 
-require('../dist/index.js').init(process.argv)
+const version = process.version.slice(1)
+const [major, minor] = version.split('.')
+if (major < 14 || (major == 14 && minor < 18)) {
+  console.error(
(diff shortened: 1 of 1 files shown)


logan-anderson (changes requested):
@Phoenix-Alpha instead of this being added in the bin directly I think it would make more sense to add it as a part of the chain for each "Command". 

[See the `checkOption` chain](https://github.com/tinacms/tinacms/blob/main/packages/@tinacms/cli/src/cmds/baseCmds.ts#L105) for an example

Phoenix-Alpha:
@logan-anderson , the error happens when `cli/dist/index.js` is imported. Inside `cli/dist/index.js` there are some code which imports `node:path`. i.e.
image
I tried to put version checker inside the chain but `node:path` importing is called before my code as it is called as soon as the `cli/dist/index.js` is imported. So I needed to put version checking part before `cli/dist/index.js` is imported.

Please let me know your opinion.

logan-anderson:
@Phoenix-Alpha that makes sense to me. @jeffsee55 What do you think?

jeffsee55:
Makes sense to me.

jeffsee55 (approved):

PR #3352, opened 2022-10-31
Update config and schema types

jeffsee55:
These type updates are only external-facing for now, with hopes to deprecate (or at least alias) the other types we export from schema tools.


logan-anderson:
Seeing an error in the tina-cloud-starter-iframe:

Screen Shot 2022-11-01 at 9 13 20 AM

logan-anderson (approved):

PR #3360, opened 2022-11-01
Simplify formify resolve

jeffsee55:
The PR removes all of the logic around generating blueprints, and instead just calls the GraphQL `graphql` function client-side. This is the same function from GraphQL that we use on the backend, but instead of resolving the content from the database, we resolve it from the Tina forms. I realized I was basically recreating this logic with the blueprint stuff, and this is a much simpler approach which I think is a lot less error-prone. 

This should improve performance on formifying pages, which isn't really noticeable locally but makes a big difference on Tina cloud. Previously we were making the following queries before the user could see the forms in the sidebar:

Old pipeline
- Introspection query (we need the schema from the server)
- Optimize the query (this basically inlines fragments so it's one big query for easier parsing logic)
- Get the "formified" query (so the original content with the extra metadata we need)
- For each node we find in the query, go and refetch that

New pipeline
- Introspection query (we need the schema from the server)
- Get the "formified" query (so the original content with the extra metadata we need)

---

In the near future we can also skip the introspection query since we actually already build that during `tina build` (`.tina/__generated__/_graphql.json`) but I've left that out for now because that only runs on an actuall window page load, so once the SPA mounts it doesn't run again.

### DIFF
--- a/.changeset/eleven-donkeys-allow.md
+++ b/.changeset/eleven-donkeys-allow.md
@@ -1,0 +1,5 @@
+---
+'@tinacms/app': patch
+---
+
+Simplify the way form values are resolved into the data payload on iframe
--- a/examples/kitchen-sink/.tina/queries/documents.gql
+++ b/examples/kitchen-sink/.tina/queries/documents.gql
@@ -14,6 +14,17 @@ query CollectionQuery {
   }
 }
 
+query NodeQuery($id: String!) {
+  node(id: $id) {
+    ...on Document {
+      ...PageParts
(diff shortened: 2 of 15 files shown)


logan-anderson:
I think I found an issue with templates (it also could be something wrong with my implementation)

Loom: https://www.loom.com/share/dd62ce2c6fe94d3b80059d394bb43a02
Branch: https://github.com/tinacms/tinacms/tree/formify-resolve-show-error

logan-anderson (approved):

PR #3366, opened 2022-11-02
Feat/nest default item in UI

logan-anderson:
Nest `defaultItem` under UI instead of at the root. This also enables defaultItem with templates


jeffsee55:
Closing because I'm manually pulling these into another PR

PR #3373, opened 2022-11-03
Include node_modules as part of the build in @tinacms/app

jeffsee55:
Previously we were running a special `npm install` command whenever we started the Tina dev server to get the `@tinacms/app` dependencies in the right location. However that command has failed for some users for reasons that aren't super clear. It also feels a little bit brittle to run `npm install` and aside from a larger bundle for the `@tinacms/app` I don't see any downsides.

### DIFF
--- a/examples/kitchen-sink/package.json
+++ b/examples/kitchen-sink/package.json
@@ -4,6 +4,7 @@
   "private": true,
   "scripts": {
     "dev": "MONOREPO_DEV=true tinacms dev -c \"next dev\"",
+    "dev2": "tinacms dev -c \"next dev\"",
     "dev-sandbox": "MONOREPO_DEV=true tinacms dev -c \"next dev\" --sandbox",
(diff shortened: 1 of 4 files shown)


examples/kitchen-sink/package.json
```diff
@@ -4,6 +4,7 @@
   "private": true,
   "scripts": {
     "dev": "MONOREPO_DEV=true tinacms dev -c \"next dev\"",
+    "dev2": "tinacms dev -c \"next dev\"",
```
kldavis4, line 7: commit?

kldavis4 (approved):

examples/kitchen-sink/package.json
jeffsee55: Yeah it's nice to debug without the `MONOREPO_DEV` variable, but usually during dev that's what you want.

Issue #3371, opened 2022-11-03
Nested  rich-text 3 levels deep fails to resolve data


logan-anderson:
## Description

Using 3 rich-text that is nested 3 levels deep tina will work durring edit mode (with a ton of error message to the console) but will fail to resolve the data.

### Steps to reproduce

Add the following field to a collection in the schema

```
{
          type: "rich-text",
          label: "Body",
          name: "_body",
          templates: [
            {
              name: "Test",
              label: "Two Column Layout",
              fields: [
                {
                  name: "leftColumn",
                  label: "Left Column",
                  type: "rich-text",
                },
                {
                  name: "rightColumn",
                  label: "Right Column",
                  type: "rich-text",
                  templates: [
                    {
                      name: "Highlight",
                      label: "Highlight block",
                      fields: [
                        { name: "content", label: "Body", type: "rich-text" },
                      ],
                    },
                  ],
                },
              ],
            },
```

add a new "TwoColumn Layout" > "Right Column" > Add a new "Highlight block"

Edit the text in the highlight block (See error message in the console )

Click save

Refresh page and see tina does not resolve the data

See video for more details: https://www.loom.com/share/6b8d3e888a41455e95332e6f5fa824c2

### Expected result

It should resolve the data like in edit mode

### Actual result

It does not resolve the data

### Environment

Mono repo: Here is the branch that has the error recreated: https://github.com/tinacms/tinacms/tree/bug/nested-markdown-issue

cc @jeffsee55

raphaeltm:
So, this is now fixed for me when I run the development server, but it's still breaking if I run a production build.

raphaeltm:
> So, this is now fixed for me when I run the development server, but it's still breaking if I run a production build.

Nevermind. I literally added some arbitrary text somewhere (a `package.json` script that just echoes "hello") and pushed (the diff shoed that was the only thing that changed) and now it works. I suspect Vercel does some caching in their builds which may have thrown things off? (the builds that made me think this was a problem only took ~2min and were run not long after I upgraded to the fixed packages... my latest build took ~5min and fixed the issue)

PR #3375, opened 2022-11-03
Fix issue with deeply nested rich-text parsing

jeffsee55:
Extract raw string for nested rich-text elements and use that for recursive parsing

Closes https://github.com/tinacms/tinacms/issues/3371


logan-anderson (approved):
Looks good

PR #3384, opened 2022-11-05
fix OverflowMenu of collection items not correctly placed on top of other elements

breizh24:
## Issue description

The `OverflowMenu` of collection items is not correctly placed on top of the other elements.
When the popover is open, some of the elements that should be under it are visible and clickable.

You can see the issue here: https://www.loom.com/share/f0fcb48284374cbe889da8bda0ae1ba2

## Solution

I have fixed the issue by moving the classes that are defining the position of the element from the div that is setting its appearance to the `Popover.Panel` element.

You can see the working solution here: https://www.loom.com/share/c6011fe901794771a784419d8e45bc9a

This issue was similar to the one I have already solved in this PR #3317

### DIFF
--- a/.changeset/three-phones-fix.md
+++ b/.changeset/three-phones-fix.md
@@ -1,0 +1,5 @@
+---
+'@tinacms/toolkit': patch
+---
+
+Fix elements displayed above the OverflowMenu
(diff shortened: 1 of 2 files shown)


jeffsee55 (approved):
Thanks @breizh24 !

PR #3394, opened 2022-11-10
improve error handling while creating a collection file

### LINKED ISSUE #3362: Creating a post with a filename that already exists doesn't handle error gracefully
## Description

When creating a post with a filename that already exists, creation fails without showing an error. It also disables the flename generation, and the ability to save

https://www.loom.com/share/9cb22b7c4bc84b13abab5123aac49a97

### Expected result

Error message should appear. User should be able to correct the error and re-save.

#### Tinacms Package Versions

├── @tinacms/cli@0.62.0
├── tinacms@0.70.0

Phoenix-Alpha:
Closes ENG-65 and ENG-674
Fix #3362

### DIFF
--- a/packages/@tinacms/graphql/src/database/index.ts
+++ b/packages/@tinacms/graphql/src/database/index.ts
@@ -234,7 +234,7 @@ export class Database {
       }
       return true
     } catch (error) {
-      throw new TinaFetchError(`Error in PUT for ${filepath}`, {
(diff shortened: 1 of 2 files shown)


jamespohalloran:
Thanks for this, but looks like it was fixed by an external contributor here: https://github.com/tinacms/tinacms/pull/3383

PR #3408, opened 2022-11-18
Pass `config.formifyCallback` to `<QueryMachine>`

coreyaus:
## Implementation notes

The `formifyCallback` is expected by the `<QueryMachine />` machine as you can see on this line, which accesses the `props.formifyCallback` value: https://github.com/tinacms/tinacms/blob/main/packages/@tinacms/app/appFiles/src/preview.tsx#L86. And that `formifyCallback` function is a valid option on the `Config` object: https://github.com/tinacms/tinacms/blob/main/packages/tinacms/src/index.ts#L96-L109, however it wasn't being passed through to the component that was attempting to use it. This update fixes that bug by ensuring the `formifyCallback` function is properly passed through, making it available within the `context` for the `query-machine` and `document-machine` respectively.

## Problem description

Using the latest version of TinaCMS, you can see this issue by setting a `formifyCallback` like the following within the config object passed to Tina's `defineConfig` function:

```typescript
formifyCallback: (
  { formConfig, createForm, createGlobalForm, skip },
  cms
) => {
  console.log('This message will never get logged cos the formifyCallback isn't being passed through!')
  console.log('formConfig', formConfig)

  return createForm(formConfig)
}
```

We would expect to see a message printed to the console, but it doesn't happen as the custom `formifyCallback` is not passed through to the `<QueryMachine />` component, so it never gets invoked (the default `formify` function runs instead)

Let me know if you need me to make any changes to this PR. Given it's a bug fix I figured we probably don't need to update the README or docs, but happy to tweak the implementation or make other updates where you need 👍 

PS: Amazing work on Tina - it's an incredibly powerful tool! 🎉

### DIFF
--- a/packages/@tinacms/app/appFiles/src/preview.tsx
+++ b/packages/@tinacms/app/appFiles/src/preview.tsx
@@ -14,6 +14,7 @@ import React from 'react'
 import { useMachine } from '@xstate/react'
 import { queryMachine, initialContext } from './lib/machines/query-machine'
 import { useCMS, defineConfig } from 'tinacms'
+import type { formifyCallback as FormifyCallback } from 'tinacms/dist/hooks/use-graphql-forms'
 
 type Config = Parameters<typeof defineConfig>[0]
 
@@ -48,6 +49,7 @@ export const Preview = (
           key={activeQuery.id}
           payload={activeQuery}
           iframeRef={props.iframeRef}
+          formifyCallback={props.formifyCallback}
         />
       )}
       <div className="h-full overflow-scroll">
@@ -72,6 +74,7 @@ export const Preview = (
 const QueryMachine = (props: {
   payload: PostMessage
   iframeRef: React.MutableRefObject<HTMLIFrameElement>
+  formifyCallback: FormifyCallback
 }) => {
   const cms = useCMS()
 


jeffsee55:
Thanks @coreyaus ! We are wondering how people are using `formifyCallback` and what we need to keep in mind to support it going forward. Now that you can mark global templates with the `global: true` flag, what else are you using it for?

jeffsee55 (approved):

coreyaus:
Thanks for looking at this Jeff (and for merging in these changes!)

There are a few reasons we like the `formifyCallback` function - I'll outline a couple of them in case it's useful:

1. Naming global forms
    - It's handy to have an easy way to control the label/name for the sidebar links that open global forms
    - This is particularly true for documents that come from a collection with templates (so the ideal label/name for the form in the sidebar depends on the template, not just the collection name or file name)

2. Skipping particular forms
    - Again this is particularly handy when requesting documents in a collection with templates (where we want to skip the global form for documents with a particular template but keep the global form for other documents with a different template in the same collection)

3. Updating the `formConfig` in programatic ways
    - The `formifyCallback` allows for some customisation that either can't be achieved directly in the Tina schema or would require a bunch of duplication
    - There are a couple of ideas here I plan to test in the coming days so I'm not sure how important the `formifyCallback` will be in the final version (I might be able to do what I want with a custom Tina field component). But I think the `formifyCallback` could still be really helpful when customising the component for a bunch of Tina fields, e.g. to hide/display them conditionally. I saw you proposed a great idea in [this Discord thread](https://discord.com/channels/835168149439643678/985928508356313128) that could help with some of these types of changes if implemented (passing the original `Component` as a prop to the custom `component` function) - it would mean being able to write code in the Tina schema like the screenshot below. In the meantime I reckon the `formifyCallback` will provide a handy way to achieve some creative use cases like that - particularly for experimentation given it makes it easier to apply the same customisation to a bunch of fields on bulk, and speeds up the reload time if we don't need to regenerate the schema to test each iteration.

customising-a-tina-field-component

PR #3411, opened 2022-11-20
ui.itemProps on collections

newash:
Hello!

I created this PR for #3389 

I tried it with `experimental-examples/tina-cloud-starter` adding the `draft` field and the `ui.itemProps` to the collection schema:

```js
...
const schema = defineSchema({
  collections: [
    {
      label: "Blog Posts",
      name: "posts",
      path: "content/posts",
      format: "mdx",
      ui: {
        router: ({ document, collection }) => {
          return `/${collection.name}/${document._sys.filename}`;
        },
        itemProps: (item) => ({
          label: item.draft ? `🚧 ${item.title}` : item.title,
          className: item.draft ? 'bg-blue-50' : ''
        })
      },
      fields: [
        {
          type: "boolean",
          name: "draft",
          label: "Draft",
        },
...
```

and could produce this:

2022-11-20_15h58_05

---

I'm not sure though if both `packages/@tinacms/schema-tools/src/types.ts` and `packages/@tinacms/schema-tools/src/types/SchemaTypes.ts` file needed to be updated, or one would have been enough.

I also haven't looked into your tests, so I'm not sure about the necessary tests either.

### DIFF
--- a/packages/tinacms/src/admin/api.ts
+++ b/packages/tinacms/src/admin/api.ts
@@ -97,6 +97,7 @@ export class TinaAdminApi {
                     filename
                     extension
                   }
+                  _values
                 }
               }
             }
@@ -142,6 +143,7 @@ export class TinaAdminApi {
                     filename
                     extension
                   }
+                  _values
                 }
               }
             }
--- a/.changeset/shiny-steaks-joke.md
+++ b/.changeset/shiny-steaks-joke.md
@@ -1,0 +1,6 @@
+---
+'@tinacms/schema-tools': minor
+'tinacms': minor
+---
+
+add ui.itemProps on collections
--- a/packages/@tinacms/schema-tools/src/types.ts
+++ b/packages/@tinacms/schema-tools/src/types.ts
@@ -73,6 +73,15 @@ export interface UICollection {
     create?: boolean
     delete?: boolean
   }
+  /**
+   * An optional function to override the properties of the collecion table
+   * row component. Except for the `label` property:
+   * that is used as the title of the collection element.
+   */
+  itemProps?: (item: Record<string, any>) => {
+    key?: string
+    label?: string
+  } & Record<string, any>
 }
 export type Option =
   | string
--- a/packages/@tinacms/schema-tools/src/types/SchemaTypes.ts
+++ b/packages/@tinacms/schema-tools/src/types/SchemaTypes.ts
@@ -131,6 +131,15 @@ export interface UICollection {
     document: Document
     collection: TinaCloudCollection<true>
   }) => string | undefined
+  /**
+   * An optional function to override the properties of the collecion table
+   * row component. Except for the `label` property:
+   * that is used as the title of the collection element.
+   */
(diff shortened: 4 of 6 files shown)


packages/tinacms/src/admin/api.ts
```diff
@@ -97,6 +97,7 @@ export class TinaAdminApi {
                     filename
                     extension
                   }
+                  _values
```
newash, line 100: While the PR is nicely small, you might agree that it's not very bandwidth-efficient.

What I could change here, is to introduce a `_sys.fields` attribute
in `packages/@tinacms/graphql/src/resolver/index.ts` instead:
an object that would have only the subset of fields.

The question is: what that subset should contain?

- everything but the one with `isBody: true` ?
- everything but those with `type: "rich-text"`?
- only those with simple types? (not `rich-text` and not `object`)

That is a design decision, and you probably in a better position than me to answer that.

jeffsee55: Yeah this is a good point, it has the potential to de-optimize that query, it might be best to provide a property on field definitions that would populate it to the `_sys.fields` property, as you noted. 

It's not so much a matter of the size of payload, but also if we can avoid providing the `_values` then we can skip retrieving the document entirely (basically the same thing we do with `_sys.title`, except it'd be a `JSON` type probably.

jeffsee55:
Thanks for the PR @newash - we'll discuss the ramifications of including the `_values` key as you mention, but this is a solid approach!

newash:
Do you have any updates on this @jeffsee55 ?

jeffsee55:
We'll be discussing this next week and we'll get back to you, thanks @newash !

packages/tinacms/src/admin/api.ts
jamespohalloran: Just thinking ahead a bit, If the user can specify which fields they want available to that list-view, maybe we should allow them to create new columns for a given field?
In that case, instead of using itemProps to control the label field, they'd be able to control the UI for any given column, as opposed to just the title-label.

Screen Shot 2023-01-17 at 9 39 00 AM

(Another common example might be including a custom "author" field in this list).

newash:
Any update lads?

jeffsee55:
Hey @newash - sorry for the delay, this PR is definitely something we'd like to have but it introduces more complexity than we'd thought. We don't have the bandwidth right now to approach it properly, but do plan to address it.

newash:
I still need this

jeffsee55:
not stale

newash:
still need this

newash:
will it ever be addressed?

scottgallant:
Hey @newash, I've been thinking a lot about the suggestion here and would like to go that route:

https://github.com/tinacms/tinacms/pull/3411#discussion_r1072221450

What do you think? 

Our roadmap is fully booked for the next few weeks, but I'll try to slot this in shortly after that.

newash:
Of course, I'm not going to argue against your obviously neater solution. 🙂

PR #3413, opened 2022-11-22
use env vars when building admin app

logan-anderson:
Fixes issue described in https://tina-cms.slack.com/archives/C0349J7LAFK/p1669142134141619

cc @jeffsee55


jeffsee55:
@logan-anderson have you tested this for both `dev` and `build` mode? Only potential issue is what we ran into with the `process.env.PRODUCTION` getting set and messing everything up.

logan-anderson:
> PRODUCTION



> @logan-anderson have you tested this for both `dev` and `build` mode? Only potential issue is what we ran into with the `process.env.PRODUCTION` getting set and messing everything up.

I did not run into any error. I think that this would only happen if we had the `mode` incorrect?

jeffsee55 (approved):

jeffsee55:
If this turns out to be buggy we can always just be explicit about allowing the `tinaio` config variables through.

PR #3427, opened 2022-11-29
Fix issue resolving alias fields when passing through the `useTina` hook

jeffsee55:
How we're using graphql in the client is unconventional, and we actually need to look at `info.fieldNodes` to determine where the data might actually be. This isn't usually the way things work, but since we're resolving data from a _previous_ query (that ran before being passed to the `useTina` hook, our needs are a bit different, hence the awkward workaround.

### DIFF
--- a/packages/@tinacms/app/appFiles/src/lib/machines/query-machine.ts
+++ b/packages/@tinacms/app/appFiles/src/lib/machines/query-machine.ts
@@ -399,7 +399,6 @@ export const queryMachine =
               return value
             },
           })
-          console.log(newData)
           if (missingForms.length > 0) {
             // Only run this one at a time
             const missingForm = missingForms[0]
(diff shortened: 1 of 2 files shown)


jeffsee55:
This PR introduced a query that we didn't support, that should now be working with this PR - https://github.com/tinacms/tina.io/pull/1440

packages/@tinacms/app/appFiles/src/lib/machines/query-machine.ts
```diff
@@ -372,9 +396,10 @@ export const queryMachine =
                   return null
                 }
               }
-              return source[fieldName]
+              return value
             },
           })
+          console.log(newData)
```
jamespohalloran, line 402:
```suggestion
```

jamespohalloran (changes requested):

jamespohalloran (approved):

PR #3428, opened 2022-11-30
added table component to mdx

Phoenix-Alpha:


jeffsee55 (commented):
I'm not sure this is worth pursuing at the moment, getting everything to work from the backend to the rich-text editor and the `<TinaMarkdown>` component isn't something we've scoped and admittedly I know I haven't provided enough context to get this done properly. A lot of what you have here is a good start, but this shouldn't be merged in its current form. Might be best to plan this one out a bit more. I think we should close this and use it as a reference when we're ready to tackle table support.

jeffsee55:
Closingg as this needs to be scoped more

PR #3436, opened 2022-12-01
Feat/merge updates

logan-anderson:
Content added is now merged with its existing content.

CC @jamespohalloran 
CC @jeffsee55  Is this a good way to implement this?


jamespohalloran:
> CC @jamespohalloran IS this a good API?

Looks good to me! I'm curious down the road if this config should be the default (or always on?) Are there reasons we'd want to delete un-modelled fields on save?

packages/@tinacms/graphql/src/resolver/index.ts
```diff
@@ -340,6 +341,9 @@ export class Resolver {
     isAddPendingDocument: boolean
     isCollectionSpecific: boolean
   }) => {
+    const doc = await this.getDocument(realPath)
+
+    const oldDoc = doc?._rawData || {}
```
jeffsee55, line 346: Why aren't we able to use `_values` for this?

logan-anderson: I didn't think of using `_values` but I just tested it out and it looks like `_values` does not have the additional fields. 

For example:
`_values`
```js
{
     _collection: 'posts',
    _template: 'posts',
    title: 'Just Another Blog Post asdf',
    heroImg: undefined,
    excerpt: { type: 'root', children: [Array] },
    author: 'content/authors/napolean.md',
    date: '2021-07-12T07:00:00.000Z',
    _body: { type: 'root', children: [Array] }
}
```

`_rawvalues`

```diff
{
+  foo: 'bar',
    title: 'Just Another Blog Post asdf',
    excerpt: 'Lorem markdownum evinctus ut cape adhaeret gravis licet progenies ut haesit maxima ille. Est scorpius, mori vel in visaeque Haemoniis viperei furoris e ad vasti, distulit. Crudus sub coniuge iam: dea propera sive\n',
    author: 'content/authors/napolean.md',
    date: '2021-07-12T07:00:00.000Z',
    _body: '\n' +
      '# Vote For Pedro\n' +
      '\n' +
      'Hello, the current date and time is <DateTime format="iso" />. Quis semper [vulputate](https://example.com) aliquam venenatdis egestas sagittis quisque orci. Donec commodo sdit viverra aliquam porttitor ultrices gravida eu. Tincidunt leo, elementum mattis elementum ut nisl, justo, amet, mattis. Nunc purus, diam cdommodo tincidunt turpis. Amet, duis sed elit interdum dignissim\n' +
      '\n' +
      '![](http://res.cloudinary.com/deuzrsg3m/image/upload/v1634858384/tina-upload_yxevty.jpg)\n' + "..."
     _collection: 'posts',
    _keepTemplateKey: false,
    _template: 'posts',
    _relativePath: 'anotherPost.mdx',
    _id: 'content/posts/anotherPost.mdx'
}
```

packages/@tinacms/mdx/src/stringify/acorn.ts
```diff
@@ -64,7 +64,7 @@ export function stringifyProps(
   Object.entries(element.props).forEach(([name, value]) => {
     const field = template.fields.find((field) => field.name === name)
     if (!field) {
-      if (name === 'children') {
```
logan-anderson, line 67: @jeffsee55  I am guessing this can be removed?

kldavis4 (commented):
@logan-anderson this looks good to me (except the changeset message) - only thing I am unclear on is the reason for the change to the kitchen-sink tests

.changeset/thirty-cheetahs-speak.md
```diff
@@ -0,0 +1,6 @@
+---
+'@tinacms/graphql': patch
+'@tinacms/schema-tools': patch
+---
+
+Contented is not merged with existing content. This means if you have a field that is not defined in the schema it will not be overridden.
```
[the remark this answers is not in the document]
logan-anderson: @kldavis4 I updated the changeset

logan-anderson:
> @logan-anderson this looks good to me (except the changeset message) - only thing I am unclear on is the reason for the change to the kitchen-sink tests



The reason the kichen-sink tests had to be updated is because the order of the keys is some cases is a bit different now. Also, some of the kitchen-sink tests where testing the old behaviour (where any key now in the mutation would get deleted)

kldavis4:
> > @logan-anderson this looks good to me (except the changeset message) - only thing I am unclear on is the reason for the change to the kitchen-sink tests
> 
> The reason the kichen-sink tests had to be updated is because the order of the keys is some cases is a bit different now. Also, some of the kitchen-sink tests where testing the old behaviour (where any key now in the mutation would get deleted)

The kitchen sink tests switched from using updateDocument to updatePage: https://github.com/tinacms/tinacms/pull/3436/files#diff-205dbaa814e14382091e8f7f69adca13b80cfb7e95e4e46549ce932b481c2446

That doesn't seem related to the key order / mutation behavior

logan-anderson:
> > > @logan-anderson this looks good to me (except the changeset message) - only thing I am unclear on is the reason for the change to the kitchen-sink tests
> > 
> > 
> > The reason the kichen-sink tests had to be updated is because the order of the keys is some cases is a bit different now. Also, some of the kitchen-sink tests where testing the old behaviour (where any key now in the mutation would get deleted)
> 
> The kitchen sink tests switched from using updateDocument to updatePage: https://github.com/tinacms/tinacms/pull/3436/files#diff-205dbaa814e14382091e8f7f69adca13b80cfb7e95e4e46549ce932b481c2446
> 
> That doesn't seem related to the key order / mutation behavior

I can switch those back. 

All of those where updated automatically buy running  [`jest -u`](https://jestjs.io/docs/cli#--updatesnapshot) as per jeffs instructions.

kldavis4 (approved):

kldavis4 (approved):

PR #3439, opened 2022-12-05
add more framework options

Phoenix-Alpha:
closes ENG-689


logan-anderson:
@Phoenix-Alpha This is great! But Since we know the asset folder based on the framework they choose is there anyway we could just skip choosing the asset folder if they choose a framework.

 If they choose "other" they should still be asked to choose a static folder folder.
 
 Thanks @Phoenix-Alpha

logan-anderson:
closing in favour of #3455

PR #3443, opened 2022-12-05
added onLogout hook function

logan-anderson:
Added `onLogout` hook function and a logout redirect page in the admin.

```ts
export default defineConfig({
   admin: {
      auth: {
        onLogout: () => {
          console.log('Logged out!')
          // hook function to be called when the user logs out
        },
      },
    },
  // ...
})
```


packages/@tinacms/schema-tools/src/types.ts
```diff
@@ -534,6 +534,7 @@ export interface Config<
   admin?: {
     auth?: {
       onLogin?: (args: { token: TokenObject }) => Promise<void>
+      onLogout: () => Promise<void>
```
kldavis4, line 537: should this be optional?

packages/@tinacms/toolkit/src/packages/react-sidebar/components/Nav.tsx
```diff
@@ -129,9 +129,13 @@ export const Nav = ({
                             sidebarWidth: null,
                             resizingSidebar: false,
                           })
-                          setEdit(false)
-                          if (cms?.api?.tina?.onLogout) {
-                            cms.api.tina.onLogout()
+                          if (cms?.api?.tina?.logout) {
+                            cms.api.tina.logout().then(async () => {
```
kldavis4, line 133: should the way this is called in LogoutPage be consistent (await vs then())

packages/tinacms/src/internalClient/index.ts
```diff
@@ -425,7 +427,7 @@ mutation addPendingDocumentMutation(
     return !!(await this.getUser())
   }
 
-  async onLogout() {
+  async logout() {
```
kldavis4, line 430: would it make sense to call the event handlers from these functions instead of in Nav & LogoutPage?

logan-anderson: That does make sense but we might allow the user to pass a `login` function as well so I kept them separate.

packages/@tinacms/schema-tools/src/types.ts
logan-anderson: yes it should

packages/@tinacms/toolkit/src/packages/react-sidebar/components/Nav.tsx
logan-anderson: yup, makes sense! I will fix it.

logan-anderson: Fixed

packages/tinacms/src/internalClient/index.ts
logan-anderson: We could make a `templateLogin` function that calls both of them.

packages/@tinacms/schema-tools/src/types.ts
logan-anderson: Fixed (I fixed it yesterday but forgot to push it up)

kldavis4 (approved):

PR #3444, opened 2022-12-05
Use new types for schema-tools internally

jeffsee55:
We recently updated the types for `defineConfig`, but were still using the older ones internally. We also had a bunch of different names for stuff like `TinaCloudCollection/Collection/TinaCloudCollectionInternal`. This uses new types in `@tinacms/schema-tools` for throughout the app.
 
- [x] Clean up the types in `@tinacms/schema-tools`
- [x] Move schema-related logic into the `TinaSchema` class
  - [x] `addNamespaceToSchema` is now called in the constructor
  - [x] `resolveForm` is now `tinaSchema.resolveForm` (eg [resolveForm](https://github.com/tinacms/tinacms/blob/983adace1e50902200b4f83ac0cc82fc99518e22/packages/@tinacms/app/appFiles/src/lib/machines/document-machine.ts#L181)). It also only requires the template name, rather than the entire template.
  - [x] Remove some checks for "global templates" which aren't supported, but were present in our types.
  - [x] Includes React types hardcoded into `@tinacms/schema-tools` to remove the React dependency on it (we were only using it for it's types).
  - [x] Removes some of the duplicate logic from `@tinacms/graphql` and `@tinacms/schema-tools`
  - [x] No longer exports schema types from `@tinacms/cli`, since those were for external use and `defineConfig` now comes from `tinacms`.

NOTE: there's a difference in the schema shape for what is defined in `defineConfig` and what the type actually ends up being. Basically we have a `namespace` property on a schema has been passed into the `TinaSchema` class (via the `addNamespaceToSchema` function. For the most part, we're dealing with the namespaced version of these types, so those are exported from `@tinacms/schema-tools`, while the non-namespaced types are only exported from `@tinacms/schema-tools/dist/types` _and are re-exported from `tinacms` for public use. Basically, we shouldn't be using these anywhere except for the external facing config.


packages/@tinacms/app/appFiles/src/lib/machines/document-machine.ts
```diff
@@ -183,27 +178,11 @@ export const documentMachine =
...
-              `Unable to find template for node ${node._internalSys.path}`
-            )
-          }
-          const resolvedForm = resolveForm({
+          const resolvedForm = schema.resolveForm({
             collection,
             basename: node._internalSys.filename,
             schema,
-            template,
+            templateName: node._internalSys.template,
```
jeffsee55, line 185: Now called with just the template name, so no need to find the template

packages/@tinacms/cli/src/index.ts
```diff
@@ -19,14 +19,6 @@ import { baseCmds as baseCommands } from './cmds/baseCmds'
...
 export { defineSchema } from './cmds/compile'
 import { logger } from './logger'
-export type {
-  TinaCloudSchema,
-  TinaSchema,
-  TinaCloudCollection,
-  TinaCollection,
-  TinaField,
-  TinaTemplate,
-} from '@tinacms/graphql'
```
jeffsee55, line 29: No longer exported since types in the .tina/config.ts should come from tinacms.

packages/@tinacms/graphql/src/resolver/filter-utils.ts
```diff
@@ -126,27 +113,17 @@ const collectConditionsForChildFields = (
...
-    for (const template of field.templates) {
-      if (typeof template === 'string') {
-        globalTemplates[template] = 1
-      }
-    }
-
     for (const [filterKey, childFilterNode] of Object.entries(filterNode)) {
-      if (filterKey in globalTemplates) {
-        throw new Error('Global templates not yet supported for queries')
-      }
```
jeffsee55, line 145: No longer runs the global template check

packages/@tinacms/graphql/src/resolver/filter-utils.test.ts
```diff
@@ -365,36 +365,6 @@ describe('resolveReferences', () => {
       `"Template nonExistentTemplate not found"`
     )
   })
-
-  it('resolves reference in object with template where filter references global template', async () => {
```
jeffsee55, line 369: Remove tests related to global template check

packages/@tinacms/graphql/src/schema/createSchema.ts
```diff
@@ -10,20 +10,20 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
...
 
 export const createSchema = async ({
   schema,
   flags = [],
 }: {
-  schema: TinaCloudSchemaBase
+  schema: Schema | SchemaWithNamespace
   flags?: string[]
 }) => {
-  const validSchema = await validateSchema(schema)
```
jeffsee55, line 26: There's another `validateSchema` function in `@tinacms/schema-tools`, and it's getting called from `defineConfig`, so I don't think this is necessary anymore

packages/@tinacms/schema-tools/src/types/index.ts
```diff
@@ -1,19 +1,705 @@
...
-
 Unless required by applicable law or agreed to in writing, software
 distributed under the License is distributed on an "AS IS" BASIS,
 WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 See the License for the specific language governing permissions and
 limitations under the License.
-
 */
+// Easier to copy this from hardcoded react file than keeping it as a dependency
+import type React from './react'
```
jeffsee55, line 14: Just copied the types from react and hardcoded them, I think our needs are simple enough for that to work

packages/@tinacms/schema-tools/src/schema.ts
```diff
@@ -103,16 +99,20 @@ export class TinaSchema {
...
+    // const globalTemplate = this.schema.templates?.find(
+    //   (template) => template.name === templateName
+    // )
+    // if (!globalTemplate) {
+    //   throw new Error(
+    //     `Expected to find global template of name ${templateName}`
+    //   )
+    // }
+    // return globalTemplate
+    return null
```
jeffsee55, line 115: Will remove this in a future PR

packages/@tinacms/schema-tools/src/schema.ts
```diff
@@ -10,17 +10,15 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
...
-  TinaFieldEnriched,
-} from '../types/index'
-import { lastItem, assertShape } from '../util'
+  Schema,
+  Collection,
+  SchemaField,
+  Template,
+  ObjectField,
+} from './types/internal'
+import { Schema as SchemaBase } from './types'
```
jeffsee55, line 20: The ones from types are non-namespaced

packages/tinacms/src/admin/pages/CollectionCreatePage.tsx
```diff
@@ -145,19 +143,19 @@ const RenderForm = ({ cms, collection, templateName, mutationInfo }) => {
...
+  const formInfo = schema.resolveForm({
     collection: schemaCollection,
     basename: schemaCollection.name,
     schema: schema,
-    template,
+    templateName: template.name,
   })
 
-  let slugFunction = template?.ui?.filename?.slugify
+  let slugFunction = schemaCollection?.ui?.filename?.slugify
```
jeffsee55, line 158: `template` should never have these properties on it, but `getTemplateForData` returns a template _or_ a collection depending on if the collection has templates or fields. This just makes it more clear that we're only getting this value off of the collection, if it's present.

packages/tinacms/src/internalClient/index.ts
```diff
@@ -82,7 +79,7 @@ export class Client {
       const enrichedSchema = new TinaSchema({
         version: { fullVersion: '', major: '', minor: '', patch: '' },
         meta: { flags: [] },
-        ...addNamespaceToSchema({ ...options.schema }, []),
```
jeffsee55, line 85: No need to add namespace here since this is now done in the constructor

PR #3453, opened 2022-12-07
Field Improvements

spbyrne:
Work-in-progress form component improvements. Just patching some issue that I've noticed.

- Adds ability to remove label by setting label to `false`, leverages this for the string list field: 
Screenshot 2022-12-07 at 4 13 17 PM

- Fixes Group, Block and List label UI. Block labels/descriptions will now be properly aligned, a gap will prevent text from running into the add button, and labels will be truncated:
Screenshot 2022-12-07 at 5 02 42 PM

- Textareas get a proper border:
Screenshot 2022-12-07 at 5 05 36 PM

- Improved Boolean toggle clarity:
Screenshot 2022-12-08 at 11 39 43 AM

- Improved Checkbox group styles, added 'direction' property to match Radio group:
Screenshot 2022-12-08 at 3 42 56 PM

- Improved Radio group styles:
Screenshot 2022-12-08 at 3 38 36 PM

- Adds toggle button field:
Screenshot 2022-12-09 at 12 20 08 PM

- Improved tags field UX/styles:
Screenshot 2022-12-09 at 11 30 21 AM
Screenshot 2022-12-09 at 11 31 05 AM

- Adds hidden field (`ui: { component: 'hidden' }`) 

TODO: Docs


packages/@tinacms/schema-tools/src/schema/resolveField.ts
```diff
@@ -77,11 +77,26 @@ export const resolveField = (
...
+            options: field.options,
+          }
+        }
         return {
           component: 'select',
           ...field,
           ...extraFields,
-          options: [{ label: `Choose an option`, value: '' }, ...field.options],
+          options:
+            field.ui && field.ui.component !== 'select'
```
logan-anderson, line 97: @spbyrne What is the motivation for adding this check for component !== 'select' ?

logan-anderson (changes requested):
Just a couple of small things

packages/@tinacms/schema-tools/src/schema/resolveField.ts
spbyrne: Things like radio buttons & checkbox groups shouldn't have an option labelled 'Choose an option', but it makes sense in the case of select otherwise the first option would be displayed even though it's not technically selected.

logan-anderson: Makes sense!

logan-anderson (dismissed):

logan-anderson (approved):

PR #3462, opened 2022-12-12
fix useEdit state from requiring querys to be active

logan-anderson:
Previously, `useEditState` would not work if called before `useTina`. This PR fixes it so that `useEditState` work no matter what.


examples/basic-iframe/pages/index.js
```diff
@@ -1,14 +1,17 @@
...
+import { useEditState, useTina } from 'tinacms/dist/react'
 
 const query = `query PageQuery {
   page(relativePath: "home.mdx"){
     body
   }
 }`
 export default function Home(props) {
+  const { edit } = useEditState()
+  console.log('edit', edit)
```
kldavis4, line 13: remove?

jamespohalloran (approved):

Issue #3400, opened 2022-11-15
Support shortcode syntax for the target SSGs (Hugo, 11ty)


newash:
## Description

This is partly a bug report but also a feature request.
I elaborate below.

### Steps to reproduce
1. you name your shortcode, like `WarningCallout`
2. you pick a starting character sequence, like `{%`
3. you also pick an ending sequence, like `%}`
4. then you define a single field that must be named `text`, that is a string

### Expected result
The docs says it is gonna be in the Markdown like `{% WarningCallout text="value of the field" %}`.

### Actual result
It is `{% value of the field %}`.

Beside the docs being incorrect, there are also some non-obvious issues with the implemented behaviour.
1. It's not possible to define multiple shortcodes with the same starting or ending sequence with the ending sequence being the bigger issue. It's possible to hack the setup with `{% WarningCallout` and `{% OtherCallout` as the starters (expecting `{% WarningCallout value of the field %}` and `{% OtherCallout value of the field %}`) but using the same ending will **CRASH** Tina.
2. Even the different starters are a hack.
3. I'm not sure about other SSG frameworks but the 11ty I use [supports 3 different template engines built-in](https://www.11ty.dev/docs/shortcodes/), and neither can understand the `{% WarningCallout text="value of the field" %}` format. And those are not some weird template engines but pretty standard ones, like Handlebars. These are the formats they understand:
   1. `{{{ WarningCallout "value of the field" }}}`
   2. `{% WarningCallout "value of the field" %}`
   3. or if you want to bother with multiple fields: `{% WarningCallout "value of the field", "other field value" %}`

I believe your goal is to support the SSGs, so it would be really nice if you would change the (target, documented) logic to support these formats. 🙏

#### Tinacms Package Versions
```text
├── @tinacms/cli@1.0.0
├── tinacms@1.0.0
```

newash:
Let me know if you need any additional info about the bug or the suggested syntax. Thank you!

hithismani:
This would be so helpful!

I would also add, that some frameworks like Hugo support shortcodes that are enclosed. 

I.e {{< shortcode attribute="anything" >}}SOME INNER TEXT THAT GOES ON 
FOR MULTIPLE LINES 
{{< /shortcode >}}


I see there's an `inline:true/false` key for the configuration in Tina, but it doesn't do anything from what I see.

newash:
Ah so, it's Hugo where the named parameter syntax is coming from. 👍
[...]

PR #3475, opened 2022-12-15
fix: Allow shortcodes to be specified by name

jamespohalloran:
Leaving this as a draft as it needs more user-testing & tests. 

The user can now model a shortcode like so:

```
    name: "rimg",
    label: "rimg",
    match: {
      start: "{{<",
      end: ">}}",
    },
    fields: [
      {
        name: "myprop",
        label: "my prop",
        type: "string",
      },
    ],
```

outputs `<<% rimg myprop="" %>>`

https://www.loom.com/share/70a391b8920e4951a62add4066a591e0

For shortcodes with with non-standard graphql node names, you can specify the match name:

```
match: {
  start: "<<%",
  name: "foo-bar",
  end: "%>>
}
```

outputs: `<<% foo-bar %>>`

Users no longer need to include the shortcode name in the match (E.g `{{< rimg`), so missing/extra spaces shouldn't impact the matching. The user can also define fields for each shortcode, and they are individually editable in the UI (instead of using a catchall `text` field).

For cases like `{{< signature "vf, il, al" >}}`, the user can optionally define a `text` field, which allows the editor to edit the shortcode attributes as a global string, instead of editing key/value pairs. 

You can also supply `children` as a field to text content in a shortcode.

fixes ENG-686

Seems to be a few TODO's to sort out.
- [x] Special characters like '-' in template name seem to break mdx parsing (Update: You can specifiy a "name" in match to account for special cases like a `-` in the template name)
- [x] We don't have a way to parse shortcodes without key'ed string, eg: `{{< signature "vf, il, al" >}}` (Update: This works now if you use a "text" key. Otherwise, we render fields for each key)
- [x] Test that "children" is working as expected
- [x]  Add tests
- [x] Update docs: https://github.com/tinacms/tina.io/pull/1460

Nice to haves:
- [ ] UPDATE: We'll handle this later: why do shortcodes/templates NEED to have at least one field? (Probably doesn't need to be apart of this PR, but I think it would make simple shortcodes / templates more usable)
- [x] Can/should we use a different key for inline raw-strings than `text`? I took a quick crack at modifying that logic but might need @jeffsee55's assistance with that.

fixes https://github.com/tinacms/tinacms/issues/3400

### DIFF
--- a/packages/@tinacms/mdx/src/parse/parseShortcode.test.ts
+++ b/packages/@tinacms/mdx/src/parse/parseShortcode.test.ts
@@ -16,12 +16,12 @@ limitations under the License.
 
 */
 import { it, expect, describe } from 'vitest'
-import { parseTemplateMatch } from './parseTemplateMatch'
+import { parseShortcode } from './parseShortcode'
 
-describe('parseTemplateMatch', () => {
+describe('parseShortcode', () => {
   describe('with keyed field', () => {
     it('parses attributes', () => {
-      const result = parseTemplateMatch('{{< signature foo="bar123">}}', {
+      const result = parseShortcode('{{< signature foo="bar123">}}', {
         name: 'signature',
         label: 'Signature',
         match: {
@@ -42,7 +42,7 @@ describe('parseTemplateMatch', () => {
 
   describe('with unkeyed attributes', () => {
     it('parses attributes', () => {
-      const result = parseTemplateMatch('{{< signature "bar123" >}}', {
+      const result = parseShortcode('{{< signature "bar123" >}}', {
         name: 'signature',
         label: 'Signature',
         match: {
@@ -63,7 +63,7 @@ describe('parseTemplateMatch', () => {
 
   describe('with children', () => {
     it('parses children field', () => {
-      const result = parseTemplateMatch(
+      const result = parseShortcode(
         '{{< signature >}}\n# FOO\n##Bar\n{{< /signature >}}',
         {
           name: 'signature',
--- a/packages/@tinacms/mdx/src/parse/parseShortcode.test.ts
+++ b/packages/@tinacms/mdx/src/parse/parseShortcode.test.ts
@@ -51,13 +51,13 @@ describe('parseShortcode', () => {
         },
         fields: [
           {
-            name: 'text',
-            label: 'text',
+            name: '_value',
+            label: 'Value',
             type: 'string',
           },
         ],
       })
-      expect(result).toEqual('<signature text="bar123">\n</signature>')
+      expect(result).toEqual('<signature _value="bar123">\n</signature>')
     })
   })
 
--- a/packages/@tinacms/mdx/src/parse/parseShortcode.test.ts
+++ b/packages/@tinacms/mdx/src/parse/parseShortcode.test.ts
@@ -76,7 +76,7 @@ describe('parseShortcode', () => {
             {
               name: 'children',
               label: 'children',
-              type: 'string',
+              type: 'rich-text',
             },
           ],
         }
--- a/.changeset/odd-pigs-unite.md
+++ b/.changeset/odd-pigs-unite.md
@@ -1,0 +1,6 @@
+---
+'@tinacms/mdx': minor
+'@tinacms/schema-tools': minor
+---
+
+fix: Shortcodes need to be specified by name to match with match-start / match-end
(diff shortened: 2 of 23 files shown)


newash:
Thanks @jamespohalloran for this PR!
Please-please reviewers, I badly need this feature to be merged (#3400)

jamespohalloran:
> Thanks @jamespohalloran for this PR! Please-please reviewers, I badly need this feature to be merged (#3400)

There's still a few tweaks / cleanup to do (there's some extra spaces added to the shortcode output), but this can be tested locally with package version `@tinacms/cli@shortcode-specifity`. (This pr will need to be merged to be working in production

packages/@tinacms/mdx/src/parse/parseShortcode.test.ts
```diff
@@ -0,0 +1,87 @@
...
+          label: 'Signature',
+          match: {
+            start: '{{<',
+            end: '>}}',
+          },
+          fields: [
+            {
+              name: 'children',
+              label: 'children',
+              type: 'string',
```
jeffsee55, line 79: I think we'll want to advise that this should be `rich-text` so `## FOO\n##Bar` are parsed as H2 elements.

jeffsee55 (approved):
This looks great, the only remaining question I'd have which might be nice to cover in the tests is for shortcodes with no fields at all (eg. `{{< ad-panel >}}`. I think it will probably just work, but would clicking on it take you to an empty form list?

packages/@tinacms/mdx/src/parse/parseShortcode.test.ts
jamespohalloran: > This looks great, the only remaining question I'd have which might be nice to cover in the tests is for shortcodes with no fields at all (eg. `{{< ad-panel >}}`. I think it will probably just work, but would clicking on it take you to an empty form list?

Oh right, hadn't considered that. I don't think it's too too weird to have it link to an empty form. Maybe if there's no fields it becomes not clickable at all (if that's not too much more effort)? 

Also, did you have any thoughts on this?:
> Can/should we use a different key for inline raw-strings than "text"? I took a quick crack at modifying that logic but might need @jeffsee55's assistance with that.

I'm just a bit scared that "text" might be too common of a prop name for existing shortcodes so it might conflict.

jeffsee55: > I'm just a bit scared that "text" might be too common of a prop name for existing shortcodes so it might conflict.

Maybe `_value`? Or something like that?

Also, this might need for of a deep refactor because the name of the field can't contain any dashes, so for example `ad-panel` as the _name_ will break everything. This worked previously because the text field was just open-ended

jamespohalloran: Yeah _value would work! Is that something you'd be able to help with? Not 100%, but I think `text` may be used by mdx logic outside this PR, so I wasn't sure if it would be a trivial change.

> Also, this might need for of a deep refactor because the name of the field can't contain any dashes, so for example ad-panel as the name will break everything. This worked previously because the text field was just open-ended

I have an optional `name` property under the `match` prop that you can use to supply names with special characters like dashes:

```
name: "adPanel",
match: {
  start: "<<%",
  name: "ad-panel",
  end: "%>>
}
// ...
```
outputs: `<<% ad-panel %>>`

jeffsee55: Ok cool, that works. I might take stab at those changes.

packages/@tinacms/mdx/src/parse/parseShortcode.test.ts
```diff
@@ -0,0 +1,87 @@
...
+        },
+        fields: [
+          {
+            name: 'text',
+            label: 'text',
+            type: 'string',
+          },
+        ],
+      })
+      expect(result).toEqual('<signature text="bar123">\n</signature>')
```
jeffsee55, line 60: Won't this break some user's shortcodes? I would expect that when it's unkeyed going _in_ it remains unkeyed going _out_

jamespohalloran: I believe this is more or less doing what it did before for unkeyed strings. parseShortcode translates `<% foobar "123" %>` to `<foobar text="123"` just so that it is usable by our rich-text field.

Then when it's output back to .md, stringifyShortcode translates it back to:
`<% foobar "123" %>`

This only happens when a "text" field is defined on a template. That special logic is kind of why I'd like to make that field name something less common (like _values)

PR #3482, opened 2022-12-19
Automate Creating PRs to bump the Tina version

logan-anderson:
This PR adds a Github action that automates the creation of PRs to the starters. I am not 100% sure that this works (I am not sure of a good way to test it) but we can test it on the next release.


jeffsee55 (approved):
@logan-anderson fyi merging this so we'll see how it goes

PR #3493, opened 2023-01-01
CMD/CTRL+S to save forms

Maxobat:
To make the editing experience more seamless, keybindings for saving the form were added. Now, Command+s or Control+s will trigger a submit on the form, just clicking the save button. Each action follows the states of the save button; if it's disabled/busy/etc., the key bindings will no-op.


packages/@tinacms/toolkit/src/packages/form-builder/FormBuilder.tsx
```diff
@@ -151,9 +173,23 @@ export const FormBuilder: FC<FormBuilderProps> = ({
         dirtySinceLastSubmit,
         hasValidationErrors,
       }) => {
+        const cannotSubmit =
```
jamespohalloran, line 176: Small suggestion, positive boolean names's are a bit easier to read:
```
const canSubmit = 
!pristine && // ...
```

packages/@tinacms/toolkit/src/packages/form-builder/FormBuilder.tsx
```diff
@@ -254,8 +285,18 @@ export const FullscreenFormBuilder: FC<FormBuilderProps> = ({
         onSubmit={tinaForm.onSubmit}
       >
         {({ handleSubmit, pristine, invalid, submitting }) => {
+          const cannotSubmit = pristine || submitting || invalid
```
jamespohalloran, line 288: This conditional is a bit different than the above cannotSubmit. Is that intentional?

jamespohalloran (changes requested):
Thanks for this! Will be a handy addition. Just left a few comments.

packages/@tinacms/toolkit/src/packages/form-builder/FormBuilder.tsx
Maxobat: Yes, follows the disabled state of each save button from the two forms. FullscreenFormBuilder seems to have a slightly simpler version:
Screenshot 2023-01-03 at 9 19 36 AM
Screenshot 2023-01-03 at 9 19 21 AM

packages/@tinacms/toolkit/src/packages/form-builder/FormBuilder.tsx
Maxobat: Thanks for the suggestion. Resolved!

https://github.com/tinacms/tinacms/pull/3493/commits/41c88fe486dfe907c197ebc372ed41bf9132dab7

Maxobat:
@jamespohalloran This is my first PR on Tina; for some reason the license/cla check is failing, even though I signed it. Thoughts as to how I resolve that? Thanks!

jamespohalloran (dismissed):

jamespohalloran:
Pr approved! Thanks again for this. For the license/cla check, it looks like the commits may have a different email than what you've signed the agreement with.
https://stackoverflow.com/questions/750172/how-do-i-change-the-author-and-committer-name-email-for-multiple-commits

Maxobat:
@jamespohalloran Sorry, tried to fix the email thing, but it did some unintentional changes. I'll fix and get back to you. But hey, at least the license thing is fixed lol.

PR #3497, opened 2023-01-05
Feat/support toml

logan-anderson:
This PR adds two things. 

## 1. Frontmatter format updates

Can now specify { frontmatterFormat: 'toml' | 'yaml' | 'json' } and {frontmatterDelimter: string | [string, string] }.

This allows the user to have more conrtol of how their frontmatter is parsed.  

## 2. Adds `yaml` and `toml` format support for a collection

In a collection `format` can now be `yaml` or `toml`.

## ~~Tina cloud updates~~

~~We will need to have access to the config/collection when we call parseFile here: https://github.com/tinacms/tina-cloud/blob/main/js/content-api/src/database/store/github/index.ts#L150~~

~~We will also need to have access config/collection when we call stringily file here: https://github.com/tinacms/tina-cloud/blob/d5ca6ab39f06911693477fba934b31410cee0db5/js/content-api/src/database/store/github/index.ts#L209~~


kldavis4:
@logan-anderson we should verify, but I think most projects (maybe all active at this point) are using the dynamodb store, so I don't _think_ that the github store should be an issue for this

jeffsee55:
Awesome! I think `markdownFrontmatterFormat` as a name is fine 🤷

jamespohalloran:
Just recapping a few notes from chatting through this in happy hour:
- maybe changing the name from `markdownFrontmatterFormat` to `frontmatterFormat`
- sounds like we may need an optional config for `frontmatterDelimiter` (Hugo uses `+++` for TOML)

packages/@tinacms/schema-tools/src/validate/schema.ts
```diff
@@ -16,7 +16,7 @@ import { name } from './properties'
 import { findDuplicates } from '../util'
 import { TinaFieldZod } from './fields'
 import { tinaConfigZod } from './tinaCloudSchemaConfig'
-const FORMATS = ['json', 'md', 'markdown', 'mdx'] as const
+const FORMATS = ['json', 'md', 'markdown', 'mdx', 'toml', 'yaml'] as const
```
kldavis4, line 19: @logan-anderson does it make sense to have all of these formats together? The existing ones apply to the format property while the latter apply to the frontmatterFormat property

logan-anderson: @kldavis4 I was able to add support for those formats as well. I will update the changeset and the description.

But doing { format: 'toml' } will now allow you to have toml files.

kldavis4: Ok, thanks for the clarification 👍🏻

kldavis4 (dismissed):

logan-anderson:
> Just recapping a few notes from chatting through this in happy hour:
> 
> * maybe changing the name from `markdownFrontmatterFormat` to `frontmatterFormat`
> * sounds like we may need an optional config for `frontmatterDelimiter` (Hugo uses `+++` for TOML)

This have been added.

jamespohalloran (approved):

PR #3499, opened 2023-01-06
List Field Meta Wrapper, Form Improvements

spbyrne:
### Video Demo -- Shows both list errors and min/max:
https://user-images.githubusercontent.com/5075484/211587949-a19681e1-f8c9-4f91-820a-f7f55eeb0e6d.mp4

## Fixes Errors

The List, Group List and Block fields weren't displaying errors. I've created a list meta wrapper component that provides the label, description, and error UI. It accepts an 'actions' prop that allows each field to control that logic.

This also updates the error styles to give it a bit more whitespace and a quick slide-in animation.

Fixes #3495

## Adds Min/Max Support

The field UI object now accepts a min and max when the field is `list: true`. If the list/group list/block field is at the max or min value, the add or remove buttons are disabled. If the min and max values are the same, the field is considered 'fixed length' and the add/remove buttons are removed entirely, unless the data is invalid (you can still add items if there's not enough to satisfy the validation).

Screenshot 2023-01-06 at 2 20 47 PM

Here's a field with min and max set to the same number: (no add/remove buttons)
Screenshot 2023-01-10 at 10 39 19 AM

### Fix Form Whitespace

Adds padding at the bottom of forms and removes unnecessary margin from the MDX component (this is provided by the field wrapper). 

Screenshot 2023-01-10 at 10 59 17 AM

PR #3506, opened 2023-01-09
default min should be 0

logan-anderson:
Fixes https://github.com/tinacms/tinacms/pull/3499#issuecomment-1376316884


spbyrne (approved):
Perfect!

PR #3507, opened 2023-01-10
feat: upgrade to graphql 16

jayalfredprufrock:
This was the bare minimum change to upgrade CLI package to graphql v16 and was enough to unblock myself. I imagine we'll want to get the rest of the monorepo updated as well so I'll follow up with some other commits. I can make sure tests pass, but I won't be able to verify behavior beyond that.


jayalfredprufrock:
Updated the app package as well. Project builds and tests pass. Not sure about those generated schema files. I can remove them from the commit if they weren't supposed to be there, but if thats the case, they should probably be gitignored.

logan-anderson (changes requested):
@jayalfredprufrock Looks like the types build is failing (That will have to be fixed before the changes get merged). 

I also think we should update the entire monrepo to v16 and not only the CLI since we could run into the same issue of having more then one version of graphQL in a project.

It looks like the following packages will have to be updated as well.
- `@tinacms/graphql`
- `@tinacms/datalayer`
- `tinacms`
-  examples/kitchen-sink

jayalfredprufrock:
Well darn, I guess I didn’t run the correct test script. I’ll take another stab tonight.

logan-anderson:
@jayalfredprufrock Any update on this?

jayalfredprufrock:
@logan-anderson I began going through the fixes, but they quickly became overwhelming. I did make some good progress though so I went ahead and pushed what I had in case somebody else wants to pick it up. My work with TinaCMS is for a side hustle so I can't generally dedicate working hours to it but I'll continue making progress when I can.

On a related note, I noticed that @tinacms/graphql depends on [graphql-type-json](https://github.com/taion/graphql-type-json) which may not be compatible with graphql 16 (https://github.com/taion/graphql-type-json/issues/343). I really have no idea if that's going to be a problem or not. I'm not a graphql expert and navigating Tina's massive monorepo is challenging and clearly graphql is a big part of how it works. Is there somebody else on the team that can perhaps provide some guidance or assistance?

jeffsee55:
I believe this is no longer a blocker for people using other GraphQL versions. @jayalfredprufrock please advise but closing this for now. We do hope to move to GraphQL 16 in the future but as long as it's not blocking others it's not currently a priority.

PR #3510, opened 2023-01-10
Feat/forestry migration updates

logan-anderson:
## This PR updates the forestry migration tool in a couple of ways.
- It now migrates templates correctly
- It can migrate when no templates are provided
- It add a root level `_tempalte` key to documents 
-  updates Error messages to have red text and warning message to have yellow
- Adds a spinner

## Steps to test. 

Run
```bash
yarn add --dev @tinacms/cli@forestry-migration-updates
```

and then run

```
yarn tinacms init
```

I am going to work on some docs that we can link to for certain errors that occur.


packages/@tinacms/cli/src/cmds/forestry-migrate/util/errorSingleton.ts
```diff
@@ -0,0 +1,80 @@
...
+
+    return ErrorSingleton.instance
+  }
+
+  private allErrorNames: NameError[]
+
+  public addErrorName(error: NameError) {
+    this.allErrorNames.push(error)
+  }
+  public printErrorNames() {
```
jamespohalloran, line 56: I don't know if I get what's happening in this function. It looks like we'd log out several different error types when called.

packages/@tinacms/cli/src/cmds/forestry-migrate/util/index.ts
```diff
@@ -15,9 +15,34 @@ import fs from 'fs-extra'
...
 import z from 'zod'
-import type { TinaFieldInner } from '@tinacms/schema-tools'
+import type { TinaFieldInner, TinaTemplate } from '@tinacms/schema-tools'
 import { logger } from '../../../logger'
+import { warnText } from '../../../utils/theme'
+import { ErrorSingleton } from './errorSingleton'
 
+const errorSingletonInstance = ErrorSingleton.getInstance()
+
+export const stringifyName = (name: string, template: string) => {
```
jamespohalloran, line 25: Are we changing the name of their fields in their content pages along with this? I have a bit of concern it might be some overhead for the user to update their site to make sure it doesn't break the build (as opposed to, if we were to just support characters like dashes in our field names)

packages/@tinacms/cli/src/cmds/forestry-migrate/util/errorSingleton.ts
logan-anderson: This is only for "name errors". 

For example if you had a template name "my-template" or a field name "my-field".

jamespohalloran, line 56:
```suggestion
  public printNameErrors() {
```
(You'll have to update the function calls as well)

jamespohalloran (approved):
Left one small comment on the naming of a function.

Overall this looks great. One thing though, I think we might be leaving the user in a bit of a confusing state by renaming the fields. Out of the box, the fields will look like they are working, however when editing they'll just update a new field that isn't used by their site. 
We can address this in a next step, but I think supporting dashes in field names may be worth the investment.

jamespohalloran (approved):

PR #3516, opened 2023-01-13
Demo

jeffsee55:
WIP


packages/@tinacms/cli/src/buildTina/attachDatabase.ts
```diff
@@ -45,7 +45,8 @@ export const attachDatabase = async (
     filename: 'database',
     allowedTypes: ['js', 'jsx', 'tsx', 'ts'],
   })
-  if (inputFile && !_options.useLocalDatabase) {
+  // if (inputFile && !_options.useLocalDatabase) {
```
jeffsee55, line 48: Not sure what other reasons there are for this check, just disabled i for now

packages/@tinacms/graphql/src/mdx/index.ts
```diff
@@ -16,7 +16,14 @@ limitations under the License.
...
 */
 
-import { parseMDX, stringifyMDX } from '@tinacms/mdx'
+// import { parseMDX, stringifyMDX } from '@tinacms/mdx'
 
-export { parseMDX }
-export { stringifyMDX }
+// export { parseMDX }
+// export { stringifyMDX }
+// MDX is bundled in a way that Nextjs doesn't like, fake it out for now
```
jeffsee55, line 23: This is a temporary workaround since the MDX bundle seems problematic for Next, so any saves will write this dummy content to all rich-text fields, and reads of rich-text show up empty.

packages/@tinacms/cli/src/buildTina/attachDatabase.ts
kldavis4: The useLocalDatabase option is to ensure that when running `yarn dev`, the user defined database is not used. When you run build or start and database.ts is present, then that database is used.

examples/basic-iframe/.tina/database.js
```diff
@@ -0,0 +1,17 @@
...
+config()
+
+const level = new MongodbLevel({
+  mongoUri: process.env.MONGO_URI,
+  dbName: 'tinacms',
+  collectionName: 'tinacms',
+})
+
+export default createDatabase({
+  bridge: new FilesystemBridge(process.cwd()),
```
jeffsee55, line 15: Having this here means the `api/gql` route has a larger bundle, including the `fs` and `path` modules, which it doesn't use. We should probably bake the `bridge` in by default, to the CLI so this isn't needed

PR #3517, opened 2023-01-13
WIP Rename document

logan-anderson:
This adds the ability to rename and delete a new document from when you are editing a new page.

#### Some things to discuss

- What do do once the user delete a document?
   -  Maybe go back to the collection page
- Is this the correct approach for relativePath
- Do we want to add this to the collection list page ?

#### UI updates
At some point this PR will need some UI love. This is the current (and terrible) UI

Screen Shot 2023-01-13 at 4 21 16 PM


jeffsee55 (commented):
Since the filename is often used to derive the visual editing path, I think this capability should be limited to only being avaible in the CMS (ie. non-visual mode)

logan-anderson:
Not sure if we want this to be a minor or a patch to the GraphQL packages. It adds a new key to the params.

packages/@tinacms/graphql/src/resolver/index.ts
```diff
@@ -469,16 +471,39 @@ export class Resolver {
           isAddPendingDocument,
         })
       }
-      if (isDeletion) {
+      // if we are deleting a document or updating its name we should check if it exists
+      if (isDeletion || isUpdateName) {
         if (!alreadyExists) {
           throw new Error(
             `Unable to delete document, ${realPath} does not exist`
```
jeffsee55, line 478: This error message should reflect the action that is failing (ie. `delete` or `update`)

packages/@tinacms/graphql/src/builder/index.ts
```diff
@@ -1176,17 +1175,24 @@ export class Builder {
...
+    const fields = await sequential(
+      this.tinaSchema.getCollectionsByName(field.collections),
+      async (collection) => {
+        return astBuilder.InputValueDefinition({
+          name: collection.name,
+          type: NAMER.dataMutationTypeName([collection.name]),
+        })
+      }
+    )
+    fields.push(
```
jeffsee55, line 1187: `_buildReferenceMutation` is used for both `createDocument` and `updateDocument` and looking at the resolve code below this would error if you provided it during the `createDocument`. Right now the mutations for create and update share the same GraphQL type but we should split these up into separate types and only have `relativePath` for the update mutation.

packages/@tinacms/graphql/src/resolver/index.ts
logan-anderson: Makes sense!

packages/@tinacms/graphql/src/builder/index.ts
logan-anderson: Good catch. This has been fixed. It now has its own param type.

spbyrne:
Just made a small tweak, otherwise looks good.

jeffsee55 (approved):

PR #3526, opened 2023-01-18
fix: use the db status endpoint for checking credentials

kldavis4:
Instead of making a graphql query to validate token/clientId/branch, switch to the checking the db status endpoint. This should fix an issue where, when a new branch is created and indexing is still in progress, the query used in checkClientInformation fails because the index hasn't been created yet. This was happening before the waitForDb check and  resulting in Vercel build failures.


jamespohalloran (dismissed):
Awesome!

jamespohalloran:
Testing this out in a PR deploy and found a few things:
The spinner logic we use doesn't really work well in CI:
Screen Shot 2023-01-20 at 9 23 00 AM

jeffsee55:
> The spinner logic we use doesn't really work well in CI:

We can probably check for `process.env.CI` before showing those types of animations.

jamespohalloran:
> > The spinner logic we use doesn't really work well in CI:
> 
> We can probably check for `process.env.CI` before showing those types of animations.

Just made that fix!

jamespohalloran (approved):

PR #3528, opened 2023-01-19
revert styling that broke tests

logan-anderson:
There was some styling that added some padding to the bottom that prevented the e2e tests from finding the element. 

I tried to update the tests to be more robust but was having difficulty getting it all to work properly. We can undo the small style update so the tests pass and look into updating the e2e tests in the future.


jamespohalloran (approved):

spbyrne:
Wait, why is the padding breaking the test? Is it using a css class selector that can be updated?

spbyrne (approved):
👍  Looks good.

Issue #3536, opened 2023-01-20
Deleting all items in an array doesn't work


jeffsee55:
When we introduced behavior to `patch` items instead of `put` we made it so that you can't fully delete an array because the old values are shallow merged

https://www.loom.com/share/0230e01fea0b40e4bc0e4c60b14adf67

PR #3548, opened 2023-01-23
Allow user to save an empty array

logan-anderson:
Fixes an issue that @jeffsee55 found where a user could not save a top level array that was empty.

Potential fix for #3536

### DIFF
--- a/.changeset/smooth-jokes-work.md
+++ b/.changeset/smooth-jokes-work.md
@@ -1,0 +1,5 @@
+---
+'@tinacms/graphql': patch
+---
+
+Fix usse where user could not save top level empty array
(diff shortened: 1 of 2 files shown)


jeffsee55 (approved):
Ok so if you have no values in an array, this would be saved as:
```md
---
title: Hello
blocks: []
---
```
And previously you would have gotten:
```md
title: Hello
```

Not sure on the implications of that, I don't believe it will affect the response of the GraphQL API, but for sites using the markdown files directly it'll make `blocks` truthy, which may lead to some unexpected issues. Still, seems like it's something we can fix if it becomes an issue.

PR #3549, opened 2023-01-23
Kldavis4/poc/abstract level store many level wrapped

logan-anderson:
Wrapped up the many-level client into something that can easily be consumed.


examples/basic-iframe/.tina/database.js
```diff
@@ -15,67 +15,66 @@ const octokit = new Octokit({
...
-        data: { sha: existingSha },
-      } = await octokit.repos.getContent({
-        owner,
-        repo,
-        path: key,
-        branch,
-      })
-      sha = existingSha
-    } catch (e) {}
+    const filePath = path.join(process.cwd(), key)
```
kldavis4, line 24: did you mean to change these? if you are specifying the filesystem bridge, its already going to be updating the local fs, so you could just leave onPut/onDelete out here

packages/@tinacms/cli/src/tinaLocalLevel/index.ts
```diff
@@ -0,0 +1,13 @@
...
+import { pipeline } from 'readable-stream'
+import { connect } from 'net'
+
+export class TinaLocalLevel extends ManyLevelGuest {
+  constructor() {
+    super()
+    const socket = connect(9000)
+    pipeline(socket, this.createRpcStream(), socket, () => {
+      // Disconnected
+    })
```
kldavis4, line 11: this probably should be in a `connect()` function instead

examples/basic-iframe/.tina/database.js
logan-anderson: When I was running the code I was getting errors from Github (404 errors) when saving. When I swapped out onput and onDelete it worked.

kldavis4: yeah, owner/repo/token/branch need to be configured for onPut/onDelete to work

packages/@tinacms/cli/src/tinaLocalLevel/index.ts
logan-anderson: Where would connect get called?

logan-anderson: Would we call it when we are indexing? and then it could be called when the user wants to fetch data?

kldavis4: Same place we call the existing connect function, so you'd just instantiate the object and call connect on it:
```
const localLevelStore = new TinaLevelClient()
localLevelStore.connect()
```

kldavis4: I'm not sure - the database would need to accept any level implementation - so I don't think we could put it in `connect()`. Maybe overriding open() and putting it in there makes sense (assuming open() gets called automatically somewhere before its actually needed)

logan-anderson: Hmm when I put it in open I get this error.
```
Error: Only one rpc stream can be active
```

logan-anderson: Added a new function. I called it `openConnection` as to not have any naming conflicts

examples/basic-iframe/.tina/database.js
```diff
@@ -0,0 +1,84 @@
...
+  auth: token,
+})
+
+const localLevelStore = new TinaLevelClient()
+const mongodbLevelStore = new MongodbLevel({
+  collectionName: 'tinacms',
+  dbName: 'tinacms',
+  mongoUri: process.env.MONGODB_URI,
+})
+localLevelStore.openConnection()
```
kldavis4, line 24: Only call if isLocal == true?

packages/@tinacms/cli/src/buildTina/attachDatabase.ts
```diff
@@ -105,8 +126,12 @@ export const attachDatabase = async (
         )
       : fsBridge
 
-    const level = new MemoryLevel<string, Record<string, any>>({
-      valueEncoding: 'json',
+    const level = new ManyLevelGuest()
```
kldavis4, line 129: I think we can use the TinaLevelClient here now

packages/@tinacms/cli/src/cmds/start-server/index.ts
```diff
@@ -42,6 +42,7 @@ interface Options {
   command?: string
   watchFolders?: string[]
   noWatch?: boolean
+  noServer?: boolean
```
kldavis4, line 45: we should confirm with @jeffsee55, but I believe the noServer option was related to the file locking issues when using ClassicLevel, which should no longer be an issue with many-level, so we may want to drop this change.

packages/@tinacms/cli/src/index.ts
```diff
@@ -18,6 +18,7 @@ import { Command } from './command'
 import { baseCmds as baseCommands } from './cmds/baseCmds'
 import { logText } from './utils/theme'
 export { defineSchema } from './cmds/compile'
+export { TinaLevelClient } from './tinaLocalLevel'
```
kldavis4, line 21: change the file name to `tinaLevel` or `tinaLevelClient`?

examples/kitchen-sink/pages/page/[filename].tsx
```diff
@@ -1,7 +1,7 @@
 import { InferGetStaticPropsType } from 'next'
 import { Json } from '../../components/json'
 import { useTina } from 'tinacms/dist/react'
-import client from '../../.tina/__generated__/client'
+import { dbConnection } from '../../lib/db'
```
Comment (kldavis4) [inline examples/kitchen-sink/pages/page/[filename].tsx:4]: We don't have parallel changes like this for basic-iframe. I attempted to, but ran into issues with the shape being returned. 

Would it make sense to pull the changes in basic-iframe into kitchen-sink (and revert basic-iframe)? The only reason I put them in basic-iframe and not kitchen-sink was because I wanted to provide an example while avoiding stomping on Jeff's classiclevel changes, which we are now discarding.

examples/basic-iframe/.tina/database.js
logan-anderson: Yeah that makes sense!

packages/@tinacms/cli/src/index.ts
logan-anderson: Would it make more sense of this being in `@tinacms/graphql` ?

`@tinacms/cli` is typically a dev dependency which makes importing it sorta awkward (even if its only used locally)

Also, I was getting some Error when trying to import `@tinacms/cli` in pages/api/gql. (only in the mono repo and not in npm version).

Thought @kldavis4 and @jeffsee55 ?

logan-anderson: I updated it

packages/@tinacms/cli/src/buildTina/attachDatabase.ts
logan-anderson: Updated!

packages/@tinacms/cli/src/cmds/start-server/index.ts
logan-anderson: Remvoed

examples/basic-iframe/package.json
```diff
@@ -14,7 +14,6 @@
     "dotenv": "^16.0.3",
     "js-base64": "^3.7.4",
     "mongodb": "^4.13.0",
```
kldavis4, line 16: remove this too?

examples/kitchen-sink/.env.sample
```diff
@@ -0,0 +1,2 @@
+MONGO_URI="mongodb+srv://<username>:<password>@cluster0.wffq71a.mongodb.net/?retryWrites=true&w=majority"
```
kldavis4, line 1: does this need to go in the new starter?

examples/kitchen-sink/.gitignore
```diff
@@ -34,3 +34,5 @@ yarn-error.log*
 # typescript
 *.tsbuildinfo
 next-env.d.ts
+
+.env
```
kldavis4, line 38: move to the new starter?

examples/kitchen-sink/next.config.js
```diff
@@ -2,7 +2,7 @@
 const nextConfig = {
   reactStrictMode: true,
   swcMinify: true,
-  pageExtensions: ['tsx'],
```
kldavis4, line 5: remove?

examples/kitchen-sink/.env.sample
logan-anderson: removed

examples/kitchen-sink/.gitignore
logan-anderson: added to new starter

kldavis4 (approved):

PR #3557, opened 2023-01-26
Set input value even with errors in rawmode

jamespohalloran:
Currently, when you make a change in raw-mode with an error, when you switch to the rich-text editor, your changes are quietly wiped. Saving also quietly omits any changes after you introduced a rich-text editor. 

This changes that behaviour a bit:
- Changes to raw-mode / rich-text-editor are always persisted in the form state (even with an error)
- The form now gets a validation error when we can't parse rich-text (to block saving)
- I added a button to jump to the raw mode if there's an error (@spbyrne may want to take a look at the styling)

https://www.loom.com/share/040cbb84408a442b93f60c9b0731cca6

fixes ENG-778
fixes ENG-784
fixes ENG-776
fixes ENG-781


jeffsee55 (approved):
Awesome!

PR #3570, opened 2023-02-01
Fix/children undefined

logan-anderson:
cc @jamespohalloran 

I noticed a bug that would break content creation. If you create a document with multiple rich-text components you would not be able to type or create a document.


jamespohalloran (approved):
Good catch

PR #3574, opened 2023-02-02
Kldavis4/poc/abstract level store move bridge

logan-anderson:
Move @tinacms/datalayer into @tinacms/graphql.  This means that`@tinacms/datalayer` is essentially a user facing facade. I made it so thats its backwards compatible so that we _shouldn't_ need to do updates on Tina Cloud if we don't want to.


packages/@tinacms/graphql/src/database/index.ts
```diff
@@ -43,10 +43,10 @@ type IndexStatusEvent = {
   error?: Error
 }
 type IndexStatusCallback = (event: IndexStatusEvent) => Promise<void>
-type OnPutCallback = (key: string, value: any) => Promise<void>
-type OnDeleteCallback = (key: string) => Promise<void>
+export type OnPutCallback = (key: string, value: any) => Promise<void>
+export type OnDeleteCallback = (key: string) => Promise<void>
 
-type CreateDatabase = {
+export type CreateDatabase = {
```
kldavis4, line 49: why are we exporting these? I don't think they need to be

logan-anderson: Just in case someone wants to do.

```ts
const githubOnPut : OnPutCallback = () =>{

//...
}
```

kldavis4 (approved):

PR #3584, opened 2023-02-06
no long express all env vars to the frontend

### LINKED ISSUE #3579: Exposed ENV vars

While converting our Jekyll site from Forestry to TinaCMS I noticed that compiling the admin exposes all ENV vars. Supplied is the link to current production index.js for tina.io, search for `TINA_`. If for whatever reason you require some private credentials for your build step then its time to roll your credentials.

I believe this is the offending line of code:
https://github.com/tinacms/tinacms/blob/main/packages/%40tinacms/app/src/index.ts#L146

It should only be outputting ENV that TinaCMS requires and it should be noted in the docs that you only want to use public non destructive / read-only ENV vars as they will be publicly exposed.

### Reproduction

https://tina.io/admin/assets/

### Steps to reproduce

- Add an env var to your build steps
- Build the admin
- Inspect the generated index.js for the exposed env var

### System Info

```shell
N/A
```

### Validations

logan-anderson:
Fixes #3579 by exposing `NODE_ENV`, `NEXT_PUBLIC_*`, `TINA_PUBLIC_*`, and `HEAD` (for integration with Netifly) environment vars in the admin.

### DIFF
--- a/.changeset/nasty-tigers-own.md
+++ b/.changeset/nasty-tigers-own.md
@@ -1,0 +1,5 @@
+---
+'@tinacms/app': patch
+---
+
+Fix issue where all env vars where send to the admin
--- a/packages/@tinacms/app/src/index.ts
+++ b/packages/@tinacms/app/src/index.ts
@@ -119,6 +119,19 @@ export const viteBuild = async ({
     TINA_IMPORT: configPrebuildPath,
   }
 
+  // TODO: make this configurable
+  const publicEnv: Record<string, string> = {}
+  Object.keys(process.env).forEach((key) => {
+    if (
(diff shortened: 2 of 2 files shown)


jeffsee55 (approved):

jsntv200:
Will this need some documentation for the whitelist? Looks like `TINA__CLIENT_ID` and `TINA__TOKEN` are now `TINA_PUBLIC_CLIENT_ID` and `TINA_PUBLIC_TOKEN` which I assume would be a breaking change for many.

Also unclear what your policies are around security leaks, but as a user I would want to know that my secure build credentials may have leaked to publicly accessible files so I could deal with it and roll credentials if required.

PR #3590, opened 2023-02-08
feat: ENG-806: configurable dos media store base directory

kldavis4:
Adds a new optional configuration property to the DO media handler to specify a `mediaRoot` to store media.

Demo

[Alt text](https://www.youtube.com/watch?v=0kpnurlbk2w)

### DIFF
--- a/packages/next-tinacms-dos/src/handlers.ts
+++ b/packages/next-tinacms-dos/src/handlers.ts
@@ -40,7 +40,10 @@ export const mediaHandlerConfig = {
 export const createMediaHandler = (config: DOSConfig, options?: DOSOptions) => {
   const client = new S3Client(config.config)
   const bucket = config.bucket
-  const mediaRoot = config.mediaRoot || ''
+  let mediaRoot = config.mediaRoot || ''
+  if (mediaRoot && !mediaRoot.endsWith('/')) {
+    mediaRoot = mediaRoot + '/'
+  }
   let cdnUrl =
     options?.cdnUrl ||
     config.config.endpoint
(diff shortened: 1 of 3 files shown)


packages/next-tinacms-dos/src/handlers.ts
```diff
@@ -115,19 +123,46 @@ async function uploadMedia(
...
     res.status(500).send(findErrorMessage(e))
   }
 }
 
+function stripMediaRoot(mediaRoot: string, key: string) {
+  if (!mediaRoot) {
+    return key
+  }
+  const mediaRootParts = mediaRoot.split('/').filter((part) => part)
+  if (!mediaRootParts || !mediaRootParts[0]) {
```
jamespohalloran, line 147: Haven't tested this yet, but would this return key if media root starts with `/`? ( mediaRootParts[0] would be '' )

packages/next-tinacms-dos/src/handlers.ts
```diff
@@ -115,19 +123,46 @@ async function uploadMedia(
...
+    return key
+  }
+  const keyParts = key.split('/').filter((part) => part)
+  // remove each part of the key that matches the mediaRoot parts
+  for (let i = 0; i < mediaRootParts.length; i++) {
+    if (keyParts[0] === mediaRootParts[i]) {
```
jamespohalloran, line 153: Wondering is this would cause an issue if you had two nested directories with the same name:

```
mediaRoot: 'blah/nested/blah'
```

jamespohalloran (changes requested):
Still plan on testing this out, but a few quick questions

packages/next-tinacms-dos/src/handlers.ts
kldavis4: good catch - pushed a fix

packages/next-tinacms-dos/src/handlers.ts
kldavis4: I don't think so - just tested it and it seems fine. Did find another issue while testing it, though, and fixed that

jamespohalloran (approved):

PR #3591, opened 2023-02-08
refetch collections when the filter is reset

logan-anderson:
Fixes an issue that @jamespohalloran found where if you selected and used a filter but then selected "None" the results would not update.


spbyrne:
Added an empty state when there's nothing to display.

Screenshot 2023-02-08 at 2 01 43 PM

spbyrne (approved):
Your fix worked 👍

jamespohalloran (approved):
Nice!

jeffsee55 (approved):

PR #3600, opened 2023-02-09
fix: handle LEVEL_NOT_FOUND when fetching from level

kldavis4:
Fix `Could not get value` error when querying non-existent content

### DIFF
--- a/packages/@tinacms/graphql/src/database/index.ts
+++ b/packages/@tinacms/graphql/src/database/index.ts
@@ -203,12 +203,19 @@ export class Database {
     } else {
       const tinaSchema = await this.getSchema(this.level)
       const extension = path.extname(filepath)
-      const contentObject = await this.level
-        .sublevel<string, Record<string, any>>(
-          CONTENT_ROOT_PREFIX,
-          SUBLEVEL_OPTIONS
-        )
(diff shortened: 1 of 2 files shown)


logan-anderson (approved):

packages/@tinacms/graphql/src/database/index.ts
```diff
@@ -203,12 +203,19 @@ export class Database {
...
+      let contentObject
+      try {
+        contentObject = await this.level
+          .sublevel<string, Record<string, any>>(
+            CONTENT_ROOT_PREFIX,
+            SUBLEVEL_OPTIONS
+          )
+          .get(normalizePath(filepath))
+      } catch (e: any) {
+        if (e.code !== 'LEVEL_NOT_FOUND') {
```
jamespohalloran, line 215: Seems we have this try/catch a few times. Should we be swallowing this error at a different level to avoid needing to always wrap it in this try catch?

jamespohalloran (approved):

packages/@tinacms/graphql/src/database/index.ts
kldavis4: I'll make a follow-up PR to wrap the level implementation and centralize this

PR #3616, opened 2023-02-15
Fix Create Branch

spbyrne:
https://user-images.githubusercontent.com/5075484/219146400-9e5f9b2f-f194-4c59-ad20-e83be5536464.mp4

### DIFF
--- a/packages/@tinacms/toolkit/src/plugins/branch-switcher/BranchSwitcher.tsx
+++ b/packages/@tinacms/toolkit/src/plugins/branch-switcher/BranchSwitcher.tsx
@@ -79,7 +79,7 @@ export const BranchSwitcher = ({
   const cms = useCMS()
   const isLocalMode = cms.api?.tina?.isLocalMode
   const [listState, setListState] = React.useState<ListState>('loading')
-  const [branchList, setBranchList] = React.useState([])
+  const [branchList, setBranchList] = React.useState([] as Branch[])
   const { currentBranch } = useBranchData()
   const initialBranch = React.useMemo(() => currentBranch, [])
   // when modal closes, refresh page is currentBranch has changed
@@ -117,6 +117,33 @@ export const BranchSwitcher = ({
   React.useEffect(() => {
     refreshBranchList()
   }, [])
+
+  // Keep branch list up to date
+  React.useEffect(() => {
+    if (listState === 'ready') {
+      // update all branches that have indexing status of 'inprogress' or 'unknown'
+      branchList
+        .filter(
+          (x) =>
+            x?.indexStatus?.status === 'inprogress' ||
+            x?.indexStatus?.status === 'unknown'
+        )
+        .map(async (x) => {
+          const indexStatus = await cms.api.tina.waitForIndexStatus({
+            ref: x.name,
+          })
+          setBranchList((prev) => {
+            const newList = Array.from(prev)
+            const index = newList.findIndex((y) => y.name === x.name)
+            newList[index].indexStatus = indexStatus
+            return newList
+          })
+          console.log({ indexStatus })
+        })
+
+      // return
+    }
+  }, [listState, branchList.length])
 
   return (
     <div className="w-full flex justify-center p-5">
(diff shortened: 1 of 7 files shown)


packages/tinacms/src/internalClient/index.ts
```diff
@@ -520,12 +645,67 @@ mutation addPendingDocumentMutation(
...
+            return Promise.resolve({
+              done: false,
+            })
+          }
+        } catch (err) {
+          return Promise.reject(err)
+        }
+      },
+      // interval is 1000ms
+      1000, // interval
```
kldavis4, line 670: I would increase this to 5 or 10 seconds

packages/tinacms/src/internalClient/index.ts
```diff
@@ -520,12 +645,67 @@ mutation addPendingDocumentMutation(
...
+    )
+    return result
+  }
+
+  async getIndexStatus({ ref }: { ref: string }) {
+    const url = `${this.contentApiBase}/db/${this.clientId}/status/${ref}`
+    const res = await this.fetchWithToken(url)
+    const result = await res.json()
+    const parsedResult = IndexStatusResponse.parse(result)
+    console.log('getIndexStatus', parsedResult, result)
```
kldavis4, line 682: delete

packages/tinacms/src/internalClient/index.ts
```diff
@@ -520,12 +645,67 @@ mutation addPendingDocumentMutation(
...
-    })
-    return res.json()
+    try {
+      const url = `${this.contentApiBase}/github/${this.clientId}/list_branches`
+      const res = await this.fetchWithToken(url, {
+        method: 'GET',
+      })
+      const branches = await res.json()
+      const parsedBranches = ListBranchResponse.parse(branches)
+      console.log('listBranches', parsedBranches)
```
kldavis4, line 694: delete

packages/@tinacms/toolkit/src/plugins/branch-switcher/BranchSwitcher.tsx
```diff
@@ -225,20 +264,41 @@ const BranchSelector = ({
...
+                    onChange(branch.name)
+                  }
+                }}
               >
                 {isCurrentBranch && (
                   <BiGitBranch className="w-5 h-auto text-blue-500/70" />
                 )}
                 {branch.name}
+                {(indexingStatus === 'inprogress' ||
+                  indexingStatus === 'unknown') && (
```
kldavis4, line 290: I would treat unknown as a separate state - its not necessarily an error state but the user probably should know if it is in that state for longer than a few seconds, because that means something is wrong with that branch

packages/tinacms/src/internalClient/index.ts
```diff
@@ -520,12 +645,67 @@ mutation addPendingDocumentMutation(
...
+            })
+          }
+        } catch (err) {
+          return Promise.reject(err)
+        }
+      },
+      // interval is 1000ms
+      1000, // interval
+      //  timeout is 10 min
+      600000 // timeout
```
kldavis4, line 672: after thinking about this some more, I think given the lambda timeout is 15 mins, that is probably a good number here (so 900000)

kldavis4:
The timeout doesn't seem to get handled properly. I see this error in the console but the dialog state doesn't change:

Screen Shot 2023-02-15 at 16 10 11

Screen Shot 2023-02-15 at 16 10 17

(Ignore that foobar branch - it has status unknown)

logan-anderson:
> The timeout doesn't seem to get handled properly. I see this error in the console but the dialog state doesn't change:
> 
> Screen Shot 2023-02-15 at 16 10 11
> 
> Screen Shot 2023-02-15 at 16 10 17
> 
> (Ignore that foobar branch - it has status unknown)

What should happen in this case?

logan-anderson:
Screenshot 2023-02-15 at 4 58 30 PM
@kldavis4  Updated so that this happens when it times out.

spbyrne:
Looks good to me, @logan-anderson!

kldavis4 (approved):

PR #3627, opened 2023-02-21
Aliases in field names

jamespohalloran:
fixes ENG-754
fixes ENG-120

Allow nameOverrides to be defined on a field, to allow us to use special field names. E.g, if you had a legacy field named "foo-bar", you could define your field like so:
```
  {
    type: "string",
    name: "foo_bar",
    nameOverride: "foo-bar",
  }
```

I'm applying the transformation to the fieldName before I index, and applying the opposite transformation when writing to the file. 

Next steps (probaby in follow up PRs):
- [Handle _template](https://linear.app/tina/issue/ENG-755/improve-support-for-blocks-with-template-instead-of-template) (we have hardcoded references to it in database/index.ts)
- add aliases automatically in forestry migration
- auto-apply aliases for special characters (fieldnames with dashes, fields named "id", etc)
- Give better error messages for invalid field names, prompting the user to use a nameOverride


packages/@tinacms/schema-tools/src/types.ts
```diff
@@ -483,6 +484,7 @@ export interface Template {
 export interface FieldCollection {
   label?: string
   name: string
+  nameOverride?: string
```
jeffsee55, line 487: @jamespohalloran this type is for the collection definition, is this needed here?

jeffsee55: I removed it when I merged with the schema types update so can add back if necessary

jamespohalloran: Oops yeah looks like it should just to be on the field: https://github.com/tinacms/tinacms/pull/3627/files#diff-88b24591b0c36f508270e9ae261e94ea2770198c387d66a95b2fe1620810bde3R147

PR #3629, opened 2023-02-21
Fix issue with registering form subscriptions too often

jeffsee55:
We were registered even subscribers on value change, which likely caused a performance issue. I'm not 100% sure this fixes the performance issue we've had reported intermittently, but seems like a good place to start.


kldavis4 (dismissed):

logan-anderson:
@jeffsee55 Should a Changeset be added?

logan-anderson (approved):

kldavis4 (approved):

PR #3631, opened 2023-02-22
Add checks to make sure local schema matches the server

logan-anderson:
This PR adds checks on both the frontend and back and that checks to make sure the schema is the same.

Logic of the schema check
- Make a GQL request to the server to get the GraphQL schema
- Pass the local GraphQL schema to the admin
- Use a graphQL diff tool to make sure they are the same. 

What is looks like like at build time
- https://www.loom.com/share/2daf0aecb1b44510ace8cf3071ef990b

What it looks like on the client
- https://www.loom.com/share/79598b4b00124bbbadf02350500dda97


examples/rich-text-sandbox/src/mdx.js
```diff
@@ -1,12 +1,12 @@
-// ../mdx/src/parse/index.ts
```
logan-anderson, line 1: I am not sure why when I ran build these files changed. @jeffsee55  Any ideas?

jeffsee55: Yeah I deleted this in another PR, my mistake. You can ignore this

jeffsee55 (approved):
This is great, and you mentioned we may not need it on the client but I think we have a few users who are building locally in CI, and so are only noticing mismatches once they enter edit mode, so this is perfect for them.

PR #3632, opened 2023-02-22
Schema types cleanup

jeffsee55:
Deprecate and reconfigure internal types. Most user-facing types should remain unchanged aside from a couple of bug fixes:

- Fix missing `indexed` and `indexes` properties on `collection` and `field` configs.

For the most part anything like `TinaCloud<Something>` or `<Something>Inner` and `<Something>Enriched` and `<Something>Base` has been removed in favor of just `<Something>`, which accepts an optional generic `WithNamespace` (`false` by default), that should usually be `true` internally because it indicates that we've processed the schema and added the `namespace` property to all of the fields.

Deprecations

```
CollectionFieldsWithNamespace
CollectionTemplates
CollectionTemplatesWithNamespace
GlobalTemplate
ObjectType
ReferenceType
ReferenceTypeInner
ReferenceTypeWithNamespace
RichTextType
RichTypeWithNamespace
TinaCloudCollection
TinaCloudCollectionBase
TinaCloudCollectionEnriched
TinaCloudSchema
TinaCloudSchemaBase
TinaCloudSchemaConfig
TinaCloudSchemaEnriched
TinaCloudSchemaWithNamespace
TinaCloudTemplateBase
TinaCloudTemplateEnriched
TinaFieldBase
TinaFieldEnriched
TinaFieldInner
TinaTemplate
```

Removals

```
Templatable
TinaIndex
ResolveFormArgs
```


packages/@tinacms/graphql/src/builder/index.ts
```diff
@@ -552,6 +545,7 @@ export class Builder {
             ],
           })
         }
+      // TODO: Should we throw here?
```
logan-anderson, line 548: We I think we could throw an error here. It theoretically should never happen since we check to make sure the length of fields or templates is greater then 1.

packages/tinacms/src/index.ts
```diff
@@ -63,10 +59,6 @@ export type TinaCloudCollection = Collection
  * @deprecated use `Collection` instead
  */
 export type TinaCollection = Collection
-/**
- * @deprecated use `Schema` instead
- */
```
logan-anderson, line 68: Just a note: I think we might still use some of these deprecated types in the tina-cloud-starter. We will have to update the starter once this PR goes though.

logan-anderson:
@jeffsee55 This looks good. Any idea why the E2E tests are all failing? I don't see any changes that update the UI.

packages/tinacms/src/tina-cms.tsx
```diff
@@ -245,23 +240,7 @@ export const TinaCMSProvider2 = ({
         schema={{ ...schema, config: { ...schema.config, ...props } }}
       >
         <style>{styles}</style>
-        <ErrorBoundary>
-          <DocumentCreator documentCreatorCallback={documentCreatorCallback} />
-          <TinaDataProvider formifyCallback={formifyCallback}>
```
logan-anderson, line 250: It is this change here that is breaking the tests. The example is out of date and is still using the "old way". I can update the example

packages/tinacms/src/index.ts
jeffsee55: I deleted it because `TinaSchema` is actually the class so this was actually wrong. I don't see it over in our starter though anyway

Issue #3621, opened 2023-02-17
Objects with list requires array return value for validate function, causing error message to not be shown


jono-vidyard:
### Describe the issue that you're seeing. Any Loom videos or screenshots usually help a lot!

Possible followup to: 

- https://discord.com/channels/835168149439643678/1059921499508723812/1059926309175828480
- https://github.com/tinacms/tinacms/issues/3495

For some reason, the `validate` function seems to require a return value of `Type[]` when being used with an object with list, according to `node_modules/@tinacms/schema-tools/dist/types.d.ts` for `UIField`:

![Screen Shot 2023-02-17 at 10 48 48 AM](https://user-images.githubusercontent.com/88905506/219758674-be79a3db-0945-43a4-b904-d28c09d53e66.png)

This seems counterintuitive - to me, the function should always return `string | void | undefined` regardless. As expected, when the function returns `Type[]`, it does not display the message in the UI.

### Reproduction

N/A (in the code)

### Steps to reproduce

The following appears to be valid and generates no TypeScript errors.

```
const innerFields: ObjectField = {
  type: 'object',
  label: 'Column Items',
  name: 'items',
  list: true,
  ui: {
    itemProps: (item) => {
      return {
        label: item?.title,
      };
    },
    validate: (items) => {
      if (items?.length < 2 || items?.length > 4) {
        return ['Columns must have between 2-4 items'];
      }
    },
    defaultItem: {
      ...defaultColumn,
    },
  },
  fields: [
    {
      type: 'string',
      label: 'Title',
      name: 'title',
    },
    {
      type: 'string',
      label: 'Text',
      name: 'text',
      ui: {
        component: 'textarea',
      },
    }
  ]
}

const defaultColumn = {
  title: 'Test Column',
  text: 'Description',
};

export const columnBlockSchema: Template = {
  name: 'columns',
  label: 'Columns (limit 2-4)',
  ui: {
    previewSrc: '/blocks/columns.png',
    defaultItem: {
      items: [defaultColumn, defaultColumn, defaultColumn],
    },
  },
  fields: [
    innerFields
  ]
};
```

You'll notice that the return value for the `validate` function is: `['Columns must have between 2-4 items'];`

Attempting to return as a string as expected: `'Columns must have between 2-4 items';` gives the following TypeScript error indicating the message should be `string[] | void`:

[...]

PR #3634, opened 2023-02-23
Fix issue where return type for  was incorrect

jeffsee55:
Closes https://github.com/tinacms/tinacms/issues/3621

### DIFF
--- a/packages/@tinacms/schema-tools/src/types.ts
+++ b/packages/@tinacms/schema-tools/src/types.ts
@@ -186,7 +186,7 @@ type UIField<Type, List extends boolean> = {
     allValues: { [key: string]: any },
     meta: Meta,
     field: UIField<Type, List>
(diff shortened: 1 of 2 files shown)


packages/@tinacms/schema-tools/src/types.ts
```diff
@@ -186,7 +186,7 @@ type UIField<Type, List extends boolean> = {
     allValues: { [key: string]: any },
     meta: Meta,
     field: UIField<Type, List>
-  ): (List extends true ? Type[] : Type) | undefined | void
```
logan-anderson, line 189: What is the issue here? 

For a string field `Type` should be `string`

logan-anderson (changes requested):

logan-anderson (approved):
Disregard my previous comment.

PR #3640, opened 2023-02-24
Feat/reference field selection

### LINKED ISSUE #3572: reference field, autogenerated graphQL

When using the `reference field` the autogenerated graphQL contains the hole website's data. I don't think this field  should be autogenerated or at least it should only contain  `_sys` data by default.

As for now i made a custom query. But I would love to include `...GlobalParts`. But for now it generates to much data. I can fetch the hole website with `...GlobalParts` and a `reference field`. 
 
My use case is for a menu:

```
query globalQuery($relativePath: String!) {
  global(relativePath: $relativePath) {
    menu {
      pageRef {
        ... on Page {
          title
          _sys {
            breadcrumbs
          }
        }
      }
    }
  }
}
```

I'm considering making a link module in my builder with the `reference field`, but then I have to write the hole page query with a builder myself and that wouldn't be that dynamic 

### Reproduction

none

### Steps to reproduce

console.log() the output of a `reference field`

### System Info

```shell
System:
    OS: macOS 13.0.1
    CPU: (16) x64 Intel(R) Core(TM) i9-9880H CPU @ 2.30GHz
    Memory: 117.13 MB / 16.00 GB
    Shell: 5.8.1 - /bin/zsh
  Binaries:
    Node: 16.3.0 - ~/.nvm/versions/node/v16.3.0/bin/node
    Yarn: 1.22.19 - ~/sites/idrottsrehab/node_modules/.bin/yarn
    npm: 8.3.0 - ~/.nvm/versions/node/v16.3.0/bin/npm
  Browsers:
    Chrome Canary: 112.0.5572.0
    Firefox Developer Edition: 110.0
    Firefox Nightly: 110.0a1
    Safari: 16.1
  npmPackages:
    @tinacms/cli: ^1.0.0 => 1.0.7 
    tinacms: ^1.0.0 => 1.1.4
```

### Validations

italoiung:
Fix #3572

Had to change this behaviour cause I was having trouble setting internal links within global blocks, that's got a collection for itself, and the maxDepth value needed for that feature would be impracticable.

I think I'd seen another issue requesting for the reference field option to select the fields exposed from the referenced collection, which is what this pull request is about, but it may also fix another circular references issues like the one mentioned, as it allows the user to prevent it within the API in certain cases.


jeffsee55:
@italoiung I'm not sure what problem this is tackling, are you saying the autogenerated query is overfetching data? In that case why can't you write your own custom query in the `.tina/queries` folder?

mmorkt:
Thanks for the fix!
@jeffsee55 please have a look at my issue https://github.com/tinacms/tinacms/issues/3572. I'm fetching the hole site from a single reference tag.

PR #3649, opened 2023-02-28
Fix/better name error

logan-anderson:
Fixes ENG-836

The error message looks like this.

```
TinaSchemaValidationError: Error name, "sub-title" must be alphanumeric and can only contain underscores. (No spaces, dashes, special characters, etc.)
If you only wan to display this value in the CMS UI, you can use the label property to customize it.

If you need to use this value in your content you can use the `nameOverride` property to customize the value. For example:
\```
{
  "name": sub_title,
  "nameOverride": sub-title,
  // ...
}
\```
```

### DIFF
--- a/packages/@tinacms/schema-tools/src/validate/properties.ts
+++ b/packages/@tinacms/schema-tools/src/validate/properties.ts
@@ -13,7 +13,7 @@ export const name = z
       ctx.addIssue({
         code: 'custom',
         message: `name, "${val}" must be alphanumeric and can only contain underscores. (No spaces, dashes, special characters, etc.)
-If you only wan to display this value in the CMS UI, you can use the label property to customize it.
+If you only want to display this value in the CMS UI, you can use the label property to customize it.
 
 If you need to use this value in your content you can use the \`nameOverride\` property to customize the value. For example:
 \`\`\`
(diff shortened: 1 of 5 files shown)


packages/@tinacms/schema-tools/src/validate/properties.ts
```diff
@@ -8,9 +8,21 @@ export const name = z
...
-    (val) => ({
-      message: `name, "${val}" must be alphanumeric and can only contain underscores`,
-    })
-  )
+  .superRefine((val, ctx) => {
+    if (val.match(/^[a-zA-Z0-9_]*$/) === null) {
+      ctx.addIssue({
+        code: 'custom',
+        message: `name, "${val}" must be alphanumeric and can only contain underscores. (No spaces, dashes, special characters, etc.)
+If you only wan to display this value in the CMS UI, you can use the label property to customize it.
```
jamespohalloran, line 16:
```suggestion
If you only want to display this value in the CMS UI, you can use the label property to customize it.
```

jamespohalloran (changes requested):
Looks good! just one typo

jamespohalloran (approved):

PR #3654, opened 2023-03-01
Feat/match in collection

logan-anderson:
> This will be a draft until the work is done on Tina Cloud to support this.

Adds back the match property.  The glob that is used looks like this.

<collection.path>/<collection.match>.<collection.format> 

Demo: https://www.loom.com/share/e1e7bc3523164b029c09d7d9d81be80c

### DIFF
--- a/.changeset/tricky-crabs-scream.md
+++ b/.changeset/tricky-crabs-scream.md
@@ -1,0 +1,5 @@
+---
+'@tinacms/graphql': patch
+---
+
+Added match property for collections
--- a/packages/@tinacms/graphql/src/database/bridge/filesystem.ts
+++ b/packages/@tinacms/graphql/src/database/bridge/filesystem.ts
@@ -19,10 +19,12 @@ export class FilesystemBridge implements Bridge {
   public addOutputPath(outputPath: string) {
     this.outputPath = outputPath
   }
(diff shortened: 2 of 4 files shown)


kldavis4:
@logan-anderson if you do the matching in the database and not the bridge, then I think there doesn't need to be any update in tina cloud

logan-anderson:
> @logan-anderson if you do the matching in the database and not the bridge, then I think there doesn't need to be any update in tina cloud

Would this have any performance concerns? If we do it it this way we would have to Still get all the the content in all of of the sub folders. (I could see this being an issue if we use the Github Bridge).

I don't think its much of a concern with isomorphic (or maybe it is) or the filesystem.

kldavis4:
It's not getting all the content. You get the full list of paths to all the content, and then filter on that list.

kldavis4:
The issue I see with doing it in the bridge is that for every bridge implementation you need to add the match logic vs adding it once in the database

logan-anderson:
> It's not getting all the content. You get the full list of paths to all the content, and then filter on that list.

I see. yes that makes perfect sense! I am going to make a new PR

logan-anderson:
Closing in favour of #3656

PR #3655, opened 2023-03-01
support pdf uploads

jamespohalloran:
Support pdf uploads:

https://www.loom.com/share/0add2b2f8da049368b3dd161025779b1

- Updated default mime-type of ImageProviders to include PDF
- Added non-image icon in media library
- Tweaked UX so we open media library immediately upon clicking media toolbar icon

closes ENG-19

TODO in future PR:
- Spike on making the "accept" argument configurable


packages/next-tinacms-cloudinary/src/cloudinary-media-store.ts
```diff
@@ -16,7 +16,7 @@ export class CloudinaryMediaStore implements MediaStore {
   fetchFunction = (input: RequestInfo, init?: RequestInit) => {
     return fetch(input, init)
   }
-  accept = 'text/*,application/*,image/*'
+  accept = 'text/*,application/pdf,image/*'
```
kldavis4, line 19: did you intend to restrict accept from application/* to application/pdf (ie was it incorrect before?)

logan-anderson: Maybe we could make accept configurable @jamespohalloran ?

jamespohalloran: > did you intend to restrict accept from application/* to application/pdf (ie was it incorrect before?)

@kldavis4 I'm not sure if this is a bug with react-dropzone, but unless I specified application/pdf specifically, PDF's weren't selectable with the file-browser. I can poke around and see if there's a workaround

Screen Shot 2023-03-01 at 4 10 22 PM

> Maybe we could make accept configurable @jamespohalloran ?

Yeah that might be nice. I think we'd probably want to do it on a per-field level though, rather than here at the media provider level

jamespohalloran: @kldavis4 I hardcoded a list of some common potential `application/` mime-types for now, since `application/*` doesn't seem to be supported.

@logan-anderson I was thinking maybe we could make this configurable in a future PR? I think if this is configurable on a per-field basis, it might not be super trivial

logan-anderson:
Code looks good but it seems like this might add some strange unexpected behaviour. 

See video for more details: https://www.loom.com/share/df6dab77414b4184ac60faa247d1ee1e

jamespohalloran:
> Code looks good but it seems like this might add some strange unexpected behaviour.
> 
> See video for more details: https://www.loom.com/share/df6dab77414b4184ac60faa247d1ee1e

Just tested this on main and seeing the same bug (inserting a link, and then inserting an image) in the same line

jamespohalloran:
> Code looks good but it seems like this might add some strange unexpected behaviour.
> 
> See video for more details: https://www.loom.com/share/df6dab77414b4184ac60faa247d1ee1e

Create a new issue here: https://github.com/tinacms/tinacms/issues/3661

logan-anderson (approved):

logan-anderson (changes requested):
@jamespohalloran I just ran `pnpm install` and the lock file was updated. Can you run `pnpm install` and push up the changes?

PR #3656, opened 2023-03-01
Add `match` for  collection from db

logan-anderson:
Adds back the match property.  The glob that is used looks like this.

<collection.path>/<collection.match>.<collection.format> 

Demo: https://www.loom.com/share/e1e7bc3523164b029c09d7d9d81be80c

Fixes: ENG-43


kldavis4:
Should there be a check in the Database.put function against the match pattern (if present)?

logan-anderson:
> Should there be a check in the Database.put function against the match pattern (if present)?

Yes good idea!

kldavis4:
code looks good to me - I assume the demo is based on the previous PR?

logan-anderson:
@kldavis4  Yes I just added the matches Key word to a collection to test.

logan-anderson:
The only thing I can think of that this PR does not cover is when we have two collections with the same path and differnt overlapping matches property. We talked about this should raise a warning or error. Not sure the best way to handle this. Thoughts @kldavis4 and @jeffsee55  ?

logan-anderson:
> The only thing I can think of that this PR does not cover is when we have two collections with the same path and differnt overlapping matches property. We talked about this should raise a warning or error. Not sure the best way to handle this. Thoughts @kldavis4 and @jeffsee55 ?

One way to do this is to check while we are indexing the content.


Could add this in `_indexAllContent` function

```ts
filteredPaths.forEach((path) => {
        if (filesSeen.has(path)) {
          // TODO: link to docs
          console.warn(
            `"${path}" Found in multiple collections. This can cause unexpected behavior. We recommend updating the \`match\` property to limit the scope of a collection.`
          )
        } else {
          filesSeen.add(path)
        }
      })

```

logan-anderson:
I update the indexing process to check for duplicate files. It prints out a warning like this.

```
Indexing completed with 1 warning(s)
"content/test/a/c.md" Found in collections multiple collections: col1, col2. This can cause unexpected behavior. We recommend updating the `match` property of those collections so that each file is in only one collection.
```

packages/@tinacms/graphql/src/database/index.ts
```diff
@@ -1000,16 +1020,46 @@ export class Database {
...
+            )}. This can cause unexpected behavior. We recommend updating the \`match\` property of those collections so that each file is in only one collection.`
+        )
+      })
+
+      await _indexContent(this, level, filteredPaths, enqueueOps, collection)
     })
     while (operations.length) {
       await level.batch(operations.splice(0, 25))
     }
+    return { warnings }
```
kldavis4, line 1062: should we have a return type on the functions in this that return the warnings? Also wondering if it would be better to error out rather than emit warnings? @jamespohalloran or @jeffsee55 have thoughts on that?

jamespohalloran: Hmm, if the behaviour's wonky for overlapping collections, an error does seem appropriate to me.
It is tricky the amount of overlap we'll likely face for Forestry sites (especially if we're going to use the `match` field to handle single document collections 🤔)

logan-anderson: Only reason I am leaning towards warnings is to remain backwards compatible

logan-anderson: Also, I _think_ the return type is inferred.

kldavis4: what do you mean by backwards compatible? are there sites using `match` currently?

logan-anderson:
Docs: https://github.com/tinacms/tina.io/pull/1500/files

logan-anderson:
I think this is good to go. I think we have two options in regards to warning VS error

1. Review and merge as is (Leave as warning)
- This is backwards compatible
- Can leave as a patch
- Does not have to wait for backups updates

2. Switch to Error 
- Is not backwards compatible 
-  Must to a minor update to GraphQL
- Will have to wait for Tina Cloud updates

packages/@tinacms/graphql/src/schema/index.test.ts
```diff
@@ -9,10 +9,14 @@ describe('TinaSchema', () => {
     test('fetches correct collection with getCollectionByFullPath', async () => {
       const schema = createSchema()
 
-      const collection1 = schema.getCollectionByFullPath('content/test/foobar')
```
logan-anderson, line 12: @jeffsee55 Is this function ever called without the file extension?

jeffsee55: I think that's safe to assume.

> Disclaimer- I either didn't write this originally or remember almost nothing about it.

jeffsee55 (approved):

PR #3666, opened 2023-03-06
fix: When current branch doesn't exist, don't blow up

jamespohalloran:
Previously, when our locally cached branch doesn't exist, things kind of blow up in the CMS.
Screen Shot 2023-03-06 at 2 56 51 PM
I added a minimal error screen, but more importantly, the toolbar stays accessible (so the user can change branches and resolve the issue).

closes ENG-802

### DIFF
--- a/.changeset/sour-gifts-double.md
+++ b/.changeset/sour-gifts-double.md
@@ -1,0 +1,6 @@
+---
+"tinacms": patch
+"@tinacms/toolkit": patch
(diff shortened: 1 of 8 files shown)


spbyrne:
Looks good! I just added a couple icons & replaced the button with the `@tinacms/toolkit` button component.

logan-anderson (approved):

PR #3660, opened 2023-03-03
fix: previous directory link in Tina MediaManager

coreyaus:
## Bug description:

- Previously the `dirname` function had a return value of type `RegExpMatchArray | null` - the result of calling `String.prototype.match()`
- However the `prevDir` variable must be of type `string` (or undefined or null) as it is passed directly to the `setDirectory` function
- Previously the media manager crashed when clicking the back link and threw a `TypeError` with the message `directory.replace is not a function` (given you can't call `replace` on a `RegExpMatchArray`)
- This fix extends the regex in the `dirname` function to use a capturing group and ensure it always returns a `string` or `undefined`

PR #3667, opened 2023-03-06
Media Back Button

spbyrne:
This fixes the breadcrumb back button behaviour when you're more than a single folder deep.

Screenshot 2023-03-06 at 2 01 50 PM

### DIFF
--- a/.changeset/stupid-apes-film.md
+++ b/.changeset/stupid-apes-film.md
@@ -1,0 +1,5 @@
+---
+'@tinacms/toolkit': patch
+---
+
+Fix media manager back button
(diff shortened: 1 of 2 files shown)


logan-anderson:
@spbyrne You might want to be aware of the regex issue here: https://github.com/tinacms/tinacms/pull/3660 (these changes look like they conflict with your changes here)

spbyrne:
Ah, I didn't see that fix. We're both basically fixing the same thing. I can test that code to see if it fixes when you're nested more than a single folder deep like this PR does.

jamespohalloran:
Merged here: https://github.com/tinacms/tinacms/pull/3660

PR #3693, opened 2023-03-14
Image List Component

spbyrne:
This adds list support for the image field type and improves the UI for the base image field. 

Screenshot 2023-03-14 at 10 32 59 AM

closes ENG-845

### DIFF
--- a/packages/@tinacms/toolkit/src/packages/fields/components/ImageUpload/ImageUpload.tsx
+++ b/packages/@tinacms/toolkit/src/packages/fields/components/ImageUpload/ImageUpload.tsx
@@ -12,6 +12,7 @@ import {
   DEFAULT_MEDIA_UPLOAD_TYPES,
   dropzoneAcceptFromString,
 } from '../../../../components/media/utils'
+import { BiFileBlank } from 'react-icons/bi'
 
 interface ImageUploadProps {
   onDrop: (_acceptedFiles: any[]) => void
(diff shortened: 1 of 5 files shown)


packages/@tinacms/toolkit/src/packages/fields/components/ImageUpload/ImageUpload.tsx
```diff
@@ -22,25 +23,29 @@ interface ImageUploadProps {
   loading?: boolean
 }
 
+const isStringImage = (src: string) => {
```
jamespohalloran, line 26: Can we use this existing function for this?
https://github.com/tinacms/tinacms/blob/a464f90e3e4e0bb6da0a716979807166a6f30a32/packages/@tinacms/toolkit/src/components/media/utils.ts#L22

@logan-anderson modified the Regex to work with Tina Cloud images, so not sure if that's something we'll need here or not

jamespohalloran (changes requested):
UI looks great! one comment on the isMedia fn, which might lead to a bug on prod

packages/@tinacms/toolkit/src/packages/fields/components/ImageUpload/ImageUpload.tsx
spbyrne: Ah! I see. This was mainly a quick stab at improving the `src && (src.includes('png') || src.includes('svg'))` logic.

jamespohalloran (approved):
Looks good!

PR #3700, opened 2023-03-14
Skip tokenizing html when parser is configured for skipping it

jeffsee55:
HTML is tokenized in a way that treats the next line as part of the html block. So the following code:

```
<div>
Item
```
Is parsed as 
```
{
  "type": "root",
  "children": [
    {
      "type": "html",
      "value": "<div>\nItem"
    }
  ],
}
```
This creates an issue when child elements of shortcodes are html, since the next line is the closing shortcode:
```
{{ myshortcode }}
<div>
{{ /myshortcode }}
```
The shortcode tokenizing logic doesn't even get a chance to run on the line `{{ /myshortcode }}`, it's consumed as HTML. The result is `<div>\n{{ /myshortcode }}` is shown in the rich-text editor for that node, and we print an extra closing tag during stringify:
```
{{ myshortcode }}
<div>
{{ /myshortcode }}
{{ /myshortcode }}
```

For now I think it makes sense to turn off the html tokenizer entirely when `skipEscaping` is turned on (which is what this PR does). And it might make sense for us to just turn off html tokenizing altogether, or replace it with our own more forgiving solution.

### DIFF
--- a/.changeset/curly-rules-flow.md
+++ b/.changeset/curly-rules-flow.md
@@ -1,0 +1,5 @@
+---
+'@tinacms/mdx': patch
+---
+
+Skip html tokenization when skipEscaping is enabled.
--- a/examples/kitchen-sink/.tina/config.tsx
+++ b/examples/kitchen-sink/.tina/config.tsx
@@ -750,6 +750,26 @@ export default defineConfig({
                 ],
               },
               {
+                name: 'center',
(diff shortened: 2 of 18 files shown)


kldavis4 (approved):
I'm not totally clear on the nuts and bolts of how this works, but not seeing any issues with the code

PR #3703, opened 2023-03-15
Media UI

spbyrne:
- Updates grid/list toggle UI
- Adds error modal for failed uploads
- Adds support for more complex formatting in alerts when needed (line breaks, bold text..etc)

The errors work, and most of the time I'm sure they'll be accurate (most often it will fail with an unsupported file) but they don't actually know what caused the failure, so if there's a better way to handle this please let me know.

### DIFF
--- a/.changeset/nasty-seals-cheer.md
+++ b/.changeset/nasty-seals-cheer.md
@@ -1,0 +1,5 @@
+---
+'@tinacms/toolkit': patch
+---
+
+Add error modal for failed media uploads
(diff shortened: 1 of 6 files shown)


kldavis4 (dismissed):
lgtm and works as expected

spbyrne:
Jeff added code to improve the errors in the media manager dropzone; I'm going to try copying this to the image field so both are handled in the same way.

kldavis4 (approved):

PR #3709, opened 2023-03-16
Only stop dev process when config is invalid at the start of the command

jeffsee55:
@logan-anderson this fixes one issue, but it does look like it's reindexing on schema changes. Perhaps too often, but it's working for me. I wonder if the new matches property introduces something that I wasn't able to see on `main`.


kldavis4 (approved):

logan-anderson:
> but it does look like it's reindexing on schema changes

This does not seem to be happening for me. 

If I 
1. add a new called `content/new-folder/test.md`
2. Start dev server.
3. Add a new collection that has path: content/new-folder
4. Get an error "No index definition for content/new-folder/test.md"

logan-anderson:
@jeffsee55  Here is a video for context: https://www.loom.com/share/ceda5bf2c14b4ba9a8fc49da7f9ca1b7

logan-anderson:
@jeffsee55 Where you able to solve the issues that I ran into?

jeffsee55:
Yes my last commit seems to have fixed it

logan-anderson (approved):
Looks good to me!

PR #3721, opened 2023-03-21
Improve forestry migration import

jamespohalloran:
Improves a few aspects of the forestry migration:
- A bit of refactoring
- branch off of [this](https://github.com/tinacms/tinacms/pull/3672) for match importing
- Skip blocks import, since it has some issues
- Handle image gallery import
- Handle a nested forestry directory through `--forestryPath` arg (It places the package.json, and .tina directory at the root, but looks in "forestryPath" for the .forestry folder)

closes ENG-755


packages/@tinacms/cli/src/next/commands/init-command/index.ts
```diff
@@ -4,8 +4,9 @@ import { initStaticTina } from '../../../cmds/init'
 
 export class InitCommand extends Command {
   static paths = [['init']]
-  rootPath = Option.String('--rootPath', {
-    description: 'Specify the root directory to run the CLI from',
+  rootPath = Option.String('--forestryPath', {
```
logan-anderson, line 7: We should might want to consider calling the variable `forestryPath` since we use `rootPath` in other places and it has a different meaning.

logan-anderson:
Another small issue: https://www.loom.com/share/4bea971f25924ec8be6cebf3e8dbbb7d

logan-anderson (changes requested):
This looks good to me besides the two comments I made.

packages/@tinacms/cli/src/next/commands/init-command/index.ts
jamespohalloran: Updated!

jeffsee55:
@logan-anderson I think your comments have been addressed

logan-anderson (approved):

PR #3728, opened 2023-03-22
Ensure errors during dev are thrown properly and exit the process

jeffsee55:


jamespohalloran:
Might need more context on this one, but I'd kind of expect it to kill the process when first running tinacms dev if there's an error, but if I make a change, break the schema, fix the schema, it might be nice if the server stayed running?

jeffsee55:
> but if I make a change, break the schema, fix the schema, it might be nice if the server stayed running?

That's how it's working right now, but if you do something that results in an invalid schema, it shows the issue in the CLI and you may not notice it, and then if you stop the server and restart things break. 🤔 Maybe this needs to just be louder in the CLI

PR #3733, opened 2023-03-23
Move noSDK into config

jeffsee55:
This shows a warning if `--noSDK` is provided, but that property still works for now. 

```js
{
  client: {
    skip: true
  }
}
```

The work to add this to part of the forestry migration init hasn't been done yet.

Closes ENG-869


packages/@tinacms/app/src/Playground.tsx
```diff
@@ -73,7 +73,7 @@ const Playground = () => {
         <div className="graphiql-doc-explorer-content">
           <div className="graphiql-markdown-description">
             {noAutoQueries
-              ? 'No autoqueries found, the Tina dev or build command was likely run with the "--noSDK" option. Learn about how to generate autoqeries.'
+              ? 'No autoqueries found, the Tina config is likey set to client.skip = true. Learn about how to generate autoqeries.'
```
jamespohalloran, line 76:
```suggestion
              ? 'No autoqueries found. The Tina config is likely set to "client.skip = true". Learn about how to generate autoqueries.'
```
^ Not new to this, but kind of seems like we meant to include a link at the end of that. I also don't think we refer to "autoqueries" in our docs.

packages/@tinacms/cli/src/next/commands/build-command/index.ts
```diff
@@ -29,7 +29,7 @@ export class BuildCommand extends Command {
     description: 'increase verbosity of logged output',
   })
   noSDK = Option.Boolean('--noSDK', false, {
-    description: "Don't generate the generated client SDK",
+    description: "DEPRECATED - Don't generate the generated client SDK",
```
jamespohalloran, line 32: Might be good to mention to new solution here, or link to a doc

jamespohalloran (changes requested):

packages/@tinacms/app/src/Playground.tsx
jeffsee55: Will follow up with a link in another PR. There are some other playground improvements needed.

packages/@tinacms/cli/src/next/commands/build-command/index.ts
jamespohalloran: I don't see why not

jeffsee55, line 32:
```suggestion
    description: "DEPRECATED - This should now be set in the config at config.skip = true'". Don't generate the generated client SDK",
```

packages/@tinacms/cli/src/next/commands/dev-command/index.ts
```diff
@@ -46,7 +46,7 @@ export class DevCommand extends Command {
     description: "Don't regenerate config on file changes",
   })
   noSDK = Option.Boolean('--noSDK', false, {
-    description: "Don't generate the generated client SDK",
+    description: "DEPRECATED - Don't generate the generated client SDK",
```
jeffsee55, line 49:
```suggestion
    description: "DEPRECATED - This should now be set in the config at config.skip = true'. Don't generate the generated client SDK",
```

logan-anderson (approved):

PR #3736, opened 2023-03-23
Improve error message when a _template property isn't one of the templates in the collection

jeffsee55:
Message will now say:
```
Error: Unable to find template "blockPag". Possible templates are: "showcase", "blockPage".
```

Closes ENG-865

### DIFF
--- a/.changeset/witty-turkeys-dress.md
+++ b/.changeset/witty-turkeys-dress.md
@@ -1,0 +1,5 @@
+---
+'@tinacms/graphql': patch
+---
+
+Improve error message when a \_template property isn't one of the templates in the collection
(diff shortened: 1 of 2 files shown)


jeffsee55:
This is where it's coming from, the error shows up further down but it originates here, `template` is `null` because the name in the markdown `_template` isn't one the collection has.

jamespohalloran:
> This is where it's coming from, the error shows up further down but it originates here, `template` is `null` because the name in the markdown `_template` isn't one the collection has.

@jeffsee55 
What I mean is, if I have a document `about.md`, that has frontmatter like:
```
// ...
_template: invalid-value
```

I just mean it would be helpful to give the error message more context. Something like:

```
Error in **about.md**
`invalid-value` is not a valid template for field `my-block`
```

Otherwise, It may be hard to know where to actually go about fixing the issue.

The template might also even be a real template, but it might just not be registered as a template for a given block.

jeffsee55:
Oh yeah that would be good, will try to follow up with this.

PR #3434, opened 2022-12-01
Include class names from `.tina/config` file in stylesheet

jeffsee55:
Fix issue where tailwind styles from within the user config weren't being picked up

PR #3740, opened 2023-03-24
Allow user files in .tina to be searched for Tailwind classes.

Maxobat:
Related to this: [https://github.com/tinacms/tinacms/pull/3434](https://github.com/tinacms/tinacms/pull/3434), a fix was added at some point for incorporating components defined in `.tina` into the Tailwind compilation. This works, mostly. It does this by including `.tina/config.{js,jsx,ts,tsx}` into the Tailwind config. But, this ignores any other files in `.tina`, like user-defined custom fields. Thus, Tailwind classes defined in some custom components don't get included in the final build. So, this change configures Tailwind to to include `.tina/**/*.{vue,js,ts,jsx,tsx,svelte}`.


jeffsee55 (approved):
Awesome! Thanks. I think we're only using 1 of these files now but I'll delete that in a follow up PR

PR #3749, opened 2023-03-27
Handle ambiguous templates from indexing

jamespohalloran:
Skip ambiguous from indexing, so we log a warning instead of failing.
This led to some friction when importing Forestry sites with missing templates

closes https://linear.app/tina/issue/ENG-874/spike-can-we-handle-ambiguous-templates-a-bit-better-when-importing

### DIFF
--- a/.changeset/late-peaches-change.md
+++ b/.changeset/late-peaches-change.md
@@ -1,0 +1,5 @@
+---
+"@tinacms/graphql": patch
+---
+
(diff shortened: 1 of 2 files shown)


logan-anderson:
I am ok with this behaviour. Thoughts @jeffsee55  ?

jeffsee55 (approved):

PR #3750, opened 2023-03-27
feat: datalayer folders

kldavis4:
# what

Implements folder indexing with collections and updates the admin user interface to enable a folder mode.

Screenshot 2023-04-13 at 4 18 18 PM


spbyrne:
Here's where I'm at with the UI. It sounds like people like to remove the toggle, though. It'll use a folder view unless the user is using search (if I remember correctly from the discussion).

Screenshot 2023-04-04 at 3 28 25 PM

logan-anderson (dismissed):
Looks good to me! I think its good to merge as is but do you think we should we also add the folder to Generated queries? (Maybe a separate  issue)

EX
```
get<Collection>Connection( folder: "foo")
...
```
```

**Comment by jeffsee55** [comment]
> but do you think we should we also add the folder to Generated queries

I think we're treating this as more of an internal API for now so that's not necessary. It's also a little bit strange because the results of the query are all of type `Document`, when it reality it's a directory. Might be something we want to clean up beforehand.

**Comment by kldavis4** [comment]
I think the contextual editing links are not working for documents in sub-folders in the kitchen-sink example. If I click on a root level post it opens fine. If I click on a document in a folder, it routes to `~/:collection/:filename` and throws an error: `Unable to find record content/:collection/:filename.mdx`. If I manually update the url to include the folder name (`~/:collection/:folder/:filename`) I get an empty form and a 404 error.

**Comment by kldavis4** [comment]
> > but do you think we should we also add the folder to Generated queries
> 
> I think we're treating this as more of an internal API for now so that's not necessary. It's also a little bit strange because the results of the query are all of type `Document`, when it reality it's a directory. Might be something we want to clean up beforehand.

I'm not sure I'm following this 🤔

**Comment by logan-anderson** [comment]
Re: @kldavis4 
> I'm not sure I'm following this 🤔

I am talking about potentially adding the file filter to `get<CollectionName>Connections`. 

EX
```
getPostConnnections(folder: "nested") {
   ...
}
```

I agree with @jeffsee55  It can stay as an internal API for now but it can remain on the radar.

 > Is this something that the route mapper is responsible for handling (if configured)?

@kldavis4  I think this is an issue with the implemtion in getStaticProps. I can take a look.

kldavis4:
> Re: @kldavis4
> 
> > I'm not sure I'm following this 🤔
> 
> I am talking about potentially adding the file filter to `get<CollectionName>Connections`.
> 
> EX
> 
> ```
> getPostConnnections(folder: "nested") {
>    ...
> }
> ```
> 
> I agree with @jeffsee55 It can stay as an internal API for now but it can remain on the radar.

^ Not this. This:

> It's also a little bit strange because the results of the query are all of type Document, when it reality it's a directory. Might be something we want to clean up beforehand.

logan-anderson:
> > Re: @kldavis4
> > > I'm not sure I'm following this 🤔
> > 
> > 
> > I am talking about potentially adding the file filter to `get<CollectionName>Connections`.
> > EX
> > ```
> > getPostConnnections(folder: "nested") {
> >    ...
> > }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > I agree with @jeffsee55 It can stay as an internal API for now but it can remain on the radar.
> 
> ^ Not this. This:
> 
> > It's also a little bit strange because the results of the query are all of type Document, when it reality it's a directory. Might be something we want to clean up beforehand.

@kldavis4  I am not sure either.

logan-anderson:
@kldavis4  I made [pr into this branch](https://github.com/tinacms/tinacms/pull/3833) that updates the example to work with folders.

kldavis4:
thanks @logan-anderson, can you update kitchen-sink as well?

jeffsee55:
> I'm not sure I'm following this 🤔

@kldavis4 what I mean is that if we add the `folder` param won't that change the results of user's queries if one of the results was a folder?

kldavis4:
> > I'm not sure I'm following this 🤔
> 
> @kldavis4 what I mean is that if we add the `folder` param won't that change the results of user's queries if one of the results was a folder?

right, ok that makes sense. Do you have an idea on how to resolve that?

jeffsee55:
> Do you have an idea on how to resolve that?

When I tried to add the proper GraphQL type to support this it would have broken the current way of querying so I don't think there's much we can do. Maybe we should have a separate field on the `collection` for the feature specifically 🤔

kldavis4:
> @kldavis4 I made [pr into this branch](https://github.com/tinacms/tinacms/pull/3833) that updates the example to work with folders.

@logan-anderson can you make this change in the kitchen-sink starter?

logan-anderson:
@kldavis4 I made a PR: https://github.com/tinacms/tinacms/pull/3840 to add that

logan-anderson:
@kldavis4 Found a small issue: https://www.loom.com/share/a2e114f1e6e343aa9fc7c9a7eec2e9d0

kldavis4:
> @kldavis4 Found a small issue: https://www.loom.com/share/a2e114f1e6e343aa9fc7c9a7eec2e9d0

@logan-anderson seems to be a bit of a weird race condition that I'm not sure the best way to resolve. I inserted a 10ms delay and it seems to have fixed it, but not sure if there is a better solution

logan-anderson (approved):
LGTM.

logan-anderson (approved):

PR #3767, opened 2023-03-31
Better match for Forestry migrate

logan-anderson:
Fixes two issue brought up in office hours
- Migration was not probably matching when path is `""` or `"/"`
- Better message for Asking about typescript

### DIFF
--- a/packages/@tinacms/cli/src/cmds/forestry-migrate/index.ts
+++ b/packages/@tinacms/cli/src/cmds/forestry-migrate/index.ts
@@ -110,151 +110,160 @@ const generateCollectionFromForestrySection = (
 ) => {
   if (section.read_only) return
 
-  {
-    // TODO: What should we do with read only sections?
-    if (section.read_only) return
+  // TODO: What should we do with read only sections?
+  if (section.read_only) return
 
-    const baseCollection: Omit<Collection, 'fields' | 'templates'> = {
-      label: section.label,
-      name: stringifyLabel(section.label),
-      path: section.path,
+  const baseCollection: Omit<Collection, 'fields' | 'templates'> = {
+    label: section.label,
+    name: stringifyLabel(section.label),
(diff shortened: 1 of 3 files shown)


packages/@tinacms/cli/src/cmds/forestry-migrate/index.ts
```diff
@@ -110,151 +110,160 @@ const generateCollectionFromForestrySection = (
...
-    // Add include and exclude if they exist
-    if (section.match) {
-      baseCollection.match = {
-        ...(baseCollection?.match || {}),
-        include: transformForestryMatchToTinaMatch(section.match),
-      }
+  const baseCollection: Omit<Collection, 'fields' | 'templates'> = {
+    label: section.label,
+    name: stringifyLabel(section.label),
+    path: section.path || '/',
```
jamespohalloran, line 119: Collections with both an empty string, and path: "/" seems to result in the same error FYI, so I don't know if this will fix the underlying issue

logan-anderson: Fixed to ignore section like this.

packages/@tinacms/cli/src/cmds/init/index.ts
```diff
@@ -183,7 +180,8 @@ const chooseTypescript = async () => {
     name: 'selection',
     type: 'confirm',
     initial: true,
-    message: 'Would you like to use Typescript?',
+    message:
+      'Would you like to use Typescript for your Tina Configuration (Recommend)?',
```
jamespohalloran, line 184:
```suggestion
      'Would you like to use Typescript for your Tina Configuration (Recommended)?',
```

jamespohalloran (changes requested):

jamespohalloran (approved):

PR #3800, opened 2023-04-12
fix: update internal client to only set Authorization header when token is available

kldavis4:
# what

Previously, if token was not set, Authorization header was being set with value of `Bearer null` which will generate an incorrect 403 status. The correct status should be 401.

### DIFF
--- a/packages/tinacms/src/internalClient/index.ts
+++ b/packages/tinacms/src/internalClient/index.ts
@@ -385,12 +385,16 @@ mutation addPendingDocumentMutation(
     query: ((gqlTag: typeof gql) => DocumentNode) | string,
     { variables }: { variables: object }
   ): Promise<ReturnType> {
+    const token = await this.getToken()
+    const headers = {
+      'Content-Type': 'application/json',
+    }
+    if (token?.id_token) {
+      headers['Authorization'] = 'Bearer ' + token.id_token
+    }
     const res = await fetch(this.contentApiUrl, {
       method: 'POST',
-      headers: {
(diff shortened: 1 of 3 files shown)


logan-anderson (approved):

jeffsee55 (dismissed):

packages/tinacms/src/internalClient/index.ts
```diff
@@ -565,10 +569,14 @@ mutation addPendingDocumentMutation(
     init?: RequestInit
   ): Promise<Response> {
     const headers = init?.headers || {}
+    const token = await this.getToken()
+    if (!token?.id_token) {
```
logan-anderson, line 573: @kldavis4  I am unsure about throwing an error here. There might be cases where you want to not pass the token. For example;
- Using cloudinary locally
- Self hosted locally or want to pass a token some other way (maybe a query param)

If we checked that we where hitting tina.io first then I could definitly see throwing an error.

Sudo code

```ts
if (!token?.id_token && url.contains('tina.io') {
     throw new Error('Authentication required')
}
```

logan-anderson (changes requested):
Looks good other then my one comment.

packages/tinacms/src/internalClient/index.ts
kldavis4: If that's the case, that is not a good name for that function

logan-anderson: Agreed it could definitely be renamed in the future.

logan-anderson (approved):

PR #3803, opened 2023-04-13
Handle root path

logan-anderson:
## How to test
In a collection in one of the examples projects make the path `/`, or `""`. You can update the match.include property so that the collection can still find the correct documents. 

Check admin collection list page and edit document page and make sure you can create update and delete documents. 

Fixes ENG-898


packages/@tinacms/schema-tools/src/schema/TinaSchema.ts
```diff
@@ -452,14 +452,15 @@ export class TinaSchema {
         ? this.getCollection(collectionOrString)
         : collectionOrString
     const normalPath = normalizePath(collection.path)
+    const slash = normalPath ? '/' : ''
```
jamespohalloran, line 455: I don't know if I follow why `slash` may be an empty string?

logan-anderson: If normalPath is `""` slash is empty otherwise its a /. I will make a better name for this.

logan-anderson: Updated

jeffsee55 (approved):

PR #3809, opened 2023-04-13
fix: normalize windows paths in config-manager

kldavis4:
Discord user reported error on save: https://discord.com/channels/835168149439643678/1096106617184985240

Issue was due to config-manager not transforming the windows path separators

### DIFF
--- a/.changeset/tricky-colts-live.md
+++ b/.changeset/tricky-colts-live.md
@@ -1,0 +1,5 @@
+---
+'@tinacms/cli': patch
+---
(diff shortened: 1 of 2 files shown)


jeffsee55 (approved):
Thanks for the fix, I'm not sure why it's failing silently 🤔

kldavis4:
> 

Yeah, I tried unsuccessfully to figure that out. I had to use the node debugger to see where it was failing.

PR #3815, opened 2023-04-14
feat: Add ability to duplicate documents

jamespohalloran:
Add the ability to duplicate documents from collection list

https://www.loom.com/share/a052bbb2463945139c85c8978aef8afe

### DIFF
--- a/packages/tinacms/src/admin/pages/CollectionListPage.tsx
+++ b/packages/tinacms/src/admin/pages/CollectionListPage.tsx
@@ -725,12 +725,11 @@ const CollectionListPage = () => {
                                               )
                                             },
                                           },
-                                          {
+                                          allowCreate && {
                                             name: 'duplicate',
                                             label: 'Duplicate',
                                             Icon: <BiCopy size="1.3rem" />,
                                             onMouseDown: () => {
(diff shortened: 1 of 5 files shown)


packages/tinacms/src/admin/pages/CollectionListPage.tsx
```diff
@@ -724,6 +725,20 @@ const CollectionListPage = () => {
                                               )
                                             },
                                           },
+                                          {
```
logan-anderson, line 728: We should probably only show this if the user allows creation of documents. 

```
 allowCreate && { ...
```

packages/tinacms/src/admin/pages/CollectionListPage.tsx
```diff
@@ -724,6 +725,20 @@ const CollectionListPage = () => {
...
+                                            name: 'duplicate',
+                                            label: 'Duplicate',
+                                            Icon: <BiCopy size="1.3rem" />,
+                                            onMouseDown: () => {
+                                              console.log('document', document)
```
logan-anderson, line 733: Remove console log?

logan-anderson (approved):

PR #3822, opened 2023-04-18
Fix/skip client checks for self hosted

logan-anderson:
Skips some checks when the user is doing self hosted.


jeffsee55 (approved):
Looks good, in the future it would be nice if we could perform these checks against the self-hosted config.

PR #3830, opened 2023-04-19
Simplify bridge interface

logan-anderson:
Implements suggestions in https://linear.app/tina/issue/ENG-916/spike-on-simplifying-bridge

cc @jeffsee55


kldavis4:
@logan-anderson here's my proposal for getting rid of that updateConfig handler from the database: https://github.com/tinacms/tinacms/pull/3842

packages/@tinacms/cli/src/next/codegen/index.ts
```diff
@@ -11,25 +14,61 @@ export class Codegen {
...
     this.queryDoc = queryDoc
     this.fragDoc = fragDoc
+    this.lookup = lookup
+  }
+
+  async writeConfigFile(fileName: string, data: string) {
+    const tinaFolderPath = await this.configManager.getTinaFolderPath(
+      this.configManager.rootPath
+    )
+    const filePath = path.join(tinaFolderPath, GENERATED_FOLDER, fileName)
```
jeffsee55, line 58:
```suggestion
    const filePath = path.join(this.configManager.generatedFolderPath, fileName)
```

jeffsee55 (approved):
Small comment but no big deal, looks good

packages/@tinacms/cli/src/next/codegen/index.ts
kldavis4: @jeffsee55 what about the one on line 68 below it?

jeffsee55: Yes sorry same thing in both places.

logan-anderson: @jeffsee55 Do we have a generatedFolder path for both the rootPath and the contentRootPath? id like to keep it consistent and leave it as is if we don't

jeffsee55: Oh sorry, I misunderstood that. No we don't have that, but would be handy to have as part of the config manager.

logan-anderson:
@kldavis4  and @jeffsee55  I updated the ConfigManager based on @jeffsee55  Feedback.

kldavis4 (approved):

PR #3840, opened 2023-04-21
Kldavis4/feat/datalayer folders update kitchen sink

logan-anderson:
Updates kitchen skink example to handle folders for post and documentation collection.


kldavis4 (approved):
thanks!

PR #3842, opened 2023-04-21
refactor: pull out the lookup map creation so the database / bridge don't need to be concerned with it

kldavis4:
Builder keeps an in memory lookup map. buildDotTinaFiles returns the lookup,  along with the schema files. codegen handles writing the config file (_lookup.json) (instead of having that handled separately in bridge / database)


logan-anderson (approved):
Thanks @kldavis4

PR #3855, opened 2023-04-24
fix: bug in auto-edit of single document collections

kldavis4:
Fixes issue where auto-edit of single document collections was not updated to use the new `/collections/edit` route

### DIFF
--- a/packages/tinacms/src/admin/components/GetCollection.tsx
+++ b/packages/tinacms/src/admin/components/GetCollection.tsx
@@ -10,7 +10,6 @@ import { FilterArgs, TinaAdminApi } from '../api'
 import LoadingPage from '../components/LoadingPage'
 import type { CollectionResponse } from '../types'
 import { FullscreenError } from './FullscreenError'
-import { handleNavigate } from '../pages/CollectionListPage'
 
 export const useGetCollection = (
   cms: TinaCMS,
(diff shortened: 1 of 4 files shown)


packages/tinacms/src/admin/components/GetCollection.tsx
```diff
@@ -32,8 +31,10 @@ export const useGetCollection = (
   const [resetState, setResetSate] = useState(0)
 
   useEffect(() => {
+    let cancelled = false
```
jeffsee55, line 34: This makes sense but just curious why it's necessary in the scope of this issue? Are we not handling a race condition properly somewhere?

kldavis4: I think it probably snuck in when this feature (the redirect to the edit page of a single item collection) was first implemented. There is a `Can't perform a React state update on an unmounted component` warning in the console without this, because I think it is still executing when it navigates from the collection list url to the edit url. I thought it would be good to clean up here along with this.

jamespohalloran (changes requested):
I think this is still an issue in the document-list for documents using the basic editor:
https://www.loom.com/share/26bb3e40b3104b76b9b38a715edd8f0e

kldavis4:
@logan-anderson do you want to update the changelog to encompass all the bug fixes in this PR (maybe make it more generic)?

packages/tinacms/src/admin/components/GetCollection.tsx
jeffsee55: Ah great, I've seen that error running around too.

jamespohalloran (approved):
Working fine for me now!

PR #3861, opened 2023-04-25
fix issue with backwards navigation in a collection update page

logan-anderson:
cc @kldavis4 

Fixes https://github.com/tinacms/tinacms/pull/3855#issuecomment-1521866712

### DIFF
--- a/packages/tinacms/src/admin/pages/CollectionUpdatePage.tsx
+++ b/packages/tinacms/src/admin/pages/CollectionUpdatePage.tsx
@@ -1,7 +1,3 @@
-/**
-
-*/
-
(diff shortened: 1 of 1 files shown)


packages/tinacms/src/admin/pages/CollectionUpdatePage.tsx
```diff
@@ -153,7 +148,7 @@ const RenderForm = ({
           <div className="mb-2">
             <span className="block text-sm leading-tight uppercase text-gray-400 mb-1">
               <Link
-                to={`/collections/${collection.name}${folder.parentName}`}
```
logan-anderson, line 156: @kldavis4 Everything in folder was empty because there is no `~` in the edit link

PR #3872, opened 2023-04-27
fix issue with build.host in defineConfig

### LINKED ISSUE #3871: Add back `host` option
Hey there, I'm late to the party - sorry for that! Never had my hands on tinacms before the migration from forest.io

While attempting the migration, an issued occurred which - I think - originated in this pull request. Previously there was an Option to expose the Vite server externally (introduced by #3484 by @logan-anderson) - which doesn't seem to be there anymore.

When we run `tinacms dev --verbose -c nuxt` (which is configured to listen to Port `3000`), I have the following processes on those open ports in the container:

```
# netstat -nltp
Active Internet connections (only servers)
Proto Recv-Q Send-Q Local Address           Foreign Address         State       PID/Program name
tcp        0      0 127.0.0.11:43609        0.0.0.0:*               LISTEN      -
tcp        0      0 0.0.0.0:3000            0.0.0.0:*               LISTEN      68/node
tcp        0      0 127.0.0.1:4001          0.0.0.0:*               LISTEN      31/node
tcp6       0      0 :::9000                 :::*                    LISTEN      31/node
```

to toy around with and potentially get tinacms up and running, we need the `4001` port configured the same way. I see how the default `4001` port gets passed down in the `dev-command` ([reference](https://github.com/tinacms/tinacms/blob/a982b105d685fc343b989462c882d56c72a04007/packages/%40tinacms/cli/src/next/commands/dev-command/index.ts#L195)) - but the `host` for [Vite needs `server.host` to be passed as part of their Configuration Object](https://vitejs.dev/config/server-options.html#server-host).

So far for my analysis - any suggestions on how to take it further @jeffsee55? I'm not very familiar with tinacms, but would like to help .. if you let me know _how_. 

Thanks
-Stefan

_Originally posted by @steffkes in https://github.com/tinacms/tinacms/issues/3659#issuecomment-1520077320_

logan-anderson:
Fixes  #3871

Fixes regression where the host from config was not getting passed to vite. 

cc @wommy @steffkes

### DIFF
--- a/.changeset/witty-timers-attack.md
+++ b/.changeset/witty-timers-attack.md
@@ -1,0 +1,5 @@
+---
+'@tinacms/cli': patch
+---
+
+Fix regression where `build.host` was not passed correctly
--- a/packages/@tinacms/cli/src/next/vite/index.ts
+++ b/packages/@tinacms/cli/src/next/vite/index.ts
@@ -106,6 +106,7 @@ export const createConfig = async ({
       include: ['react/jsx-runtime', 'react/jsx-dev-runtime'],
     },
     server: {
+      host: configManager.config?.build?.host ?? false,
       watch: noWatch
         ? {
             ignored: ['**/*'],


kldavis4 (approved):

wommy:
im just confused because theres no [build.host](https://vitejs.dev/config/build-options.html)
yet theres a [server.host](https://vitejs.dev/config/server-options.html#server-host) and a [preview.host](https://vitejs.dev/config/preview-options.html#preview-host)

logan-anderson:
@wommy The Tina Config is different from the vite config. The `build.host` tina config sets the `server.host` vite config. 

We may consolidate these configs in the future to allow someone to override any part of the vite config but right now we only allow to override certain parts of the vite config.

Issue #3870, opened 2023-04-27
Collection templates stopped working after updating from 1.4.6 to 1.5.0


kslstn:
### Describe the issue that you're seeing. Any Loom videos or screenshots usually help a lot!

For one collection, I've defined templates [as described in the docs](https://tina.io/docs/reference/templates/#example-using-templates-instead-of-fields).

Tina CMS builds the dev server without error, but when I use the admin to navigate to the collection, it won't load, saying "Looks like there's nothing to edit on this page." I can't create new pages there either.

When I look at the browser I see the following errors:
Uncaught TypeError: fields is undefined [...]
The above error occurred in the <GetCollection> component: [...]

### Reproduction

https://github.com/kslstn/tina-issues/tree/templates

### Steps to reproduce

Update:

dependencies:
- tinacms 1.4.6
+ tinacms 1.5.0

devDependencies:
- @tinacms/cli 1.5.6
+ @tinacms/cli 1.5.7

Then define a template [as described in the docs](https://tina.io/docs/reference/templates/#example-using-templates-instead-of-fields).

### System Info

```shell
System:
    OS: macOS 13.3.1
    CPU: (4) x64 Intel(R) Core(TM) i5-7500 CPU @ 3.40GHz
    Memory: 322.47 MB / 16.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 16.14.2 - /usr/local/bin/node
    npm: 8.5.0 - /usr/local/bin/npm
  Browsers:
    Chrome: 112.0.5615.137
    Edge: 112.0.1722.58
    Firefox: 112.0.1
    Safari: 16.4
    Safari Technology Preview: 16.4
  npmPackages:
    @tinacms/cli: ^1.5.7 => 1.5.7 
    tinacms: ^1.5.0 => 1.5.0
```

### Validations

- [X] Follow our [Code of Conduct](https://github.com/tinacms/tinacms/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [docs](https://tina.io/docs).
- [X] Check that there isn't [already an issue](https://github.com/tinacms/tinacms/issues) that reports the same bug.

logan-anderson:
@kslstn I was not able to reproduce this issue but I did notice a couple of other issues.

You can see this video for more info: https://www.loom.com/share/d28ec20d43f94f22a243953aa028ccc8

kslstn:
[...]

PR #3881, opened 2023-04-28
Fix Database.get on folders within collections that have templates

kldavis4:
The hydrator in the database was failing on folders when calling getCollectionAndTemplateByFullPath() when templates are defined on a collection. This was being surfaced as a TinaQueryError but the folder object was present, just missing the _templates key.


logan-anderson (approved):
Does not fix https://github.com/tinacms/tinacms/issues/3870#issuecomment-1527789152 but looks good!

PR #3894, opened 2023-05-02
fix: folder root collection issues

kldavis4:
- glob was matching "-[ext]". A script named `node_modules/.bin/concat-md` was matching when a collection was configured at the root which was not a valid markdown. Updated to match literally ".[ext]".
- When configured with a root level collection, tina was scanning node_modules in dev mode and was causing a `ENAMETOOLONG` error. This change excludes node_modules folders entirely.
- the folder construction logic doesn't properly handle root level paths causing a '.' folder to be created instead of '' (root)

### DIFF
--- a/.changeset/large-kids-sin.md
+++ b/.changeset/large-kids-sin.md
@@ -1,0 +1,5 @@
+---
+'@tinacms/graphql': patch
+---
+
+Fix root level collection issues including a folder related bug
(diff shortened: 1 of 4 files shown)


logan-anderson (approved):

jamespohalloran (approved):
Tested on a local site and seems to be working well!

PR #3901, opened 2023-05-03
feat: add support for filtering on list fields

kldavis4:
See #3897

### DIFF
--- a/packages/@tinacms/graphql/src/database/datalayer.ts
+++ b/packages/@tinacms/graphql/src/database/datalayer.ts
@@ -5,6 +5,7 @@
 import { JSONPath } from 'jsonpath-plus'
 import sha from 'js-sha1'
 import {
+  ARRAY_ITEM_VALUE_SEPARATOR,
   BatchOp,
   CONTENT_ROOT_PREFIX,
   INDEX_KEY_FIELD_SEPARATOR,
@@ -30,6 +31,7 @@ export type BinaryFilter = {
   operator: OP.EQ | OP.GT | OP.LT | OP.GTE | OP.LTE | OP.STARTS_WITH | OP.IN
   type: string
   pad?: PadDefinition
+  list: boolean
 }
 export type TernaryFilter = {
   pathExpression: string
@@ -39,12 +41,14 @@ export type TernaryFilter = {
   rightOperator: OP.LT | OP.LTE
   type: string
   pad?: PadDefinition
+  list: boolean
 }
 export type IndexDefinition = {
   fields: {
     name: string
     type?: string
     pad?: PadDefinition
+    list: boolean
   }[]
 }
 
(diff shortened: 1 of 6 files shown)


packages/@tinacms/graphql/src/database/datalayer.ts
```diff
@@ -132,14 +136,26 @@ const makeKeyForField = <T extends object>(
...
       // TODO I think these dates are ISO 8601 so I don't think we need to convert to numbers
       const rawValue = data[field.name]
-      const resolvedValue = String(
-        field.type === 'datetime'
-          ? new Date(rawValue).getTime()
-          : field.type === 'string'
-          ? stringEscaper(rawValue as string | string[])
-          : rawValue
-      ).substring(0, maxStringLength)
-      valueParts.push(applyPadding(resolvedValue, field.pad))
```
kldavis4, line 142: Note that this change is basically just making explicit what was already happening when an array of strings was cast to a string (ie ['foo','bar'] -> 'foo,bar')

packages/@tinacms/graphql/src/database/datalayer.ts
```diff
@@ -214,6 +230,105 @@ export const coerceFilterChainOperands = (
   return result
 }
```
kldavis4, line 232: Extracting these functions out so they can be re-used for both the list true/false cases

logan-anderson:
The only thing that I was a bit confused on was the equal value being a string and not an array. 

I would expect `tags: { eq: ['foo', 'bar'] }` but we have `tags: {eq: "foo"}`. Not sure what the correct behaviour is but it did catch me off guard at first.

logan-anderson:
Side note: 

While testing I noticed this bug. Not sure if its related to this PR or not. We can probably don't need to generate filter args for type: 'rich-text'.

Screenshot 2023-05-03 at 2 00 24 PM

kingRayhan:
Is there any update on this PR?

kldavis4:
> Side note:
> 
> While testing I noticed this bug. Not sure if its related to this PR or not. We can probably don't need to generate filter args for type: 'rich-text'.
> 
> Screenshot 2023-05-03 at 2 00 24 PM

not related to this pr. currently it looks like string, reference, number, datetime and boolean are filterable.

logan-anderson:
> Can you elaborate on what you mean? If you had an array of values that you wanted to match against a list value, you'd probably use in and not eq.

What If I wanted to search all of the lists that had exactly ["foo, "bar"].

This might be a bit nuanced and I am totally fine leaving as is.

kldavis4:
> > Can you elaborate on what you mean? If you had an array of values that you wanted to match against a list value, you'd probably use in and not eq.
> 
> What If I wanted to search all of the lists that had exactly ["foo, "bar"].
> 
> This might be a bit nuanced and I am totally fine leaving as is.

I don't think that would be supported and not something I'd recommend supporting unless someone really had a good use case

logan-anderson (approved):

PR #3902, opened 2023-05-03
Support form deep-linking

jeffsee55:
Sorry for the wall of text, just want to leave this here for posterity.

The only user-facing changes we should see with this PR is the new "FormList" view and a slightly crumby animation when moving between form panels.

This PR does _not_ do anything for our quick editing feature that I've previewed, that will follow in another PR, and should be less invasive than this one.

## Introduces `cms.state` and `cms.dispatch`

[See code](https://github.com/tinacms/tinacms/pull/3902/files#diff-f7ac1376a8b7e12354116767332a22a8ff7a0471dbfe3b65bc62330f6fd1738cR23-R24)

Most of Tina is not reactive, and changing properties on classes always requires us to also set up subscriptions, or listen for events. As complexity grows, this has become pretty brittle, and pain points were felt while building out the logic for deep-linking. The need to effectively switch the "active" form meant that looking for it in `cms.forms` (a plugin) the values wouldn't always be there. This issue, along with many others necessitated a move towards state that is kept in React-land. 

I'm not sure exactly how far we want to lean into the top-level React state, but I think when top-level React state doesn't fit, local React state should, and we can probably start to treat the CMS object as more of a large config class that shouldn't be mutated once initialized (eg. move Events/Alerts and possible Forms into React state).

## `FormBuilder` handles responsibilities that `GroupForm`, `GroupListForm` and `Blocks` were handling

When you have nested properties, we've previously been mounting a React portal over top of the previous group of properties. There were performance issues with this approach, not to mention an endless battle for the highest z-index for each subsequent portal that was mounted (since they all mount at the same insertion point). But also, this strategy prevented us from being able to "deep-link" into a form. Say you want to see the fields at `blocks.0.cta.actions.2`. Previously we would need to mount and traverse through each layer of fields. Now, the `FormBuilder` is given the form object, and the active field name (`blocks.0.cta.actions.2`), and it grabs the appropriate fields. Deep-linking is trivial at this point and I expect some performance improvements, but the z-index issues for things that are still using the form portal like rich-text nested forms and the blocks selector have been avoided for now, since I think we're hoping to restructure things in a more sane manner in the near future.

cc @spbyrne I tried to keep _some_ animation intact, but it kind of sucks. I'm sure we can do something better but every time I tried it ended up being the beginnings of a large overhaul, which I want to avoid in this scope.

I've left much of the code in `GroupForm`, `GroupListForm` and `Blocks` untouched, just [commented out](https://github.com/tinacms/tinacms/pull/3902/files#diff-f972affe284a8c490486748af11bb1c33b6ee11379d127906afb616927ddbdd5R51) things that needed to be, hope to remove large chunks of it soon. 

## A new UI for the "FormList"

[See code](https://github.com/tinacms/tinacms/pull/3902/files#diff-16c2cada385faea4e6a18f445f9dad99417e3829bb85fff0826e2df3c36f7dc7R87)

Screenshot 2023-05-03 at 4 49 21 PM

This view is what you see when you have a list of forms and click the "Return to Form List" text above a form. Previously, it was something we'd mostly avoided, and moving forward, I dont expect it to get much attention. But it now shows all forms on the page, and keeps track of where they are in the query hierarchy. It's not the prettiest, but at least it should help people navigate once we land "quick editing". This view is updated from the `graphql-reducer`, and managing the hierarchy is separate from managing the forms themselves, both info is stored in the top-level `cms.state`.


logan-anderson:
Noticed a UX issue/regression: https://www.loom.com/share/e4ebe6030dff4b639947fc105dcbda4e

logan-anderson:
Contextual editing is broken if you go to a document with an other -> edit in CMS -> go to a different collection 

See: https://www.loom.com/share/0f275794279e490c96eff7e2400c86d4

logan-anderson:
@jeffsee55  Seeing an issue with drag and drop

Note: this is **not** happening on main

https://www.loom.com/share/465f8a29e38a42e0a8027c4b4efa8719

packages/@tinacms/toolkit/src/packages/form-builder/FormBuilder.tsx
```diff
@@ -184,15 +186,28 @@ export const FormBuilder: FC<FormBuilderProps> = ({
...
-
               <FormPortalProvider>
-                <FormWrapper id={tinaForm.id}>
-                  {tinaForm && tinaForm.fields.length ? (
-                    <FieldsBuilder form={tinaForm} fields={tinaForm.fields} />
-                  ) : (
-                    <NoFieldsPlaceholder />
-                  )}
-                </FormWrapper>
+                <Transition
```
jeffsee55, line 190: The breaking drag-n-drop is due to this element. Will look into another solution

jeffsee55:
Thanks @logan-anderson - this was due to the `<Transition>` component, which DnD doesn't like https://github.com/atlassian/react-beautiful-dnd/issues/128. 

cc @spbyrne the `<Transition>` I added sucked, and now knowing that it breaks other stuff I'm inclined to leave animations out of this PR altogether and we can tackle it in the future.

logan-anderson (approved):
Not seeing any issues!

PR #3903, opened 2023-05-04
Fix/single doc collection navigation

logan-anderson:
Fixes issue where single document collections would not open contextual editing.


kldavis4:
@logan-anderson what is the __typename that is not Document & not Folder?

logan-anderson:
> @logan-anderson what is the __typename that is not Document & not Folder?


While looking at it `__typename` appeared to be the name of the document type in came cases. For example "Post" or "Global".

kldavis4 (approved):

PR #3922, opened 2023-05-09
Add placeholder screen to admin

jamespohalloran:
Add admin placeholder screen if assets fail to load.

Screen Shot 2023-05-09 at 2 37 52 PM

closes #ENG-972


jeffsee55 (approved):
yea, that `setTimeout` seems good enough to me!

PR #3923, opened 2023-05-09
fix: Bitwise OR operator to Logical OR operator in media store limit

jasperpcode:
Because of the use of a Bitwise OR operator instead of a Logical OR operator, unexpected behavior occurred when changing the pageSize of the media manager. This PR fixes this operator so it works as expected.

Before:
Screenshot from 2023-05-09 16-25-00

After:
Screenshot from 2023-05-09 16-27-26


logan-anderson (commented):
This looks good to me! Unless there was any reason for the logical or @kldavis4 ?

kldavis4 (approved):

PR #3929, opened 2023-05-10
Add media root to s3

logan-anderson:
Adds the `mediaRoot` option to the s3 media store. 

EX
```ts
// pages/api/s3/[...media].ts

import {
  mediaHandlerConfig,
  createMediaHandler,
} from 'next-tinacms-s3/dist/handlers'

import { isAuthorized } from '@tinacms/auth'

export const config = mediaHandlerConfig

export default createMediaHandler({
  config: {
    credentials: {
      accessKeyId: process.env.NEXT_PUBLIC_S3_ACCESS_KEY || '',
      secretAccessKey: process.env.S3_SECRET_KEY || '',
    },
    region: process.env.NEXT_PUBLIC_S3_REGION,
  },
  mediaRoot: 'tina',
  bucket: process.env.NEXT_PUBLIC_S3_BUCKET || '',
  authorized: async (req, _res) => {
    if (process.env.NODE_ENV === 'development') {
      return true
    }
    try {
      const user = await isAuthorized(req)

      return user && user.verified
    } catch (e) {
      console.error(e)
      return false
    }
  },
})
```

### DIFF
--- a/.changeset/soft-ladybugs-joke.md
+++ b/.changeset/soft-ladybugs-joke.md
@@ -1,0 +1,5 @@
+---
+'next-tinacms-s3': patch
+---
+
+Add the `mediaRoot` option to the s3 media store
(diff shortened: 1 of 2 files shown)


kldavis4:
See #3590 for comparison

kldavis4 (approved):

PR #3911, opened 2023-05-04
Add support for sites deployed to sub-paths

jeffsee55:
Closes https://github.com/tinacms/tinacms/issues/3906

~~TODO: Check links work correctly in the admin~~

PR #3932, opened 2023-05-11
Add basePath prefix to files in media manager

Albert-Mv:
In addition to [this](https://github.com/tinacms/tinacms/pull/3911) PR I made some changes to media manager to add basePath support.

Before these changes files were loaded in the media manager without the basePath prefix. Links to images were incorrect.
Now, if we have basePath, links will have prefix.

### DIFF
--- a/.changeset/neat-candles-greet.md
+++ b/.changeset/neat-candles-greet.md
@@ -1,5 +1,0 @@
----
-'@tinacms/graphql': patch
----
-
-Fix error "Cannot return null for non-nullable field" on boolean required fields
--- a/examples/basic-iframe/CHANGELOG.md
+++ b/examples/basic-iframe/CHANGELOG.md
@@ -4,6 +4,14 @@
 
 ### Patch Changes
 
+- Updated dependencies [e731ab0c5]
+  - @tinacms/graphql@1.4.14
+  - @tinacms/datalayer@1.2.14
+
+## null
+
+### Patch Changes
+
 - Updated dependencies [ca74add40]
 - Updated dependencies [5a6018916]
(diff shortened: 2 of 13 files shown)


logan-anderson:
@Albert-Mv what is the difference between this and `mediaRoot`?

jamespohalloran (approved):

logan-anderson (changes requested):
@Albert-Mv They way this is working right now it will not work in production. 

See this video for a bit more context:  https://www.loom.com/share/d17dc35ef23743ac877761a149987948

We are happy to take it over from here since it will be a bit harder to test. But if you want to take it on, your more then welcome.

Albert-Mv:
@logan-anderson thank you for the feedback. I got it, I'll modify this two functions to add Tina Cloud url support

Albert-Mv:
> @Albert-Mv They way this is working right now it will not work in production.
> 
> See this video for a bit more context: https://www.loom.com/share/d17dc35ef23743ac877761a149987948
> 
> We are happy to take it over from here since it will be a bit harder to test. But if you want to take it on, your more then welcome.

Could you please help me a bit: I can't understand how did you run admin panel on localhost in production mode (I don't see the alert "You are currently in Local mode")

logan-anderson:
@Albert-Mv its a bit award right now but the best way is to create a new Tina Cloud project with the tina-cloud-starter. Set that up in app.tina.io and then add the token, clientId and branch to the examples/tina-cloud-starter in the mono repo. 

Then you can run `pnpm build` and `pnpm start`. 

However for testing this feature in particular the site has to use `basePath`. You may have to setup some nextjs redirects to ensure that works.

Issue #3609, opened 2023-02-13
Better support self hosted Indexing and bridge.


logan-anderson:
> 1. What would happen for projects that use Tina's new feature for separate content repositories? 

This is a great question. This was not really taken into consideration when doing our first version of this. 

For context, behind the scene TinaCMS  uses a "Bridge" to get the content for indexing. For self hosted we use the filesystem bridge by default  but for this use case we would need a "Github Bridge" we had one at one point but I think it was removed from the codebase. We can look at adding it back to the codebase for those who want to use a separate content repo.

Although its not available right now I think that this would look something like this.

```ts
export default createDatabase({
  bridge: new GithubBridge({owner, ref, repo, token }),
  level: isLocal ? localLevelStore : mongodbLevelStore,
  onPut: isLocal ? localOnPut : githubOnPut,
  onDelete: isLocal ? localOnDelete : githubOnDelete,
});
```

This would index your content from another source but would still do it only at build time.

> What would be the best process to ensure re-indexing occurs in that case?

> A)

If you editing from TinaCMS, this should not be an issue because the Datalayer should automatically be updated on save. But if you push commits to the content repo yourself (not through TinaCMS) that could cause an issue. The Datalayer would not know about these changes and would not be updated
 
Again this was not prioritized, so there won't be any examples of how to do it but I can provide an overview of how we could support this.

I think the could look something like this

Commit in content repo -> Github action to get all changed files in content repo -> Sends updated files to a api route in your main repo. 

The api route would look something like this.

```ts
import { NextApiHandler } from "next";

import database from "../../.tina/database";

const indexHandler: NextApiHandler = async (req, res) => {
  // use your own auth logic
  const isAuthorized = true;
  if (!isAuthorized) {
    // get the paths from the request body instead of hardcoding them
    const paths: string[] = ["content/home.json"];
    await database.indexContentByPaths(paths);
    return res.status(200).json({ success: true });
  } else {
    res.status(401).json({ success: false });
  }
};

export default indexHandler;

```

This works right now but unfortunately currently looks at the filesystem which is not super helpful (and would not work for separate content repos)

I made a quick demo video that may help clarify: https://www.loom.com/share/c7d1c4f5bda04ec3a1dacff2a03efac1 
 

> B)

Reindexing the entire repo is supported but I was looking at the code and I think there is some cleanup required to make it easier for folks who are self hosting.

I think the API should look like this (not currently supported)

```ts
import { NextApiHandler } from "next";

[...]

PR #3936, opened 2023-05-15
Recreate GithubBridge to support separate content repos in self-hosted projects

coreyaus:
## Summary

- This adds a `GithubBridge` in order to better support separate content repositories for self-hosted projects
- This will also help provide a path to better support reindexing for self-hosted projects (i.e. keeping the data layer updated when changes are pushed directly via git, rather than made through the Tina UI)

This PR draws heavily from the original `GithubBridge` that previously existed - i.e. [this historical file](https://github.com/tinacms/tinacms/blob/930260c839cbd1908e9e6902734c94f6d64d3282/packages/%40tinacms/datalayer/src/database/bridge/github.ts) that was removed in [this commit](https://github.com/tinacms/tinacms/commit/fb74f4a129164e3b40ff7cd38928682dadab945c), so full credit to whoever wrote the original implementation!

## Background

- See the excellent information @logan-anderson provided in this discussion thread: https://github.com/tinacms/tinacms/discussions/3589#discussioncomment-4960323
- And the further details in this issue: https://github.com/tinacms/tinacms/issues/3609, including this video overview of the key components to consider: https://www.loom.com/share/c7d1c4f5bda04ec3a1dacff2a03efac1
- Follows on from the updates in this PR: #3830

### DIFF
--- a/.changeset/lovely-forks-talk.md
+++ b/.changeset/lovely-forks-talk.md
@@ -1,0 +1,7 @@
+---
+'@tinacms/datalayer': patch
+'@tinacms/graphql': patch
+'@tinacms/cli': patch
+---
+
+This adds a GithubBridge in order to better support separate content repositories for self-hosted projects. This will also help provide a path to better support reindexing for self-hosted projects (i.e. keeping the data layer updated when changes are pushed directly via git, rather than made through the Tina UI). See further notes in this discussion: https://github.com/tinacms/tinacms/discussions/3589#discussioncomment-4960323 and this issue: https://github.com/tinacms/tinacms/issues/3609
--- a/packages/@tinacms/cli/src/next/database.ts
+++ b/packages/@tinacms/cli/src/next/database.ts
@@ -54,7 +54,10 @@ export async function createAndInitializeDatabase(
     configManager.config.contentApiUrlOverride
   ) {
     database = (await configManager.loadDatabaseFile()) as Database
-    database.bridge = bridge
+    // Use the bridge specified in the database file if one exists
+    if (!database.bridge) {
+      database.bridge = bridge
+    }
   } else {
     if (
       configManager.hasSelfHostedConfig() &&
--- a/packages/@tinacms/github-bridge/README.md
+++ b/packages/@tinacms/github-bridge/README.md
@@ -1,0 +1,15 @@
+# TinaCMS GitHub Bridge
+
+This package provides a `GithubBridge` for [TinaCMS](https://tina.io/) in order to better support separate content repositories for self-hosted projects.
+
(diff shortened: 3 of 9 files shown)


Perni1984:
@coreyaus many thanks for putting this together, it will help us a lot as we're currently struggling with this topic.

logan-anderson (changes requested):
Hey @coreyaus changes look good! Thanks for this. Just one small thing. Can the `GithubBridge` be moved to its own package?

This is so that folks who are not using it do not have to install @octokit/rest.

The package could be named `@tinacms/github-bridge`

If you don't want to move it we would be happy to (it just could take a bit as we are working on some other things at the momant ).

If you want to do this yourself feel free. The general steps are
1. Create a new folder called `packages/@tinacms/github-bridge`
2. Copy the package.json from `@tinacms/graphql` and remove all deps that are not needed. Make sure to keep `@tinacms/scripts`, `typescript` and `jest`. 
3. update the version in the package.json to 0.0.1 and the name to  `@tinacms/github-bridge` 
4. Create a `src/index.ts` file and add github bridge to it

coreyaus:
Thanks for the quick review @logan-anderson and great to hear it'll be useful @Perni1984!

That makes sense about moving this to its own package to avoid the `@octokit/rest` dependency for users who don't need it - I'm definitely happy to give that a go and should hopefully be able to take a look at it this weekend 👍 I'll post an update when there's a revised version you can look at, or if I have any questions along the way.

Cheers gents, and have a great day!

coreyaus:
Thanks for the early feedback and guidance on this PR. The `github-bridge` worked well but the main issue I ran into when I created it a couple of months ago was that handling the sync to GitHub via the `put` function in the `github-bridge` made the `onPut` and `onDelete` handlers in the `database.ts` file redundant. When I initially ran the tinacms build from this PR locally in my test project I saw 2 commits for each saved change - the commit with the changes made by the `github-bridge` and an empty commit made via the `onPut` handler in the `database.ts` file.

To be honest I thought I'd written at the time to get feedback on how to handle that behaviour but obviously dreamt that so apologies I didn't provide an update earlier! I've been following some of the recent draft PRs related to self-hosting improvements (e.g. #4095 and #4106). It looks like those will address that point above and also add a bunch of powerful features so I will close this PR now, given it will be superseded when #4106 is merged to `main`

Let us know if there's any useful way we can contribute to that current work on self-hosted improvements (e.g. testing, docs, etc). In the meantime l'll give the latest code on the `feat/self-hosted-improvements` branch a spin locally - looks like that's possible by installing version `0.0.0-20230901181730` of the core packages.

Happy Friday, and thanks for all the epic work you're doing!

coreyaus:
Hey all - just a quick follow up that now that the latest self-hosting features have been officially launched I've created a new PR around this same concept of re-introducing the GitHubBridge to support self-hosted repositories with separate content repositories. See the full details in PR #4384. Happy Friday!

PR #3943, opened 2023-05-17
Remove use of GlobalForm plugin

jeffsee55:
Remove use of GlobalForm plugin, instead just using cms.state.forms to allow clicking on a global form in the nav, which will now just open the form in the sidebar

https://www.loom.com/share/df8533ed5d6846c6a465126897d7ed18

Fixes issue @logan-anderson found. 

TODO - a lot of cleanup if we're no longer relying on Global Forms, but will follow up with that in another PR


logan-anderson:
@jeffsee55 Getting this error when running yarn dev in the tina-cloud-starter and going into the admin
`Cannot destructure property 'setMenuIsOpen' of 'React.useContext(...)' as it is null.`

Screenshot 2023-05-18 at 12 19 58 PM

PR #3949, opened 2023-05-18
feat: search

kldavis4:
### Summary

This PR implements search functionality using a library called `search-index`

### Changes

- Adds automatic search indexing on CLI build when search is configured. Use `--skip-search-index` to disable indexing
- Adds `search-index` CLI command for building a new search index
- Adds automatic local searching indexing to dev CLI
- Adds /searchIndex routes to local dev api providing query, delete and put documents into search index
- Implements [SearchIndexer](https://github.com/tinacms/tinacms/blob/kldavis4/feat/search/packages/%40tinacms/search/src/indexer/index.ts#LL22C16-L22C16) class in @tinacms/search which is used by the CLI to build search indexes. This class will scan the local filesystem and then transform each document into a form that can be used for search indexing. This is the same form returned by the graphql resolver. Note that during search indexing, I chose to avoid using the database directly (to avoid the unnecessary local indexing), but this requires quite a bit of refactoring in @tinacms/grapqhl to export functions used by both the Database and the SearchIndexer (which accounts for a significant chunk of the changes in this PR)
- Add `searchable` property to fields to allow disabling indexing of specific fields. By default, all fields are indexed, except for 'image' types
- Add `maxSearchIndexFieldLength` on text type fields to change how much of a long form text field is used for indexing. By default, this length is 100 characters. It can be configured both globally, and on a per field basis using this property.
- Add `search` property to config. Under this property, `search.searchClient` allows a custom search client implementation, `search.indexBatchSize` controls the number of documents per indexer request (default 100), `maxSearchIndexFieldLength` controls the default string length to index (default 100). `search.tina.indexerToken` allows the search index token to be set. `search.tina.stopwordLanguages` allows additional [stopword](https://github.com/fergiemcdowall/stopword) languages to be specified (default `['eng']`)
- Implement LocalSearchIndexClient used by the SearchIndexer to perform local indexing in dev mode
- Implement TinaCMSSearchIndexClient used by the SearchIndexer to perform indexing to tina cloud in non-dev-mode
- Implement frontend search index client which is used to send requests to the search api
- Updates admin UI to add search functionality (including search index updates / deletes)


packages/@tinacms/search/src/client/index.ts
```diff
@@ -0,0 +1,147 @@
...
+    // @ts-ignore
+    this.searchIndex = await si({
+      db: this.memoryLevel,
+      stopwords: lookupStopwords(this.stopwordLanguages),
+    })
+  }
+
+  async onFinishIndexing() {
+    const headers = new Headers()
+    headers.append('x-api-key', this.indexerToken || 'bogus')
```
kldavis4, line 85: remove default

logan-anderson (approved):
Looks good to me. I was able to test locally and with Tina Cloud. I also tested the new command. Everything worked as expected.

packages/tinacms/src/auth/TinaCloudProvider.tsx
```diff
@@ -170,6 +175,30 @@ export const TinaCloudProvider = (
...
+      } else {
+        searchClient = props.schema.config?.search?.searchClient
+      }
+    }
+
+    if (searchClient) {
+      cms.registerApi('search', searchClient)
+    }
+  }
+  setupSearch(props)
```
jamespohalloran, line 200: Should this be in a useEffect? As it, it will be recalled anytime this is rerendered

packages/tinacms/src/admin/components/GetCollection.tsx
```diff
@@ -66,10 +67,85 @@ export const useGetCollection = (
...
+        setLoading(false)
+      }
+    }
+
     if (cancelled) return
 
     setLoading(true)
-    fetchCollection()
+    if (search) {
+      searchCollection()
```
jamespohalloran, line 145: Not sure I understand what we're doing here. Won't this keep looping

packages/tinacms/src/auth/TinaCloudProvider.tsx
jamespohalloran, line 200:
```suggestion
  React.useEffect(() => {
  setupSearch(props)
  },[]);
```

packages/tinacms/src/auth/TinaCloudProvider.tsx
```diff
@@ -170,6 +175,30 @@ export const TinaCloudProvider = (
...
+    } else {
+      const hasTinaSearch = Boolean(props.schema.config?.search?.tina)
+      if (hasTinaSearch) {
+        searchClient = new TinaCMSSearchClient(cms.api.tina)
+      } else {
+        searchClient = props.schema.config?.search?.searchClient
+      }
+    }
+
+    if (searchClient) {
```
jamespohalloran, line 196: Do we need this check?

packages/tinacms/src/admin/components/GetCollection.tsx
```diff
@@ -66,10 +67,85 @@ export const useGetCollection = (
       }
     }
 
+    const searchCollection = async () => {
```
jamespohalloran, line 70: There's kind of a lot going on inside this useEffect. Might make sense to pull out some of these functions so we don't have to keep declaring them.

jamespohalloran (changes requested):

packages/tinacms/src/auth/TinaCloudProvider.tsx
kldavis4: I think they specify searchClient (and not tina) but leave it undefined, `searchClient` could end up null.

packages/tinacms/src/auth/TinaCloudProvider.tsx
kldavis4: yeah.. I'm not sure why useEffect isn't used in here. I updated this to put the search initialization in a useEffect as suggested: https://github.com/tinacms/tinacms/pull/3949/commits/43e928114e313b3032e44a9ff006bd4462baec1f

packages/tinacms/src/admin/components/GetCollection.tsx
kldavis4: I don't _think_ it will loop - its not changing any of the useEffect dependencies

packages/tinacms/src/admin/components/GetCollection.tsx
kldavis4: I split the hooks up which hopefully makes things clearer

packages/tinacms/src/admin/components/GetCollection.tsx
jamespohalloran: Oops! My bad, thought we were calling searchCollection recursively

jamespohalloran (approved):

PR #3950, opened 2023-05-19
Add basepath support in media manager

Albert-Mv:

### DIFF
--- a/.changeset/modern-bears-wash.md
+++ b/.changeset/modern-bears-wash.md
@@ -1,0 +1,7 @@
+---
+'@tinacms/datalayer': patch
+'@tinacms/cli': patch
+'@tinacms/graphql': patch
+---
+
+Adds basepath to files in media manager
--- a/packages/@tinacms/cli/src/next/commands/dev-command/server/media.ts
+++ b/packages/@tinacms/cli/src/next/commands/dev-command/server/media.ts
@@ -98,6 +98,7 @@ interface ListMediaRes {
   error?: string
 }
 export interface PathConfig {
+  basePath: string
   rootPath: string
   apiURL: string
   publicFolder: string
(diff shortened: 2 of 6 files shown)


logan-anderson (changes requested):
@Albert-Mv Looking good overall just a couple of things.

1. I am getting this error when running `yarn dev` in the `experimental-examples/tina-cloud-starter`

```
TypeError: Cannot read properties of undefined (reading 'basePath')
```

2. Can you add a changeset by running `pnpm changeset` ? I think this would be a patch update for all effected packages. 

I have not yet tested in Tina Cloud but this is what I saw initially. Thanks again for the contribution.

Albert-Mv:
I tried to reproduce the error different ways, but couldn't do it
`TypeError: Cannot read properties of undefined (reading 'basePath')`

I also changed the files on the server side - maybe this is the problem (although I cannot check this code on my computer without a server)

If you can provide more information, I would appreciate it and try to reproduce again

logan-anderson:
> I tried to reproduce the error different ways, but couldn't do it
> 
> ```
> TypeError: Cannot read properties of undefined (reading 'basePath')
> ```
> 
> I also changed the files on the server side - maybe this is the problem (although I cannot check this code on my computer without a server)
> 
> If you can provide more information, I would appreciate it and try to reproduce again Probably I missed somethig in pipeline you do. I use node v20, clean fork from tinacms(with only my changes) and also try it in the `experimental-examples/tina-cloud-starter`

Hmm not seeing the error anymore. Something must have not been build properly. Sorry about that, ill let you know if I see it again.

logan-anderson:
@Albert-Mv  Did some more digging around and testing with Tina Cloud and here is how to reproduce the error in the mono repo. 

Video: https://www.loom.com/share/d80419d63598496ba57c6f5858186093

This is the code you have to replace

```
config: {
         assetsHost: 'assets.example.com',
         clientId: 'test-client-id',
         useRelativeMedia: false,
},
```

This is the file: https://github.com/tinacms/tinacms/blob/main/packages/@tinacms/cli/src/next/vite/plugins.ts

PR #3957, opened 2023-05-19
fix: #3952: use uri decoder on folder name from url

### LINKED ISSUE #3952: Problem with special chars in content subfolder names.
Possibly only an issue on Windows.

*What Happens*
Sub folders do not support certain characters. Most importantly `space` chars. If a sub folder has a space in the folder name then its contents are displayed empty.

Screenshot from the reproduction repo of "Page Contents" section/collection:
image

Screenshot of the contents of `sub folder`:
image

Same issue happens when emojies are used. I havn't tested other special chars, but `-` and `_` are working fine.

### Reproduction

https://github.com/arthuratmeetbreeze/tinacms-subfolder-issue-reproduction

### Steps to reproduce

1. Clone reproduction repo
2. Run `yarn install`
3. Run `yarn dev`
4. Navigate to Admin
5. Navigate to `Page Content` collection
6. Open `sub folder` or `👍subfolder` (contents are empty) 
7. Open `subfolder`, `sub-folder` or `sub_folder` (contents are loaded properly) 

### System Info

```shell
System:
    OS: Windows 10 10.0.22621
    CPU: (16) x64 AMD Ryzen 7 6800H with Radeon Graphics
    Memory: 4.98 GB / 15.26 GB
  Binaries:
    Node: 18.14.2 - C:\Node\node.EXE
    npm: 9.5.0 - C:\Node\npm.CMD
  Browsers:
    Edge: Spartan (44.22621.1702.0), Chromium (113.0.1774.35)
    Internet Explorer: 11.0.22621.1
```

### Validations

kldavis4:
Fix #3952 

Special characters in the folder name from the url were url encoded, causing queries to use incorrect folder name

### DIFF
--- a/.changeset/red-peaches-move.md
+++ b/.changeset/red-peaches-move.md
@@ -1,0 +1,5 @@
+---
+'tinacms': patch
+---
+
+Fix to decode the folder name from the url
--- a/packages/tinacms/src/admin/pages/utils.tsx
+++ b/packages/tinacms/src/admin/pages/utils.tsx
@@ -32,9 +32,10 @@ export const useCollectionFolder = () => {
   useEffect(() => {
     // set folder using the pathname
     const match = loc.pathname.match(folderRegex)
+    const folderName = match ? decodeURIComponent(match[1]) : ''
     const update = {
-      name: match ? match[1] : '',
-      fullyQualifiedName: match ? (match[1] ? `~/${match[1]}` : '~') : '',
+      name: folderName,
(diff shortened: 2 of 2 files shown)


raakvlakraakt:
I think this is related. I have the problem that when there's a special character in the filename (ö), the url doesn't match and the page throws an error. When changing the 'ö' to a regular 'o' fixes the problem. When making a file, is it possible that the filename automatically removes special characters?

PR #3965, opened 2023-05-23
Index the content before a build when self hosting

logan-anderson:
When debugging some self hosting issues I noticed that we don't index the self hosted DB when building. We need to index on build so that they DB can get the newest markdown data. 

This may have been a regressing with our larger CLI PR a month or two ago.

### DIFF
--- a/packages/@tinacms/cli/src/next/commands/baseCommands.ts
+++ b/packages/@tinacms/cli/src/next/commands/baseCommands.ts
@@ -92,11 +92,14 @@ export abstract class BaseCommand extends Command {
     database,
     graphQLSchema,
     tinaSchema,
+    text,
   }: {
     database: Database
     graphQLSchema: DocumentNode
     tinaSchema: TinaSchema
+    text?: string
   }) {
+    const textToUse = text || 'Indexing local files'
     const warnings: string[] = []
     await spin({
       waitFor: async () => {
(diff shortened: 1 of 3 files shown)


packages/@tinacms/cli/src/next/commands/baseCommands.ts
```diff
@@ -106,7 +109,7 @@ export abstract class BaseCommand extends Command {
         })
         warnings.push(...res.warnings)
       },
-      text: 'Indexing local files',
```
kldavis4, line 109: @logan-anderson why are we distinguishing in the status text between 'local files' and 'self-hosted content'. It seems like in both cases, the content is local. Isn't it more about the destination? Like 'Indexing content to local data layer' & 'Indexing content to self-hosted data layer'?

logan-anderson: @kldavis4 I like that better! ill update it

logan-anderson: Just updated to "Indexing to self-hosted data layer" and "Indexing local content" what do you think?

kldavis4: sounds good

kldavis4 (approved):

PR #3970, opened 2023-05-25
add type: module to tinacms package

logan-anderson:
Adds `"type": "module"` to tinacms package.


jeffsee55:
What's the reason for this? I recall us getting some strange behavior when setting this before.

jeffsee55:
I think changing out build output from `.es.js` to `.mjs` would also fix this. I'll try to see if I can find the issue with `"type": "module"`, but I think it was related to the CLI. Changing the output extension would be pretty easy and I think someone else asked for that as well.

jeffsee55:
I ran into other issues moving `.es.js` to `.mjs` while testing, I honestly have no idea why/when it's supposed to one over the other and I think changing "type" to "module" is sometimes also not what we want, but that may have been before we updated the CLI. I didn't run into any errors with this PR so lets just move ahead with it, sorry for the delay

PR #3978, opened 2023-05-31
Fix issue where activeField search param would be called twice

jeffsee55:
Fix issue where activeField search param would be called twice, resulting in the incorrect form being focused.

### DIFF
--- a/packages/@tinacms/app/src/lib/graphql-reducer.ts
+++ b/packages/@tinacms/app/src/lib/graphql-reducer.ts
@@ -179,6 +179,8 @@ export const useGraphQLReducer = (
   >([])
   const [operationIndex, setOperationIndex] = React.useState(0)
 
+  const activeField = searchParams.get('active-field')
+
   React.useEffect(() => {
     const run = async () => {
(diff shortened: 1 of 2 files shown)


packages/@tinacms/app/src/lib/graphql-reducer.ts
```diff
@@ -435,10 +437,10 @@ export const useGraphQLReducer = (
             { id: payload.id, data: result.data },
           ])
         }
-        const activeField = searchParams.get('active-field')
         if (activeField) {
           setSearchParams({})
           const [queryId, eventFieldName] = activeField.split('---')
+          console.log('calling set active field name', eventFieldName)
```
kldavis4, line 443: keep this?

jeffsee55, line 443:
```suggestion
```

jeffsee55: Thanks, removed it

kldavis4 (approved):

PR #4014, opened 2023-06-16
Fix visual editing active form

jeffsee55:
Fixes 2 issues:

- The cleanup process during unmount didn't clear the active form.
- Also, when choosing an active form we were choosing the last form item instead of the first.


packages/@tinacms/cli/src/next/commands/dev-command/index.ts
```diff
@@ -70,70 +71,77 @@ export class DevCommand extends BaseCommand {
     const setup = async ({ firstTime }: { firstTime: boolean }) => {
```
jamespohalloran, line 71: Rather than pass in "firstTime", would it be cleaner to just check is database is undefined?

jeffsee55: Sorry this was mixed in with my branch, it's not part of the PR, will fix

jamespohalloran: Alternatively, if part of this is only executed on the first time, maybe we should split this into 2 functions? (`init`, and something else)?

PR #4029, opened 2023-06-22
Update deps 2

jeffsee55:
Notable changes

- Upgrade Plate from 14 to 21. These versions move quickly but there seemed to only be one breaking change that affected us (since we don't use Plate for UI) which was the deprecation of `getPreventDefaultHandler`
- ESLint upgrade 2 major versions. There were some autofixes, mostly changing `let` to `const` where possible
- `onClickOutside` was behaving strangely, I've run into issues with this module before and wasn't able to pin it down. Updated import path to explicitly use es.js https://github.com/tinacms/tinacms/pull/4029/files#diff-1fa4294738ff3e00628bdb4ccee48e78c0b9127f0d72ba9f0501467dfc401e70R12
- Upgrades to Jest 2 major versions


packages/@tinacms/app/package.json
```diff
@@ -9,6 +9,7 @@
     "typescript": "^4.6.4"
   },
   "dependencies": {
+    "@codemirror/language": "6.0.0",
```
kldavis4, line 12: was this meant to be added?

jeffsee55: Yeah it's used by the GraphiQL package, though they weren't declaring it properly

kldavis4 (approved):

kldavis4:
did some testing in the monorepo and changes seem good 👍🏻

Issue #3976, opened 2023-05-30
Nested objects in rich text templates showing blank UI


HenrySkup:
### Describe the issue that you're seeing. Any Loom videos or screenshots usually help a lot!

Hi, I have several templates within the TinaMarkdown component which have within them object fields.  These were working before, but on the latest release, when navigating the CMS to them, all I get is a blank sidebar.

<img width="1572" alt="Screenshot 2023-05-30 at 17 36 46" src="https://github.com/tinacms/tinacms/assets/5592703/5145349f-c0a1-46bd-9330-ebe571c6ece0">
<img width="1569" alt="Screenshot 2023-05-30 at 17 36 50" src="https://github.com/tinacms/tinacms/assets/5592703/945884b1-07b0-4209-9ae1-a991e6eb210b">
<img width="1564" alt="Screenshot 2023-05-30 at 17 36 53" src="https://github.com/tinacms/tinacms/assets/5592703/ab7dfb4c-5264-4c28-b774-83e2597f2dd9">
<img width="1573" alt="Screenshot 2023-05-30 at 17 36 55" src="https://github.com/tinacms/tinacms/assets/5592703/c3134d0a-f484-4ecf-a7a7-7b838a55c7e9">

### Reproduction

n/a

### Steps to reproduce

the page schema is :

```
...
fields: [
    { label : 'Title',
      name      : 'title',
      type      : 'string',
      isTitle   : true,
      required  : true 
    },
    { label: "Show Title",
      name: "showTitle",
      type: "boolean",
    },
    { label: "Main Content",
      name: "body",
      type: "rich-text",
      isBody: true,
      templates : [ 
        { label : 'Thing',
          name  : 'Thing',
          fields : [
            { label : 'Title',
              name      : 'title',
              type      : 'string',
            },
            { label : 'More Stuff',
              name      : 'stuff',
              type      : 'object',
              list : true,
              fields : [
                { label : 'A',
                  name      : 'a',
                  type      : 'string',
                },
                { label : 'B',
                  name      : 'b',
                  type      : 'string',
                },
              ]
            },
          ]
        } ],
    },

...
```

### System Info

```shell
System:
    OS: macOS 13.3.1
    CPU: (8) arm64 Apple M2
    Memory: 168.02 MB / 16.00 GB
    Shell: 3.2.57 - /bin/sh
  Binaries:
    Node: 18.12.1 - /usr/local/bin/node
[...]

PR #4044, opened 2023-06-26
Fix issue where deeply nested rich-text fields weren't selectable

jeffsee55:
Closes https://github.com/tinacms/tinacms/issues/3976


kldavis4 (approved):
tested and confirmed the issue is fixed

PR #4045, opened 2023-06-27
feat: self-hosted improvements

kldavis4:
# what

- fix frontend to not call events or billing state endpoints when a custom content api is configuring
- Add optional `--partial-reindex` CLI flag that can be used to reduce unnecessary data layer requests in a self-hosted environment

Partial reindexing works by getting the current git tree sha and comparing it to the last sha since indexing was performed. If there is not previous sha, it will do a full reindex. If there is a previous sha, the getChangedFiles function receives a filter object generated using the collection metadata to determine what files relevant to indexing were changed, if any. Then any added or modified files are indexed and and removed files are deleted from the index.


packages/tinacms/src/tina-cms.tsx
```diff
@@ -150,9 +150,9 @@ export const TinaCMSProvider2 = ({
     )
   }
   const apiURL = props?.client?.apiUrl || props?.apiURL
-
+  console.log('auth', schema?.config?.admin?.auth)
```
logan-anderson, line 153: Remove Logs?

kldavis4:
@logan-anderson its opt in functionality, so just want to make sure existing build is not broken. Here is the tagged version for @tinacms/cli & tinacms: 0.0.0-20230628131052

You can also try it on self-hosted https://github.com/tinacms/tina-self-hosted-demo/pull/85/ though that is currently broken until #4053 is merged

logan-anderson:
@kldavis4 Getting this error on the tina-cloud-starter when installing`0.0.0-20230628131052` version.

```
Starting Tina Dev Server
Indexing local files ⠋Error: Unable to find generated lookup file
If you are using the .tina folder. Please set {tinaDirectory: ".tina"} in your createDatabase options or migrate to the new tina folder: https://tina.io/blog/tina-config-rearrangements/
Unexpected end of JSON input
Unable to start dev server, please fix your Tina config and try again
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

kldavis4:
@logan-anderson it seems to be an intermittent thing. I saw it yesterday as well but it would just go away when re-running. Not really sure what that might be...

logan-anderson (approved):
Looks good to me. (That error went away after I re-ran. Not sure why it would come up in the first place)

PR #4053, opened 2023-06-28
Upgrade prism

jeffsee55:
Fixes Next build error:
```
Element type is invalid: expected a string (for built-in components) or a class/function (for composite components) but got: [missing argument].[missing argument]
```
I got mixed results here, have no idea why but I was able to break this on previous tina versions as well as the latest beta. Ultimately upgrading to a more modern build of the prism library seems to have fixed things. I suspect our `cjs` builds aren't playing well with default exports in some cases but can't pin it down with this

Bonus is that we can now provide a `theme` prop:

```jsx
  code_block: (props) => <Prism {...props} theme="palenight" />,
```

### DIFF
--- a/.changeset/forty-readers-begin.md
+++ b/.changeset/forty-readers-begin.md
@@ -1,0 +1,5 @@
+---
+'tinacms': patch
+---
+
+Upgrade prism react renderer, allow for theming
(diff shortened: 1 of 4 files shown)


kldavis4:
@jeffsee55 testing now

PR #4095, opened 2023-07-18
improvements to Self Hosted Tina. 

logan-anderson:
fixes: ENG-1053

Changes
- deprecate onPut, onDelete and the level args to `createDatabase`
- `createDatabase` code moves from `@tinacms/graphql` to `@tinacms/datalayer`
- adds `databaseAdapter` instead of `level`
- adds `gitProvider` instead of onPut and onDelete. 
- adds `createLocalDatabase` to  `@tinacms/datalayer`
- adds `GitHubProvider` to  `@tinacms/datalayer`

TODO:
- add generated db connection file

---

The database.ts file can now look like this.
```ts
import {
  createDatabase,
  createLocalDatabase,
  GitHubProvider,
} from '@tinacms/datalayer'
import { MongodbLevel } from 'mongodb-level'

const isLocal = process.env.TINA_PUBLIC_IS_LOCAL === 'true'

export default isLocal
  ? createLocalDatabase()
  : createDatabase({
      gitProvider: new GitHubProvider({
        branch: process.env.GITHUB_BRANCH,
        owner: process.env.GITHUB_OWNER,
        repo: process.env.GITHUB_REPO,
        token: process.env.GITHUB_PERSONAL_ACCESS_TOKEN,
      }),
      databaseAdapter: new MongodbLevel<string, Record<string, any>>({
        collectionName: process.env.GITHUB_BRANCH,
        dbName: 'tinacms',
        mongoUri: process.env.MONGODB_URI,
      }),
    })

```


kldavis4:
where is this change coming from?

> createDatabase code moves from @tinacms/graphql to @tinacms/datalayer

I don't recall that from our conversation(s) or is it documented in the ticket. And my understanding is that since we are dropping the fair source license, there is no longer a good reason for the @tinacms/datalayer lib. Regardless of that, its unclear why we want to do any of that refactoring in this pr.

packages/@tinacms/cli/src/next/config-manager.ts
```diff
@@ -166,6 +168,14 @@ export class ConfigManager {
       this.generatedFolderPath,
       'client.js'
     )
+    this.generatedDatabaseClientTSFilePath = path.join(
+      this.generatedFolderPath,
+      'backendClient.ts'
```
kldavis4, line 173: my vote would be to call this `databaseClient` as it is more specific than 'backend' - and also whenever you import it, the import ends up being called databaseClient

examples/tina-self-hosted-demo/pages/[filename].tsx
```diff
@@ -32,7 +32,8 @@ export const getStaticProps = async ({ params }) => {
 }
 
 export const getStaticPaths = async () => {
-  const pagesListData = await dbConnection.queries.pageConnection()
+  const pagesListData = await databaseClient.queries.pageConnection()
+  console.log(pagesListData)
```
Comment (kldavis4) [inline examples/tina-self-hosted-demo/pages/[filename].tsx:36]:
```suggestion
```

packages/@tinacms/datalayer/src/gitProviders.ts
```diff
@@ -0,0 +1,84 @@
+import { Octokit } from '@octokit/rest'
```
kldavis4, line 1: might be good to use the same pattern as bridge, with a `provider` directory, and then each implementation in a separate module: https://github.com/tinacms/tinacms/tree/main/packages/%40tinacms/graphql/src/database/bridge

packages/@tinacms/datalayer/src/gitProviders.ts
```diff
@@ -0,0 +1,84 @@
...
+  constructor(args: GitProviderOptions) {
+    this.owner = args.owner
+    this.repo = args.repo
+    this.branch = args.branch
+    this.octokit = new Octokit({
+      auth: args.token,
+    })
+  }
+
+  async onPut(key: string, value: string) {
```
kldavis4, line 31: The getContent calls here need to use `ref` and not `branch`. See https://github.com/tinacms/tina-self-hosted-demo/blob/main/tina/database.ts#L39

packages/@tinacms/cli/src/next/config-manager.ts
logan-anderson: Yeah I like `databaseClient`  better as well.

kldavis4:
> > where is this change coming from?
> > createDatabase code moves from @tinacms/graphql to @tinacms/datalayer
> > I don't recall that from our conversation(s) or is it documented in the ticket. And my understanding is that since we are  dropping the fair source license, there is no longer a good reason for the @tinacms/datalayer lib. Regardless of that, its unclear why we want to do any of that refactoring in this pr.
> 
> @kldavis4
> 
> We are adding the concept of a `gitProvider` into `@tinacms/dayalayer`. `createDatabase` needs access to the `gitProvider` interface. Since we cant import the gitProvider interface in `@tinacms/graphql` the `createDatabse` function needs to move to the same place as the `gitProvider` type.
> 
> Alternatively, we could also move the `gitProvider` interface to`@tinacms/graphql`.
> 
> See vid for a better explanation: https://www.loom.com/share/285c3096a1ba4690a8e28d20d376bddd

I think keeping it in @tinacms/graphql and re-exporting is fine since we are doing that with the existing stuff

logan-anderson:
@kldavis4 I have updated based on all of your suggestions.

kldavis4:
@logan-anderson did you push up your changes?

packages/@tinacms/datalayer/src/providers/github.ts
```diff
@@ -0,0 +1,94 @@
+import { Octokit } from '@octokit/rest'
```
kldavis4, line 1: I thought you were moving all this to @tinacms/graphql?

.changeset/great-seahorses-yell.md
```diff
@@ -0,0 +1,173 @@
+---
+'@tinacms/datalayer': minor
+'@tinacms/self-hosted-starter': patch
+'@tinacms/graphql': patch
+'@tinacms/cli': patch
+---
+
+## Changes
+- deprecate onPut, onDelete and the level args to `createDatabase`
+- `createDatabase` code moves from `@tinacms/graphql`t `@tinacms/datalayer`
```
kldavis4, line 10: outdated?

packages/@tinacms/datalayer/src/providers/github.ts
logan-anderson: I thought you meant the `createDatabase` function and the gitProvider interface? 

I am OK with moving this but the reason it is here so that folks who are not using self hosted don't need to install the deps of the GithubProvider. If we move it into `@tinacms/graphql` everyone would have to install those extra deps

kldavis4: I guess that's a fair point, but I'm not sure how significant that would be and I'm unclear on whether we are going to have a separate self-hosted only dependency or consolidate in either @tinacms/graphql or @tinacms/datalayer. I've been under the impression that we are consolidating, which is the reason why I've been commenting about this stuff getting added to datalayer.

Maybe @octokit/rest should be a peer dependency, since if we have other providers in the future, we may not want to include those either, and they would be peer deps as well.

logan-anderson: Peers deps does not feel right either since it will still require users to install octokit even if they don't use it. 

Maybe it should be its own package?

kldavis4: I think we can mark it as an optional peer dep: https://docs.npmjs.com/cli/v9/configuring-npm/package-json#peerdependenciesmeta

logan-anderson: We could do that. But if we are going to get the usr to install a bunch of packages would it make sense just to put this in iits own package? That way would could ensure they have the right version of all the dependencies installed

logan-anderson: Yup makes sense to me! That way we can have one provider per package and we don't have to mix deps.

logan-anderson: @kldavis4  Just pushed up a change to moves the GithubProvider to its own package called `tinacms-provider-github`

packages/tinacms-provider-github/README.md
```diff
@@ -0,0 +1,38 @@
+he Github Git Provider handles saving and deleting content to github. It can be used as a prop to the `createDatabase` function.
```
kldavis4, line 1: @logan-anderson should this package be named *-gitprovider-* or *-git-provider-*?

.changeset/great-seahorses-yell.md
```diff
@@ -0,0 +1,173 @@
...
+'@tinacms/self-hosted-starter': patch
+'@tinacms/graphql': patch
+'@tinacms/cli': patch
+---
+
+## Changes
+- deprecate onPut, onDelete and the level args to `createDatabase`
+- adds `databaseAdapter` instead of `level`
+- adds `gitProvider` instead of onPut and onDelete. 
+- adds `GitHubProvider` to  `@tinacms/datalayer`
```
kldavis4, line 12: remove?

.changeset/great-seahorses-yell.md
```diff
@@ -0,0 +1,173 @@
+---
+'@tinacms/datalayer': minor
```
kldavis4, line 2: does this need an entry for the provider package?

packages/tinacms-provider-github/README.md
logan-anderson: I am ok with either one

.changeset/great-seahorses-yell.md
logan-anderson: Since we are still telling users to use the data layer package and this does break the types in the data layer package (upgrading will break a type check) I minor makes sense to me so users do not auto update on an `install`.

packages/tinacms-provider-github/README.md
kldavis4: ok - so am I, you pick :) I just felt provider was too generic...

logan-anderson: Pushing up an update that renames it to -gitprovider-

packages/tinacms-gitprovider-github/README.md
```diff
@@ -0,0 +1,38 @@
+he Github Git Provider handles saving and deleting content to github. It can be used as a prop to the `createDatabase` function.
```
kldavis4, line 1:
```suggestion
The Github Git Provider handles saving and deleting content to GitHub. It can be used as a prop to the `createDatabase` function.
```

packages/tinacms-gitprovider-github/README.md
```diff
@@ -0,0 +1,38 @@
...
+
+## Adding the Github Git Provider
```
kldavis4, line 3:
```suggestion
## Adding the GitHub Git Provider
```

kldavis4 (approved):
Here's the branch: feat/self-hosted-improvements

Can you squash this and merge there?

PR #4104, opened 2023-07-24
Add support for static mode on tina media

jeffsee55:
Add support for "static" setting in Tina media, which preprocesses the available media files and disables uploads and deletions of media from the CMS.

```tsx
    tina: {
      publicFolder: 'public',
      mediaRoot: 'uploads',
      static: true,
    },
```

Screenshot with no insert or delete actions:

Screenshot 2023-07-24 at 12 55 07 PM


logan-anderson (changes requested):
Just one piece of feedback. 

When the folder does not exist we get an error message and `tinacms dev` fails

Screenshot 2023-07-25 at 9 19 21 AM

Instead could we either 
1. Create an empty folder
2. Just create an empty JSON array

In both cases it would be nice to print a warning message saying that the folder does not exist.

Other then that, this looks good to me.

logan-anderson (approved):
Looks like there is some merge conflicts but looks good otherwise.

logan-anderson (approved):

PR #4106, opened 2023-07-25
improvements to Self Hosted Tina

logan-anderson:
improvements to self hosted tina.


logan-anderson:
> Getting this error when running basic-iframe in dev mode:
> 
> ```
> error - ./tina/__generated__/client.js
> Error:
>   x Expected ',', got '{'
>    ,----
>  3 | import type { TinaClient } from "tinacms/dist/client";
>    :             ^
>    `----
> 
> Caused by:
>     0: failed to process input file
>     1: Syntax Error
> ```

Fixed in https://github.com/tinacms/tinacms/pull/4144

kldavis4:
Does the createDatabase function need to be backwards compatible? It looks like databaseAdapter and gitProvider are not optional, but the deprecated properties are.

logan-anderson:
> Does the createDatabase function need to be backwards compatible? It looks like databaseAdapter and gitProvider are not optional, but the deprecated properties are.

I believe other then the types it is backwards compatible. Since this is a minor, I think its OK to make a breaking changes in the types and if folks don't want to update they can add a ts-ignore

packages/@tinacms/cli/src/cmds/init/configure.ts
```diff
@@ -0,0 +1,446 @@
...
+
+  if (config.token !== undefined) {
+    envFile += `TINA_TOKEN=${config.token}\n`
+  }
+  await fs.writeFile('.env.tina', envFile)
+}
+
+async function configure(
+  env: InitEnvironment,
+  opts: { debug?: boolean; isBacked?: boolean }
```
kldavis4, line 44: @logan-anderson is this supposed to be `isBackend`?

logan-anderson: 😅 Yes. Good catch

logan-anderson: Fixed!

packages/@tinacms/cli/src/cmds/init/apply.ts
```diff
@@ -0,0 +1,795 @@
...
+      })
+    }
+  }
+
+  if (!env.forestryConfigExists && !env.tinaConfigExists) {
+    // add /content/posts/hello-world.md
+    await addContentFile({ config, env })
+  }
+
+  if (config.framework.reactive && !env.tinaConfigExists) {
```
kldavis4, line 166: @logan-anderson I am working on tests for init and running into issues with this change: https://github.com/tinacms/tinacms/commit/f9bf419e1bca49431569e739fa918a29308f332c

The change adds skipping the generation of files if the folder tina is already present. When you run the backend init and self-host, in the addReactiveFile function, this results in the demo blog file not having the databaseClient and the `setup:users` script won't ever get added to the package.json because the vercel-kv-credentials-provider option only gets set during `init backend`.

If we make the blog post file a GeneratedFile and use writeGeneratedFile, then it will prompt the user whether to overwrite it, and during backend init the correct template will get generated (assuming the user says yes to overwrite). Additionally the extendNextScripts() function can be safely re-run on the same package.json. So I think the `!env.tinaConfigExists` guard on the call to addReactiveFile should probably be removed and the demo blog post should become a GeneratedFile.

packages/@tinacms/graphql/src/database/index.ts
```diff
@@ -73,7 +74,93 @@ export type CreateDatabase = {
...
+      ...config,
+      level: config.level,
+    })
+  }
+
+  if (!config.gitProvider) {
+    throw new Error(
+      'createDatabase requires a gitProvider. Please provide a gitProvider.'
+    )
+  }
```
kldavis4, line 146: I discovered that createDatabase is blowing up in tina cloud whenever gitProvider is not specified even when a bridge is passed in. I don't think gitProvider should be required if bridge is specified.

logan-anderson:
Use https://github.com/tinacms/tinacms/pull/4310 instead.

PR #4107, opened 2023-07-25
Remove @tinacms/toolkit

jeffsee55:
EDIT: this now absorbs `@tinacms/toolkit` into `tinacms`. Had a hard time satisfying our build process and thought this might be easier to knock out all at once.

Following along from the dependency cleanup, trying to add some sane structure to things. Following this would be a similar update to `tinacms`, and possibly some movement between those 2 modules, or this module being absorbed by `tinacms` if it makes sense to.

- use kebab-case for all file and folder names
- use "@/" path alias
- move from jest to vitest
- move deeply nested files to a more flattened structure


logan-anderson:
This is awesome! I wonder if we should do this after the self hosted improvements PR just to make sure we don't have any lost changes?

jeffsee55:
I didn't see any code overlap between the two @logan-anderson. They seem pretty separate between front/backend?

logan-anderson:
@jeffsee55 Yeah you are correct. There does not seem to be any overlap from what I can tell.

logan-anderson:
Looks good to me. One small comment is that[ vitest runs in watch mode by default in non production environments](https://vitest.dev/guide/features.html#watch-mode). Could we updated the test command to run `vitest run` and the test-watch command could be just `vitest` ?

PR #4120, opened 2023-08-03
DO NOT MERGE: Add onSubmit hook function

logan-anderson:
This adds an `onSubmit` hook function that allows users to respond run code when a document is created or updated. It also allows users to modify or add values sent to the GraphQL endpoint.


jeffsee55 (commented):
This might be more accurately called `onBeforeSubmit` or something like that?

logan-anderson:
> This might be more accurately called `onBeforeSubmit` or something like that?

Yes `onSubmit` is not a good name. Maybe just `beforeSubmit` ?

logan-anderson:
Closing if favour of https://github.com/tinacms/tinacms/pull/4122

PR #4121, opened 2023-08-04
Allow absolute paths

logan-anderson:
Fixes ENG-1082

Updated so a user can add an absolute path to the filename

Before all files where created relative to the users current folder and we gave an error if the filename started with a `/`.

Now we check if the filename starts with a `/` and if it does we use that as the absolute path to the file.

Demo: https://www.loom.com/share/5256114d1ce648eda69881e33f8f6bd4?sid=3eafb588-c4da-49eb-ace2-d6b02313e14c


packages/tinacms/src/admin/pages/CollectionCreatePage.tsx
```diff
@@ -37,7 +37,14 @@ const createDocument = async (
...
 
-  const relativePath = `${folder ? `${folder}/` : ''}${filename}.${
+  if (typeof filename !== 'string') {
+    throw new Error('Filename must be a string')
+  }
+
+  // Append the folder if it exists and the filename does not start with a slash
+  const appendFolder = folder && !filename.startsWith('/') ? `/${folder}` : ''
+
+  const relativePath = `${appendFolder ? `${folder}/` : ''}${filename}.${
```
jamespohalloran, line 47:
```suggestion
  const relativePath = `${appendFolder}${filename}.${
```
I think you're already handing the empty case on the line above.

packages/tinacms/src/admin/pages/CollectionCreatePage.tsx
```diff
@@ -192,11 +199,19 @@ export const RenderForm = ({
...
+            form.relativePath =
+              schemaCollection.path +
+              `${values?.values?.filename}.${schemaCollection.format || 'md'}`
+          } else {
+            // keeps the forms relative path in sync with the filename
+            form.relativePath =
+              schemaCollection.path +
+              folderName +
+              `/${values?.values?.filename}.${schemaCollection.format || 'md'}`
+          }
```
jamespohalloran, line 214:
```suggestion
    form.relativePath =
      schemaCollection.path +
      appendFolder +
      `${values?.values?.filename}.${schemaCollection.format || 'md'}`
```
^ I think you can remove the conditional like so

jamespohalloran (dismissed):

packages/tinacms/src/admin/pages/CollectionCreatePage.tsx
logan-anderson: Yeah! I think so. Ill push up this change.

packages/tinacms/src/admin/pages/CollectionCreatePage.tsx
logan-anderson: Yup this makes more sense to me

jeffsee55 (approved):

PR #4122, opened 2023-08-04
Add Before Submit function

logan-anderson:
Fixes ENG-1081

Add a `beforeSubmit` hook function on a collection.ui. This give users the ability to run a function before the form is submitted. 

If the function returns values those values will be used will be submitted instead of the form values.

If the function returns a falsy value the original form values will be submitted.

### Example

```js
// tina/config.{ts.js}

export default defineConfig({
  schema: {
    collections: [
      {
        ui: {
          // Example of beforeSubmit
          beforeSubmit: async ({ values }) => {
            return {
              ...values,
              lastUpdated: new Date().toISOString(),
            };
          },
          //...
        },
        //...
      },
      //...
    ],
  },
  //...
});
```

### DIFF
--- a/.changeset/soft-glasses-watch.md
+++ b/.changeset/soft-glasses-watch.md
@@ -1,0 +1,40 @@
+---
+'@tinacms/schema-tools': patch
+'tinacms': patch
+---
+
+Add a `beforeSubmit` hook function on a collection.ui. This give users the ability to run a function before the form is submitted. 
+
+If the function returns values those values will be used will be submitted instead of the form values.
+
+If the function returns a falsy value the original form values will be submitted.
+
+### Example
+
+
+
+```js
+// tina/config.{ts.js}
+
+export default defineConfig({
+  schema: {
+    collections: [
+      {
+        ui: {
+          // Example of beforeSubmit
+          beforeSubmit: async ({ values }) => {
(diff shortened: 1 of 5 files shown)


logan-anderson:
> The API looks handy! is this called for both onCreate, and onUpdate? The one thing I'm wondering is if users will eventually need unique behaviour for both onCreate vs onUpdate (to set something like a "createdDate"). Not that we need that now, but should that impact the name we use ?(beforeSubmit vs beforeUpdate)

@jamespohalloran  This can be used for both by looking at `form.crudType`. 

Which has this type.

```ts
crudType?: 'create' | 'update'
```

jamespohalloran:
> > The API looks handy! is this called for both onCreate, and onUpdate? The one thing I'm wondering is if users will eventually need unique behaviour for both onCreate vs onUpdate (to set something like a "createdDate"). Not that we need that now, but should that impact the name we use ?(beforeSubmit vs beforeUpdate)
> 
> @jamespohalloran This can be used for both by looking at `form.crudType`.
> 
> Which has this type.
> 
> ```ts
> crudType?: 'create' | 'update'
> ```

Cool, that make senses to me

jeffsee55 (approved):

PR #4132, opened 2023-08-09
Filterable templates

jamespohalloran:
Previously, collections with templates instead of fields weren't getting the option to filter/search/sort
https://www.loom.com/share/9fa1adec08504274855a9216f0574f69?sid=524d5849-29ae-4e5b-9467-4fad42d17afe

fixes ENG-1103

### DIFF
--- a/packages/tinacms/src/admin/pages/CollectionListPage.tsx
+++ b/packages/tinacms/src/admin/pages/CollectionListPage.tsx
@@ -176,6 +176,22 @@ export const handleNavigate = (
   }
 }
 
+function getUniqueTemplateFields(collection: Collection<true>): TinaField[] {
+  const fieldSet: TinaField[] = []
+
+  collection.templates.forEach((template) => {
+    template.fields
+      .filter((f) => {
+        return fieldSet.find((x) => x.name === f.name) === undefined
+      })
+      .forEach((field) => {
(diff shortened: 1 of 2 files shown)


logan-anderson (approved):

packages/tinacms/src/admin/pages/CollectionListPage.tsx
```diff
@@ -284,10 +300,17 @@ const CollectionListPage = () => {
                 const documents = collection.documents.edges
                 const admin: TinaAdminApi = cms.api.admin
                 const pageInfo = collection.documents.pageInfo
-                const fields = collectionExtra.fields?.filter((x) =>
+
+                // get unique fields from all templates
+                const fields = (
```
jeffsee55, line 305: I'm surprised these fields don't need to be namespaced by their template name. 🤷 if it works, 👍

jeffsee55 (approved):

packages/tinacms/src/admin/pages/CollectionListPage.tsx
jamespohalloran: I wonder if they may need to if you supplied a strict "starts_with" in the search? Otherwise filtering/search seemed to work fine for me but I'll do some extra testing on the beta release.

PR #4133, opened 2023-08-10
Fix before submit types

logan-anderson:
Update the before submit types to not pass the finalForm form since it is contained in the TinaForm

### DIFF
--- a/packages/@tinacms/schema-tools/src/types/index.ts
+++ b/packages/@tinacms/schema-tools/src/types/index.ts
@@ -817,8 +817,7 @@ export interface UICollection<Form = any, CMS = any, TinaForm = any> {
   beforeSubmit?: (arg: {
     values: Record<string, unknown>
     cms: CMS
-    form: Form
-    tinaForm: TinaForm
+    form: TinaForm
   }) => Promise<void | Record<string, unknown>>
 }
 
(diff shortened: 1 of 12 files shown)


packages/@tinacms/schema-tools/src/types/index.ts
```diff
@@ -817,8 +817,7 @@ export interface UICollection<Form = any, CMS = any, TinaForm = any> {
   beforeSubmit?: (arg: {
     values: Record<string, unknown>
     cms: CMS
-    form: Form
-    tinaForm: TinaForm
+    form: TinaForm
   }) => Promise<void | Record<string, unknown>>
```
jeffsee55, line 821: What I was saying the other day is that we _could_ type this as:

```ts
// leave CMS out
interface BeforeSubmit {
  values: Record<string, unknown>
  form: TinaForm
}
```
And then in `tinacms` we do:


```ts
declare module "@tinacms/schema-tools" {
  interface BeforeSubmit { 
    cms: TinaCMS
  }
}
```
Doesn't need to be done in this PR, and maybe some better ways to do it, but that's one approach.

jeffsee55 (approved):

packages/@tinacms/schema-tools/src/types/index.ts
logan-anderson: Yeah I think we should leave it out for now and look into it in a separate PR

PR #4147, opened 2023-08-16
Filtering subitems to avoid duplicates

karelnagel:
Filtering all the nav items to avoid duplicate references, that caused errors in console.
The error: 
```
Warning: Encountered two children with the same key, `content/colors/foreground.json`. Keys should be unique so that components maintain their identity across updates. Non-unique keys may cause children to be duplicated and/or omitted — the behavior is unsupported and could change in a future version.

```

### DIFF
--- a/.changeset/funny-coins-help.md
+++ b/.changeset/funny-coins-help.md
@@ -1,0 +1,5 @@
+---
+'tinacms': patch
+---
+
+filtering subitems to avoid duplicates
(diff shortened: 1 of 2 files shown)


jeffsee55:
Thanks for the contribution @karelnagel - I'm surprised there were two items with the same form id though. Do you have a specific query where you notice this ocurring?

logan-anderson:
@karelnagel Can you provide a reproduction of this error? (PR looks good but we would like to see how the issue occurred before it is merged )

PR #4159, opened 2023-08-22
Fix self-hosted-init issues with existing tina config

jeffsee55:
When running the backend init with an existing `tina/config.ts` file, we don't ask them if they want to use Typescript, but we can infer it. Also, we weren't asking them for a package manager, which meant that we later would skip package installation.

This doesn't address a slightly more awkward issue, which is that we fail to detect `tina/config.tsx` or `tina/config.jsx`. The `makeGeneratedFile` might need some updates to address that, https://github.com/tinacms/tinacms/blob/dbe4367905a9ec580a0e5b8c758cf932b6f757c4/packages/%40tinacms/cli/src/cmds/init/detectEnvironment.ts#L19-L20


logan-anderson (approved):
LGTM. We can do further testing on the PR deploy

PR #4162, opened 2023-08-23
Always ask for pacakge manager

logan-anderson:
I was noticing that when doing a backend init on a site that already has tina setup it was not doing an install (Adding new self hosting deps). 

This PR updates the questions so that it always asks for the package manager so that it has that info to use to run the install command.

### DIFF
--- a/packages/@tinacms/cli/src/cmds/init/apply.ts
+++ b/packages/@tinacms/cli/src/cmds/init/apply.ts
@@ -338,7 +338,11 @@ const addDependencies = async (
 ) => {
   const tagVersion = params.tinaVersion ? `@${params.tinaVersion}` : ''
   const { dataLayerAdapter, nextAuth, packageManager } = config
-  let deps = [`tinacms`, '@tinacms/cli']
+  let deps = []
+  if (!env.tinaConfigExists) {
+    deps.push('tinacms')
+    deps.push('@tinacms/cli')
+  }
   let devDeps = []
   if (nextAuth) {
     deps.push('tinacms-next-auth', 'next-auth')
@@ -371,7 +375,7 @@ const addDependencies = async (
     yarn: `yarn add ${deps.join(' ')}`,
   }
 
-  if (packageManagers[packageManager]) {
+  if (packageManagers[packageManager] && deps.length > 0) {
     logger.info(logText('Adding dependencies, this might take a moment...'))
     logger.info(indentedCmd(`${logText(packageManagers[packageManager])}`))
     await execShellCommand(packageManagers[packageManager])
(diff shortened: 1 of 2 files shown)


packages/@tinacms/cli/src/cmds/init/apply.ts
```diff
@@ -338,7 +338,11 @@ const addDependencies = async (
 ) => {
   const tagVersion = params.tinaVersion ? `@${params.tinaVersion}` : ''
   const { dataLayerAdapter, nextAuth, packageManager } = config
-  let deps = [`tinacms`, '@tinacms/cli']
+  let deps = []
+  if (!env.tinaConfigExists) {
```
kldavis4, line 342: under what condition would you run the package manager install and not want to install `[`tinacms`, '@tinacms/cli']`? It seems like maybe if the deps were already present then you would not include, but `env.tinaConfigExists` shouldn't be indicating that.

logan-anderson: I added this because when a user is doing backend init they already have Tina setup on there site and it felt weird to update the users Tina packages and install packages that exist. This also means an install will happen when doing a backend init with Tina Cloud.

It would be more technically correct to look at the users packages.json and see what deps they have installed but I can't really see why a user would have a `tina/config` file and not have tinacms already  installed. 

Alternately we can leave it how it was and they will always update to the latest tina version when they run backend init.

kldavis4: can we not use backend init mode as the flag for whether to add the tina deps (instead of tinaConfigExists)?

logan-anderson: We can't do that because it would be nice to support doing the backend init on a site that does not have tina setup yet

kldavis4: We should just throw an error and tell them to use init in that case

logan-anderson: The benefit to using backend init projects that do not have Tina already is that the user does not have to copy and past some of the setup in the config file. 

I don't think that parsing the package.json should be to much work we might already be doing it somewhere else. 

If its going to be to much work to do that though we can throw an error asking the user use init first.

logan-anderson: @kldavis4 I updated the code to look at the package.json. It will only add tinacms and @tinacms/cli if they are both not present in the users package.json.

kldavis4 (approved):

PR #4171, opened 2023-08-29
Wrap up auth provider

jamespohalloran:


packages/tinacms/src/admin/index.tsx
```diff
@@ -237,6 +237,10 @@ export const TinaAdmin = ({
                         }
                       />
                     )}
+                    <Route
```
logan-anderson, line 240: @jamespohalloran What is this used for?

jamespohalloran: Oops! This must have slipped in with something else I was playing with. Removing now

kldavis4 (approved):

Issue #4187, opened 2023-09-05
"Expected Template value" error when navigating into nested rich-text template


jamespohalloran:
### Describe the issue that you're seeing. Any Loom videos or screenshots usually help a lot!

When navigating into a nested MDX component, you get: 

<img width="1346" alt="Screenshot 2023-09-05 at 1 35 32 PM" src="https://github.com/tinacms/tinacms/assets/3323181/42922228-7207-4663-80a3-55a105f291c1">

### Reproduction

https://github.com/tinacms/tinacms/tree/nested-template-bug

### Steps to reproduce

To see the bug:
- Checkout https://github.com/tinacms/tinacms/tree/nested-template-bug
- Run examples/basic-iframe project
- Go to: http://localhost:3000/admin/index.html#/collections/edit/post2/b
- In the rich-text field, add a `Tabs` component, then a sub-`Tab Item` component, then navigate with the Tab Item.

You should see the above error

### System Info

```shell
System:
    OS: macOS 13.4.1
    CPU: (16) x64 Intel(R) Core(TM) i9-9880H CPU @ 2.30GHz
    Memory: 470.97 MB / 32.00 GB
    Shell: 3.2.57 - /bin/bash
  Binaries:
    Node: 16.18.0 - ~/.nvm/versions/node/v16.18.0/bin/node
    Yarn: 1.22.19 - ~/.nvm/versions/node/v16.18.0/bin/yarn
    npm: 9.1.2 - ~/.nvm/versions/node/v16.18.0/bin/npm
  Browsers:
    Chrome: 116.0.5845.140
    Firefox: 116.0.3
    Safari: 16.5.2
```

### Validations

- [X] Follow our [Code of Conduct](https://github.com/tinacms/tinacms/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [docs](https://tina.io/docs).
- [X] Check that there isn't [already an issue](https://github.com/tinacms/tinacms/issues) that reports the same bug.

PR #4189, opened 2023-09-05
Fix issue where deeply nested mdx fields were not picking up the correct template

jeffsee55:
When an MDX component is nested 2 levels deep, we were picking up the index of the first item while trying to resolve the second. Closes https://github.com/tinacms/tinacms/issues/4187

### DIFF
--- a/.changeset/swift-pants-return.md
+++ b/.changeset/swift-pants-return.md
@@ -1,0 +1,5 @@
+---
+'tinacms': patch
+---
+
+Fix issue where deeply nested mdx fields were not picking up the correct template"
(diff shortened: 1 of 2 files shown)


jamespohalloran (approved):

wkelly17:
@jeffsee55  - I'm still getting this error it seems. I didn't exactly clone the whole repo, but I'm running this section here - https://github.com/tinacms/tinacms/blob/nested-template-bug/examples/basic-iframe/tina/config.js#L187-L251 - and its still throwing the same error.

PR #4191, opened 2023-09-05
Only preserve legacy values if no field is defined for them

jeffsee55:
Update mutation logic so that existing markdown values aren't overwritten by the Tina form, only if the schema has not defined that field. 

Eg:

```
title: Hello
tags:
  - a
  - b
```
And a schema:
```
{
  //...
  fields: [
    {type: 'string', name: 'title'},
  ]
}
```
`tags` won't be changed, but if there is a field defined for `tags`, whatever value is sent in from the Tina form will be used, even `null`:
```
title: Hello
tags: null
```

### DIFF
--- a/.changeset/late-vans-wash.md
+++ b/.changeset/late-vans-wash.md
@@ -1,0 +1,5 @@
+---
+'@tinacms/graphql': patch
+---
+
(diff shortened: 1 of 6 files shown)


logan-anderson (approved):
LGTM

PR #4205, opened 2023-09-15
feat: Tina nextauth integration

kldavis4:


examples/tina-self-hosted-demo/tina/config.tsx
```diff
@@ -1,35 +1,29 @@
...
-    //   },
-    //   authenticate: async () => {
-    //     localStorage.setItem(TINA_TOKEN_KEY, 'some-token')
-    //     return true
-    //   },
-    //   getUser: async () => {
-    //     return localStorage.getItem(TINA_TOKEN_KEY)
-    //   },
-    // },
+    authHooks: {}, // TODO why is this required?
```
logan-anderson, line 17: @kldavis4 This should not be required I can make a PR to update the types.

examples/tina-self-hosted-demo/tina/config.tsx
```diff
@@ -53,6 +47,30 @@ const config = defineStaticConfig({
   },
   schema: {
     collections: [
+      {
+        applicationData: true,
```
logan-anderson, line 51: What is this flag used for?

examples/tina-self-hosted-demo/tina/config.tsx
```diff
@@ -1,35 +1,29 @@
...
-    //   authenticate: async () => {
-    //     localStorage.setItem(TINA_TOKEN_KEY, 'some-token')
-    //     return true
-    //   },
-    //   getUser: async () => {
-    //     return localStorage.getItem(TINA_TOKEN_KEY)
-    //   },
-    // },
+    authHooks: {}, // TODO why is this required?
+    authCollection: 'user',
```
logan-anderson, line 18: It doesn't _really_ matter to me but it might be nice to handle this the same way we handle isTitle, isBody, ect. 

This is just a suggestion but we could do this instead.

Currently we have

```ts
authCollection: 'user',
```

Would it make more sense to have

```ts
  collections: [
      {
        applicationData: true,
        isAuthCollection: true,
        label: 'Users',
        name: 'user',
        path: 'content/users',
        format: 'md',
        fields: [
          {
            type: 'string',
            label: 'Name',
            name: 'name',
          },
          {
            type: 'string',
            label: 'Email',
            name: 'email',
          },
          {
            type: 'password',
            label: 'Password',
            name: 'password',
          },
        ],
      },
```

We could throw an error if the user has  `isAuthCollection:  true` for more then one collection.

logan-anderson (commented):
Just a couple of questions

examples/tina-self-hosted-demo/tina/config.tsx
kldavis4: thanks - I updated it to be optional

examples/tina-self-hosted-demo/tina/config.tsx
kldavis4: That controls whether the collection writes to git or not. Haven't come up with a better name for it yet

examples/tina-self-hosted-demo/tina/config.tsx
kldavis4: Sounds good - I'll update it

examples/tina-self-hosted-demo/tina/config.tsx
logan-anderson: is it read only then?

kldavis4: No.. it writes to the DB only

logan-anderson:
@kldavis4 Making a PR to fix the build: https://github.com/tinacms/tinacms/pull/4208

logan-anderson:
@kldavis4 Another build related fix: https://github.com/tinacms/tinacms/pull/4209

logan-anderson:
@kldavis4 Looks like the @tinacms/graphql tests are failing

examples/tina-self-hosted-demo/tina/config.tsx
```diff
@@ -1,35 +1,24 @@
...
-    //     return localStorage.getItem(TINA_TOKEN_KEY)
-    //   },
-    // },
-  },
   clientId: process.env.NEXT_PUBLIC_TINA_CLIENT_ID!,
+  authProvider:
+    !isLocal &&
+    new UsernamePasswordNextAuthProvider({
+      name: 'Credentials',
+      callbackUrl: '/admin/index.html',
```
logan-anderson, line 20: Could we make `/admin/index.html` the default and let user's override if they have something different?

examples/tina-self-hosted-demo/tina/config.tsx
```diff
@@ -1,35 +1,24 @@
...
-    //   getUser: async () => {
-    //     return localStorage.getItem(TINA_TOKEN_KEY)
-    //   },
-    // },
-  },
   clientId: process.env.NEXT_PUBLIC_TINA_CLIENT_ID!,
+  authProvider:
+    !isLocal &&
+    new UsernamePasswordNextAuthProvider({
+      name: 'Credentials',
```
logan-anderson, line 19: What is `name: 'Credentials'`used for and could "Credentials" be a default value?

logan-anderson (commented):
Just a couple of nitpicky questions / suggestions

logan-anderson:
@kldavis4 Found a bug in the hosted version: https://www.loom.com/share/748d269ba4734b459ef32faaddfa7c72?sid=d2361a80-df65-41b7-9b7f-66650c10664b

Not sure what the fix is. Seems like a ES module/ CJS error. We may have to update our lodash imports

logan-anderson:
@kldavis4 Looks liked that fixed the import error I was seeing but now I am getting a new error: https://www.loom.com/share/80b8186109d4417ea125c74ee76c734a?sid=2d274795-8d26-4447-8c2f-30a491a3ee4c

logan-anderson:
having issues updating the password: https://www.loom.com/share/3811468bf8964e87a9979f863a284deb?sid=c9f5e326-9567-4f91-8717-49ab86a1ffec

i am using the local in memory DB if thats an issue.

logan-anderson:
@kldavis4 What is the motivation for not using NextAuth's sign in page?

 It seemed to work for me by going to /api/auth/signin.
 
 Also, this is looking amazing and feels like it will be a really good solution!

kldavis4:
> having issues updating the password: https://www.loom.com/share/3811468bf8964e87a9979f863a284deb?sid=c9f5e326-9567-4f91-8717-49ab86a1ffec
> 
> i am using the local in memory DB if thats an issue.

I'm not sure I understand.. I updated the main branch on that and used dev:prod with a mongodb database and the password update screen worked. I don't think the password update screen should be showing in local dev mode (isLocal=true) so that might be a bug if that is what you are doing. If not, let me know more details on how you are running it.

kldavis4:
@logan-anderson it does look like you were accessing the change password screen in local mode (at least I was able to replicate the behavior you were seeing by doing that). I made this change https://github.com/tinacms/tinacms/pull/4205/commits/ca23d0bf219c1c2119909f2d2ad496de530fffe3 to hide the account management in local mode - the only issue is that the isLocalMode prop is set based on the contentApiUrlOverride, so I had to change the config to account for that.

logan-anderson:
> @logan-anderson it does look like you were accessing the change password screen in local mode (at least I was able to replicate the behavior you were seeing by doing that). I made this change [ca23d0b](https://github.com/tinacms/tinacms/commit/ca23d0bf219c1c2119909f2d2ad496de530fffe3) to hide the account management in local mode - the only issue is that the isLocalMode prop is set based on the contentApiUrlOverride, so I had to change the config to account for that.
> 
> (Here's the PR tag: `0.0.0-20230926124302`)

I wonder if we could have it show up based on the authProvider?

logan-anderson:
@kldavis4 I am not able to update the password correctly (using the same demo site, MongoDB level store) 

See this video: https://www.loom.com/share/f81cbe8495b94ab4bf91abbf269801ec

for more details

logan-anderson:
Another slightly strange behavior.

I just updated the deps and re-deployed my site and it erased all of my users I had created previously.

Screenshot 2023-09-26 at 10 16 30 AM

kldavis4:
> Another slightly strange behavior.
> 
> I just updated the deps and re-deployed my site and it erased all of my users I had created previously.
> 
> Screenshot 2023-09-26 at 10 16 30 AM

thanks - looking into it

kldavis4:
> Another slightly strange behavior.
> 
> I just updated the deps and re-deployed my site and it erased all of my users I had created previously.
> 
Looks like a bug in the mongodb level implementation. I'll work on a fix. I _think_ that it should work correctly if you switch to the redis kv database adapter

kldavis4:
@logan-anderson I fixed the bug in mongodb-level but wasn't able to publish it to npm - I think the last version might have been published manually (any ideas @jeffsee55 ?) - I created a namespaced version that you can test with: `@kldavis4/mongodb-level`

logan-anderson:
@kldavis4 I have not run into any more issues. I am OK with merging this into `feat/self-hosted-improvments` when you think its ready

PR #4223, opened 2023-09-28
Partial queries

jamespohalloran:
Support partial queries.

Note: I made this a `minor` release, since it turns the data response to be optional. Might want to discuss and make sure that's what makes sense here.

closes ENG-1092


logan-anderson:
@jamespohalloran Is there a way to use this when using the generate client. 

EX
```ts
 const test = await client.queries.CustomQuery()
```

jamespohalloran:
> @jamespohalloran Is there a way to use this when using the generate client?
> 
> EX
> 
> ```ts
>  const test = await client.queries.CustomQuery()
> ```

It's currently only setup on .request (Just because the main intent want to acommodate partial queries).

packages/@tinacms/graphql/src/builder/index.ts
```diff
@@ -438,7 +438,16 @@ export class Builder {
...
-    return astBuilder.FieldDefinition({ type, name, args, required: true })
+    // Is errorPolicy set to include?
+    const errorPolicyInclude =
+      this.config.tinaSchema.config?.config?.client?.errorPolicy === 'include'
+    return astBuilder.FieldDefinition({
+      type,
+      name,
+      args,
+      // If errorPolicy is set to include, we don't want to throw an error so we set required to false
+      required: !errorPolicyInclude,
```
jeffsee55, line 449: Is this change is necessary? If we don't throw from the client it will get stored the error in the `errors` array, which seems like the desired behavior. This would make it so that it's just `null` with no error message anywhere.

packages/tinacms/src/unifiedClient/index.ts
```diff
@@ -73,15 +99,16 @@ export class TinaClient<GenQueries> {
       )
     }
     const json = await res.json()
-    if (json.errors) {
+    console.log({ errorPolicyDefined })
```
jamespohalloran, line 102:
```suggestion
```

packages/@tinacms/graphql/src/builder/index.ts
jamespohalloran: cc @logan-anderson 

Also, I don't know about having the errorPolicy be apart of the client. Seems like something you may want at the request-level depending on what your fetching.

jeffsee55: I think it makes sense to be available in both spaces. Having said that, I'd almost wonder if we're better off just never throwing and just return the errors array.

> we can update that do be required: false

That seems fine, though I think leaving it as-is is also ok. If you have an `errors` array then it's safe to say that all bets are off for your Typescript types, IMO. Mainly, I just think we should alter between the 2 behaviors based on the client config.

packages/tinacms/src/unifiedClient/index.ts
```diff
@@ -27,20 +31,42 @@ export type TinaClientURLParts = {
...
+  // }>
+  // public async request<DataType extends Record<string, any> = any>(
+  //   args: Omit<TinaClientRequestArgs, 'errorPolicy'> & {
+  //     errorPolicy: 'include'
+  //   }
+  // ): Promise<{
+  //   data: DataType
+  //   query: string
+  //   errors: GraphQLError[]
+  // }>
```
jamespohalloran, line 64:
```suggestion
```

packages/@tinacms/graphql/src/builder/index.ts
jamespohalloran: > That seems fine, though I think leaving it as-is is also ok.
> I just think we should not alter between the 2 behaviors based on the client config (though I get why you did that)

So are you saying that we should implement what @logan-anderson is suggesting here? https://github.com/tinacms/tinacms/pull/4223#discussion_r1347453912

Just want to be clear on next steps to get this through

jeffsee55: Sorry @jamespohalloran yes I was saying that seems good to do. This all looks good to me 👍

jeffsee55 (approved):

Issue #4220, opened 2023-09-26
Undo / Redo is not working in the rich text editor


logan-anderson:
### Describe the issue that you're seeing. Any Loom videos or screenshots usually help a lot!

https://www.loom.com/share/b2f46ee9459c4577a97af2342a7d7a77?sid=abf48bab-81f7-48ac-90ef-e65c56ad7d41

### Validations

- [X] Follow our [Code of Conduct](https://github.com/tinacms/tinacms/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [docs](https://tina.io/docs).
- [X] Check that there isn't [already an issue](https://github.com/tinacms/tinacms/issues) that reports the same bug.

logan-anderson:
Working in "tinacms": "1.5.6",   "@tinacms/cli": "1.5.13",

Not working in   "tinacms": "^1.5.20",   "@tinacms/cli": "^1.5.29",

logan-anderson:
@jeffsee55 Any idea why this would work in the monrepo but not on the beta tagged release?

jeffsee55:
It looks like it's due to a esm/cjs issue, so probably started when I pushed that out a couple of weeks ago.

PR #4226, opened 2023-09-29
Use named import for is-hotkey package

jeffsee55:
Fixes https://github.com/tinacms/tinacms/issues/4220


logan-anderson (approved):
Works! Thanks @jeffsee55

PR #4231, opened 2023-10-04
 add next backend functions

logan-anderson:
add next backend functions


kldavis4:
I think this should be a node backend (not next) since it doesn't use any next specific features

logan-anderson:
> I think this should be a node backend (not next) since it doesn't use any next specific features

That makes sense to me as long as we are ok with two things

- use the [node types](https://nodejs.org/api/http.html#httprequestoptions-callback)
- not using any [next.js features](https://nextjs.org/docs/pages/api-reference/functions/next-server#nextrequest) that they add to req and res. (I dont think we are using any yet)

logan-anderson:
> > I think this should be a node backend (not next) since it doesn't use any next specific features
> 
> That makes sense to me as long as we are ok with two things
> 
> * use the [node types](https://nodejs.org/api/http.html#httprequestoptions-callback)
> * not using any [next.js features](https://nextjs.org/docs/pages/api-reference/functions/next-server#nextrequest) that they add to req and res. (I dont think we are using any yet)

@kldavis4 Looking into this further there _is_ some next specific code that we are using. 

EX
```js
const [req, res] = params;
const { routes } = req.query;
```

```js
res.json( 
// ...
)
```

```js
res.status(404);
```

kldavis4:
> > > I think this should be a node backend (not next) since it doesn't use any next specific features
> > 
> > 
> > That makes sense to me as long as we are ok with two things
> > 
> > * use the [node types](https://nodejs.org/api/http.html#httprequestoptions-callback)
> > * not using any [next.js features](https://nextjs.org/docs/pages/api-reference/functions/next-server#nextrequest) that they add to req and res. (I dont think we are using any yet)
> 
> @kldavis4 Looking into this further there _is_ some next specific code that we are using.
> 
> EX
> 
> ```js
> const { routes } = req.query;
> ```
> 
> ```js
> res.json( 
> // ...
> )
> ```
> 
> ```js
> res.status(404);
> ```

I think those are convenience functions/props that we don't _need_ to use

kldavis4:
I believe also James suggested we explicitly pass in the `req, res` objects the way the nextauth api handler [does](https://github.com/tinacms/tinacms/blob/feat/self-hosted-improvements/examples/tina-self-hosted-demo/pages/api/auth/%5B...nextauth%5D.ts#L5)

logan-anderson:
> > > > I think this should be a node backend (not next) since it doesn't use any next specific features
> > > 
> > > 
> > > That makes sense to me as long as we are ok with two things
> > > 
> > > * use the [node types](https://nodejs.org/api/http.html#httprequestoptions-callback)
> > > * not using any [next.js features](https://nextjs.org/docs/pages/api-reference/functions/next-server#nextrequest) that they add to req and res. (I dont think we are using any yet)
> > 
> > 
> > @kldavis4 Looking into this further there _is_ some next specific code that we are using.
> > EX
> > ```js
> > const { routes } = req.query;
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > ```js
> > res.json( 
> > // ...
> > )
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > ```js
> > res.status(404);
> > ```
> 
> I think those are convenience functions/props that we don't _need_ to use

@kldavis4 Looking into it more it doesn't look like node Response and Request is supported in Next.js with the pages router. (Looks like it is supported in the app router)

See https://www.loom.com/share/0c15e259b39a44e0bd30065cac5f60b6?sid=5b30a949-2cf4-4f15-8fa1-2aeb5a4b08e8

Note:
- This works in the app router
- I think it also works if you set runtime to edge (GQL endpoint won't work in edge from what I rememebr)

kldavis4:
I was not suggesting that we implement a backend that supports edge runtimes (yet). The node request object is [IncomingMessage](https://nodejs.org/api/http.html#class-httpincomingmessage) and response object is [OutgoingMessage](https://nodejs.org/api/http.html#class-httpoutgoingmessage).

See https://nextjs.org/docs/pages/building-your-application/routing/api-routes#parameters - if we stick to this api, the same backend can be used in an express route and nextjs.

To support the app router we'll need a different backend (see https://nextjs.org/docs/app/building-your-application/routing/route-handlers)

logan-anderson:
@kldavis4 This PR is ready for testing / review now. The example is up to date.

Here is a quick video explaining some of the build config changes: https://www.loom.com/share/5e2e2bd87788443dbf78047984c4a6cc

examples/tina-self-hosted-demo/pages/api/tina/[...routes].ts
```diff
@@ -0,0 +1,16 @@
...
+  databaseClient: databaseClient,
+  secret: process.env.NEXTAUTH_SECRET,
+})
+
+const tinaNodeBackend = TinaNodeBackend({
+  authentication: NextAuthAuthentication({ nextAuthOptions }),
+  databaseClient,
+})
+
+export default tinaNodeBackend
```
Comment (kldavis4) [inline examples/tina-self-hosted-demo/pages/api/tina/[...routes].ts:16]: See https://github.com/tinacms/tinacms/pull/4231#issuecomment-1747092012

Comment (logan-anderson) [reply examples/tina-self-hosted-demo/pages/api/tina/[...routes].ts:16]: @kldavis4 all of the next backend specific features have been removed.  Are you talking about this comment https://github.com/tinacms/tinacms/pull/4231#issuecomment-1747386131 ?

Comment (kldavis4) [reply examples/tina-self-hosted-demo/pages/api/tina/[...routes].ts:16]: huh?

Screen Shot 2023-10-05 at 13 55 04

kldavis4:
@logan-anderson shouldn't the self-hosted demo also be updated to remove auth.ts and update the contentApiUrlOverride?

logan-anderson:
> @logan-anderson shouldn't the self-hosted demo also be updated to remove auth.ts and update the contentApiUrlOverride?

Yup! I can push up a commit that removes the`auth.ts` file. 

It looks like `contentApiUrlOverride: '/api/tina/gql',` has been set correctly from your previous PR

packages/tinacms-authjs/package.json
```diff
@@ -27,11 +27,11 @@
     "next": "12.3.4",
     "next-auth": "4.22.1",
     "react": "^18",
-    "react-dom": "^18",
-    "tinacms": "workspace:*"
+    "react-dom": "^18"
   },
   "devDependencies": {
     "@tinacms/scripts": "workspace:*",
+    "tinacms": "workspace:*",
```
kldavis4, line 34: why is this now in devDependencies - we are still using AbstractAuthProvider form 'tinacms' in src/tinacms?

logan-anderson: @kldavis4 tinacms really a peer dependency. I  added it as a devDependencies so that we can get the correct types in the mono repo.  

It is assumed that anyone using `tinacms-authjs` will have tinacms installed. If if leave it as a dependency, the user could get two versions of tinacms.  (The version from `tinacms-authjs` and there own version)

kldavis4: ok

examples/tina-self-hosted-demo/pages/api/tina/[...routes].ts
```diff
@@ -0,0 +1,16 @@
...
+
+export const nextAuthOptions = TinaAuthJSOptions({
+  databaseClient: databaseClient,
+  secret: process.env.NEXTAUTH_SECRET,
+})
+
+const tinaNodeBackend = TinaNodeBackend({
+  authentication: NextAuthAuthentication({ nextAuthOptions }),
+  databaseClient,
+})
```
Comment (kldavis4) [inline examples/tina-self-hosted-demo/pages/api/tina/[...routes].ts:14]: can you update this to show this is setup running in dev mode (is local true/false)?

packages/tinacms-authjs/src/index.ts
```diff
@@ -133,10 +133,46 @@ const TinaCredentialsProvider = ({
...
+
+      // @ts-ignore
+      if (!req.session) {
+        Object.defineProperty(req, 'session', {
+          value: session,
+          writable: false,
+        })
+      }
+
+      if (!session?.user?.name) {
```
kldavis4, line 153: I think this should be isAuthorized and should actually check the role. Refer to https://github.com/tinacms/tinacms/blob/feat/self-hosted-improvements/packages/tinacms-authjs/src/index.ts#L98

packages/tinacms-authjs/src/index.ts
```diff
@@ -133,10 +133,46 @@ const TinaCredentialsProvider = ({
       authenticate(databaseClient, credentials.username, credentials.password),
   })
 
+const NextAuthAuthentication = ({
```
kldavis4, line 136: Naming should be consistent with the other packages - AuthJS instead of NextAuth. Also thinking it would be good to disambiguate frontend (auth providers) from backend. Maybe call it <*>BackendAuthentication?

examples/tina-self-hosted-demo/pages/api/tina/[...routes].ts
Comment (logan-anderson) [reply examples/tina-self-hosted-demo/pages/api/tina/[...routes].ts]: Updated

packages/tinacms-authjs/src/index.ts
logan-anderson: Updated

packages/@tinacms/datalayer/src/node.ts
```diff
@@ -0,0 +1,141 @@
+import type { IncomingMessage, ServerResponse } from 'http'
```
kldavis4, line 1: I am hesitant to revisit this again... but it feels like given that @tinacms/datalayer is currently just a facade for @tinacms/graphql this should probably go into @tinacms/graphql for now (in a backends directory). And actually this would probably be better in a separate self-hosting package, since it is only intended for that (whereas the other stuff in @tinacms/datalayer & @tinacms/graphql gets shared with tina cloud)

examples/tina-self-hosted-demo/pages/api/tina/[...routes].ts
Comment (logan-anderson) [reply examples/tina-self-hosted-demo/pages/api/tina/[...routes].ts:22]: The comment you linked to is

> I think this should be a node backend (not next) since it doesn't use any next specific features

I believe this is already done?

From the context it looked like you might be talking about the comment I linked to 

> I believe also James suggested we explicitly pass in the req, res objects the way the nextauth api handler [does](https://github.com/tinacms/tinacms/blob/feat/self-hosted-improvements/examples/tina-self-hosted-demo/pages/api/auth/%5B...nextauth%5D.ts#L5)


This can be done by doing 

```ts
const handler = TinaNodeBackend(
  // ...
)
export default (req, res) =>{
   handler(req, res)
}
``` 

IMO we should not do it this way since I think its a bit more cumbersome and not the standard that other libraries are doing. 

See examples of TRPC and NextAuth 
- https://github.com/nextauthjs/next-auth-example/blob/main/app/api/auth/%5B...nextauth%5D/route.ts
- https://github.com/trpc/examples-next-prisma-starter/blob/main/src/pages/api/trpc/%5Btrpc%5D.ts

But If @jamespohalloran and yourself wan't to be more explicit I don't have an issue with it.

packages/tinacms-authjs/src/index.ts
```diff
@@ -150,7 +150,10 @@ const NextAuthAuthentication = ({
         })
       }
 
-      if (!session?.user?.name) {
+      if (!session?.user) {
```
kldavis4, line 153: should this maybe be writing a response or returning an error that the caller can use so that we can distinguish between 401 and 403 like we do here? https://github.com/tinacms/tinacms/blob/feat/self-hosted-improvements/packages/tinacms-authjs/src/index.ts#L107

Also isAuthenticated should probably be called isAuthorized

packages/@tinacms/datalayer/src/node.ts
logan-anderson: It was my understanding that `@tinacms/datalayer` would be that package. Is it still used on TinaCloud?

examples/tina-self-hosted-demo/pages/api/tina/[...routes].ts
Comment (kldavis4) [reply examples/tina-self-hosted-demo/pages/api/tina/[...routes].ts:22]: Here's the reference for this form of initialization in next-auth: https://next-auth.js.org/configuration/initialization#advanced-initialization

I think I prefer it in the more explicit form as we don't have to answer the question of how to modify the request object (or having a page like this to document it) but I'll defer to James if he thinks otherwise

packages/@tinacms/datalayer/src/node.ts
kldavis4: > It was my understanding that @tinacms/datalayer would be that package.

I don't recall us having any discussion about that, other than when were using @tinacms/datalayer to wrap up everything that was not open source (before we changed licensing)

It looks like only the older versions of @tinacms/graphql in tina cloud have the older @tinacms/datalayer dependency, so it it is probably ok to put the backend stuff here (though a bit confusing how it is morphing over time..). If we leave it there, I would move it to a sub folder called backend or api instead of leaving it at the root of the source tree.

packages/tinacms-authjs/src/index.ts
logan-anderson: Totally agree. I updated the name to `isAuthorized` it returns weather there was an error or not and the proper message and status code.

packages/tinacms-authjs/src/index.ts
```diff
@@ -133,10 +134,61 @@ const TinaCredentialsProvider = ({
       authenticate(databaseClient, credentials.username, credentials.password),
   })
 
+const AuthJsBackendAuthentication = ({
+  authOptions,
+}: {
+  authOptions: AuthOptions
+}) => {
+  const nextBackendAuthentication: BackendAuthentication = {
```
kldavis4, line 142:
```suggestion
  const backendAuthentication: BackendAuthentication = {
```

packages/tinacms-authjs/src/index.ts
```diff
@@ -133,10 +134,61 @@ const TinaCredentialsProvider = ({
...
+  }
+  return nextBackendAuthentication
+}
 export {
   TinaCredentialsProvider,
   TinaAuthJSOptions,
   withAuthJSApiRoute,
+  AuthJsBackendAuthentication,
   // TODO: This probably needs a better name?
   createAuthJSApiRoute,
```
kldavis4, line 193:
```suggestion
```

kldavis4: also remove `withAuthJSApiRoute` and `createAuthJsApiRoute`

packages/@tinacms/datalayer/src/node.ts
```diff
@@ -0,0 +1,154 @@
...
+      }
+    | {
+        isAuthorized: false
+        errorMessage: string
+        errorCode: number
+      }
+  >
+  extraRoutes?: {
+    [key: string]: {
+      isAuthRequired?: boolean
```
kldavis4, line 26:
```suggestion
      protected?: boolean
```
or `secure`

packages/@tinacms/datalayer/src/node.ts
```diff
@@ -0,0 +1,154 @@
+import type { IncomingMessage, ServerResponse } from 'http'
```
kldavis4, line 1: move `node.ts` to a subfolder like `backend` or `api`

packages/@tinacms/datalayer/src/node.ts
logan-anderson: @kldavis4 `protected` is a reserved word when using strict mode. [see this web page for more info](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Strict_mode#future-proofing_javascript). I think it makes more sense to go with `isProtected` or `secure`

packages/tinacms-authjs/src/index.ts
logan-anderson: removed

packages/@tinacms/datalayer/src/node.ts
logan-anderson: updated to use `secure`

packages/@tinacms/datalayer/src/node.ts
logan-anderson: Moved all of the code from `node.ts` into `backend/index.ts`. The root `node.ts` will need to stay there and re-export all of the code to make the import path easier to understand. 

Ex: `import { TinaNodeBackend } from '@tinacms/datalayer/dist/node'`  instead of `import { TinaNodeBackend } from '@tinacms/datalayer/dist/backend/index'`

kldavis4: I'm not sure I understand why we aren't just exporting `TinaNodeBackend` from index.ts so the import is `import { TinaNodeBackend } from '@tinacms/datalayer'`

kldavis4 (approved):

PR #4233, opened 2023-10-04
feat: custom next-auth api endpoint path

kldavis4:
support custom next-auth path (`/api/tina/auth/*`) to support consolidated api without rewrite


kldavis4:
@logan-anderson allows the next-auth client to use the /api/tina/auth/* endpoint

logan-anderson (approved):

PR #4247, opened 2023-10-12
Table support

jeffsee55:
Add support for basic markdown tables.

### Usage

```ts
// tina/config.ts
import `tinaTableTemplate` from `tinacms`

// add it to the rich-text template
  {
    type: 'rich-text',
    label: 'Body',
    name: '_body',
    templates: [
      tinaTableTemplate
    ///
```

Customize the `table`, `tr`, `th` and `td` fields in the `<TinaMarkdown>` component:

```tsx
<TinaMarkdown content={props.body} components={{
  th: (props) => <th className="bg-gray-100 font-bold" {...props} />,
  td: (props) => <td className="bg-gray-100" {...props} />,
}} />
```


logan-anderson:
Everything worked as expected for me. 

Is there any reason why we are not including table support out of the box? (it could be in the dropdown with title, bold, etc)

This would just be nice for folks that have tables already in there site. They would be able to edit tables right away.

jeffsee55:
> Is there any reason why we are not including table support out of the box

I expect we will, I just wanted to get it out in a somewhat opt-in manner for now while it's still new

logan-anderson (approved):

PR #4248, opened 2023-10-12
feat: warn nextauth custom provider

kldavis4:
Add a warning on backend api initialization when using a custom provider with the /api/tina/* catch-all

Demo: https://www.loom.com/share/af508f26eb6543e5a5012923e8724c6d?sid=cf83351e-99e6-4673-800f-de93443ce9e2

Note that this doesn't know whether a separate next-auth api handler has been configured, so the warning will show whenever the tina backend is initialized.


logan-anderson (approved):
Awesome!

PR #4280, opened 2023-10-23
add secure: true to gql endpoint

logan-anderson:
Make sure the gql endpoint as auth turned on.


kldavis4:
@logan-anderson does this fix the update password issue?

PR #4284, opened 2023-10-24
feat: cli init update tina config

kldavis4:
- update the backend init to update the tina config using the typescript compiler api
- fix a bug in the overwrite logic for the api routes

### DIFF
--- a/packages/@tinacms/cli/src/cmds/init/codegen/index.ts
+++ b/packages/@tinacms/cli/src/cmds/init/codegen/index.ts
@@ -356,15 +356,22 @@ export const addSelfHostedTinaAuthToConfig = async (
     fs.readFileSync(pathToConfig, 'utf8'),
     config.typescript ? ts.ScriptTarget.Latest : ts.ScriptTarget.ESNext
   )
+  const { configImports, configAuthenticationClass, extraTinaCollections } =
+    config.authenticationProvider
+
+  const importMap: Record<string, string[]> = {
+    // iterate over configImports and add them to the import map
+    ...configImports.reduce((acc, { from, imported }) => {
+      acc[from] = imported
+      return acc
+    }, {} as Record<string, string[]>),
+  }
 
   const transformedSourceFileResult = ts.transform(
     sourceFile,
     [
       makeImportsVisitor(sourceFile, {
-        'tinacms-authjs/dist/tinacms': [
-          'TinaUserCollection',
-          'UsernamePasswordAuthJSProvider',
-        ],
+        ...importMap,
         tinacms: ['LocalAuthProvider'],
       }),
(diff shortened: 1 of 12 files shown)


logan-anderson:
**Doesn't need fixed to be in this PR**

We need to generate the users content/users/index.json.

packages/@tinacms/cli/src/cmds/init/codegen/index.ts
```diff
@@ -0,0 +1,403 @@
...
+  const sourceFile = ts.createSourceFile(
+    pathToConfig,
+    fs.readFileSync(pathToConfig, 'utf8'),
+    config.typescript ? ts.ScriptTarget.Latest : ts.ScriptTarget.ESNext
+  )
+
+  const transformedSourceFileResult = ts.transform(
+    sourceFile,
+    [
+      makeImportsVisitor(sourceFile, {
```
logan-anderson, line 363: You _could_ get these values from ` config.authenticationProvider.configImports` instead of hard coding them.

packages/@tinacms/cli/src/cmds/init/codegen/index.ts
```diff
@@ -0,0 +1,403 @@
...
+        ...parseVariableStatement(
+          "const isLocal = process.env.TINA_PUBLIC_IS_LOCAL === 'true'"
+        )
+      ),
+      makeUpdateObjectLiteralPropertyVisitor(
+        sourceFile,
+        'defineConfig',
+        'authProvider',
+        ...parseExpression(
+          'isLocal ? new LocalAuthProvider() : new UsernamePasswordAuthJSProvider()'
```
logan-anderson, line 381: Same with this you _could_ get this value from ` config.authenticationProvider.configAuthenticationClass`

packages/@tinacms/cli/src/cmds/init/codegen/index.ts
```diff
@@ -0,0 +1,403 @@
...
+        )
+      ),
+      makeUpdateObjectLiteralPropertyVisitor(
+        sourceFile,
+        'defineConfig',
+        'contentApiUrlOverride',
+        ...parseExpression("'/api/tina/gql'")
+      ),
+      makeAddExpressionToSchemaCollectionVisitor(
```
logan-anderson, line 390: This _could_ be only applied if `config.authenticationProvider.name === 'next-auth'`

logan-anderson (approved):
This is amazing and blew my mind! I am **definitely** ok with merging as is (Was not able to find any issues in testing). Just left a couple of comments that you can optionally add if you want to make it more future proof for when we want to add more authentication providers to the init. (Like maybe clerk for example)

PR #4286, opened 2023-10-25
Fix clerk auth

logan-anderson:
Updates the `tinacms-clerk` package to use the new auth patterns

### DIFF
--- a/packages/tinacms-clerk/src/tinacms.ts
+++ b/packages/tinacms-clerk/src/tinacms.ts
@@ -1,0 +1,74 @@
+import { AbstractAuthProvider } from 'tinacms'
+
+import type Clerk from '@clerk/clerk-js'
+
+export class ClerkAuthProvider extends AbstractAuthProvider {
+  clerk: Clerk
+  allowedList?: string[]
+  constructor({
+    clerk,
+    allowedList,
+  }: {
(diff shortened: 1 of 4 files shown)


kldavis4 (approved):

packages/tinacms-clerk/src/tinacms.ts
```diff
@@ -0,0 +1,74 @@
...
+          socialButtonsBlockButton: 'px-4 py-2 border border-gray-100',
+          formFieldInput: `px-4 py-2`,
+          formButtonPrimary: 'bg-blue-600 text-white p-4',
+          formFieldInputShowPasswordButton: 'm-2',
+          dividerText: 'px-2',
+        },
+      },
+    })
+  }
+  async getUser() {
```
kldavis4, line 55: instead of doing this, we should implement authorize like here: https://github.com/tinacms/tina-cloud-starter-self-hosted-clerk/blob/kldavis4/clerk_users_collection/tina/config.tsx#L17

That way, you aren't immediately logged out and get a helpful not authorized message

logan-anderson: Updated

kldavis4 (approved):

PR #4287, opened 2023-10-26
add users/index.json file when doing init

logan-anderson:
When doing init and using self hosted, add a content/users/index.json file

### DIFF
--- a/packages/@tinacms/cli/src/cmds/init/apply.ts
+++ b/packages/@tinacms/cli/src/cmds/init/apply.ts
@@ -121,6 +121,25 @@ async function apply({
       config,
       generatedFile: env.generatedFiles['next-api-handler'],
     })
+    // add pages/api/tina/users.json file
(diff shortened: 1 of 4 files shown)


packages/@tinacms/cli/src/cmds/init/apply.ts
```diff
@@ -121,6 +121,25 @@ async function apply({
       config,
       generatedFile: env.generatedFiles['next-api-handler'],
     })
+    // add pages/api/tina/users.json file
```
kldavis4, line 124:
```suggestion
    // add content/users/index.json file
```

kldavis4 (approved):
one of the comments is wrong but otherwise looks good to me

logan-anderson:
> one of the comments is wrong but otherwise looks good to me

That was codepilot 😂 Good catch @kldavis4

kldavis4 (approved):

PR #4310, opened 2023-11-03
feat: self-hosting improvements

logan-anderson:
improvements to self hosted tina.

See the [changeset](https://github.com/tinacms/tinacms/blob/self-host/.changeset/calm-fireants-suffer.md) for a full list of changes


examples/kitchen-sink/tina/config.tsx
```diff
@@ -16,27 +16,26 @@ const router = ({ document, collection }) => {
...
-      },
-      authenticate: async () => {
-        localStorage.setItem(TINA_TOKEN_KEY, 'some-token')
-        return true
-      },
-      getUser: async () => {
-        return localStorage.getItem(TINA_TOKEN_KEY)
-      },
-    },
+    auth: customAuthProvider,
```
kldavis4, line 38: is this correct?

examples/tina-self-hosted-demo/tina/config.tsx
```diff
@@ -1,37 +1,26 @@
...
 import { ColorPickerInput } from '../components/fields/color'
 import { iconSchema } from '../components/util/icon'
+import {
+  TinaUserCollection,
+  UsernamePasswordAuthJSProvider,
+} from 'tinacms-authjs/dist/tinacms'
+
+const isLocal = process.env.TINA_PUBLIC_IS_LOCAL === 'true'
+// process.env.NEXTAUTH_URL = process.env.NEXT_PUBLIC_NEXTAUTH_URL
+// console.log(process.env.NEXTAUTH_URL)
```
kldavis4, line 16:
```suggestion
```

examples/kitchen-sink/tina/config.tsx
logan-anderson: I think this could be removed and it would work the same.

Passing an object to the admin provider is correct but a class is preferred. 

Should we remove he customAuthProvider from this example?

kldavis4: I think its fine as-is if it works - maybe better to set it via authProvider since admin.auth is deprecated, though

kldavis4 (approved):

PR #4315, opened 2023-11-07
fix self hosted types

logan-anderson:
Fix self hosted types so we don't get any errors in strict mode.


logan-anderson:
Tested in the SSG repo and everything seems to be working.

kldavis4 (approved):

PR #4324, opened 2023-11-10
Minor Improvements

logan-anderson:
A couple of small QOL updates that I noticed when going threw the demo with Scott.

- Updates verbiage on some of the questions (makes it a bit more clear with examples)
- removes looking at app router
   -  I realized that  if the user is using the app router, adding the pages router example will still work because both the app router and pages router can be used at the same time. We ran into an issue where the entire init backend would break if the pages router is being used.

### DIFF
--- a/packages/@tinacms/cli/src/cmds/init/apply.ts
+++ b/packages/@tinacms/cli/src/cmds/init/apply.ts
@@ -522,18 +522,22 @@ const logNextSteps = ({
 }) => {
   if (isBackend) {
     logger.info(focusText(`\n${titleText(' TinaCMS ')} backend initialized!`))
-    logger.info(
-      'Please add the following environment variables to your .env file'
-    )
-    logger.info(
-      indentText(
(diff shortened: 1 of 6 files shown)


packages/@tinacms/cli/src/cmds/init/apply.ts
```diff
@@ -538,11 +538,13 @@ const logNextSteps = ({
       )
       logger.info(indentText(envFileText))
     }
+    logger.info(
+      'Before you can start your site you will need to update your site to use the backend client.\nSee docs for more info: https://tina.io/docs/self-hosted/querying-data/'
```
kldavis4, line 542: "your site" is repetitive

logan-anderson: Update to "'Before you can run your site you will need to update to use the backend client"

kldavis4 (approved):

PR #4326, opened 2023-11-12
Allow each `AuthProvider` to render a custom login screen, not just redirect to a login page

coreyaus:
## Notes

- This is an extension to the latest in-progress work on the `self-host` branch
    - It includes small tweaks to the `tinacms` and `@tinacms/schema-tools` packages
- Given it's not a PR to the `main` branch, I haven't run `pnpm changeset` to generate a changeset file
    - If you're happy to merge these changes into the `self-host` branch then I assume they can be included as part of the changeset that PR will generate? Please let me know if that's incorrect and I can add a changeset - I just didn't think it made sense to attempt any version bumps when pushing into a feature branch other than `main`

## Feature description

**Add `getLoginScreen` to `AbstractAuthProvider`**

- This extends the existing `LoginStrategy` type to include a new `LoginScreen` option
- When the `LoginStrategy` is set to `LoginScreen` and a `getLoginScreen` function is provided to return a `React.ReactNode` the main `AuthWallInner` will display the custom login screen provided, rather than showing the modal popups and forcing a redirect or displaying the username and password form

## Next steps

If you're happy with this idea, I could help refactor the `UsernamePassword` login strategy to use the LoginScreen, which would make also make it more flexible to be extended in future (or for users to bring their own authentication form components)

**UPDATE:** I've opened a follow-up PR that demonstrates what it would look like to update the existing `UsernamePasswordAuthJSProvider` to use the login screen: #4328. I'd love to see the `getLoginScreen` function (or something similar) added to TinaCMS and really only created the second PR as a demo in case it's helpful - there's definitely no reason we couldn't add the `getLoginScreen` function but leave the `UsernamePasswordAuthJSProvider` form untouched

PR #4328, opened 2023-11-12
Use LoginScreen for UsernamePasswordAuthJSProvider, not modal (follows #4326)

coreyaus:
## Notes

- Follows on from #4326 
- This relates to the `self-host` branch that @logan-anderson and @kldavis4 have been developing
- Let me know if you'd like me to add a changeset - I assumed it wasn't necessary given it's a PR suggestion into the unreleased `self-host` feature branch but if that's incorrect I can easily add a changeset
- I'd love to see the `getLoginScreen` function from PR 4326 (or something similar) added to TinaCMS and really only created this second PR as a demo in case it's helpful
    - Definitely no worries at all if you'd prefer to close this PR and leave the `UsernamePasswordAuthJSProvider` form untouched, but hopefully you'll consider the `getLoginScreen` function from PR 4326 regardless 🤞  

## Feature description

- This refactors the form code for the `UsernamePasswordAuthJSProvider` so it now appears on a login screen, rather than appearing in a modal
- It also includes a suggestion to hide the TinaCMS sidebar until the user has signed into the admin portal, though this second commit could easily be reverted if you'd prefer to always show the TinaCMS sidebar alongside the login screen

image

### DIFF
--- a/packages/@tinacms/schema-tools/src/types/index.ts
+++ b/packages/@tinacms/schema-tools/src/types/index.ts
@@ -1,4 +1,3 @@
-import type { FC } from 'react'
 import type React from 'react'
 
 type Meta = {
@@ -442,6 +441,12 @@ type TokenObject = {
 
 export type LoginStrategy = 'UsernamePassword' | 'Redirect' | 'LoginScreen'
 
+type LoginScreenProps = {
+  handleAuthenticate: () => Promise<void>
+  authProps: Record<string, string>
+  setAuthProps: React.Dispatch<React.SetStateAction<Record<string, string>>>
+}
+
 export interface AuthProvider {
   /**
    *  Used for getting the token from the custom auth provider
@@ -480,8 +485,8 @@ export interface AuthProvider {
   isAuthorized: (context?: any) => Promise<boolean>
   isAuthenticated: () => Promise<boolean>
   getLoginStrategy: () => LoginStrategy
-  getLoginScreen: () => FC | null
-  getSessionProvider: () => FC<{ basePath?: string }>
+  getLoginScreen: () => React.FC<LoginScreenProps> | null
+  getSessionProvider: () => React.FC<{ basePath?: string }>
 }
 
 interface AuthHooks {
@@ -876,7 +881,7 @@ export type Option =
   | string
   | {
       label?: string
-      icon?: FC
+      icon?: React.FC
       value: string
     }
 
--- a/packages/@tinacms/schema-tools/src/types/index.ts
+++ b/packages/@tinacms/schema-tools/src/types/index.ts
@@ -439,12 +439,9 @@ type TokenObject = {
   refresh_token?: string
 }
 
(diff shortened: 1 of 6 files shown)


kldavis4:
@coreyaus what is the context for this change?

coreyaus:
Hey @kldavis4 

Thanks for taking a look a this - I'll explain below but please let me know if anything is unclear

### Initial PR (4326): Allowing each `AuthProvider` to use the `loginScreen` feature

- The easier and more useful place to start is probably #4326. I created that PR because I noticed the `AuthWallInner` component exposed an optional `loginScreen` prop that was never actually being used and I realised it could provide a really handy feature for anyone creating their own AuthProvider (e.g. I've created my own for SupabaseAuth)
- So in that first PR I added a `getLoginScreen` function to the `AuthProvider` class, modelled off the `getSessionProvider` function.
    - That would allow us to display a login screen directly within the Tina Admin Portal, rather than always redirecting to a separate `/login` route (which can be a little jarring)
- I think PR 4326 is a useful (and simple) feature addition that would add value to people creating their own AuthProvider, and also simplify the process of doing so. I'd love to see that one get merged (while caring less about this current PR).

### This current PR: Using the `loginScreen` approach for the `UsernamePasswordAuthJSProvider`

As an optional extension (in case it's useful but no dramas if not), I added this follow up PR to demonstrate how the existing `UsernamePassword` provider could be refactored to use a login screen. In part it stands to serve as a demo in case the initial PR (that I actually care more about) was unclear or too abstract (it only changes a few lines of code and in the interests of brevity doesn't explicitly show an example implementation of how the `getLoginScreen` method would be used)

- This current PR doesn't make any functionality changes (besides the small item in the final dot point below), but aims to bring small some quality of life improvements to the code, particularly if you need to add other login strategies in the future (e.g. extensions to the username and password provider to support multi-factor authentication or whatever else people dream up). Obviously that's subjective though so feel free to dismiss these attempted improvements if they're not useful!
    - You can see a couple of example simplifications this PR makes in the screenshots below (though obviously happy to defer to others on whether they're actually improvements - very much appreciate this is your code and you might have different reasons for wanting the forms to appear in a modal!)
- The main difference is that the username and password form now uses the existing `loginScreen` feature of the `AuthWallInner` component (which was previously unused), rather than opening the forms in a modal.
- It would also make it possible for developers to use the `UsernamePasswordAuthJSProvider` but supply their own custom login form (e.g. if they have a preferred design for that form).

As a final point, [this commit](https://github.com/tinacms/tinacms/pull/4328/commits/5d902d33fbe2573031202235c9148e58d2ef9b6d) adds a couple of lines to hide the empty TinaCMS sidebar from the admin portal until the user has authenticated (a separate but related idea). The main result is that the final login page UX is made to feel more like a traditional app sign in experience.

**Nested conditional logic is no longer required**

image

**Form logic and design can now live in its own component**

This code now lives in its own `UsernamePasswordLoginScreen.tsx` component, rather than appearing in the generic `AuthWallInner` component:

image

PR #4344, opened 2023-11-17
fix: remove namespace from db adapter

kldavis4:
namespace is only specified when calling createDatabase and not set on the adapter

### DIFF
--- a/packages/@tinacms/cli/src/cmds/init/prompts/databaseAdapter.ts
+++ b/packages/@tinacms/cli/src/cmds/init/prompts/databaseAdapter.ts
@@ -10,7 +10,6 @@ const supportedDatabaseAdapters: {
 } = {
   ['upstash-redis']: {
     databaseAdapterClassText: `new RedisLevel({
-        namespace: branch,
         redis: {
           url: process.env.KV_REST_API_URL || 'http://localhost:8079',
           token: process.env.KV_REST_API_TOKEN || 'example_token',
@@ -27,8 +26,7 @@ const supportedDatabaseAdapters: {
   },
   mongodb: {
     databaseAdapterClassText: `new MongodbLevel({
-          // If you are not using branches you could pass a static collection name. ie: "tinacms"
-          collectionName: \`tinacms-\${branch}\`,
+          collectionName: 'tinacms',
           dbName: 'tinacms',
           mongoUri: process.env.MONGODB_URI,
         })`,


logan-anderson:
@kldavis4  What happens if the namespace is not passed? Will all indexes go to the same namespace?

kldavis4:
> @kldavis4 What happens if the namespace is not passed? Will all indexes go to the same namespace?

There is a default namespace (for createDatabase)

logan-anderson:
> > @kldavis4 What happens if the namespace is not passed? Will all indexes go to the same namespace?
> 
> There is a default namespace (for createDatabase)

If you are not using the namespace, switched branches,  changed content, and pushed up to Github it would build and over ride the content on `main` (Production deployment) Correct? 

I think it might be nice to keep passing the branch name as the namespace as it would avoid this issue.

kldavis4:
> > > @kldavis4 What happens if the namespace is not passed? Will all indexes go to the same namespace?
> > 
> > 
> > There is a default namespace (for createDatabase)
> 
> If you are not using the namespace, switched branches, changed content, and pushed up to Github it would build and over ride the content on `main` (Production deployment) Correct?
> 
> I think it might be nice to keep passing the branch name as the namespace as it would avoid this issue.

I think you are misunderstanding the change - we already pass namespace to the createDatabase function. See: https://github.com/tinacms/tinacms/blob/self-host/examples/tina-self-hosted-demo/tina/database.ts#L21

If you set the collectionName to the branch on the mongo adapter, the user collection will be different for each branch.

logan-anderson:
> > > > @kldavis4 What happens if the namespace is not passed? Will all indexes go to the same namespace?
> > > 
> > > 
> > > There is a default namespace (for createDatabase)
> > 
> > 
> > If you are not using the namespace, switched branches, changed content, and pushed up to Github it would build and over ride the content on `main` (Production deployment) Correct?
> > I think it might be nice to keep passing the branch name as the namespace as it would avoid this issue.
> 
> I think you are misunderstanding the change - we already pass namespace to the createDatabase function. See: https://github.com/tinacms/tinacms/blob/self-host/examples/tina-self-hosted-demo/tina/database.ts#L21
> 
> If you set the collectionName to the branch on the mongo adapter, the user collection will be different for each branch.

Gotcha, I thought that the namespace was done at the database / collection level. My bad

logan-anderson (approved):

kldavis4:
> > > > > @kldavis4 What happens if the namespace is not passed? Will all indexes go to the same namespace?
> > > > 
> > > > 
> > > > There is a default namespace (for createDatabase)
> > > 
> > > 
> > > If you are not using the namespace, switched branches, changed content, and pushed up to Github it would build and over ride the content on `main` (Production deployment) Correct?
> > > I think it might be nice to keep passing the branch name as the namespace as it would avoid this issue.
> > 
> > 
> > I think you are misunderstanding the change - we already pass namespace to the createDatabase function. See: https://github.com/tinacms/tinacms/blob/self-host/examples/tina-self-hosted-demo/tina/database.ts#L21
> > If you set the collectionName to the branch on the mongo adapter, the user collection will be different for each branch.
> 
> Gotcha, I thought that the namespace was done at the database / collection level. My bad

Yeah, it used to be until the user collection was added

PR #4349, opened 2023-11-17
Allow auth providers to display a custom login form in the Tina admin portal

coreyaus:
- This change allows each `AuthProvider` to display a custom login form within the Tina admin portal, rather than showing the modal popups and forcing a redirect or displaying the default username and password form. This will hopefully simplify the process of creating custom auth providers and handling user authentication when self-hosting
- On a technical level this PR:
    - extends the existing `LoginStrategy` type to include a new `LoginScreen` option
    - adds a `getLoginScreen` function to `AbstractAuthProvider`, which will return the login form
- This PR replaces the original version here: https://github.com/tinacms/tinacms/pull/4326

### DIFF
--- a/.changeset/loud-panthers-check.md
+++ b/.changeset/loud-panthers-check.md
@@ -1,0 +1,6 @@
+---
+'@tinacms/schema-tools': patch
+'tinacms': patch
+---
+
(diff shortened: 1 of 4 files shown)


kldavis4:
@coreyaus I made a few tweaks / fixes in a separate PR based on this one: #4361 - going to close this one in favor of that one

PR #4388, opened 2023-12-02
Support custom VitePlugins in the Tina config

coreyaus:
## Notes

- This is intended to provide advanced ways to customise the Admin Portal build output.
- One example use case is to provide a custom tailwind plugin (similar to the TinaTailwind plugin) and build CSS for any custom React components we load in the Tina Admin portal via Tina field plugins.
    - Otherwise the Tailwind classes we use in Tina field plugins do not appear in the Tina Admin portal CSS.
- Hopefully this will help requests like the one here: https://github.com/tinacms/tinacms/discussions/3912


coreyaus:
Not stale 🙂

PR #4395, opened 2023-12-11
Add ContentType to headers sent to AWS S3

arzafran:
ContentType header was missing in `next-tinacms-s3` package, making AWS serve SVGs as `application/octet-stream`.

This PR solves it by pulling the `mimetype` from the file and if it's not available it fallbacks to `application/octet-stream`


pnpm-lock.yaml
kldavis4, line 1: there should be no changes to the lock file for this PR

packages/next-tinacms-s3/src/handlers.ts
```diff
@@ -3,21 +3,21 @@
...
 } from '@aws-sdk/client-s3'
-import { Media, MediaListOptions } from 'tinacms'
-import path from 'path'
 import fs from 'fs'
-import { NextApiRequest, NextApiResponse } from 'next'
 import multer from 'multer'
+import { NextApiRequest, NextApiResponse } from 'next'
+import path from 'path'
+import { Media, MediaListOptions } from 'tinacms'
 import { promisify } from 'util'
```
kldavis4, line 21: This looks like just re-ordering - if that's the case, please revert these changes

kldavis4 (changes requested):

kldavis4:
Moved these changes to https://github.com/tinacms/tinacms/pull/4742

PR #4408, opened 2023-12-29
add rtl support for field labels and descriptions

saharzal:
We are using tinacms for our site and because our native language is **persian** ,our content has **RTL** direction.  but the dashboard does not look good enough when we have rtl content.
In this PR i've added rtl support for labels ,descriptions and also in value of textfields and textareas.

before these changes, admin dashboard fields looked like this:

before

here is how result looks like:

 please tell me if you think there is better way to improve this issue.

### DIFF
--- a/packages/tinacms/src/toolkit/fields/components/text-field.tsx
+++ b/packages/tinacms/src/toolkit/fields/components/text-field.tsx
@@ -1,4 +1,5 @@
 import * as React from 'react'
+import { getDirection } from '../field-utils'
 
 type a = React.DetailedHTMLProps<
   React.InputHTMLAttributes<HTMLInputElement>,
@@ -23,6 +24,7 @@ export const BaseTextField = React.forwardRef<
       className={`${textFieldClasses} ${
         disabled ? disabledClasses : ''
       } ${className}`}
+      dir={getDirection(rest.value)}
       {...rest}
     />
   )
--- a/packages/tinacms/src/toolkit/fields/components/textarea.tsx
+++ b/packages/tinacms/src/toolkit/fields/components/textarea.tsx
@@ -1,4 +1,5 @@
 import * as React from 'react'
+import { getDirection } from '../field-utils'
 
 type a = React.DetailedHTMLProps<
   React.InputHTMLAttributes<HTMLInputElement>,
@@ -19,6 +20,7 @@ export const TextArea = React.forwardRef<
       className="shadow-inner text-base px-3 py-2 text-gray-600 resize-y focus:shadow-outline focus:border-blue-500 block w-full border border-gray-200 focus:text-gray-900 rounded-md"
       ref={ref}
       style={{ minHeight: '160px' }}
+      dir={getDirection(props.value)}
     />
   )
 })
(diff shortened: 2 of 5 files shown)


saharzal:
@jeffsee55
@kldavis4

hello,
It's just a reminder to check whether this PR can be merged or not and if not how we can fix that.
I was wondering if I am missing a step to get a review, checked the contribution.md, I didn't see anything about the PR process.

Thanks in advance

packages/tinacms/src/toolkit/fields/components/text-field.tsx
```diff
@@ -23,6 +24,7 @@ export const BaseTextField = React.forwardRef<
       className={`${textFieldClasses} ${
         disabled ? disabledClasses : ''
       } ${className}`}
+      dir={getDirection(rest.value)}
```
JackDevAU, line 27: I don't think we need this since we are setting the direction in the `wrap-field-with-meta.tsx` file.

packages/tinacms/src/toolkit/fields/components/textarea.tsx
```diff
@@ -19,6 +20,7 @@ export const TextArea = React.forwardRef<
       className="shadow-inner text-base px-3 py-2 text-gray-600 resize-y focus:shadow-outline focus:border-blue-500 block w-full border border-gray-200 focus:text-gray-900 rounded-md"
       ref={ref}
       style={{ minHeight: '160px' }}
+      dir={getDirection(props.value)}
```
JackDevAU, line 23: Same with this one. This shouldn't be needed.

packages/tinacms/src/toolkit/fields/field-utils.ts
```diff
@@ -0,0 +1,12 @@
...
+}
+
+export const getDirection = (
+  value: number | string | boolean | readonly string[]
+) => {
+  if (typeof value === 'string') {
+    return hasRTLChar(value) ? 'rtl' : 'ltr'
+  }
+  return 'unset'
+}
```
JackDevAU, line 12: Can we instead use the browser `Intl.Locale` to do the work for us? 

Something like: 

```
/*
Get the navigator language. Used to Auto format Fields direction based on the navigator language.
*/
export const getNavigatorLanguage = () =>
  navigator.languages && navigator.languages.length
    ? navigator.languages[0]
    : navigator.language || 'en'

/*
Parse the fields.lang to get the direction of the text.
*/
export const getDirection = (lang: string) => {
  const locale = new Intl.Locale(lang)
  //@ts-ignore
  return locale.textInfo.direction
}

```

Then we could add a new field config called `lang` that uses https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/Locale/language to select what direction. Off the top of my head, this could have a default set as `en`, then another options as `auto` to use the users Browser default or a specific language you set.

JackDevAU (changes requested):
Awesome start! 

I've left some comments 🙏 If you aren't already, I highly recommend joining the Discord and we can talk more in detail there!

PR #4409, opened 2023-12-30
Fix inline rich-text templates with nested object or list fields

### LINKED ISSUE #4074: Unable to edit a template with list: true inside of rich text

When using a template with list: true inside of a rich text we are unable to edit it

### Reproduction

https://www.loom.com/share/84b8826968a44f49becdc834dc5e8d64

### Steps to reproduce

1. Add the a template that looks like the following to any rich text section.
```ts
          {
                name: "imagesSection",
                label: "ImagesSection",
                fields: [
                  {
                    name: "images",
                    label: "Images",
                    type: "object",
                    list: true,
                    fields: [
                      {
                        type: "string",
                        label: "Image",
                        name: "image",
                      },
                    ],
                  },
                ],
              },
```

2. Try and edit the imagesSection and see that you get stuck in a loop

### System Info

```shell
tina-cloud-starter with latest TinaCMS versions
```

### Validations

coreyaus:
## Summary

This PR fixes an issue where it was impossible to navigate to nested fields (e.g. object fields or list items) when editing the values for a rich-text template when that template was configured to be `inline: true`

https://www.loom.com/share/085f1aa40af742a5a7d8b8168760c52b

## Additional notes

- This PR resolves #4074 and adds tests to prevent regressions
    - I think the bug from the original issue was resolved for non-inline rich-text templates by this PR: #4044. However, the bug remained for inline rich-text templates
- Specifically the issue broke the editing process for rich-text templates that:
    - were designated to be `inline: true`
    - had a nested list field OR a nested object field

### DIFF
--- a/.changeset/twenty-dodos-protect.md
+++ b/.changeset/twenty-dodos-protect.md
@@ -1,0 +1,5 @@
+---
+'tinacms': minor
+---
+
+Fixes an issue where it was impossible to navigate to nested fields (e.g. object fields or list items) when editing the values for a rich-text template when that template was configured to be inline: true
--- a/packages/tinacms/src/toolkit/forms/form.test.ts
+++ b/packages/tinacms/src/toolkit/forms/form.test.ts
@@ -217,4 +217,187 @@ describe('Form', () => {
       })
     })
   })
+  describe('#getActiveField', () => {
+    describe('when called without a field name', () => {
+      it('returns the form itself', async () => {
+        const form = new Form(DEFAULTS)
+        const fieldName = null
+        expect(form.getActiveField(fieldName)).toEqual(form)
+      })
+    })
+
+    describe('when called for an object field', () => {
+      it('returns the correct template with namespaced field names', async () => {
+        const form = new Form({
+          ...DEFAULTS,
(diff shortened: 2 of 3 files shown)


coreyaus:
In case it's useful for reference, here are the inline rich-text templates used as examples in the screen recording above

**Example rich-text template with a nested object field**

```ts
{
  inline: true,
  name: "Underline",
  label: "Underline",
  fields: [
    {
      isTitle: true,
      type: "string",
      label: "Text",
      name: "text",
      required: true,
    },
    {
      type: "object",
      label: "Some object",
      name: "objectfield",
      fields: [
        {
          type: "string",
          label: "A text field",
          name: "objecttextfield",
        },
      ],
    },
  ],
},
```

**Example rich-text template containing a list field**

```ts
{
  name: "imagesSection",
  label: "ImagesSection",
  inline: true,
  fields: [
    {
      name: "images",
      label: "Images",
      type: "object",
      list: true,
      fields: [
        {
          type: "string",
          label: "Image",
          name: "image",
        },
      ],
    },
  ],
}
```

jeffsee55 (approved):

PR #4493, opened 2024-04-18
Enable customizable tsconfig file inside tina folder

### LINKED ISSUE #4488: 🐛Vite - Internal server error: Failed to resolve import

I'm attempting to use aliases for imports, but I'm encountering an issue where Vite is unable to resolve the paths when generating `config.prebuild.jsx`.

### Reproduction

https://github.com/amankumarrr/tina-module-alias-issue

## From Reproduction URL

- `yarn run`

## Steps to Reproduce from a new Starter repo  
- Add aliases in tsconfig (https://nextjs.org/docs/app/building-your-application/configuring/absolute-imports-and-module-aliases)
- Update the imports path (i.e from `../../components/blocks/hero` to `@/components-blocks/hero`) in page.ts
- `yarn run`

### System Info

```shell
Windows 11
Browser
- Chrome 
- Edge
```
**Screenshot:**
image

**Figure: Error getting while importing templates using alias** 

### Validations

johnxu16:
Fix #4488 

Use tsconfig inside tina folder if it exists, so using customizable alias is possible

### DIFF
--- a/packages/@tinacms/cli/src/next/config-manager.ts
+++ b/packages/@tinacms/cli/src/next/config-manager.ts
@@ -24,6 +24,7 @@ export class ConfigManager {
   tinaFolderPath: string
   isUsingLegacyFolder: boolean
   tinaConfigFilePath: string
+  tinaTsConfigFilePath: string
   tinaSpaPackagePath: string
   contentRootPath?: string
   envFilePath: string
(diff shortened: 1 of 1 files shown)


jeffsee55:
I don't think our build logic will be able to pick this up, we'd also need to expose the vite config for this, which is probably a bit more nuanced.

PR #4544, opened 2024-05-24
fix watch script for window environment

Ben0189:
Fix watch script for window environment to run "pnpm watch command"

### DIFF
--- a/examples/tina-self-hosted-demo/package.json
+++ b/examples/tina-self-hosted-demo/package.json
@@ -3,7 +3,7 @@
   "version": "0.1.49",
   "private": true,
   "scripts": {
-    "dev": "tinacms dev -c \"next dev\"",
+    "dev": "TINA_PUBLIC_IS_LOCAL=true tinacms dev -c \"next dev\"",
     "dev:prod": "tinacms dev --verbose -c \"next dev\"",
     "build": "tinacms build && next build",
     "start": "tinacms build && next start",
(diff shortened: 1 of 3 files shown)


examples/tina-self-hosted-demo/package.json
```diff
@@ -3,7 +3,7 @@
   "version": "0.1.49",
   "private": true,
   "scripts": {
-    "dev": "TINA_PUBLIC_IS_LOCAL=true tinacms dev -c \"next dev\"",
+    "dev": "tinacms dev -c \"next dev\"",
```
JackDevAU, line 6: Just to double check - do we need to update the docs or add this to a `.nev` file `TINA_PUBLIC_IS_LOCAL=true`
I believe this might still be needed

packages/@tinacms/scripts/src/index.ts
```diff
@@ -116,25 +116,25 @@ const watch = () => {
...
-        {
-          ignored: ['**/spec/**/*', 'node_modules'],
-        }
+        { ignored: ['**/spec/**/*', 'node_modules'] }
       )
       .on('change', async (path) => {
         const changedPackagePath = watchPaths.find((p) => path.startsWith(p))
         await run({ dir: changedPackagePath })
-        // console.log('change', changedPackagePath)
+        // console.log('change', changedPackagePath);
```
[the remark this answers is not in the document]
Ben0189: Done removed

examples/tina-self-hosted-demo/package.json
Ben0189: the better approach will be remove this change and update the docs, so that window user will not have issue when running the self hosted demo

Changes reverted

JackDevAU (approved):
🚀

PR #4559, opened 2024-06-03
🛠️ Adds ability to re-build on tagged label

JackDevAU:
**Hey Team,** 

We have the ability to publish a build to npm if we add the label 'tagged'. However, currently we have to push a new commit after we add the 'tagged' label to actually trigger this. 

With this change, as soon as we add the label it will now re-build

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -9,7 +9,6 @@ on:
 
 jobs:
   build:
-    if: github.ref == 'refs/heads/main' || contains(github.event.pull_request.labels.*.name, 'tagged')
     strategy:
       matrix:
         os:
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -5,10 +5,11 @@ on:
     branches:
       - main
   pull_request:
-    types: [opened, synchronize, reopened]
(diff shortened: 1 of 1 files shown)


jeffsee55 (approved):
💡

wicksipedia (approved):

.github/workflows/main.yml
```diff
@@ -5,10 +5,11 @@ on:
     branches:
       - main
   pull_request:
-    types: [opened, synchronize, reopened]
+    types: [opened, synchronize, reopened, labeled]
 
 jobs:
   build:
+    if: github.ref == 'refs/heads/main' || contains(github.event.pull_request.labels.*.name, 'tagged')
```
wicksipedia, line 12: This will prevent the build from running unless the pr is tagged
We currently use these builds as required checks
Maybe just apply the logic to the publish step?

JackDevAU: As per our conversation - we are going to remove this as the GitHub doesn't have a limit on build minutes on public repos 🚀

JackDevAU, line 12:
```suggestion

```

wicksipedia (approved):
🚀

PR #4579, opened 2024-06-25
Remove yarn for pnpm

JackDevAU:
- Remove yarn from `package.json` files and replace with pnpm
- Disable cypress e2e tests as we move to Playwright 

Cypress Tests will be re-added back via: https://github.com/tinacms/tinacms/issues/4580


.github/workflows/e2e.yml
```diff
@@ -1,42 +1,73 @@
...
-          path: cypress/screenshots
-      # Test run video was always captured, so this action uses "always()" condition
-      - uses: actions/upload-artifact@v2
-        if: always()
-        with:
-          name: cypress-videos
-          path: cypress/videos
+      - name: Skip Cypress Tests
+        run: |
+          echo "Skipping Cypress tests as we transition to a new playwright"
```
ncn-ssw, line 16: **suggestion**: Should a link be included to #4580?
```suggestion
          echo "Skipping Cypress tests as we transition to Playwright as per https://github.com/tinacms/tinacms/issues/4580"
```

ncn-ssw (dismissed):

.github/workflows/e2e.yml
JackDevAU: ✅ ty

wicksipedia (approved):
lgtm

PR #4590, opened 2024-06-26
Fix issue where inline nested jsx wasn't properly parsed

jeffsee55:
TLDR; this PR improves handling rich-text inside inline element props but probably isn't a great path forward. We will likely break this behavior in the future.

This PR add support for nesting rich-text blocks inside inline elements. It does so with some admittedly awkward syntax. Essentially we'd need to ensure linebreaks are escaped so that when we save a JSX prop it doesn't break onto another line, which would break us out of our inline node.

Here's an example https://github.com/tinacms/tinacms/pull/4590/files#diff-8006d011f95ff6235e039043ede59c7631bb1682647163353ce8cd92e55384f1R1 of what that looks like.

Some of us discussed this and it's _not_ something we want to promote as a happy path for using Tina because we're going against the grain pretty hard here with what we get from our markdown parser. We will likely roll some of the ideas for improvements into a larger bucket of general improvements needed for the rich-text editor. Specifically in this case, we discussed storing nested rich-text inside a frontmatter key and just referencing it. I'll leave that for a later RFC.


packages/@tinacms/mdx/src/next/tests/mdx-inline-nested/field.ts
```diff
@@ -0,0 +1,43 @@
...
+              fields: [
+                {
+                  name: 'title',
+                  label: 'Short Description',
+                  type: 'string',
+                  required: true,
+                },
+              ],
+            },
+          ],
```
eduard-nz, line 38: This looks really good. I understand that this would be a different issue, so just as FYI: This content structure will fail in the editor when clicking on the nested `dropBox`. The first one is fine, the 2nd will throw a lot of JS errors in the browser console and render a white screen.

packages/@tinacms/mdx/src/parse/mdx.ts
```diff
@@ -63,6 +63,15 @@ export function mdxJsxElement(
...
     )
+    template.fields.forEach((field) => {
+      if (field.type === 'rich-text') {
+        const value = props[field.name]
+        if (typeof value !== 'string') {
+          // props[field.name] = toTinaMarkdown({ type: 'root', children: value }, field)
+        }
+        console.log(props[field.name])
+      }
+    })
```
eduard-nz, line 74: This doesn't seem to do anything (other than debug info)? I might be wrong - but maybe this code block shouldn't make it back to `main` ...

jeffsee55: Yep! Thanks for the callout, removed.

Ben0189 (dismissed):
LGTM, i did some clean up for the experimental-example projects, and added component for dropbox so the UI is happy😊

JackDevAU:
Closing this as a new PR for this issue was opened https://github.com/tinacms/tinacms/pull/4700

PR #4595, opened 2024-06-28
🔧 Add fallback token to publish workflow

JackDevAU:
Attempt to fix: https://github.com/tinacms/tinacms/actions/runs/9708034175/job/26794165637


wicksipedia (approved):
not sure this will make a difference... the fallback tokens have been added to steps that only run on `latest`, which should have the app token anyway

PR #4625, opened 2024-07-10
🐛 TinaCMS - Collections - Fix header styling for smaller screens

### LINKED ISSUE #4613: 💄 Collections - Fix header UI on mobile screens
Breaking down #4473 . The collections list page header isn't formatted properly for mobiles.

- [x] Fix padding with burger icon
- [x] Fix filtering section styling

Image
**Figure: Collections list page header not displaying properly on mobile**

Jord-Gui:
Fixes #4613. 
Also fixed some pages that had the h-screen issue and wasn't rendering properly on mobile.

image

**Figure: Collections header on xs screens (320px - 559px)**

image

**Figure: Collections header on sm screens (560px - 719px)**

image

**Figure: Collections header on md screens (720px - 1029px)**

image

**Figure: Collections header on lg screens (1030px - 1279px)**

image

**Figure: Collections header on xl screens (1280px + )**

### DIFF
--- a/packages/tinacms/src/admin/components/Sidebar.tsx
+++ b/packages/tinacms/src/admin/components/Sidebar.tsx
@@ -36,7 +36,7 @@ const Sidebar = ({ cms }: { cms: TinaCMS }) => {
   const [menuIsOpen, setMenuIsOpen] = React.useState(false)
 
   const isLocalMode = cms.api?.tina?.isLocalMode
-  const navBreakpoint = 1000
+  const navBreakpoint = 1029
   const windowWidth = useWindowWidth()
   const renderDesktopNav = windowWidth > navBreakpoint
   const activeScreens = screens.filter(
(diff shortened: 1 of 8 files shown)


packages/tinacms/src/admin/components/Sidebar.tsx
```diff
@@ -36,7 +36,7 @@ const Sidebar = ({ cms }: { cms: TinaCMS }) => {
   const [menuIsOpen, setMenuIsOpen] = React.useState(false)
 
   const isLocalMode = cms.api?.tina?.isLocalMode
-  const navBreakpoint = 1000
```
JackDevAU, line 39: Should we consider setting this breakpoint in a configuration file and then using it in the Tailwind config to ensure consistency with our existing breakpoints like "sm," "md," and "lg"? Or is this localised to this file only?

JackDevAU (approved):
Awesome stuff! 🚀

(I've left a comment - but I think it could be something that we can refactor later on)

packages/tinacms/src/admin/components/Sidebar.tsx
Jord-Gui: Yeah its localised to this file only. I was thinking of moving it to css but that would require extra refactoring effort

PR #4628, opened 2024-07-10
💄 UI/UX - modify form saved indicator

wicksipedia:
- removed 'no changes' text as it was misleading if you were working on a branch with some saved commits
- move save/unsaved indicator to the right
- change 'unsaved' indicator color to red to make it more prominent

CleanShot 2024-07-10 at 11 30 44


ncn-ssw (approved):
🦙 Performed correctly in manual testing

Issue #4363, opened 2023-11-21
Edge Runtime Does Not Work With TinaCMS Self Hosted.


logan-anderson:
### Describe the issue that you're seeing. Any Loom videos or screenshots usually help a lot!

Currently you can use [the tina client](https://tina.io/docs/features/data-fetching/) (with Tina Cloud) in a cloudflare worker / edge runtime because it only depends on fetch but if you wan't to use [self hosted TinaCMS](https://tina.io/docs/self-hosted/overview/) you will need to add NodeJS dependencies that will not work in the Edge Runtime. 

There are two pieces of work to get TinaCMS self hosted  working with an an edge runtime.

## 1. The database client and `@tinacms/graphql` have Node.js Dependencies 

The database client depends on `@tinacms/datalayer` and `@tinacms/graphql` both which have NodeJS dependencies. 

We would have to move all Node.js dependencies to `@tinacms/cli` 

## 2. A `TinaNodeBackend`  that uses Javascript Request and Response.

Currently to [host the graphql / auth backend](https://tina.io/docs/reference/self-hosted/tina-backend/nextjs/) a route helper is exported. We would need to add another version of this handler built on  Javascript  `Request` and `Response`.

### Reproduction

https://github.com/logan-anderson/nextjs-cloudflare/tree/self-host

### Steps to reproduce

1. Clone the repo
2. Check out the `self-host` branch
3. See errors

### System Info

```shell
N/A (All latest TinaCMS packages)
```

### Validations

- [X] Follow our [Code of Conduct](https://github.com/tinacms/tinacms/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [docs](https://tina.io/docs).
- [X] Check that there isn't [already an issue](https://github.com/tinacms/tinacms/issues) that reports the same bug.

logan-anderson:
Not stale

kulesy:
Closing — this has been labeled `wontfix` and open since November 2023. Part of backlog refinement.

PR #4629, opened 2024-07-10
Support edge runtime

fgnass:
This PR continues the effort started by @logan-anderson in #4363 and allows self-hosted TinaCMS to run on edge-runtimes.

# List of changes

## Split up @tinacms/graphql database

The `packages/@tinacms/graphql/src/database/index.ts` file is split up into `remote.ts` and `local.ts`, with `local.ts` containing all the node-specific code that accesses the file system and `remote.ts` containing all the platform-independent code.

## Add worker export to @tinacms/graphql

A new entry is added to the exports map in `package.json` with the `"worker"` key to target edge runtimes. It exports `index-edge.ts` which in turn exports the remote database part plus the `resolve` function.

## Add worker export to @tinacms/datalayer

A new entry is added to the exports map in `package.json` with the `"worker"` key to target edge runtimes. It re-exports `resolve` and `createDatabase` from `@tinacms/graphql` and provides a stub for `createLocalDatabase` that throws an Error when invoked.

## Add edge bundle to @tinacms/mdx

The build script for `@tinacms/mdx` is modified to create yet another bundle flavor, which again targets edge runtimes.

## Lax Level type

The `Level` type alias in `packages/@tinacms/graphql/src/database/level.ts` now no longer enforces a specific `TFormat`, which is the type that level implementations use internally to store data. This change allows use implementations like [d1-level](https://www.npmjs.com/package/d1-level) which uses strings internally.

# Notes

The new package exports still contain imports of Node built-ins like `url` or `process` that need to be polyfilled by the consumer. My first approach was to bundle all dependencies and include them in the `@tinacms/datalayer` edege export (just like `@tinacms/mdx` does), but I decided against this so that packages like `abstract-level` can be shared between tinacms and the consumer and don't end up twice in the final worker. In case of Cloudflare pages this can be achieved via the wranger `--node-compat` option.


JackDevAU:
Hey @fgnass 

This looks good! I'll test it out tomorrow and give it a review 🙏

twodft:
Any updates? I'm excited about this implementation for supporting edge runtime at Cloudflare Workers

ioslh:
Looking forward to this

packages/@tinacms/graphql/src/database/level.ts
```diff
@@ -8,30 +8,21 @@ import {
   AbstractSublevelOptions,
 } from 'abstract-level'
 
-export type Level = AbstractLevel<
-  Buffer | Uint8Array | string,
-  string,
-  Record<string, any>
->
+export type Level = AbstractLevel<any, string, Record<string, any>>
```
kldavis4, line 11: I'm not sure about this change. I think the d1 abstract-level should work without making this change. Have a look at https://github.com/tinacms/upstash-redis-level/blob/main/src/index.ts

packages/@tinacms/graphql/src/database/remote.ts
kldavis4, line 1: Can you explain what local/remote mean in this context?

fgnass: Local in the sense that it accesses the local filesystem and remote as it only talks to remote services.

kldavis4:
@fgnass In order to move this along I've created a new branch based off your work [here](https://github.com/tinacms/tinacms/pull/4728). I've also got a [branch](https://github.com/tinacms/tina-self-hosted-demo/tree/kd/edge-runtime) of the self hosted demo to implement it on a branch using vercel edge runtime. 

As of now I'm currently this error when building:

```
./pages/api/tina/[...routes].ts
Dynamic Code Evaluation (e. g. 'eval', 'new Function', 'WebAssembly.compile') not allowed in Edge Runtime 
Learn More: https://nextjs.org/docs/messages/edge-dynamic-code-evaluation

The error was caused by importing '@tinacms/datalayer/dist/index-edge.mjs' in './pages/api/tina/[...routes].ts'.

Import trace for requested module:
  ./pages/api/tina/[...routes].ts
```

I don't think we are doing any dynamic code evaluation, so will need to figure out why that's happening.

kldavis4:
Made a bit more progress on this. Was able to get the build to complete. Getting this error when deploying:

```
Error: The Edge Function "api/tina/[...routes]" size is 2.09 MB and your plan size limit is 1 MB. Learn More: https://vercel.link/edge-function-size
```

Added bundle analyzer to the branch in self-hosted-demo which can be run using `ANALYZE=true npx next build`. It looks like @tinacms/mdx accounts for a about 330kb but from seems like there will need to be a lot of work to reduce the size to get it under the free limits.

ncn-ssw:
Intention is to merge PR https://github.com/tinacms/tinacms/pull/4728 and create a spike to further improve support for edge.

Issue #4474, opened 2024-03-27
📦 Packages with vulnerability


t1gu1:
### Describe the issue that you're seeing. Any Loom videos or screenshots usually help a lot!

Some package contain vulnerabilities.

![image](https://github.com/tinacms/tinacms/assets/12479055/e78ab37a-493b-41de-bc04-3ef7d4f92fc5)

### Reproduction

localhost

### Steps to reproduce

- npm install
- npm audit

### System Info

```shell
- Mac OS
- Node ➜ v18.16.0
- Node ➜ 9.5.1
- tinacms ➜ 1.6.0
- tinacms/cli ➜ 1.5.42
```

### Validations

- [X] Follow our [Code of Conduct](https://github.com/tinacms/tinacms/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [docs](https://tina.io/docs).
- [X] Check that there isn't [already an issue](https://github.com/tinacms/tinacms/issues) that reports the same bug.

exofoliohq:
I see the same on my end with :
```
- Windows 11
- Node ➜ v20.12.0
- tinacms ➜ 1.6.1
- tinacms/cli ➜ 1.5.43
```
Terminal output : 
```
# npm audit report

axios  0.8.1 - 0.27.2
Severity: moderate
Axios Cross-Site Request Forgery Vulnerability - https://github.com/advisories/GHSA-wf5p-g6vw-rhxx
No fix available
node_modules/axios
  @tinacms/cli  *
  Depends on vulnerable versions of @tinacms/app
  Depends on vulnerable versions of axios
  Depends on vulnerable versions of tinacms
  node_modules/@tinacms/cli

lodash.set  *
Severity: high
Prototype Pollution in lodash - https://github.com/advisories/GHSA-p6mc-m468-83gw
No fix available
node_modules/lodash.set
  tinacms  <=0.0.0-20240328200248 || 0.4.0-dev.0 || >=0.50.0
  Depends on vulnerable versions of lodash.set
  node_modules/tinacms
    @tinacms/app  <=0.0.22 || >=1.2.0
    Depends on vulnerable versions of tinacms
    node_modules/@tinacms/app

5 vulnerabilities (1 moderate, 4 high)

Some issues need review, and may require choosing
a different dependency.
```
May bumping up these dependencies be simple, straightforward and break nothing 🤞

ncn-ssw:
Parked in Sprint 3 due to onboarding issues

[...]

PR #4647, opened 2024-07-16
fix packages critical vulnerabilities issue

Ben0189:
Update packages in example project to solve package vulnerability issue 

Override **loader-utils** and **ws** peer dependency packages with no vulnerabilities issue at the root 

Related PBI : #4474

### DIFF
--- a/.changeset/good-tools-buy.md
+++ b/.changeset/good-tools-buy.md
@@ -1,0 +1,7 @@
+---
+'@tinacms/self-hosted-starter': patch
+'create-tina-app': patch
+'tinacms': patch
+---
+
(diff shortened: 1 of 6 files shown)


.changeset/good-tools-buy.md
```diff
@@ -0,0 +1,7 @@
+---
+'@tinacms/self-hosted-starter': patch
+'create-tina-app': patch
+'tinacms': patch
+---
```
JackDevAU, line 5: we don't need to include this in the changeset 🙏

```suggestion
---
'create-tina-app': patch
'tinacms': patch
---
```

JackDevAU (approved):

PR #4700, opened 2024-08-04
Rich-text local variables

jeffsee55:


kldavis4:
@JackDevAU still need to test this with tina cloud but this should be ready to look at

kldavis4:
update: spent a load of time today testing (mostly wrestling with tina cloud local dev unfortunately) but at the moment I can't seem to get it working with tina cloud (works fine in local dev).

It looks like it is somehow an issue with the way @tinacms/mdx is getting pulled in. When I inspect the version associated with the 1.5 endpoint, it doesn't have the snapshot. Will have to investigate further unless you have a suggestion.

ncn-ssw:
@kldavis4 @JackDevAU 

If we are doing a minor-bump on the [at]tinacms/graphql package, should we also look to include https://github.com/tinacms/tinacms/pull/5219/ ?

kldavis4:
> @kldavis4 @JackDevAU
> 
> If we are doing a minor-bump on the [at]tinacms/graphql package, should we also look to include #5219 ?

I don't think this _needs_ a minor bump - as it is backwards compatible. I agree if we are going to bump minor anyway we should include that other PR

.changeset/young-chefs-tickle.md
```diff
@@ -0,0 +1,6 @@
+---
+"@tinacms/graphql": minor
```
ncn-ssw, line 2: **question**: Should this be minor or patch?

JackDevAU:
I think we might need to create an internal "embedComponent" (potentially just allow the parsing to not turn the template into a custom component) that if not overridden using `<TinaMarkdown/>` will use the default type component to render. Currently if we have nested content we get something a long the lines of `No component provided for nestedField`. Ideally if it isn't something that is overridden, we just use whatever type that is specified in the Template schema. 

image
**Figure: Current unexpected results**

JackDevAU:
Closing due to being stale - will be looked at in v4

PR #4705, opened 2024-08-07
✨ Feature - Implement infinite scrolling in the Media Manager

Jord-Gui:
Implements infinite scrolling for the Media Manager as per https://github.com/tinacms/tinacloud/issues/2035


packages/next-tinacms-cloudinary/src/handlers.ts
```diff
@@ -95,9 +95,10 @@ async function listMedia(
   try {
     const {
       directory = '""',
-      limit = 500,
+      limit = '500',
```
JackDevAU, line 98: Why did we change this to a string?

packages/next-tinacms-cloudinary/src/handlers.ts
```diff
@@ -95,9 +95,10 @@ async function listMedia(
...
       offset,
-    } = req.query as MediaListOptions
+      filesOnly = 'false',
+    } = req.query
```
JackDevAU, line 101: Should we add back `MediaListOptions`

packages/next-tinacms-cloudinary/src/handlers.ts
```diff
@@ -106,7 +107,7 @@ async function listMedia(
 
     const response = await cloudinary.search
       .expression(query)
-      .max_results(limit)
+      .max_results(parseInt(limit as string, 10))
```
JackDevAU, line 110: If we change back `limit` to a number do we still need to parseInt?

packages/next-tinacms-cloudinary/src/handlers.ts
```diff
@@ -122,34 +123,37 @@ async function listMedia(
     }
     let folders: string[] = []
     let folderRes = null
+    const loadFolders = !(filesOnly == 'true')
```
JackDevAU, line 126: Should we be using `==` instead of `===`

packages/next-tinacms-cloudinary/src/handlers.ts
```diff
@@ -122,34 +123,37 @@ async function listMedia(
...
-      folderRes = await cloudinary.api.folders(directory)
-    } catch (e) {
-      // If the folder doesn't exist, just return an empty array
-      if (e.error?.message.startsWith("Can't find folder with path")) {
-        // ignore
-      } else {
-        console.error('Error getting folders')
-        console.error(e)
-        throw e
+    if (loadFolders) {
```
JackDevAU, line 128: Can this be inverted? i.e. if(!loadFolders) return;

JackDevAU (commented):
Left some comments but other than those - Looks good 💪

packages/next-tinacms-cloudinary/src/handlers.ts
Jord-Gui: Good idea

packages/next-tinacms-cloudinary/src/handlers.ts
Jord-Gui: Having MediaListOptions caused confusion as they were actually strings, being of type NextApiRequest:
image

I had to spend a while figuring out why my conditionals were not working as expected (`filesOnly` was coming in as a `string` and not `boolean`.

If we do want to have MediaListOptions there as a easier way people to track where this query comes from, I can do a type conversion from NextApiRequest to MediaListOptions?

packages/next-tinacms-cloudinary/src/handlers.ts
Jord-Gui: This is due to any overrides from the request coming in being of type `string`. I think it would be safer to be consistent.

packages/next-tinacms-cloudinary/src/handlers.ts
Jord-Gui: Yeah we wouldn't need to. I'll spin up a better type definition above

packages/next-tinacms-cloudinary/src/handlers.ts
Jord-Gui: Good idea

packages/next-tinacms-cloudinary/src/handlers.ts
JackDevAU: ahh ok - agreed!

JackDevAU (approved):
✅

PR #4712, opened 2024-08-09
reference field styling fix

Ben0189:
fix styling issue for reference field 
migrate one example project to use app router


packages/tinacms/src/toolkit/fields/components/reference/components/popover.tsx
```diff
@@ -18,7 +18,7 @@ const PopoverContent = React.forwardRef<
       align={align}
       sideOffset={sideOffset}
       side="bottom"
-      avoidCollisions={false}
+      avoidCollisions={true}
```
JackDevAU, line 21: by default this is true! https://www.radix-ui.com/primitives/docs/components/popover#content
Let's remove it

JackDevAU (changes requested):
Looks good - Just needs a changeset :)

JackDevAU (approved):

JackDevAU (approved):

PR #4713, opened 2024-08-09
CHANGELOG - link to the module's changelogs

wicksipedia:
Looking at the root changelog, it's easy to think that it is out of date since the topmost entry was versioned at 0.43.1.

This PR deletes all the old "notable changes" - since they are pre v1 and instead link off to the individual module's changelogs.


ncn-ssw (approved):
🦙

Issue #2980, opened 2022-06-14
TinaMarkdown doesn't use alt tags from markdown.


perkinsjr:
When using `TinaMarkdown` alt text isn't inserted into the image tag even if it is included:

Example
```![example image](/img/example.jpg)```

returns 

`<a src="/img/example.jpg")`

We should include the alt text for images when supplied.

**Comment by burntcustard** [comment]
Not passing alt text to `<img>` elements when using `TinaMarkdown` is still a bug and in my opinion shouldn't have been marked as wontfix.

I believe I've tracked down at least half of the cause - [line 269 in `packages/tinacms/src/rich-text/index.tsx`](https://github.com/tinacms/tinacms/blob/50f86caed86fbc6526586a11cf66e70504c6972f/packages/tinacms/src/rich-text/index.tsx#L269) is:
```
return <img src={child.url} alt={child.caption} />
```

It's attempting to put the `caption` field as alt text, rather than the alt text field.

I've been able to get around this locally in two ways, either:

**a) Use a markdown image `title`, which gets pulled through as the caption and then is set as alt text:**

```
![Alt that doesn't work](/path/to/img.jpg "image title that's set as alt text")
```

I don't like this solution, because it's generally not how alt text is done in markdown, as far as I've seen isn't documented anywhere on the [Tina Docs](https://tina.io/docs/), and feels fragile if/when this alt text issue is "fixed". It _is_ simple though.

**b) By changing that (via line 153 in `/node_modules/tinacms/dist/rich-text/index.js`) to:**

```
return <img src={child.url} alt={child.caption || child.alt} />
```

My thought here is that if either of those are set, it will use them for the alt text of the image. Still slightly confusing, but as far as I can tell it's doesn't really break anything in my Tina projects, and wouldn't in any others either.

Unfortunately with that "fix", I see what I believe is a related console error in the admin view, because of the alt text in my markdown confusing Tinacms' inner workings somehow:
```
Warning: Extra attributes from the server: alt
img
[...]
```

Hopefully at least one of these suggestions helps out anyone who stumbles across this issue after having the same problem themselves!

burntcustard:
Almost 3 months on and this seems to still be a major accessibility / TinaCMS UX issue.

As a bit of extra info, it's not just an issue when editing markdown files, but when using the editor as well, the image `Alt` field doesn't do anything, and the `Caption` field sets the `alt` attribute of the `<img>` element.

EthanShoeDev:
[...]

PR #4715, opened 2024-08-09
WIP Fix alt tags on images

EthanShoeDev:
Fixes: https://github.com/tinacms/tinacms/issues/2980

### DIFF
--- a/packages/tinacms/src/toolkit/forms/form.ts
+++ b/packages/tinacms/src/toolkit/forms/form.ts
@@ -495,13 +495,13 @@ export class Form<S = any, F extends Field = AnyField> implements Plugin {
                 {
                   type: 'string',
                   label: 'Alt',
-                  name: [templateName, 'alt'].join('.'),
+                  name: [templateName.replace(/\.props$/, ''), 'alt'].join('.'),
                   component: 'text',
                 },
                 {
                   type: 'string',
                   label: 'Caption',
-                  name: [templateName, 'caption'].join('.'),
+                  name: [templateName.replace(/\.props$/, ''), 'caption'].join('.'),
                   component: 'text',
                 },
               ],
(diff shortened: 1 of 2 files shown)


burntcustard:
I tested this with [tina-cloud-starter](https://github.com/tinacms/tina-cloud-starter), with the dependencies listed in the last comment in #2980 -  and it seems to work!

I would like to point out that this is _sort of_ a breaking change. Anyone who was using workarounds, like putting alt text in the caption field instead of the alt field, or within the title `""` instead of the alt text `[]` in markdown (as discussed in #2980), will have alt text unexpectedly disappear from their images.

For example this:
image
i.e:
`[](/uploads/tina-cloud-authorization.png \"This is my alt text\")`

Will have it's alt-text-via-workaround removed.

Despite that concern, I'd like to see this being merged ASAP - it's already taken far too long for TinaCMS to support image alt text properly!

packages/tinacms/src/toolkit/forms/form.ts
```diff
@@ -495,13 +495,13 @@ export class Form<S = any, F extends Field = AnyField> implements Plugin {
...
                   label: 'Alt',
-                  name: [templateName, 'alt'].join('.'),
+                  name: [templateName.replace(/\.props$/, ''), 'alt'].join('.'),
                   component: 'text',
                 },
                 {
                   type: 'string',
                   label: 'Caption',
-                  name: [templateName, 'caption'].join('.'),
+                  name: [templateName.replace(/\.props$/, ''), 'caption'].join('.'),
```
EthanShoeDev, line 504: While this works, it is an ugly fix. I expect the maintainer will want to try rework this code a bit.

Jord-Gui: lgtm

Jord-Gui (approved):

Issue #4707, opened 2024-08-08
♻️ TinaCMS - Customise the reference field selector


wicksipedia:
CC @wicksipedia @tkapa @Ben0189 

Hey Team, 

### Description: 

Following https://github.com/tinacms/tinacms/pull/4701 we would now like to be able to customise the selector  

### Tasks
- [x] Should be able to style the `option` to include more data than just the filename
  - [x] Extend the `reference` field type and add the following or similar
  - [x] an `options` object that has the following or similar
    - [x] `label: ({field}) => string` - The label to be displayed    
    - [x] `search:string[]` - The field's that should be used to index/search on  
    - [x] `optionOverride` - A custom component that will replace the default option that is rendered. Maybe experiment with replacing the whole List component? 

### Screenshots

![Image](https://github.com/user-attachments/assets/9a11e570-3514-46e2-a7d9-5aed6c48cd88)
**Figure: MacOS**

![Image](https://github.com/user-attachments/assets/1e740be7-9c8d-4c61-b714-6e678ea14106)
**Figure: Windows**

Ben0189:
Progress update 
- [ ] Add display option property to schema tools for reference field 
- [ ] Validate new display option property
- [ ] Refactor graphql query to retrieve display option property
- [ ] Refactor reference field selector component to display content based on option property

JackDevAU:
Before:


![Image](https://github.com/user-attachments/assets/879b4b0c-fa15-43c3-bef2-915120876d67)



After:

![Image](https://github.com/user-attachments/assets/9533ef72-2d93-4a17-8507-557c101eb9f3)

ncn-ssw:
@Ben0189 

Please note that a Done video is required.
- [ ] Done video

Ben0189:
Done video for new reference field - https://www.youtube.com/watch?v=WjxYTxlnZMU

Ben0189:
Documentation pr still under review - https://github.com/tinacms/tina.io/pull/2041

Jord-Gui:
> Documentation pr still under review - [tinacms/tina.io#2041](https://github.com/tinacms/tina.io/pull/2041)

PR has been merged into main

PR #4732, opened 2024-08-19
Implement custom display reference field

Ben0189:
#4707 

Allow user to customize reference field selector , search function is working under given reference file path

image


packages/@tinacms/schema-tools/src/types/index.ts
```diff
@@ -214,8 +214,20 @@ export type ImageField = (
   }
 
 export type ReferenceField = (
-  | FieldGeneric<string, undefined>
-  | FieldGeneric<string, false>
+  | FieldGeneric<
+      string,
+      undefined,
+      { selectComponent: (props: any) => Element | undefined }
```
JackDevAU, line 220: How do we feel about renaming this from `select` to `option` to make it more accurate to it's implementation as it is inserted inside a `select` element

packages/@tinacms/schema-tools/src/types/index.ts
```diff
@@ -214,8 +214,20 @@ export type ImageField = (
...
+    >
+  | FieldGeneric<
+      string,
+      false,
+      {
+        selectComponent: (props: {
```
JackDevAU, line 226: Same as above comment

packages/tinacms/src/toolkit/fields/components/reference/reference-select.tsx
```diff
@@ -182,33 +183,32 @@ const ComboboxDemo: React.FC<ReferenceSelectProps> = ({
...
-                          },
-                        }) => (
+                      {edges.map(({ node }) => {
+                        const { id, _values } = node
+
+                        return (
                           <CommandItem
                             key={`${id}-option`}
-                            value={id || title}
+                            value={id}
```
JackDevAU, line 192: Can we make this better by adding another `search` prop on the reference type that the user can input for what to search on.Either a string or this could even be a function that the user controls completely

JackDevAU (changes requested):
Awesome stuff just a few comments

packages/tinacms/src/toolkit/fields/components/reference/reference-select.tsx
```diff
@@ -182,33 +183,32 @@ const ComboboxDemo: React.FC<ReferenceSelectProps> = ({
...
                             <div className="flex flex-col">
-                              <span className="font-semibold text-sm">
-                                {title || name || id}
-                              </span>
-                              {(title || name) && (
-                                <span className="text-x">{filename}</span>
-                              )}
+                              <div>
+                                {field?.ui?.selectComponents && _values ? (
+                                  field.ui.selectComponents(_values, {
```
jeffsee55, line 201: As we discussed, it might be nice to give the callback value a `values` object but also all of the `_sys` values we give you in the GraphQL response.

```ts
{
  values: Record<srtring, unknown>,
  _sys: {
   filename: string
   title: string // this would be used if `isTitle` is set on the collection
   breadcrumbs: string[],
   extension: string,
   // ... etc
 }
}

File: packages/tinacms/src/toolkit/fields/components/reference/reference-select.tsx (hunk 3 above)
Comment (jeffsee55) [reply packages/tinacms/src/toolkit/fields/components/reference/reference-select.tsx:192]: In addition to making search customizable, Ben and I discussed what it would look like to improve search behavior. We'll want to make this paginated in the future so we'll need to implement search from the GraphQL query. To do that we'll need to use the `filter` argument on the the documents

```graphql
query ($collection: String!, $filter: DocumentFilter){
  collection(collection: $collection) {
    documents(first: -1, filter: $filter) {
```
Where `DocumentFilter` supports querying on the colletion's fields. We can make use of the `isTitle` field to give them title search behavior by default.

packages/@tinacms/schema-tools/src/types/index.ts
jeffsee55: I think this was from the pseudo code I'd written, both of these types should be (https://github.com/tinacms/tinacms/pull/4732/files#diff-c1b9d8ea97eb7c96298291e7c8940e58d0b8f56cac01609406b23872e943e935R201))

```ts
      {
        selectComponent?: (props: {
          values: Record<string, unknown>
          _sys: Document['_sys']
        }) => any
      }
```
(if you change the code [over here](https://github.com/tinacms/tinacms/pull/4732/files#diff-c1b9d8ea97eb7c96298291e7c8940e58d0b8f56cac01609406b23872e943e935R201))

experimental-examples/kitchen-sink/tina/custom-component-reference-select/reference-field-schema.tsx
```diff
@@ -0,0 +1,32 @@
...
+import AuthorCollectionCustomReference from '../../component/custom-reference-select-author'
+import PostCollectionCustomReference from '../../component/custom-reference-select-post'
+import { CollectionProps, COLLECTIONS } from './model'
+
+const referenceField = {
+  label: 'Author',
+  name: 'author',
+  type: 'reference',
+  ui: {
+    selectComponents: (props: CollectionProps, filepath: string) => {
```
jeffsee55, line 11: This should `selectComponent`. If you fix that you'll see the types from [here](https://github.com/tinacms/tinacms/pull/4732/files#diff-88b24591b0c36f508270e9ae261e94ea2770198c387d66a95b2fe1620810bde3R220). They won't know the shape of the `values`, but you could cast them here. So if you want type safety on these I think it'd make sense to do (assuming the signature I mentioned below is used):

```ts
    selectComponent: (args) => {
      const props = args.values as CollectionProps
      const filename = args._sys.filepath
    
```

packages/@tinacms/schema-tools/src/types/index.ts
Ben0189: Done renamed to option

packages/@tinacms/schema-tools/src/types/index.ts
Ben0189: Done renamed

packages/tinacms/src/toolkit/fields/components/reference/index.tsx
```diff
@@ -14,6 +14,7 @@ export interface ReferenceFieldProps {
   component: string
   collections: string[]
   options: (Option | string)[]
+  ui?: any
```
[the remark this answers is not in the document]
Ben0189: Done - removed ui and used option component with proper type

packages/tinacms/src/toolkit/fields/components/reference/reference-select.tsx
Ben0189: As per conversation with @jeffsee55 , we will tackle the custom search in a different pr (current pr is for custom reference e component and the search still work based on filename )

packages/tinacms/src/toolkit/fields/components/reference/reference-select.tsx
Ben0189: As per our conversation, we will tackle the custom search in a different pr (current pr is for custom reference e component and the search still work based on filename )

PBI - https://github.com/tinacms/tinacms/issues/4738

experimental-examples/kitchen-sink/tina/custom-component-reference-select/reference-field-schema.tsx
Ben0189: Done - the current example here do the type safety by defined the interface for the param. They both works the same i guess

Jord-Gui (approved):

PR #4740, opened 2024-08-22
feat: add caching to client

kldavis4:

### DIFF
--- a/.changeset/rotten-maps-change.md
+++ b/.changeset/rotten-maps-change.md
@@ -1,0 +1,5 @@
+---
+'tinacms': patch
+---
+
+Client caching
--- a/packages/tinacms/src/unifiedClient/cache.ts
+++ b/packages/tinacms/src/unifiedClient/cache.ts
@@ -1,0 +1,5 @@
+export type Cache = {
+  get: (key: string) => Promise<any>
+  makeKey: (key: any) => string
+  set: (key: string, value: any) => Promise<void>
+}
--- a/packages/tinacms/src/unifiedClient/index.ts
+++ b/packages/tinacms/src/unifiedClient/index.ts
@@ -1,6 +1,7 @@
 import fetchPonyfill from 'fetch-ponyfill'
 import type { GraphQLError } from 'graphql'
 import type { Config } from '@tinacms/schema-tools'
+import type { Cache } from './cache'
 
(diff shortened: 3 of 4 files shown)


.changeset/rotten-maps-change.md
```diff
@@ -0,0 +1,6 @@
+---
+'tinacms': patch
+'@tinacms/cli': patch
+---
+
+Client caching
```
wicksipedia, line 6:
```suggestion
CLI - Adds client caching and cli flag to disable: --no-client-build-cache
```

These will need to be more details as changelogs will be cross published to https://tina.io/whats-new/tinacms/

I'd recommend tagging the PR and checking build times against https://github.com/sswconsulting/ssw.website - if there are any performance improvements, include stats as well

wicksipedia (commented):
Docs on https://tina.io/docs/cli-overview/ will need to be updated as well

kldavis4:
@wicksipedia docs update: https://github.com/tinacms/tina.io/pull/2106

packages/@tinacms/cli/src/next/codegen/index.ts
```diff
@@ -340,7 +344,11 @@ export default databaseClient;
 
     const clientString = `import { createClient } from "tinacms/dist/client";
 import { queries } from "./types";
-export const client = createClient({ url: '${apiURL}', token: '${token}', queries, ${
+export const client = createClient({ ${
+      this.noClientBuildCache == false
```
JackDevAU, line 348:
```suggestion
      this.noClientBuildCache === false
```

packages/tinacms/src/cache/node-cache.ts
```diff
@@ -0,0 +1,32 @@
+import shajs from 'sha.js'
```
JackDevAU, line 1: Should we use the inbuilt `crypto` module instead of this? 
e.g. `import { createHash } from 'crypto';`

kldavis4: I ran into issues with that but don't recall exactly what. Will try it again and let you know.

JackDevAU: thought that might've been the case :/

kldavis4: Got it to work using require()

kldavis4:
@JackDevAU ready for review

JackDevAU (approved):
🦙

PR #4742, opened 2024-08-23
feat: Add ContentType to headers sent to AWS S3 / DOS

kldavis4:
Based on https://github.com/tinacms/tinacms/pull/4395

> ContentType header was missing in `next-tinacms-s3` package, making AWS serve SVGs as `application/octet-stream`.

This PR solves it by pulling the `mimetype` from the file and if it's not available it fallbacks to `application/octet-stream`

Also makes corresponding change to DOS implementation


JackDevAU (approved):

packages/next-tinacms-dos/src/handlers.ts
```diff
@@ -109,6 +109,9 @@ async function uploadMedia(
   // eslint-disable-next-line @typescript-eslint/ban-ts-comment
   // @ts-ignore
   const filePath = req.file.path
+  // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+  // @ts-ignore
+  const fileType = req.file.mimetype
```
JackDevAU, line 114:
```suggestion
  const fileType = req.file?.mimetype
```

Does this need a `?`

packages/next-tinacms-s3/src/handlers.ts
```diff
@@ -111,6 +111,9 @@ async function uploadMedia(
     // eslint-disable-next-line @typescript-eslint/ban-ts-comment
     // @ts-ignore
     const filePath = req.file.path
+    // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+    // @ts-ignore
+    const fileType = req.file.mimetype
```
JackDevAU, line 116: same as above

kldavis4, line 116:
```suggestion
    const fileType = req.file?.mimetype
```

JackDevAU (approved):
🚀

PR #4748, opened 2024-08-26
Refactor reference field component

Ben0189:
clean up on reference field selector component 

- extract inline function 
- extract props model into models folder 
- improve readme file


Jord-Gui (changes requested):
Looks good, just needs a changeset

Jord-Gui (approved):

PR #4750, opened 2024-08-27
♻️ Media Manager - Fix media manager active item

### LINKED ISSUE #4714: ♻️ TinaCMS - Media Manager - Active item remains present on navigation
### Pain
When I select an item in the media manager, I get a preview of it. When I navigate folders, the preview is still there. 

### Expected Behaviour
Once I navigate, the preview resets and is not shown
- [x] Refactor

### Example

Image

**Figure: Active item shown within a folder**

Image

**Figure: Active item is still present even though I have navigate to another folder and the image does not exist in that folder**

Jord-Gui:
Fixes #4714. 

Ensure active item is not present on navigation and page reload

### DIFF
--- a/.changeset/hot-donkeys-lay.md
+++ b/.changeset/hot-donkeys-lay.md
@@ -1,0 +1,5 @@
+---
+'tinacms': patch
+---
+
+Media Manager - Make active item disappear on navigation and refresh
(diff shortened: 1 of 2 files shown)


ncn-ssw (approved):
🦙 Approved

PR #4758, opened 2024-08-31
fix: @tinacms/cli: prevent search config from getting written to lock file

kldavis4:
# problem

Discord user reported search indexer token is being written to tina-lock.json

# solution

Filter search config from output


ncn-ssw (approved):
🦙 Tested correct

PR #4770, opened 2024-09-07
feat: media-manager: add async error handle for media upload

kldavis4:
Leverage the new request-status endpoint in TinaCloud media manager for upload and delete requests so that async errors are surfaced to the user.

Also fixes media manager UI to close the delete confirmation modal regardless of whether an error occurs (instead of it staying open when an error occurs)

### DIFF
--- a/.changeset/tender-hounds-smell.md
+++ b/.changeset/tender-hounds-smell.md
@@ -1,0 +1,5 @@
+---
+'tinacms': patch
+---
+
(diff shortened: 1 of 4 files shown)


ncn-ssw (approved):
🦙 Tested correct

@kldavis4 - I have renamed a couple of `now`s to clarify their purpose and corrected an error message. Please verify that you are happy with the change before merging.

Issue #4720, opened 2024-08-13
♻️ TinaCMS - Refactor Examples


JackDevAU:
CC @wicksipedia @ncn-ssw @Ben0189 @Jord-Gui 

Hey Team, 

As per my conversation with @ncn-ssw the current examples in the repo aren't great, and some don't work... 

Current Examples - To be removed.

- `examples/basic-iframe`
- `examples/empty`
- `examples/hugo-quickstart`
- `examples/kitchen-sink`
- `examples/next-2024`
- `examples/tina-self-hosted-demo`
- `experimental-examples/kitchen-sink`
- `experimental-examples/tina-cloud-starter`

These will be replaced with the following, 

- next
  - `next/kitchen-sink`
  - `next/self-hosted`
 
- react
  - `react/kitchen-sink`

- astro
  - `astro/kitchen-sink`

- hugo
  - `hugo/kitchen-sink`

This approach will make the examples actually useful, to internal testers as well as the general public to play around and use as reference. 

### Task 
- [ ] Remove old examples
- [ ] Create new kitchen sink examples separated by framework
- [ ] Update docs + Readme

18-th:
Updated Empty example with the latest supported versions of packages and Tina v3 approaches in #6387 ✅

KahaMason:
Update as of 17/2/26:

PRs: 
* [NextJS - New Kitchen-Sink Example](https://github.com/tinacms/tinacms/pull/6428)

Progress on PR:
- [x] Created new Kitchen Sink app in NextJS 15 & React 18
- [x] Update all the html, components and CSS stylings (many html elements / components / CSS were missing originally)
- [x] Updated to use Biome, ESM, App routing, Server Components + generateStaticParams
- [x] Reformat the content schema for the example content.
- [x] Fix the existing tests in the new next/kitchen-sink example

Remaining work on this Issue:
- [ ] Finalize next/kitchen-sink example
- [ ] Recreate kitchen-sink example in React, Astro & Hugo
- [ ] Remove old examples
- [ ] Update the docs

KahaMason:
👀 [NextJS - New Kitchen-Sink Example](https://github.com/tinacms/tinacms/pull/6428) is ready for review
- New Kitchen-Sink example (NextJS)

kulesy:
✅ Done

PR #4780, opened 2024-09-12
Feature - customize graphql with filter

Ben0189:
A experimental on improving reference field list selector filtering using customize graphql 

related PBI [#4781](https://github.com/tinacms/tinacms/issues/4781)


experimental-examples/kitchen-sink/tina/custom-component-reference-select/reference-field-schema.tsx
```diff
@@ -49,6 +26,23 @@ const referenceField = {
...
+    //   },
+    // },
+    //Dynamic example - user can define the function they want here but need to make sure the return type match the schema
+    collectionFilter: () => {
+      return {
+        author: {
+          location: getLocationFromHost('ndcsydney.com'),
+        },
+      }
+    },
```
JackDevAU, line 45: - Can we make this function inline to make this a little clearer?

packages/@tinacms/schema-tools/src/types/index.ts
```diff
@@ -7,6 +7,8 @@ type Meta = {
   error?: any
 }
 
+export type FilterValue = string[] | string // extend this type if needed
```
JackDevAU, line 10:
```suggestion
export type FilterValue = string[] | string 
```

Ben0189: i intentionally leave the comment there, this type is extendable when we need to support other datatype in future

experimental-examples/kitchen-sink/tina/custom-component-reference-select/reference-field-schema.tsx
Ben0189: > Can we make this function inline to make this a little clearer?
Isn't this already inline ?

> Can we add another field that shows off the static example?
Intentionally leave the commented out code above in the same reference field so that we can have a clearer comparison on both method.

JackDevAU: > Isn't this already inline ?

No - I would move the function that is being called from another file into this one to make it a bit easier to see at a first glance. As per Matt's recommendation this morning 

> Intentionally leave the commented out code above in the same reference field so that we can have a clearer comparison on both method.

I guess that works - Would do 2 separate examples then on the docs page 🙏

packages/@tinacms/schema-tools/src/types/index.ts
JackDevAU: ah lets give it more info then! 
what its used for, where its used, etc

packages/tinacms/src/toolkit/fields/components/reference/utils/fetch-optioins-query-builder.ts
```diff
@@ -0,0 +1,20 @@
+export type FilterValue = string[] | string // extend this type if needed
+export type CollectionFilters =
+  | Record<string, FilterValue>
+  | (() => Record<string, FilterValue>)
+
+//Currently only support eq for filter, this function will loop thorugh the record and build the filter query
```
JackDevAU, line 6:
```suggestion
//Currently only support in for filter, this function will loop through the record and build the filter query
```

experimental-examples/kitchen-sink/content/author/Napolean.md
```diff
@@ -1,4 +1,5 @@
 ---
-name: Napolean
+name: melbourne - Napolean
```
ncn-ssw, line 2: **issue**: If this test data is to be retained, then please capitalize the placenames "Melbourne" and "Sydney".

experimental-examples/kitchen-sink/tina/custom-component-reference-select/reference-field-schema.tsx
```diff
@@ -49,6 +26,34 @@ const referenceField = {
...
+    //     title: 'hello world',
+    //   },
+    // },
+    //Dynamic example - user can define the function they want here but need to make sure the return type match the schema
+    collectionFilter: () => {
+      return {
+        author: {
+          location: (hostname: string): LocationEnum => {
+            switch (true) {
+              case hostname.includes('ndcmelbourne.com'):
```
ncn-ssw, line 44: **question**: Are we able to use example data consistent with https://www.ssw.com.au/rules/use-generic-consistent-names-on-examples/ ?

ncn-ssw:
@JackDevAU - Given your longer-term intentions with the examples, did you want the demonstration of this feature in "experimental-examples/kitchen-sink" or another example like "examples/next-2024"?

packages/tinacms/src/toolkit/fields/components/reference/model/reference-field-props.ts
```diff
@@ -1,3 +1,5 @@
+import { CollectionFilters } from '../utils/fetch-optioins-query-builder'
```
ncn-ssw, line 1: **question**: Should the imported file be "fetch-options-query-builder"?

packages/tinacms/src/toolkit/fields/components/reference/utils/fetch-optioins-query-builder.ts
```diff
@@ -0,0 +1,22 @@
+// This type is used in collectionFilter for reference field, it represent the datatype of the field in the referenced collection that will be used to do the filter
+// This type is extendable if we need to support more type for reference selection filter (e.g. boolean)
```
ncn-ssw, line 2: **suggestion**: Improved English
```suggestion
// It is extendable in case we need to support more types for the reference selection filter (e.g. boolean)
```

packages/tinacms/src/toolkit/fields/components/reference/utils/fetch-optioins-query-builder.ts
```diff
@@ -0,0 +1,22 @@
...
+export type FilterValue = string[] | string
+export type CollectionFilters =
+  | Record<string, FilterValue>
+  | (() => Record<string, FilterValue>)
+
+//Currently only support in for filter, this function will loop through the record and build the filter query
```
ncn-ssw, line 8: **question**: Does this comment add value?

I worry it will be ignored at the next refactor - see https://www.ssw.com.au/rules/code-commenting/#what-are-the-downsides-of-comments

experimental-examples/kitchen-sink/content/author/Napolean.md
JackDevAU: I guess we should also fix the spelling of "Napolean" to _Napoleon_ as well 🙏

JackDevAU:
> @JackDevAU - Given your longer-term intentions with the examples, did you want the demonstration of this feature in "experimental-examples/kitchen-sink" or another example like "examples/next-2024"?

In this case I don't think it matters that much - We are planning on revamping the examples as per https://github.com/tinacms/tinacms/issues/4720

packages/tinacms/src/toolkit/fields/components/reference/utils/fetch-optioins-query-builder.ts
Ben0189: This comment is not a TODO, it is a explanation of what the below function means, thus it does not require a PBI link to it. Also i have reword this comment 😊

packages/tinacms/src/toolkit/fields/components/reference/model/reference-field-props.ts
Ben0189: Did you mean to suggest splitting the types into a separate file? I intentionally kept them in this utils file since they are closely related, making it easier to reference. The utils file is still relatively small, so keeping them together won't cause any mess.

ncn-ssw:
**question**: Should there be automated tests with this change?

experimental-examples/kitchen-sink/content/author/Napolean.md
Ben0189: Done - fix spelling mistake

experimental-examples/kitchen-sink/tina/custom-component-reference-select/reference-field-schema.tsx
Ben0189: Done - fix to use bob-northwind

packages/tinacms/src/toolkit/fields/components/reference/model/reference-field-props.ts
Ben0189: Done - fixed file name

packages/tinacms/src/toolkit/fields/components/reference/utils/fetch-optioins-query-builder.ts
Ben0189: Done - As per our conversation , the comment does not bring much value, thus comment removed

JackDevAU:
Might need to note in the docs that any text that is compared **IS** case sensitive. So If you enter e.g. `name: "pedro"` and you have an an Author with the `name: "Pedro"` this **won't match**

packages/@tinacms/schema-tools/src/types/index.ts
```diff
@@ -7,6 +7,10 @@ type Meta = {
   error?: any
 }
 
+// This type is used in collectionFilter for reference field, it represent the datatype of the field in the referenced collection that will be used to do the filter
+// This type is extendable if we need to support more type for reference selection filter (e.g. boolean)
+type FilterValue = string[] | string
```
JackDevAU, line 12: This was giving me a type error. Can we change this to something like, 

```
type FilterValue = {[key]: string};
```

Ben0189: Done - Fixed the schema, can you pull latest changes and see if the type error still exist?

Ben0189:
> Might need to note in the docs that any text that is compared **IS** case sensitive. So If you enter e.g. `name: "pedro"` and you have an an Author with the `name: "Pedro"` this **won't match**

Done updated the document and here is the PR 

https://github.com/tinacms/tina.io/pull/2229

ncn-ssw (approved):
🦙 Tested correct.

.changeset/happy-scissors-deliver.md
```diff
@@ -0,0 +1,5 @@
+---
+'tinacms': patch
```
JackDevAU, line 2:
```suggestion
'tinacms': patch
"@tinacms/schema-tools": patch
```

JackDevAU: add schema-tools

packages/@tinacms/schema-tools/src/types/index.ts
```diff
@@ -216,6 +220,9 @@ export type ImageField = (
 type ReferenceFieldOptions = {
   optionComponent?: OptionComponent
   experimental___filter?: (list: Array<any>, searchQuery: string) => Array<any>
```
JackDevAU, line 222:
```suggestion
/**
     * @deprecated use `collectionFilter` instead as experimental___Filter will be removed in a future release
     */
  experimental___filter?: (list: Array<any>, searchQuery: string) => Array<any>
```

JackDevAU (dismissed):
🦙 Awesome stuff! 

Guess we should also note in the docs that inline functions won't work as the properties aren't called. e.g. the below code won't work and will return all authors:

```js
collectionFilter: {
          author: {
            name: () => Date.now().toString(),
          },
        },
```
Instead we would need to add this function outside as it will only be run on build (and initialisation of the tinacms panel)

packages/@tinacms/schema-tools/src/types/index.ts
```diff
@@ -215,7 +219,13 @@ export type ImageField = (
 
 type ReferenceFieldOptions = {
   optionComponent?: OptionComponent
+/**
+     * @deprecated use `collectionFilter` instead as experimental___Filter will be removed in a future release
+     */
```
ncn-ssw, line 224: **nitpick**: Identation of comment

Ben0189: ✅ Done

ncn-ssw (approved):
🦙 Good work

Ben0189:
> 🦙 Awesome stuff!
> 
> Guess we should also note in the docs that inline functions won't work as the properties aren't called. e.g. the below code won't work and will return all authors:
> 
> ```js
> collectionFilter: {
>           author: {
>             name: () => Date.now().toString(),
>           },
>         },
> ```
> 
> Instead we would need to add this function outside as it will only be run on build (and initialisation of the tinacms panel)

To clarify, inline functions will work, but you need to pass them as executable values rather than references. You can define the function as a constant and then call it when passing it to the name property.

(Notes : The reference component will not execute your nested function if you pass it inline or variable without execute)

Issue #4763, opened 2024-09-04
🐛 BUG - Certain characters break the search in the reference field. e.g. /


joshbermanssw:
### Describe the issue that you're seeing. Any Loom videos or screenshots usually help a lot!

I am trying to use the new search feature in the reference collection but it does not seem to work as i would expect.

If i type in content/docs it shows nothing
If i type in contentdocs it shows all items in the content/docs directory 
/a

https://github.com/user-attachments/assets/ddec0143-608c-46ef-92a3-7fb75476bcee

### Reproduction

n/a

### Steps to reproduce

see video 

### System Info

on current prod version of tina.io (tinacms v2.2.1) 

### Validations

- [x] Follow our [Code of Conduct](https://github.com/tinacms/tinacms/blob/main/CODE_OF_CONDUCT.md)
- [x] Read the [docs](https://tina.io/docs).
- [x] Check that there isn't [already an issue](https://github.com/tinacms/tinacms/issues) that reports the same bug.

Ben0189:
this error happen due to race condition when we release new experimental custom filter for reference field, here are a quick fix so the new release is backward compatible 😊

https://github.com/tinacms/tinacms/pull/4764

Ben0189:
As per conversation with @JackDevAU , user want to search content in a specific folder ('/' is not working). We will have a look into the search function to see if there is a quick fix for this

Ben0189:
✅Done fixes - https://github.com/tinacms/tinacms/pull/4782

PR #4782, opened 2024-09-12
Enable / in reference search

Ben0189:
Solve bug in search #4763

### DIFF
--- a/.changeset/modern-cougars-clap.md
+++ b/.changeset/modern-cougars-clap.md
@@ -1,0 +1,5 @@
+---
+'tinacms': patch
+---
+
+enable / in search
--- a/.changeset/modern-cougars-clap.md
+++ b/.changeset/modern-cougars-clap.md
@@ -2,4 +2,4 @@
 'tinacms': patch
 ---
 
-enable / in search
+enable / in search for reference field selection component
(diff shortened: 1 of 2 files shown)


packages/tinacms/src/toolkit/fields/components/reference/reference-select.tsx
```diff
@@ -177,13 +177,7 @@ const ComboboxDemo: React.FC<ReferenceSelectProps> = ({
             shouldFilter={!field.experimental___filter}
             filter={(value, search) => {
               //Replace / in the file path with empty string to make it searchable
```
ncn-ssw, line 179: **question**: Is this comment still relevant?

.changeset/modern-cougars-clap.md
```diff
@@ -0,0 +1,5 @@
+---
+'tinacms': patch
+---
+
+enable / in search
```
kldavis4, line 5: Can you provide more details for the change set?

Ben0189: Done - updated changeset

packages/tinacms/src/toolkit/fields/components/reference/reference-select.tsx
Ben0189: Done - removed redundant comment

Jord-Gui (approved):

PR #4785, opened 2024-09-14
fix: normalize cache dir to avoid unescaped characters breaking esbuild

kldavis4:
Fix #4779 caused by not escaping backslashes on windows


ncn-ssw (approved):
🦙 Approved

PR #4799, opened 2024-09-18
Add `--pkg-manager` option to `create-tina-app` cli + GitHub action to build / test starter templates

brookjeynes-ssw:
- Added `--pkg-manager` option to `create-tina-app` cli.
- Renamed `--example` option to `--template` for `create-tina-app` cli.
- `create-tina-app` cli tool only lists installed package managers.
- Refactored `create-tina-app` cli tool.
- Added GitHub action to build / test starter templates.


.github/workflows/build-starter-templates.yml
```diff
@@ -0,0 +1,55 @@
...
+          - basic
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v3
+
+      - name: Set up Node.js
+        uses: actions/setup-node@v3
+        with:
+          node-version: 22
```
JackDevAU, line 29: nitpick - should we use v4?

packages/create-tina-app/src/index.ts
```diff
@@ -76,38 +104,25 @@ export const run = async () => {
...
   //TODO: Update this?
   await telemetry.submitRecord({
     event: {
       name: 'create-tina-app:invoke',
-      example,
-      useYarn: Boolean(res.packageManager === 'yarn'),
+      example: template,
+      useYarn: Boolean(pkgManager === 'yarn'),
     },
   })
```
JackDevAU, line 124: not sure what we should do wit this... From my understanding this endpoint isn't live. I think we can remove this. Thoughts? @wicksipedia

JackDevAU (commented):
🦙 left some comments but looks good overall! - Need to test it still but code looks good 🚀

.github/workflows/build-starter-templates.yml
wicksipedia: use v4 for checkout + setup-node
also node version should be a matrix parameter.... need to support 18, 20, 22

packages/create-tina-app/src/index.ts
```diff
@@ -2,27 +2,36 @@ import { Telemetry } from '@tinacms/metrics'
...
 import { isWriteable, makeDir, isFolderEmpty } from './util/fileUtil'
 import { install } from './util/install'
 import chalk from 'chalk'
 import { tryGitInit } from './util/git'
-import { exit } from 'node:process'
-import { EXAMPLES, downloadExample } from './examples'
+import { TEMPLATES, downloadTemplate } from './templates'
 import { preRunChecks } from './util/preRunChecks'
 
+const PKG_MANAGERS = ['yarn', 'npm', 'pnpm']
```
wicksipedia, line 13:
```suggestion
const PKG_MANAGERS = ['npm', 'yarn', 'pnpm']
```
best to have npm first since we know that they'll have this installed (otherwise they wouldn't have been able to run `npx`)

.github/workflows/build-starter-templates.yml
brookjeynes-ssw: Done 
- Updated to v4
- Added `node-version` param for version 18, 20, and 22

packages/create-tina-app/src/index.ts
wicksipedia: we do get telemetry when the cli is invoked

I'd want to know which package manager is used (as opposed to just useYarn)

brookjeynes-ssw: Done - updated the metrics call to pass in the package manager instead of a boolean

wicksipedia (approved):
🦙 llgtm

JackDevAU (approved):
🦙 Test passed - Great upgrade to the cli 🚀

Issue #4773, opened 2024-09-09
[RESOLVED] Cannot find module 'tinacms/dist/cache'


chaotic-justice:
versions:
    "tinacms": "^2.2.3"
    "@tinacms/cli": "^1.6.3"

deployed to vercel, build successful, but returns "500 internal server error"

Error: Cannot find module 'tinacms/dist/cache'
Require stack:
- /var/task/.next/server/chunks/446.js
- /var/task/.next/server/webpack-runtime.js
- /var/task/.next/server/app/[lang]/page.js
- /var/task/node_modules/next/dist/compiled/next-server/server.runtime.prod.js

wicksipedia:
Are you able to provide a repo to reproduce the issue?

chaotic-justice:
> Are you able to provide a repo to reproduce the issue?

it was all good as of a month ago. last working commit was 08/12

https://github.com/chaotic-justice/cave-dweller

wicksipedia:
Thanks for that!

In the meantime you can try and change your build command to disable the cache by adding ‘--no-client-build-cache’

See https://tina.io/docs/cli-overview/#-tinacms-build-

chaotic-justice:
updated build command to     "build": "tinacms build --no-client-build-cache && next build",
still 500

kldavis4:
@chaotic-justice I deployed your `develop` branch to vercel and was not able to replicate any issues. I did run `npm install` and commit the `package-lock.json` - it looks like it hadn't been updated for 6 months. Try that and see if it solves your problem.

chaotic-justice:
> Thanks for that!
> 
> In the meantime you can try and change your build command to disable the cache by adding ‘--no-client-build-cache’
> 
> See https://tina.io/docs/cli-overview/#-tinacms-build-

it turns out that command solved the issue. i lazily changed the build command in `package.json` but if you were using vercel you had to manually override vercel's build command. thanks! @wicksipedia

kldavis4:
key point: these errors appear in the vercel runtime logs

kldavis4:
PR to resolve issue https://github.com/tinacms/tinacms/pull/4803

wicksipedia:
closing as resolved

Issue #4788, opened 2024-09-15
Dependency Issues


mcpeapsUnterstrichHD:
### Describe the issue that you're seeing. Any Loom videos or screenshots usually help a lot!

I start the dev server but ran in to dependecy errors in vite with no way to fix
```bash
/opt/homebrew/bin/yarn run dev
yarn run v1.22.21
warning ../../package.json: No license field
$ tinacms dev -c "next dev --turbo"
Starting Tina Dev Server
Indexing local files ⠋
Building search index ⠙
│
○  Tina Dev Server is running... ──────────────────────────────────────────╮
│                                                                          │
│  🦙 Tina Config                                                          │
│     CMS:                <your-dev-server-url>/admin/index.html           │
│     API playground:     <your-dev-server-url>/admin/index.html#/graphql  │
│     API url:            http://localhost:4001/graphql                    │
│                                                                          │
│  🤖 Auto-generated files                                                 │
│     GraphQL Client:     tina/__generated__/client.ts                     │
│     Typescript Types:   tina/__generated__/types.ts                      │
│                                                                          │
│                                                                          │
├──────────────────────────────────────────────────────────────────────────╯
Starting subprocess: next dev --turbo
✘ [ERROR] Failed to resolve entry for package "fs". The package may have incorrect main/module/exports specified in its package.json. [plugin vite:dep-pre-bundle]

    node_modules/vite/node_modules/esbuild/lib/main.js:1373:21:
      1373 │         let result = await callback({
           ╵                      ^

    at packageEntryFailure (file:///Users/mahd/DEV/comboom_punkt_sucht_app/node_modules/vite/dist/node/chunks/dep-41cf5ffd.js:28725:11)
    at resolvePackageEntry (file:///Users/mahd/DEV/comboom_punkt_sucht_app/node_modules/vite/dist/node/chunks/dep-41cf5ffd.js:28722:5)
    at tryNodeResolve (file:///Users/mahd/DEV/comboom_punkt_sucht_app/node_modules/vite/dist/node/chunks/dep-41cf5ffd.js:28453:20)
    at Context.resolveId (file:///Users/mahd/DEV/comboom_punkt_sucht_app/node_modules/vite/dist/node/chunks/dep-41cf5ffd.js:28212:28)
    at Object.resolveId (file:///Users/mahd/DEV/comboom_punkt_sucht_app/node_modules/vite/dist/node/chunks/dep-41cf5ffd.js:44276:64)
[...]

PR #4803, opened 2024-09-19
fix: `tinacms/dist/cache` not found error

kldavis4:
Fix issue #4773 

Switch to dynamic imports to load the cache module so that webpack loads it properly. Also fix the cache directory to use /tmp when needed.


.changeset/rare-turkeys-bow.md
```diff
@@ -0,0 +1,5 @@
+---
+'tinacms': patch
+---
+
+Fix module not found error at runtime for tinacms/dist/cache
```
kldavis4, line 5:
```suggestion
Fix module not found `tinacms/dist/cache` error at runtime due to broken webpack bundling
```

packages/tinacms/src/cache/node-cache.ts
```diff
@@ -6,12 +6,31 @@ const makeKey = (key: any) => {
   return createHash('sha256').update(input).digest('hex')
 }
 
-export const NodeCache = (dir: string, fs: any): Cache => {
+// makeCacheDir creates the cache directory if it doesn't exist
+const makeCacheDir = (dir: string, fs: any) => {
+  const path = require('path')
+  const os = require('os')
```
JackDevAU, line 12: Why do we need to use `require` here?

packages/tinacms/src/cache/node-cache.ts
```diff
@@ -6,12 +6,31 @@ const makeKey = (key: any) => {
...
+  if (!fs.existsSync(parts[0])) {
+    cacheDir = path.join(os.tmpdir(), parts[0])
+  }
+
+  fs.mkdirSync(cacheDir, { recursive: true })
+  return cacheDir
+}
+
+export const NodeCache = (dir: string): Cache => {
+  const fs = require('fs')
```
JackDevAU, line 27: Why do we need to use `require` here?

packages/tinacms/src/unifiedClient/index.ts
```diff
@@ -47,18 +50,29 @@ export class TinaClient<GenQueries> {
...
-    // if we're in node, use the fs cache
-    if (cacheDir && typeof require !== 'undefined') {
-      const { NodeCache } = require('tinacms/dist/cache')
-      this.cache = NodeCache(cacheDir, require('fs'))
+  async init() {
+    if (this.initialized) {
+      return
+    }
+    try {
+      if (this.cacheDir && typeof require !== 'undefined') {
```
JackDevAU, line 61: Why do we need to check if `require` is undefined here?

kldavis4: Detects if nodejs environment which has fs/path/os apis (versus edge and browser runtime)

packages/tinacms/src/cache/node-cache.ts
kldavis4: This is conditionally imported depending on whether it is a nodejs env.

The only alternative (without polyfilling) is dynamic imports - I wasn't able to get them working correctly though

kldavis4:
pushed some changes to allow dynamic imports to be used. There's still an issue with the cli dev/build commands on this project: https://github.com/comboomPunkTsucht/comboom.sucht-BLOG

Not sure what exactly is the cause yet though

kldavis4:
This also includes CLI updates to fix #4788

packages/tinacms/src/cache/node-cache.ts
```diff
@@ -1,17 +1,36 @@
...
-  return createHash('sha256').update(input).digest('hex')
+// makeCacheDir creates the cache directory if it doesn't exist
+const makeCacheDir = async (dir: string, fs: any) => {
+  const path = await import('path')
+  const os = await import('os')
+
+  const parts = dir.split(path.sep).filter(Boolean)
+
+  let cacheDir = dir
+  // check if the root directory exists, if not then create create in tmp and return new path
```
ncn-ssw, line 11: **nitpick**: Should the second 'create' be 'cache'?
```suggestion
  // check if the root directory exists, if not then create cache in tmp and return new path
```

ncn-ssw (approved):
🦙 Approved

JackDevAU (approved):
🦙

PR #4820, opened 2024-09-25
Refactored the `create-tina-app` cli tool

brookjeynes-ssw:
- Created `Logger`. Moved all pre-defined styles into `Logger`.
- Moved global variables such as `program` to local space.
- Updated `preRunChecks` to warn the user if they're using a non-supported version of Node.
- Replaced `throw new Error('...')` with `exit(1)` to clean up CLI error outputs.
- Added the ability for the user to `CTRL+C` or `SIGINT` early.
- Added validation on project name to conform with NPM naming standards.
- Refactored general code: 
  - Abstracted related logic into functions. 
  - Separated out logic from some functions doing multiple things. 
  - Renamed certain functions to be more clear.
- Added error handling around functions that could throw errors. Previously, many errors were being ignored.
- Added more logs to the CLI tool so the user can remain updated throughout the initialisation process.
- The `name` and `version` field within the generated `package.json` now reflect what was entered by the user.
- Linted files.


packages/create-tina-app/src/util/preRunChecks.ts
```diff
@@ -1,11 +1,19 @@
-/**
+import { log } from './logger'
 
-*/
+export const SUPPORTED_NODE_VERSIONS = ['18', '20', '22']
```
JackDevAU, line 3: Anyway to make this not static?

packages/create-tina-app/src/util/install.ts
```diff
@@ -80,12 +80,10 @@ export function install(
        */
       args = ['install']
       if (!isOnline) {
-        console.log(chalk.yellow('You appear to be offline.'))
+        log.warn('You appear to be offline.')
         if (packageManager === 'yarn') {
-          console.log(chalk.yellow('Falling back to the local Yarn cache.'))
+          console.warn('Falling back to the local Yarn cache.')
```
JackDevAU, line 85: should this be `log.warn`?
```suggestion
          log.warn('Falling back to the local Yarn cache.')
```

JackDevAU:
🦙 Looks good!! - just need to resolve the merge conflict!

JackDevAU (dismissed):
🦙

ncn-ssw (approved):
🦙 Runs correctly

PR #4826, opened 2024-09-26
♻️ Add index to key

JackDevAU:
Fixes an issue where it is possible to have multiple forms with the same key in React.


wicksipedia (approved):
keys are good

PR #4833, opened 2024-09-30
Add PNPM to the "build all templates" GitHub action

brookjeynes-ssw:
PNPM isn't installed within the GitHub action and therefore when it's used the action will fail.


ncn-ssw (approved):
🦙

PR #4834, opened 2024-09-30
Fix broken 'build all templates' workflow

brookjeynes-ssw:
The build is currently failing for multiple reasons, this PR fixes those issues


.github/workflows/build-starter-templates.yml
```diff
@@ -62,3 +60,16 @@ jobs:
               --body "The **${{ matrix.template }}** starter template failed to build using **${{ matrix.package-manager }}**.\n\nCheck the full workflow logs for more details: [View Workflow Run]($RUN_URL)" \
               --labels "bug"
           fi
+
+  check-failures:
```
wicksipedia, line 64: not sure what value this job is adding... is this because of the "continue on error" setting? if yes, add a comment to explain it

.github/workflows/build-starter-templates.yml
```diff
@@ -19,46 +20,65 @@ jobs:
...
           node-version: ${{ matrix.node-version }}
 
       - uses: pnpm/action-setup@v4
         name: Install pnpm
-        if: ${{ matrix.package-manager == pnpm }}
-        id: pnpm-install
         with:
-          package_json_file: package.json
-          run_install: false
+          version: 9
```
wicksipedia, line 35: do we have docs/create-tina-app checks for the pnpm version?

.github/workflows/build-starter-templates.yml
```diff
@@ -19,46 +20,65 @@ jobs:
...
 
       - name: Report errors
+        continue-on-error: false
+        id: report-errors
         if: failure()
         run: |
           ISSUE_TITLE="Starter template '${{ matrix.template }}' failed to build with ${{ matrix.package-manager }}"
           RUN_URL="https://github.com/${{ github.repository }}/actions/runs/${{ github.run_id }}"
 
+          echo "failed=true" >> "$GITHUB_OUTPUT"
```
wicksipedia, line 60:
```suggestion
          echo $ISSUE_TITLE >> $GITHUB_STEP_SUMMARY
          echo "failed=true" >> "$GITHUB_OUTPUT"
```
suggest adding step summaries as well

wicksipedia (approved):
test pass with changes

.github/workflows/build-starter-templates.yml
brookjeynes-ssw: We do not. I will go ahead and update tina.io to include information about the supported package managers.

wicksipedia (approved):

PR #4845, opened 2024-10-01
rebuild split object key with its paths

khnn:
When hosted on Vercel and if the query parameters contain an encoded slash (eg. when files are in a folder) they will be split into an array of strings. This leads to elements in a folder not getting deleted because the objectKey is just the name of the first folder. This patch is aiming to solve that by rebuilding the objectKey including its paths.

I tried to explain it well, but please feel free to update naming and description of the PR.

image
**Figure: ❌ Bad - What `media` looks like on Vercel**

image
**Figure: ✅ Good - What `media` looks like in local**

- [x] It must successfully build, test, and lint your branch
- [ ] It must pass the dangerjs script
- [x] It must be up-to-date with master
- [x] It must be approved by at least 1 tinacms core member
- [x] It must have conventional-commits that clearly explain what has been changed


JackDevAU:
Hey @khnn,

Thanks for this contribution! We'll be looking into it shortly. 🦙

khnn:
> Hey @khnn,
> 
> Thanks for this contribution! We'll be looking into it shortly. 🦙

If you need any information or context let me know.

Jord-Gui (approved):
I was able to recreate the issue and test that this change fixes it

PR #4886, opened 2024-10-03
Add Node version to GitHub issue title

brookjeynes-ssw:
- Fixed duplicate issue creation bug

### DIFF
--- a/.github/workflows/build-starter-templates.yml
+++ b/.github/workflows/build-starter-templates.yml
@@ -24,13 +24,6 @@ jobs:
       failed: ${{ steps.report-errors.outputs.failed }}
 
     steps:
-      - name: Get Current Job Log URL
-        uses: Tiryoh/gha-jobid-action@v1
-        id: jobs
-        with:
-          job_name: 'build (${{ matrix.distro }})' # input job.<job-id>.name and matrix here.
-          per_page: 54 # This is the amount of jobs there are: `package-manager * template * node-version`
-
       - name: Set up Node.js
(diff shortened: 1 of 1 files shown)


.github/workflows/build-starter-templates.yml
```diff
@@ -54,20 +54,24 @@ jobs:
         id: report-errors
         if: failure()
         run: |
-          ISSUE_TITLE="Starter template '${{ matrix.template }}' failed to build with ${{ matrix.package-manager }}"
+          ISSUE_TITLE="Starter template '${{ matrix.template }}' failed to build with ${{ matrix.package-manager }} using Node ${{ matrix.node-version }}"
           RUN_URL="https://github.com/${{ github.repository }}/actions/runs/${{ github.run_id }}"
+          ISSUE_BODY=$(cat <<EOF
+          The **${{ matrix.template }}** starter template failed to build with **${{ matrix.package-manager }}** using Node **${{ matrix.node-version }}**.
+
+          Check the full workflow logs for more details: [View Workflow Run]($RUN_URL)
```
JackDevAU, line 62: Possible to make this either, 

A. Go to the actual run 
B. Add the error output into the issue 🙏

Its very difficult to find the run when theres 30+ runs with the same names (not exactly)

brookjeynes-ssw: As per my conversation with Jack, getting the runners job ID is difficult when dealing with Matrices so for now we will continue to just link to the action.

JackDevAU (approved):

Issue #4795, opened 2024-09-17
🚸 TinaCMS | rich text editor - Obscured Toolbar Items in CMS Window


isaaclombardssw:
### Pain

It may appear Tina is missing a few rich-text options (e.g. the embed) as they can be obscured in the thin TinaCMS editor, and the toolbar isn't possible to scroll on.

My late afternoon brain came to the conclusion that those options just didn't exist for a page I was editing. 

I also reckon I'm not the first victim of this one. This is on Firefox / Mac.

I think we could improve the UX.

Additionally, to note, if you click on a half-obscured toolbar item the hidden items will be opened in a floating menu (but nothing in the UI really indicates this).

### Screenshots

![2024-09-12_09-10-14](https://github.com/user-attachments/assets/6eee2fb2-7292-469f-af46-30a57572a6d8)
**Figure 1 – missing toolbar items??**

![2024-09-12_09-09-44](https://github.com/user-attachments/assets/13745670-ba19-4f05-95e1-42d0125fe016)
**Figure 2 – revealed on a wider screen size**

![Screenshot 2024-09-12 at 12 54 39 pm](https://github.com/user-attachments/assets/d95ee15c-4d8f-4de7-9c39-60fac3e1f981)
**Figure 3 – clicking a half-obscured toolbar item to reveal the hidden toolbar items**

### Steps to reproduce

Open a CMS instance where a rich-text field is being edited.

Resize the subwindow.

### System Info

System:
    OS: macOS 13.6.1
    CPU: (4) x64 Intel(R) Core(TM) i5-7360U CPU @ 2.30GHz
    Memory: 60.18 MB / 8.00 GB
    Shell: 5.9 - /bin/zsh
Binaries:
    Node: 20.16.0 - ~/.nvm/versions/node/v20.16.0/bin/node
    Yarn: 1.22.22 - /usr/local/bin/yarn
    npm: 10.8.1 - ~/.nvm/versions/node/v20.16.0/bin/npm
    pnpm: 9.6.0 - /usr/local/bin/pnpm
Browsers:
    Safari: 16.6

### Validations

- [x] Follow our [Code of Conduct](https://github.com/tinacms/tinacms/blob/main/CODE_OF_CONDUCT.md)
- [x] Read the [docs](https://tina.io/docs).
- [x] Check that there isn't [already an issue](https://github.com/tinacms/tinacms/issues) that reports the same bug.

jeoffreyfischer:
[...]

PR #5004, opened 2024-10-04
Fix toolbar item cms

Ben0189:
#4795 

- Fix toolbar floating icon bug in rich text editor using container query in tailwind
- Refactor the width calculation to be maintainable approach
- Remove all magic number


Jord-Gui (approved):

Jord-Gui (changes requested):
Fix resizing

Ben0189:
Done fixing the resizing calculation, we will now safe from any width bug in the toolbar

Jord-Gui (changes requested):
There's a small bug where the three dots pop up even though there isn't anything to hide
image

Ben0189:
✅ Done fix - previous calculation does not take in consideration when embed button is not show 🥲 (now solved)

packages/tinacms/src/toolkit/fields/plugins/mdx-field-plugin/plate/components/fixed-toolbar-buttons.tsx
```diff
@@ -100,14 +101,19 @@ const toolbarItems: { [key in ToolbarOverrideType]: ToolbarItem } = {
...
-  const [itemsShown, setItemsShown] = React.useState(11)
-  const { overrides } = useToolbarContext()
+  const [itemsShown, setItemsShown] = React.useState(10)
+  const { overrides, templates } = useToolbarContext()
+  const showEmbedButton = templates.length > 0
 
   useResize(toolbarRef, (entry) => {
     const width = entry.target.getBoundingClientRect().width
-    const itemsShown = (width - EMBED_ICON_WIDTH) / ICON_WIDTH
+    const paragraphIconWidth = width < MD_BREAKPOINT ? 58 : 128
```
JackDevAU, line 110: should we add these floating numbers into variables?

Ben0189:
I had also refactor the width calculation function so it is now easier to add plugin without breaking the width again

Jord-Gui (approved):

Issue #4681, opened 2024-07-24
🔌 Rich text - update table plugin


wicksipedia:
Newer versions of plate have a great table editing experience - we should leverage that to improve the editing experience

- [x] Support https://platejs.org/docs/components/table-element

Jord-Gui:
Remaining work

- [ ] Ensure code and other blocks saved in the table cell save properly

Jord-Gui:
As per my conversation with @JackDevAU, prioritise the below to get an mvp out:
- See if I make any progress on the heading 1 - 6 support
- Support bullet list, numbered list, code block, quoting
- Support multiple lines within a table cell i.e. pressing enter

Once the above is working, double check with PO

Deprioritise the below
- Will probably need to support custom templates as well
- Table header
- Cell resizing

Jord-Gui:
Cc: @Ben0189 

As per my conversation with @jeffsee55, @JackDevAU, [mdast](https://github.com/syntax-tree/mdast?tab=readme-ov-file#phrasingcontent) does not support having block elements within table cells. To support this, we would need to write our own parser, which requires much more effort. 

As suggested by Jeff, the easiest course of action now is to disable the unsupported cell formats when inside a table. We can ship this and then if it is needed, continue the conversation of a custom parser

Jord-Gui:
As per my conversation with @wicksipedia, in the addition to the above task:
- [ ] Handle gracefully failing the ones that aren't supported (e.g. when a user edits the raw markdown)
- [ ] Update the documentation on what is supported/unsupported
- [ ] Create new PBIs and prioritise with PO for the ones we want to support custom

Jord-Gui:
Cc: @JackDevAU 

After testing what happens when you use raw markdown to edit the table, the parser treats unsupported formats as a paragraph string and displays it literally. e.g. "1. Numbered list". 

As per my conversation with @wicksipedia, try and see if we can give a warning in the editor. Something along the lines of "you've put in an unsupported file". If it is a larger effort than expected, create a new PBI for it along with the others above.

Jord-Gui:
Cc: @wicksipedia @JackDevAU 

As per my conversation with @adamcogan, once this has been released, record a Done Video.
- [ ] Record Done Video

Jord-Gui:
Changes have been merged into main, just need to complete the above admin tasks before closing this off.

Jord-Gui:
[...]

PR #5098, opened 2024-10-07
✨ Rich Text - Add table plugin

Jord-Gui:
Implements #4681 

The rich text editor now natively supports adding a table.
image
**Figure: Three tables created using the rich text editor**


packages/@tinacms/scripts/src/index.ts
```diff
@@ -30,7 +30,12 @@ interface Option {
 
 const deepMerge = (target, source) => {
   for (const key in source) {
-    if (!source.hasOwnProperty(key) || key === '__proto__' || key === 'constructor') continue;
+    if (
+      !source.hasOwnProperty(key) ||
+      key === '__proto__' ||
+      key === 'constructor'
+    )
+      continue
```
ncn-ssw, line 38: **issue**: If `continue` is to broken out to its own line, please use braces as per https://www.ssw.com.au/rules/how-to-avoid-problems-in-if-statements/

wicksipedia: better to have an eslint rules turned on for this
https://eslint.org/docs/latest/rules/brace-style
https://eslint.org/docs/latest/rules/curly

packages/tinacms/src/toolkit/fields/plugins/mdx-field-plugin/plate/index.tsx
```diff
@@ -57,7 +57,6 @@ export const RichEditor = (props: RichTextType) => {
   )
 
   // This should be a plugin customization
```
ncn-ssw, line 59: **question**: Is this comment referring to the removed line immediately below it, or the 'paragraph' of code?

packages/@tinacms/scripts/src/index.ts
Jord-Gui: I have reverted the change for now as its not relevant to this PR

packages/tinacms/src/toolkit/fields/plugins/mdx-field-plugin/plate/index.tsx
Jord-Gui: I have reverted this change as discussed as it is not relevant to this PR

ncn-ssw (approved):
🦙 Excellent work resolving the many finicky issues that we found in testing

PR #5191, opened 2024-10-11
Add project build tokens to build all templates action

brookjeynes-ssw:
Add project build tokens to build all templates action. This allows the projects to build against TinaCloud instead of locally fixing a series of NextJS issues we ran into due to the Tina `--local` build flag not working for SSR


ncn-ssw (approved):
🦙

PR #5204, opened 2024-10-15
lock mermaid to 9.3.0 to allow cjs

JackDevAU:
- Fixes issue with `mermaid` version 10+ moving to pure esm output. (can't use require to import)

```js
Module not found: ESM packages (mermaid) need to be imported. Use 'import' to reference the package instead. https://nextjs.org/docs/messages/import-esm-externals
``` 

[Read more about pure esm](https://gist.github.com/sindresorhus/a39789f98801d908bbc7ff3ecc99d99c)

### DIFF
--- a/.changeset/violet-kings-occur.md
+++ b/.changeset/violet-kings-occur.md
@@ -1,0 +1,5 @@
+---
+"tinacms": patch
+---
+
+lock mermaid to 9.4.0 to allow cjs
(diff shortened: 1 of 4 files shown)


ncn-ssw (approved):
🦙 Approved - as per my conversation with @JackDevAU, we are in not in a position to go to full ESM in the near future.

PR #5205, opened 2024-10-15
♻️ Rich Text - Table Plugin - Add mermaid button to unsupported table type

Jord-Gui:
Ensure that the mermaid button doesn't show up in the fixed-toolbar when a user is inside a table.

image
**Figure: Mermaid button not shown when user is inside a table**

image
**Figure: Mermaid button shows up when user is outside a table**


packages/tinacms/src/toolkit/fields/plugins/mdx-field-plugin/plate/plugins/core/common.tsx
```diff
@@ -59,6 +59,7 @@ export const unsupportedItemsInTable = new Set([
   'Unordered List',
```
JackDevAU, line 59: I think we should change this at a later stage, I feel its going to be forgotten and someones going to change the label/tooltip of one of the buttons and this will break.

JackDevAU (approved):

PR #5219, opened 2024-10-18
🚧 WIP - feat: reverse reference queries

kldavis4:
# what

Add query support for reverse references and update tina-self-hosted-demo to provide an example


kldavis4:
docs update: https://github.com/tinacms/tina.io/pull/2475

packages/@tinacms/graphql/src/build.ts
```diff
@@ -224,3 +234,36 @@ const _buildSchema = async (builder: Builder, tinaSchema: TinaSchema) => {
...
 }
+
+function collectionHasReferenceFields(collection: Collection<true>) {
+  let result = false
+  if (!collection.fields) {
+    return false
+  }
+  for (const field of collection.fields) {
+    if (field.type === 'reference') {
+      result = true
```
ncn-ssw, line 245: **suggestion**: As `result` cannot be returned to false after line 243, to short-circuit unnecessary loops I suggest:
- removing `result` declaration at line 239.
- returning `true` at this line (line 245)
- returning `false` at line 252.

packages/@tinacms/graphql/src/build.ts
```diff
@@ -224,3 +234,36 @@ const _buildSchema = async (builder: Builder, tinaSchema: TinaSchema) => {
...
+    } else if (field.type === 'object') {
+      if (fieldHasReferenceFields(field)) {
+        return true
+      }
+    }
+  }
+  return result
```
ncn-ssw, line 252: **suggestion**: As `result` cannot be returned to false after line 243, to short-circuit unnecessary loops I suggest:

```suggestion
  if (!collection.fields) {
    return false
  }
  for (const field of collection.fields) {
    if (field.type === 'reference') {
      return true
    } else if (field.type === 'object') {
      if (fieldHasReferenceFields(field)) {
        return true
      }
    }
  }
  return false
```

ncn-ssw:
⚠️ **Warning** - Do not merge until further notice.

This change is likely to introduce backwards-incompatibilities with previous version of the package.

mishmanners:
People on our Discord are asking for this.

oxc:
I get the appeal of a reverse query operation, but wouldn't it be easier to be able to filter reference fields by slug/path in addition to the "deep" filtering?

JackDevAU:
Closing as stale — reverse reference queries will be implemented in v4, tracked under #6921.

Issue #5226, opened 2024-10-22
"r.join is not a function" in NextJs during generateStaticParams in Production Build


abel3t:
### Describe the issue that you're seeing. Any Loom videos or screenshots usually help a lot!

While creating an production build, the following error occurs during the "Collecting page data" step:

```bash
TypeError: r.join is not a function
```

### Reproduction

```bash
npm run build
```

### Steps to reproduce

1. **Set up a Next.js 14 project using the App Router**:
   - Ensure your project is using the App Router (`app` directory).
   
2. **Implement `generateStaticParams`**:
   - Add the following code in your component or page file to generate static parameters:

   ```typescript
   export async function generateStaticParams() {
     const posts = await client.queries.postConnection();
     return (
       posts?.data?.postConnection?.edges?.map((post) => ({
         slug: post?.node?._sys.filename,
       })) || []
     );
   }
   ```

3. **Run a production build**:
   - Execute the build command to generate a static site: `npm run build` or `yarn build`.

### System Info

```bash
System:
    OS: macOS 15.0.1
    CPU: (8) arm64 Apple M1 Pro
    Memory: 425.78 MB / 16.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 20.8.1 - ~/.nvm/versions/node/v20.8.1/bin/node
    Yarn: 1.22.19 - ~/.yarn/bin/yarn
    npm: 10.1.0 - ~/.nvm/versions/node/v20.8.1/bin/npm
    pnpm: 8.6.10 - ~/Library/pnpm/pnpm
    bun: 1.0.30 - ~/.bun/bin/bun
  npmPackages:
    @tinacms/cli: ^1.5.53 => 1.6.8 
    tinacms: ^2.1.1 => 2.3.0 
```

### Validations

- [x] Follow our [Code of Conduct](https://github.com/tinacms/tinacms/blob/main/CODE_OF_CONDUCT.md)
- [x] Read the [docs](https://tina.io/docs).
- [x] Check that there isn't [already an issue](https://github.com/tinacms/tinacms/issues) that reports the same bug.

mAAdhaTTah:
I'm also experiencing this, although it doesn't seem to stop the build.

JackDevAU:
Hi @abel3t & @mAAdhaTTah 

Cheers for finding this - Taking a look now 🙏

JackDevAU:
Hi All, 

A fix has been released in the following versions, 

[...]

PR #5228, opened 2024-10-23
💚 Fix r.join is not a function

JackDevAU:
- Attempts to fix - #5226

Tested Build, 

- ✅ Locally 
- ✅ GitHub Pages
- ✅ Self hosted
- ✅ Vercel

### DIFF
--- a/.changeset/good-buses-mix.md
+++ b/.changeset/good-buses-mix.md
@@ -3,4 +3,4 @@
 "@tinacms/cli": patch
 ---
 
-⚠️  Don't Merge ⚠️  makeCacheDir Check
+🐛 fix r.join error when building pages
--- a/.changeset/good-buses-mix.md
+++ b/.changeset/good-buses-mix.md
@@ -1,5 +1,6 @@
 ---
 "tinacms": patch
+"@tinacms/cli": patch
 ---
 
 ⚠️  Don't Merge ⚠️  makeCacheDir Check
--- a/.changeset/good-buses-mix.md
+++ b/.changeset/good-buses-mix.md
@@ -1,0 +1,5 @@
+---
+"tinacms": patch
+---
+
+⚠️  Don't Merge ⚠️  makeCacheDir Check
--- a/packages/@tinacms/cli/src/next/vite/index.ts
+++ b/packages/@tinacms/cli/src/next/vite/index.ts
@@ -157,6 +157,8 @@ export const createConfig = async ({
 
   const isBrowser = typeof window !== 'undefined'
 
+  console.log('configManager IS BROWSER? - ', isBrowser)
+
   const alias = {
     TINA_IMPORT: configManager.prebuildFilePath,
     SCHEMA_IMPORT: configManager.generatedGraphQLJSONPath,
@@ -170,6 +172,9 @@ export const createConfig = async ({
       path: path.join(configManager.spaRootPath, 'src', 'dummy-client.ts'),
     }),
   }
+
+  console.log('alias - ', JSON.stringify(alias))
+
   if (configManager.shouldSkipSDK()) {
     alias['CLIENT_IMPORT'] = path.join(
       configManager.spaRootPath,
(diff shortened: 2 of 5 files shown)


.changeset/good-buses-mix.md
```diff
@@ -0,0 +1,5 @@
+---
+"tinacms": patch
+---
+
+⚠️  Don't Merge ⚠️  makeCacheDir Check
```
JackDevAU, line 5:
```suggestion
---
"tinacms": patch
"@tinacms/cli": patch
---

⚠️  Don't Merge ⚠️  makeCacheDir Check
```

.changeset/good-buses-mix.md
```diff
@@ -0,0 +1,6 @@
+---
+"tinacms": patch
+"@tinacms/cli": patch
+---
+
+⚠️  Don't Merge ⚠️  makeCacheDir Check
```
JackDevAU, line 6:
```suggestion
🐛 fix r.join error when building pages
```

packages/tinacms/src/cache/node-cache.ts
```diff
@@ -1,25 +1,41 @@
...
   }
 
-  fs.mkdirSync(cacheDir, { recursive: true })
   return cacheDir
 }
 
 export const NodeCache = async (dir: string): Promise<Cache> => {
-  const fs = await import('node:fs')
-  const { createHash } = await import('node:crypto')
+  // Change to require?
```
JackDevAU, line 35:
```suggestion

```

packages/tinacms/src/cache/node-cache.ts
```diff
@@ -1,25 +1,40 @@
...
 
+  const pathParts = dir.split(path.sep)
+  const cacheHash = pathParts[pathParts.length - 1]
   let cacheDir = dir
-  // check if the root directory exists, if not then create create in tmp and return new path
-  if (!fs.existsSync(path.join(path.sep, parts[0]))) {
-    cacheDir = path.join(os.tmpdir(), parts[parts.length - 1])
+
+  // ["Users", "jackpettit", "Developer", "tina", "tina-cloud-starter", "tina", "__generated__", "".cache", "1729659325506"]
+  // "/tmp/17.../"
```
ncn-ssw, line 19: **question**: What is the purpose of this comment?

ncn-ssw:
**question**: Given that we've had several issues with the caching system, can we automate the tests?

@JackDevAU

packages/tinacms/src/cache/node-cache.ts
```diff
@@ -1,25 +1,40 @@
...
+
+  if (!fs.existsSync(dir)) {
```
brookjeynes-ssw, line 21: This won't work as `dir` won't exist yet. We should check if `.../.cache` exists and if it doesn't, resort to `/tmp`

ncn-ssw: `dir` is a parameter to the function (line 5 after changes), and the existence of a valid value is checked on line 10 (after changes).

brookjeynes-ssw: Sorry @ncn-ssw, I should have been more clear. The system directory that `dir` references won't exist therefor `existsSync` will always result in false

packages/@tinacms/cli/src/next/vite/index.ts
```diff
@@ -155,17 +155,22 @@ export const createConfig = async ({
...
-    path: path.join(configManager.spaRootPath, 'src', 'dummy-client.ts'),
+    ...(isBrowser && {
+      crypto: path.join(configManager.spaRootPath, 'src', 'dummy-client.ts'),
+      fs: path.join(configManager.spaRootPath, 'src', 'dummy-client.ts'),
+      os: path.join(configManager.spaRootPath, 'src', 'dummy-client.ts'),
+      path: path.join(configManager.spaRootPath, 'src', 'dummy-client.ts'),
+    }),
   }
+
   if (configManager.shouldSkipSDK()) {
```
kldavis4, line 174: why are these changes needed? Isn't this only executed in the cli dev/build commands?

JackDevAU: I originally thought these could potentially be causing the issue as to why the import wasn't working. After testing this some more - I have noticed that isBrowser is always false. So Will remove these 🙏

kldavis4 (approved):

kldavis4 (approved):

brookjeynes-ssw (approved):

PR #5232, opened 2024-10-24
Fix an issue where text nodes without their 'type' properties weren't…

jeffsee55:
Fix an issue where text nodes without their 'type' properties weren't handling node marks properly

This appears to be a regression in the rich-text editor, though I'm not 100% sure. We should ensure that _all_ rich text nodes have a `type` property.


JackDevAU:
**Update:**

There’s currently an issue when using marks e.g. "bold" where adding a space after a word (e.g., `*bold test *`) results in the space being saved incorrectly in the Markdown file. Initially, the Rich Text editor strips this space and displays it as expected, but it breaks upon re-parsing.

I'm working on a solution to trim spaces within the formatting markers (e.g., `*`), moving the space outside the markers where necessary. This will prevent words from being joined incorrectly when the Markdown is rendered.

JackDevAU:
Update: 

After my chat with Jeff, we found this problem has been around for a while and instead of tackling it in the stringify parsing logic, this should be handled on parse. Ideally in a Plate/Slate Plugin.

Ben0189:
Need more time to deep dive in to rich text editor implementation and parsing logic

here are a useful recording for understanding rich text editor in [TinaCMS](https://sswcom-my.sharepoint.com/personal/benneoh_ssw_com_au/_layouts/15/stream.aspx?id=%2Fpersonal%2Fbenneoh%5Fssw%5Fcom%5Fau%2FDocuments%2FRecordings%2FCall%20with%20Jack%20Pettit%20%5BSSW%5D%2D20241113%5F162604%2DMeeting%20Recording%2Emp4&referrer=StreamWebApp%2EWeb&referrerScenario=AddressBarCopied%2Eview%2Eac9adf71%2D5494%2D4df3%2Da8c9%2D69549527f5b3&ga=1)

Ben0189:
As per conversation with @JackDevAU , after fixing "text nodes without their 'type' properties weren't handling node marks properly" we found that still does not solve the problem. We are now investigate into handling at slate level rather than stringify

JackDevAU:
Update:

After a discussion with Jeff, he will be continuing the remaining work. As discussed instead of using a Plate plugin to handle the space issues, we are instead moving this to the parsing layer. (i.e. on save)

JackDevAU:
Closing this as stale. v4 will fix this issue

PR #5248, opened 2024-10-30
feat: cli verification of tina schema mismatch between remote and local

### LINKED ISSUE #5221: 🐛 Bug - TinaCloud schema validation not checking paths
@wicksipedia @kldavis4 

As documented at https://github.com/tinacms/tinacms/discussions/5214, a developer has seen an issue where the build did *not* fail when the paths in the collections of their local schema in `tina-lock.json` did not match the version held by TinaCloud.

The build should have failed to ensure that the user was aware that they had not committed their updated schema.

See the [lnigojv/website](https://github.com/Inigojv/website/blob/main/) repository for the reproduction.

## Tasks

- [ ] Determine why the check against TinaCloud passed when it shouldn't have
- [ ] Fix

kldavis4:
# Problem

If there is a change to the tina schema that doesn't change the graphql schema (changes to a collection path for example) and that change is not pushed to the remote GitHub repository, there can be inconsistent behavior between local dev and production.

# Solution

Compare the local tina schema to the remote tina schema and build time. If differences are detected, alert the user.

Fix #5221

### DIFF
--- a/packages/@tinacms/cli/src/next/commands/build-command/index.ts
+++ b/packages/@tinacms/cli/src/next/commands/build-command/index.ts
@@ -214,7 +214,13 @@ export class BuildCommand extends BaseCommand {
         database,
         codegen.productionUrl
       )
-      await this.checkTinaSchema(configManager, database, codegen.productionUrl)
+      await this.checkTinaSchema(
+        configManager,
+        database,
+        codegen.productionUrl,
+        this.previewName,
+        this.verbose
+      )
     }
 
     await buildProductionSpa(configManager, database, codegen.productionUrl)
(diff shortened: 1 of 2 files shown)


packages/@tinacms/cli/src/next/commands/build-command/index.ts
```diff
@@ -598,6 +608,79 @@ export class BuildCommand extends BaseCommand {
...
+          prog: '❌',
+        })
+        let errorMessage = `The local Tina schema doesn't match the remote Tina schema. Please push up your changes to GitHub to update your remote tina schema.`
+        if (config?.branch) {
+          errorMessage += `\n\nAdditional info: Branch: ${config.branch}, Client ID: ${config.clientId} `
+        }
+        throw new Error(errorMessage)
+      }
+    } catch (e) {
+      throw e
```
ncn-ssw, line 681: **question**: Why catch and rethrow the error?

kldavis4: No reason - it was just leftover from copy/paste of existing function

brookjeynes-ssw (approved):

PR #5341, opened 2024-12-05
Remove `got` package from `create-tina-app`

JackDevAU:
- Removed `got` package as it isn't used

image

**Figure: got package version `14.x.x` does not work with node version < 20**


Ben0189 (approved):
LGTM

PR #5352, opened 2024-12-09
fix: set null for empty values in NumberInput and DateField

### LINKED ISSUE #2445: Number field cannot be empty again once edited; Invalid value
## Description
I'm using a number field with the GQL + Tina Cloud setup and realized that there is a bug in the number field. If you edit the field, e.g. from empty to `100` and then remove the `100` again, an error occurs when saving the value. If I do not touch the fields at all tho it does accept it when they are empty. Here's a video to illustrate:

https://user-images.githubusercontent.com/3264754/149910849-f6b038b0-a0c4-456f-9a45-ef475a910682.mov

I think what causes this is that the initial value is `undefined` while after touching the field and emptying it the value becomes `""` which fails some type check. To fix it maybe the [NumberInput component](https://github.com/tinacms/tinacms/blob/1a16ef190c3ad07c16751d87acbf27759d14d347/packages/%40tinacms/toolkit/src/packages/fields/components/NumberInput.tsx#L34)'s `onChange` handler should parse `""` to `undefined` again. Or maybe it should happen at some later stage?

### Steps to reproduce

1. Create a field in your schema like this:
```
{
  type: "number",
  label: "My Number",
  name: "myNumber",
}
```

2. Go into the admin editor and change the value to a valid number. Remove the value again and try to save the form. An error will occur.

### Expected result

The value should go back to undefined and saving should not fail. If the field is not marked required setting it back to empty should just work.

### Actual result

Saving fails.

### Environment

Local development environment

#### Tinacms Package Versions
```text
├── @tinacms/auth@0.50.2
├── @tinacms/cli@0.59.0
├── next-tinacms-cloudinary@3.4.5
├── tinacms@0.62.0
```

shuaixr:
Fix #2445

### DIFF
--- a/packages/tinacms/src/toolkit/fields/components/number-input.tsx
+++ b/packages/tinacms/src/toolkit/fields/components/number-input.tsx
@@ -20,7 +20,7 @@ export const NumberInput: React.FC<NumberProps> = ({
     value={value}
     onChange={(event) => {
       const inputValue = event.target.value
-      const newValue = inputValue === '' ? null : inputValue
+      const newValue = inputValue === '' ? undefined : inputValue
       if (onChange) {
         const syntheticEvent = {
           ...event,
--- a/packages/tinacms/src/toolkit/fields/components/number-input.tsx
+++ b/packages/tinacms/src/toolkit/fields/components/number-input.tsx
@@ -13,4 +13,24 @@ export const NumberInput: React.FC<NumberProps> = ({
   onChange,
   value,
   step,
-}) => <Input type="number" step={step} value={value} onChange={onChange} />
+}) => (
+  <Input
+    type="number"
+    step={step}
+    value={value}
+    onChange={(event) => {
+      const inputValue = event.target.value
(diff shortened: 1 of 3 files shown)


packages/tinacms/src/toolkit/fields/components/number-input.tsx
```diff
@@ -13,4 +13,24 @@ export const NumberInput: React.FC<NumberProps> = ({
...
   step,
-}) => <Input type="number" step={step} value={value} onChange={onChange} />
+}) => (
+  <Input
+    type="number"
+    step={step}
+    value={value}
+    onChange={(event) => {
+      const inputValue = event.target.value
+      const newValue = inputValue === '' ? null : inputValue
```
JackDevAU, line 23:
```suggestion
      const newValue = inputValue === '' ? undefined : inputValue
```

JackDevAU (dismissed):
Hey @shuaixr 

Thanks for this fix! Awesome stuff 🦙

JackDevAU (approved):

PR #5358, opened 2024-12-11
🚧 WIP - ⚡ Performance - Allow concurrent loading of files during queries

ncn-ssw:
This PR allows entries to be loaded concurrently in batches during a query while maintaining the required sort order in cases where the required filtering information is not held as part of the sort-key. The aim is to improve query performance for unoptimized queries when entries must be accessed using network requests, as they are from TinaCloud.

This change will require the splitting of the `Database.query` method due to the large differences in how item records are constituted from sort-key information and from retrieved entries.

### DIFF
--- a/packages/@tinacms/graphql/src/database/index.ts
+++ b/packages/@tinacms/graphql/src/database/index.ts
@@ -1,5 +1,4 @@
 import path from 'node:path'
-import fs from 'fs-extra'
 import type { DocumentNode } from 'graphql'
 import { GraphQLError } from 'graphql'
 import micromatch from 'micromatch'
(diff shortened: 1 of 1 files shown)


ncn-ssw:
⚠️ Warning - This PR requires substantial further work to bring it to an acceptable standard (e.g. type naming)

ncn-ssw:
❌ Closing:

- The more common cause of poor performance appears to be the deserialization of Markdown content on every query.
- Leaving this PR open will incur too much merge debt for it to be useful.
- The problem that this PR would have resolved can be worked around by using a [multi-field index](https://tina.io/docs/graphql/queries/advanced/performance).

PR #5366, opened 2024-12-16
POC : Support path alias in tina.config

Ben0189:
The alias path in the ts config file is now able to resolve by esbuild for tina.config file 

**Summary of work done in this pr** 
- update esbuild packages, the latest esbuild version support the path alias from ts config


kldavis4:
@Ben0189 what is the [esbuild-plugin-alias-config](https://www.npmjs.com/package/esbuild-plugin-alias-path) providing?

Ben0189:
> @Ben0189 what is the [esbuild-plugin-alias-config](https://www.npmjs.com/package/esbuild-plugin-alias-path) providing?

@kldavis4 Initial plan was to use this lib to support the alias from tsconfig
image
Figure: resources from esbuild [community](https://github.com/esbuild/community-plugins)

But looks like latest esbuild version does that for us

packages/@tinacms/cli/package.json
```diff
@@ -78,7 +78,7 @@
 		"cors": "^2.8.5",
 		"crypto-js": "^4.2.0",
 		"dotenv": "^16.4.5",
-		"esbuild": "^0.18.20",
```
ncn-ssw, line 81: **question**: Should the version of `esbuild` included here match the 1 in the [[at]tinacms/scripts](https://github.com/tinacms/tinacms/blob/67b87d8bbcdc6ff660194869e72d2a204e0d5d93/packages/%40tinacms/scripts/package.json#L25) package?

Ben0189: Done update the scripts packages to use the same version of esbuild

examples/next-2024/tsconfig.json
```diff
@@ -17,10 +21,21 @@
         "name": "next"
       }
     ],
+    "baseUrl": ".",
     "paths": {
-      "@/*": ["./*"]
+      "@lib/*": ["lib/*"],
+      "@components/header": ["components/header.ts"],
```
ncn-ssw, line 27: **question**: Are all these path aliases still in use?

examples/tina-self-hosted-demo/tina/tina-lock.json
ncn-ssw, line 1: **question**: Why delete the whole `tina-lock.json` file?

Ben0189: ah was to revert the changes to that file, but accidentally removed (using github UI)

Revert the commit, the file is now back

examples/next-2024/tsconfig.json
Ben0189: Done remove all the path, the ts config is already use in another example so we no longer need this

ncn-ssw (approved):
🦙 Modified example runs correctly

Issue #5364, opened 2024-12-13
tina-lock.json generation broken after tina > 2.5.x when sharing schema definitions with nested fields of type object between multiple templates


cod3blox:
### Describe the issue that you're seeing. Any Loom videos or screenshots usually help a lot!

Hello,

since updating to Tina > 2.5.0, I’ve been experiencing issues building my projects. 

```
Checking local Tina Schema matches server. ❌
Error: The local Tina schema doesn't match the remote Tina schema. Please push up your changes to GitHub to update your remote tina schema.
```

My schemas are relatively complex, and I reuse parts of the schema definitions. After spending a significant amount of time debugging my schema, I identified a reproducible issue that is not related to unpushed schema changes in Git.

I noticed that the generation of the tina-lock.json results to a "corrupt" schema when fields of the type "[object](https://tina.io/docs/reference/types/object
)" reach a certain level of nesting. The issue is particularly strange because the nesting depth and the bug only start causing problems when schema definitions are shared between multiple block [templates](https://tina.io/docs/reference/templates
).

To make the problem reproducible and easy to understand, I used the tina-barebone-starter as a starting point and updated it to Tina 2.5.2 / CLI 1.7.0. In this project, I extended the "page" collection with a few additional fields and implemented a simple example that consistently triggers the issue.

Please try building my example. It should consistently result in the "Checking local Tina Schema matches server" error.

Pay special attention to the file `/tina/fieldsets/image-options-effects.js`. In this file, I’ve added comments explaining the source of the issue.

### Reproduction

https://github.com/tina-cms-toolstage/tina-barebone-starter.git

### Steps to reproduce

1. clone repo (https://github.com/tina-cms-toolstage/tina-barebone-starter.git)
2. pnpm install
3. link repo to tina cloud
4. Replace the .env.example, with .env (add the needed IDs from tina cloud)
5. pnpm dev
6. push tina-lock.json to git (main branch)
7. pnpm build

### System Info

```shell
System:
    OS: Windows 11 10.0.22631
    CPU: (32) x64 AMD Ryzen 9 7950X 16-Core Processor
    Memory: 39.31 GB / 63.65 GB
  Binaries:
    Node: 20.18.0 - C:\Program Files\nodejs\node.EXE
    npm: 10.8.2 - C:\Program Files\nodejs\npm.CMD
    pnpm: 9.12.3 - C:\Program Files\nodejs\pnpm.CMD
  Browsers:
[...]

PR #5393, opened 2024-12-20
fix: issue causing #5364

kldavis4:
Fix issue causing namespace to be incorrectly generated for deeply nested fields.

This bug was due to incorrectly creating a reference to the object passed in rather than creating a new object (clone) using the spread operator.

Instead of this copy:

```
  const newNode: {
    [key in keyof T]: (T & { namespace?: string[] }) | string
  } = maybeNode
```

We are now doing a deep clone:

```
  const newNode: Node = { ...maybeNode, namespace: [...namespace] }
```

Also I noticed that there is a copy of this function in @tinacms/schema-tools so its been moved out of @tinacms/graphql (instead of copying) and test cases have been added.


kldavis4:
added test case that fails using the old function due to the bug causing #5364

packages/@tinacms/schema-tools/src/schema/addNamespaceToSchema.test.ts
```diff
@@ -0,0 +1,352 @@
...
+    // Ensure namespaces do not share references
+    expect((result.children[0] as Node).namespace).toEqual(['child1'])
+    expect((result.children[1] as Node).namespace).toEqual(['child2'])
+    expect((result.children?.[1]?.children?.[0] as Node).namespace).toEqual([
+      'child2',
+      'child1',
+    ])
+  })
+
+  test('should generate unique namespaces for nested nodes with similar names', () => {
```
brookjeynes-ssw, line 186: Should namespaces be unique? This test does not test for that.

packages/@tinacms/schema-tools/src/schema/addNamespaceToSchema.test.ts
```diff
@@ -0,0 +1,352 @@
...
+          },
+        },
+      ],
+    })
+
+    // Ensure the namespaces are not modified
+    expect(result).toStrictEqual(node) // Check if the function returns the same object
+  })
+
+  test('deeply nested schema issue #5364', () => {
```
brookjeynes-ssw, line 345: Test "should preserve existing namespaces and not modify them" says that namespaces should not be modified once created. Does that then mean this test is doing nothing since they can not be modified?

packages/@tinacms/schema-tools/src/schema/addNamespaceToSchema.test.ts
```diff
@@ -0,0 +1,352 @@
...
+
+    // Ensure nested namespaces are correctly formed
+    expect((result.children[0] as Node).namespace).toEqual(['child'])
+    expect((result.children[0].nested as Node).namespace).toEqual([
+      'child',
+      'child',
+    ])
+  })
+
+  test('should preserve existing namespaces for repeated names at different levels', () => {
```
brookjeynes-ssw, line 265: Are "should preserve existing namespaces for repeated names at different levels" and "should preserve existing namespaces and not modify them" not testing for the same thing? That existing namespaces should not be modified? Could these be reduced into a single test?

packages/@tinacms/schema-tools/src/schema/addNamespaceToSchema.test.ts
kldavis4: 👍🏻 updated with more accurate name

packages/@tinacms/schema-tools/src/schema/addNamespaceToSchema.test.ts
kldavis4: The bug involves passing a schema with namespace already on it and the namespace getting modified. There is a node with this namespace:

```
[
    "page",
    "blocks",
    "generativeImageBlock",
    "advancedImageOptions",
    "effects",
    "optimizations",
    "improve"
]
```

That gets modified to:

```
[
    "page",
    "blocks",
    "imageBlock",
    "advancedImageOptions",
    "effects",
    "optimizations",
    "improve"
]
```

So the test is validating that the namespace doesn't get modified

packages/@tinacms/schema-tools/src/schema/addNamespaceToSchema.test.ts
kldavis4: 👍🏻 removed duplicate test

packages/@tinacms/schema-tools/src/schema/addNamespaceToSchema.test.ts
```diff
@@ -0,0 +1,310 @@
...
+    })
+  })
+
+  test('should handle nodes without name or value properties', () => {
+    const node = {
+      otherProp: 'data',
+      children: [{ otherProp: 'childData' }],
+    }
+    const result = addNamespaceToSchema(node)
+    console.log(result)
```
brookjeynes-ssw, line 73: Is this a stray console.log or here for a reason?

brookjeynes-ssw (approved):

packages/@tinacms/schema-tools/src/schema/addNamespaceToSchema.test.ts
kldavis4: nope :)

PR #5397, opened 2024-12-22
fix: update sanitizeUrl in RTE to not strip query params

### LINKED ISSUE #5208: Bug: URLs with no path have their query parameters removed
When I enter a URL with no path but with a query parameter into an MDX enabled rich text editor the query parameter is removed after the MDX is parsed.

For e.g. if I enter `https://example.com/?utm_source=blog` then I get `https://example.com`.

I did some investigation and [this code](https://github.com/tinacms/tinacms/blob/00f6525871c7c6bd40091424337df72c7bfcf783/packages/%40tinacms/mdx/src/parse/remarkToPlate.ts#L584-L591) seems to be the culprit.

```
url = 'https://example.com/?utm_source=blog'
parsedUrl = new URL(url)

parsedUrl.pathname === '/'
// is true, but
url.endsWith('/')
// is false, so
parsedUrl.origin
// is returned
```

kldavis4:
Preserve query params and fragments in sanitizeUrl. Also adds test coverage.

Fix #5208


wicksipedia (approved):
tested as working

PR #5402, opened 2024-12-29
Stop mdx tests from hanging

brookjeynes-ssw:
This PR fixes an issue where the mdx tests would hang after completion due to `vitest` by default watching.


ncn-ssw (approved):
🦙 Testing tested correct

PR #5404, opened 2024-12-30
Replace find DOM node for react 19 upgrade

Ben0189:
- refactor ClickOutBase from class component to functional component 
- replace HOC from react-clickoutside with hook from react-use


JackDevAU (approved):
🦙🚀

PR #5410, opened 2025-01-04
feat: allow bare hash string to be inserted as url

kldavis4:
# Problem

jadenadams124124 on discord reports that a bare hash like `#foobar` can not be inserted as a link for links to anchor tags in the same document

# Solution

Add support for inserting bare hash

### DIFF
--- a/packages/tinacms/src/toolkit/fields/plugins/mdx-field-plugin/plate/transforms/is-url.test.ts
+++ b/packages/tinacms/src/toolkit/fields/plugins/mdx-field-plugin/plate/transforms/is-url.test.ts
@@ -1,0 +1,53 @@
+import { isUrl } from './is-url'
+
+describe('isUrl', () => {
+  test('validates full URLs with protocol', () => {
+    expect(isUrl('http://example.com')).toBe(true)
+    expect(isUrl('https://example.com')).toBe(true)
+    expect(isUrl('ftp://example.com')).toBe(true)
+  })
+
+  test('validates localhost URLs', () => {
+    expect(isUrl('http://localhost')).toBe(true)
+    expect(isUrl('http://localhost:3000')).toBe(true)
+  })
+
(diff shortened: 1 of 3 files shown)


packages/tinacms/src/toolkit/fields/plugins/mdx-field-plugin/plate/transforms/is-url.test.ts
```diff
@@ -0,0 +1,53 @@
+import { isUrl } from './is-url'
+
+describe('isUrl', () => {
+  test('validates full URLs with protocol', () => {
```
brookjeynes-ssw, line 4: Does this test and "validates non-localhost domain URLs" test for the same thing?

packages/tinacms/src/toolkit/fields/plugins/mdx-field-plugin/plate/transforms/is-url.test.ts
```diff
@@ -0,0 +1,53 @@
...
+  test('validates localhost URLs', () => {
+    expect(isUrl('http://localhost')).toBe(true)
+    expect(isUrl('http://localhost:3000')).toBe(true)
+  })
+
+  test('validates non-localhost domain URLs', () => {
+    expect(isUrl('http://example.com')).toBe(true)
+    expect(isUrl('https://example.co.uk')).toBe(true)
+  })
+
```
kldavis4, line 19:
```suggestion
```

packages/tinacms/src/toolkit/fields/plugins/mdx-field-plugin/plate/transforms/is-url.test.ts
kldavis4: removed the other test

brookjeynes-ssw (approved):

PR #5436, opened 2025-01-21
🐛 Bug - Race condition in the request cache

ncn-ssw:
This change fixes a bug where partial files were being read from the request cache during builds, resulting in failed builds.

It fixes the bug by doing the following:

- Adding a direct dependency on [async-lock](https://www.npmjs.com/package/async-lock) package. Note that there was already an indirect dependency.
- Removing the error throwing when the cache fails to read or write, to improve the resilience of builds.
- Adding locking to the request method based on the cache key generated from the query. This locking performs 2 functions:
    - Prevents the near-simultaneous transmission of 2 or more copies of the same request, by forcing later requests to wait for the result of the first request.
    - Prevents the reading of partially written cache files.

### DIFF
--- a/.changeset/plenty-dingos-tickle.md
+++ b/.changeset/plenty-dingos-tickle.md
@@ -1,0 +1,5 @@
+---
+"tinacms": patch
+---
+
+🐛 Bug - Fixes race condition in the request cache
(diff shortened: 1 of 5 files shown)


JackDevAU (commented):
Code looks good! 🦙 will test 🔜

brookjeynes-ssw (approved):

ncn-ssw:
✅ Done - released in [tinacms version 2.6.1](https://www.npmjs.com/package/tinacms/v/2.6.1) on Thursday 23 January 2025

PR #5439, opened 2025-01-22
feat: test improvements for @tinacms/graphql

kldavis4:
# what

#4700 has a number of changes to tests that can be split into a separate pull request.

Changes

- jest -> vitest
- add @tinacms/graphql/tests


ncn-ssw (approved):
🦙 Approved

I have restored the original `package.json` "test" and "test-watch" scripts. This is to keep behavior consistent with other packages, and prevent reports of the tests hanging.

PR #5441, opened 2025-01-22
fix: report error when creating pr fails

brookjeynes-ssw:
This change fixes an issue where a failed "create_pull_request" request would 
cause tina to get stuck on the "Creating PR" page. Failure to create a PR
would leave the user in a state where they could not create it again without
deleting the branch. ~This PR also fixes that issue by attempting to create a
PR on save.~

(Edit: Strikethrough by @ncn-ssw - testing demonstrated other side effects beyond saving, so that part of the PR was removed)


ncn-ssw (approved):
🦙 Approved - testing demonstrated that error message was shown, breaking loop

I have had to remove the attempt to create a PR upon save. During testing, I found that the create_pull_request endpoint was being called for reasons other than just during the saving of content.

Issue #5395, opened 2024-12-20
Browser tab suddenly closes when adding an RTE custom node


guillermodlpa:
### Describe the issue that you're seeing. Any Loom videos or screenshots usually help a lot!

I apologize for not having reproduction steps, but I think I've found the problem nonetheless thanks to this log I got one time:

![Image](https://github.com/user-attachments/assets/85038bd8-5d8d-411c-8b2f-9b691c915cbb)

I am adding a custom element on the RTE editing interface, after the rendered nodes on the page are out of sync with the ones in the editor field. What happens is that the entire browser tab closes, without opportunity to save changes. 

**The point is that I've found a line in tinacms calling** `window.close`. It appears to be accidental. I guess it was intending to close the parent popover menu, it's actually closing the entire browser window without warning to the user.

https://github.com/tinacms/tinacms/blame/67b87d8bbcdc6ff660194869e72d2a204e0d5d93/packages/tinacms/src/toolkit/fields/plugins/mdx-field-plugin/plate/components/plate-ui/templates-toolbar-button.tsx#L65

If you notice, there's no `close` prop being passed anywhere, so it results on `window.close`. Weird that the linter didn't catch this problem, no?

### Reproduction

n/a

### Steps to reproduce

I have experienced this problem live, but I don't know how to get to the situation to replicate. I've managed to replicate sometimes by:

1. On a rich text field, adding a custom component.
2. On Inspector, deleting the HTML for it.
3. Then adding another custom component.

### System Info

```shell
n/a
```

### Validations

- [x] Follow our [Code of Conduct](https://github.com/tinacms/tinacms/blob/main/CODE_OF_CONDUCT.md)
- [x] Read the [docs](https://tina.io/docs).
- [x] Check that there isn't [already an issue](https://github.com/tinacms/tinacms/issues) that reports the same bug.

muffn:
I think I have the same or at least a similar bug but can't explain what is happening. 

In dev mode, when I add a custom RTE element it instantly closes my browser tab 100% of the time while in production it works flawlessly.

guillermodlpa:
Thanks for fixing it! 🤍

PR #5447, opened 2025-01-24
fix: #5395 call setOpen instead of window.close

kldavis4:
As reported in #5395 when adding an embed in the rich text editor, the browser window sometimes closes. Instead of calling `[window.]close()`, we should be calling `setOpen(false)`


ncn-ssw (approved):
🦙 Tested correct

PR #5469, opened 2025-02-03
add node:* prefix :)

kourosh-alasti:
Hey lads, quick fix. Been running into issue with cloudflare workers and their node.js compatibility and poly-fills. 

Figured i would suggest fixing this as this has started becoming the standard. Thanks!


JackDevAU (approved):

JackDevAU:
Cheers for this!

PR #5474, opened 2025-02-06
refactor: update reference components to consume max width in sidebar

wicksipedia:
The reference picker was locked down to a width smaller than the sidebar unnecessarily.

This PR makes the dropdown trigger max width of the sidebar and makes the popover content container the same width (as determined by the css variable).

**Before**
CleanShot 2025-02-06 at 16 32 52

**After**
CleanShot 2025-02-06 at 16 26 47


JackDevAU (approved):
🦙🚀

PR #5479, opened 2025-02-10
🚧 WIP - ⚡ Reliability - Introduce concurrent connection limit

ncn-ssw:
This change introduces a concurrent connection limit to reduce the chance of the client being unnecessarily throttled. This should improve the reliability of a typical developer's experience.

As semaphores are required for concurrency control, this change will add the [async-sema](https://www.npmjs.com/package/async-sema) library to the dependencies.


ncn-ssw:
❌ Not feasible - during testing it became evident that this approach would not work as memory is not shared between concurrent page builds

PR #5481, opened 2025-02-11
fix: ensure existing indexes for doc are removed before reindexing

kldavis4:
Fixes: https://github.com/tinacms/tinacloud/issues/2603

Fix issue where existing indexes are not removed prior to reindexing via the `indexContentByPaths` database method


JackDevAU (approved):
✅🦙

PR #5497, opened 2025-02-15
fix: handle deeply nested references

### LINKED ISSUE #5471: 🐛 Bug - References not updated if wrapped within a list
@wicksipedia @JackDevAU @kldavis4 

When a referenced item is deleted, any references that are themselves wrapped within a list are not updated. This can result in users experiencing errors during build with messages such as "Error: Unable to fetch, please see our FAQ for more information..." as their process attempts to access items that are no longer in existence.

For instance, any 'author' reference specified as follows will not be updated:

```js
{
  type: "object",
  name: "subAuthorRefs",
  label: "Sub-Author References",
  list: true,
  fields: [{
    type: "reference",
    label: "Author",
    name: "author",
    collections: ["author"],
  }],
},
```

Note that you cannot have a list with a reference type directly - in this case, the references are wrapped within a list of objects.

A replication of this bug is in the [ref-integrity-bug](https://github.com/ncn-ssw/ref-integrity-bug/) repository, and its README contains instructions for replicating the bug.

This bug was found while investigating [an issue reported in the Discord channel](https://discord.com/channels/835168149439643678/1336088843799105546)

## Tasks

- [ ] Fix

kldavis4:
# Problem

The current reference checking logic relies on existing data layer indexes to query for references. If the reference is nested (ie field on a list of objects), tina is unable to execute the query because indexing nested fields is not currently supported.

# Solution

Implement a virtual references index on a collection which contains all of the references to a collection, including the jsonpath to the document being referenced and the document where the reference is located. This references index is then queried in the hasReferences & findReferences functions.

Addresses issue logged in #5471


.changeset/strange-apes-sparkle.md
```diff
@@ -0,0 +1,7 @@
+---
+'@tinacms/schema-tools': patch
+'@tinacms/graphql': patch
```
ncn-ssw, line 3: **question**: Should this be a patch or a minor, given this is adding a reference index to collections?

kldavis4: It should be a backwards compatible change, so it's not necessary for it to be a minor (and have a new version of the endpoint)

ncn-ssw: I'm finding that the TinaCloud hash test is failing if:
- TinaCloud is updated to use the new change, but
- A repo with nested lists [such as the bug replication](https://github.com/ncn-ssw/ref-integrity-bug/) is not.

Given that some sites would only be updating content with rebuilding their schemas, this change does appear to require a 'minor' bump.

ncn-ssw:
**issue**: When I test the fix using the [ref-integrity-bug](https://github.com/ncn-ssw/ref-integrity-bug) package, it appears that all references are not being updated upon delete.

To replicate:

1. Clone the ref-integrity-bug repository linked above
2. Update the repository to use a tinacms package updated with this fix - I have done this through `pnpm link`
3. Run `npm install && npm dev`
4. Go to http://localhost:3000/admin/index.html#/collections/author/~
5. Delete `content/authors/napoleon.md`
6. Go to http://localhost:3000/admin/index.html#/~/posts/learning-about-markdown

When I do this, I am seeing an error stating that it is "Unable to find record content/authors/napoleon.md", as per the screenshot below.

image  
**Figure: Full error dialog**

.changeset/strange-apes-sparkle.md
kldavis4: I ran the following test
- Create a project with the current latest version of @tinacms/graphql 
- Update tina cloud locally with the tagged version of @tinacms/graphql (`0.0.0-d9672bc-20250218033222`) - don't reindex the project after the update
- Configure a clone of the [repo](https://github.com/ncn-ssw/ref-integrity-bug/) against the local tina cloud and run the build command:

```
> tinacms build && next build

Starting Tina build
Checking clientId and token. ✅
Checking branch is on Tina Cloud. ✅
Checking indexing process in Tina Cloud... ✅

Checking local GraphQL Schema matches server. ✅
Checking local Tina Schema matches server. ✅
│
○  Tina build complete ───────────────────────────────────────────────────────────────────────────────────────────────╮
│                                                                                                                     │
│  🦙 Tina Config                                                                                                     │
│     API url:            https://kd-content.tinajs.dev/1.5/content/af7503e5-cd79-4561-a7ee-cc383a03d865/github/main  │
│                                                                                                                     │
│  🤖 Auto-generated files                                                                                            │
│     GraphQL Client:     tina/__generated__/client.ts                                                                │
│     Typescript Types:   tina/__generated__/types.ts                                                                 │
│     Static HTML file:   public/admin/index.html                                                                     │
│                                                                                                                     │
│                                                                                                                     │
├─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯
  ▲ Next.js 14.2.23
  - Environments: .env

   Creating an optimized production build ...
Browserslist: caniuse-lite is outdated. Please run:
  npx update-browserslist-db@latest
  Why you should do it regularly: https://github.com/browserslist/update-db#readme
Browserslist: caniuse-lite is outdated. Please run:
  npx update-browserslist-db@latest
  Why you should do it regularly: https://github.com/browserslist/update-db#readme
 ✓ Compiled successfully
 ✓ Linting and checking validity of types    
 ✓ Collecting page data    
 ✓ Generating static pages (12/12)
 ✓ Collecting build traces    
 ✓ Finalizing page optimization    

Route (app)                                 Size     First Load JS
┌ ○ /                                       195 B          1.45 MB
├ ○ /_not-found                             882 B          89.4 kB
├ ● /[...filename]                          194 B          1.45 MB
├   └ /about
├ ○ /posts                                  1.6 kB          486 kB
└ ● /posts/[...filename]                    29.3 kB         514 kB
    ├ /posts/june/learning-about-tinacloud
    ├ /posts/learning-about-components
    ├ /posts/learning-about-markdown
    └ [+3 more paths]
+ First Load JS shared by all               88.5 kB
  ├ chunks/453-9a0547fb7948e52c.js          31.7 kB
  ├ chunks/fe07e1a7-bbbaf61941d50a90.js     53.7 kB
  └ other shared chunks (total)             3.06 kB


○  (Static)  prerendered as static content
●  (SSG)     prerendered as static HTML (uses getStaticProps)
```

I also made some changes using tina locally to verify the graphql api. Not seeing any errors / failures.

Can you share more details?

kldavis4:
@ncn-ssw I am unable to replicate the [issue](https://github.com/tinacms/tinacms/pull/5497#issuecomment-2667118523) 

I am using the tagged release `0.0.0-d9672bc-20250218033222` so I suspect the issue may be with pnpm link

Here's what the updated files look like:
Screen Shot 2025-02-18 at 6 43 52 PM

ncn-ssw:
Hi @kldavis4 

Regarding the [references not being updated issue](https://github.com/tinacms/tinacms/pull/5497#issuecomment-2667118523), both @JackDevAU and I have both replicated the issue. I have replicated it using tagged releases 0.0.0-d9672bc-20250218033222 and 0.0.0-8d8586e-20250219031255 (the latest at the time of this post).

I have updated the [ref-integrity-bug](https://github.com/ncn-ssw/ref-integrity-bug) repo to reference the [0.0.0-8d8586e-20250219031255 tagged release](https://github.com/ncn-ssw/ref-integrity-bug/blob/main/package.json) in case I have made a mistake in how we've linked it.

image  
**Figure: The section of the diff showing that 1 reference has been removed and another remains**

.changeset/strange-apes-sparkle.md
ncn-ssw: I'm sorry @kldavis4 , I have made a mistake in my testing as I cannot replicate the hash problem using the conditions I've described above. I believe that I had not correctly committed the `tina/tina-lock.json` file.

Unfortunately, that still demonstrates an issue 😭.

I have updated the [ref-integrity-bug](https://github.com/ncn-ssw/ref-integrity-bug) repository to:
- use the latest tag release for this PR and 
- retain the `tina/tina-lock.json` file from `tinacms` package version 2.6.2.

This condition is roughly what a developer would see if they accepted a dependabot PR, and you should see that it fails on `pnpm build` with an updated TinaCloud instance. Though we know that it can be resolved by committing the lock file, I don't believe that the average developer would anticipate that a patch upgrade would require any code change.

Given that `@tinacms/graphql` package is an indirect dependency, it is also possible that older versions of `tinacms` may pull in the latest "1.x.x" version of `@tinacms/graphql`, depending on the package manager and whether a package lock file has been included. @JackDevAU believes this contributed to the issues we saw around the [tinacms v2.5.0 release](https://discord.com/channels/835168149439643678/857232913044996097/1311949372560506961)

@JackDevAU's assessment is that any change to the `tina/tina-lock.json` file output should be treated as backwards incompatible, and hence should be considered a major version bump for the `@tinacms/graphql` package.

ncn-ssw: To be clear, we're talking about a major update rather than a minor. Long term though, this would be very restrictive to Tina's continued development.

Is amending the schema check an option?

kldavis4: How do you mean amending the schema check?

ncn-ssw: Making it tolerant to the addition of fields, allowing the check to pass so long any additional fields are set to default values.

ncn-ssw: Note that this a broader question than this PR, and would not be resolved in this PR.

kldavis4: I'll look into it. Don't recall for sure where the logic is. I think it is in the cli so it's possible making it more flexible would require a library upgrade for users

kldavis4: Confirmed it is performing the comparison in the client: https://github.com/tinacms/tinacms/blob/main/packages/%40tinacms/cli/src/next/commands/build-command/index.ts#L589

So making the lock check more flexible will require a cli upgrade

kldavis4: @ncn-ssw I updated the pr to preserve the existing walkFields method when generating the schema so there should no longer be any schema changes associated with this pull request

ncn-ssw (approved):
🦙 Nice work - I have tried my best to break this change and can't.

As per my conversation with @JackDevAU, I have added a note to review any pseudo-indexes when we look to move to the next major version.

PR #5498, opened 2025-02-15
fix: correctly handle noop renames

kldavis4:
# problem

Renaming a file to the same name fails. Also, current UI doesn't make it clear whether extension is required or not.

# solution

Add handling in resolver for no-op renames and update UI to prevent renaming to same name. Also don't display extension to user in modal to prevent confusion.


ncn-ssw (approved):
🦙 Tested correct

ncn-ssw (approved):
🦙 Tested correct

PR #5509, opened 2025-02-20
Add auto format on rich text editor for strikethrough

Ben0189:


packages/tinacms/src/toolkit/fields/plugins/mdx-field-plugin/plate/plugins/core/autoformat/autoformat-marks.ts
```diff
@@ -31,4 +32,9 @@ export const autoformatMarks: AutoformatRule[] = [
     type: MARK_CODE,
     match: '`',
   },
+  {
+    mode: 'mark',
+    type: MARK_STRIKETHROUGH,
+    match: '~~',
+  },
```
JackDevAU, line 39: Just to make it consistent with the spec we should also add a match for a single `~` as well.

Ben0189: Done

JackDevAU (approved):
🦙 LGTM

Issue #5513, opened 2025-02-20
Media links previews with spaces or non-ASCII characters are broken starting from v2.6.3


itBAIT:
### Describe the issue that you're seeing. Any videos (eg. YouTube or Loom) or screenshots help!

I use AWS S3 Bucket media. Starting from 2.6.3 links containing spaces or any non-ASCII characters are broken in the media 
manager. This pull request is the issue: #5448 

### Reproduction

na

### Steps to reproduce

1. Upload an image with the name "space non-ascii-ę.png"
2. The function encodeUrlIfNeeded returns the extra-encoded pathname: 
   https://example.eu-central-1.amazonaws.com/space%2520non-ascii-%25C4%2599.png
3. The expected result is:
   https://example.eu-central-1.amazonaws.com/space%20non-ascii-%C4%99.png

If a media folder contains such characters all links are broken too

<img width="617" alt="Image" src="https://github.com/user-attachments/assets/962eb1a0-1e77-46d0-a597-a453f0b5c015" />

### System Info

```shell
System:
    OS: Windows 11 10.0.26100
    CPU: (16) x64 AMD Ryzen 7 5800H with Radeon Graphics
    Memory: 2.31 GB / 13.86 GB
  Binaries:
    Node: 20.18.0 - C:\Program Files\nodejs\node.EXE
    Yarn: 1.22.22 - ~\AppData\Roaming\npm\yarn.CMD
    npm: 11.0.0 - C:\Program Files\nodejs\npm.CMD
  Browsers:
    Chrome: 132.0.6834.160
    Firefox: 135.0.1
    Edge: Chromium (127.0.2651.86)
```

### Validations

- [x] Follow our [Code of Conduct](https://github.com/tinacms/tinacms/blob/main/CODE_OF_CONDUCT.md)
- [x] Read the [docs](https://tina.io/docs).
- [x] Check that there isn't [already an issue](https://github.com/tinacms/tinacms/issues) that reports the same bug.

kldavis4:
@JackDevAU fix https://github.com/tinacms/tinacms/pull/5517

wicksipedia:
Fix released in 2.7.1

PR #5517, opened 2025-02-21
fix: double encoding media urls

kldavis4:
# problem

Media manager urls with special characters are broken #5513

# solution

Fix encoding function. The URL object already encodes url parts, so the function can just return the url from the object


JackDevAU (approved):
🦙

PR #5522, opened 2025-02-26
Implement basePath handling in S3 media store

JackDevAU:
Attempts to fix: https://github.com/tinacms/tinacms/issues/5523
feat: implement basePath handling in S3 media store for improved URL Management

### DIFF
--- a/packages/next-tinacms-s3/src/handlers.ts
+++ b/packages/next-tinacms-s3/src/handlers.ts
@@ -65,7 +65,7 @@ export const createMediaHandler = (config: S3Config, options?: S3Options) => {
     }
     switch (req.method) {
       case 'GET':
-        if (req.url.startsWith('/api/s3/media/upload_url')) {
+        if (req.query.key) {
           const expiresIn: number =
             (req.query.expiresIn && Number(req.query.expiresIn)) || 3600
           const s3_key = req.query.key
@@ -87,9 +87,8 @@ export const createMediaHandler = (config: S3Config, options?: S3Options) => {
           )
 
           return res.json({ signedUrl, src: cdnUrl + s3_key })
-        } else {
-          return listMedia(req, res, client, bucket, mediaRoot, cdnUrl)
         }
+        return listMedia(req, res, client, bucket, mediaRoot, cdnUrl)
       case 'DELETE':
         return deleteAsset(req, res, client, bucket)
       default:
--- a/packages/next-tinacms-s3/src/s3-media-store.ts
+++ b/packages/next-tinacms-s3/src/s3-media-store.ts
@@ -19,6 +19,16 @@ export class S3MediaStore implements MediaStore {
     return fetch(input, init)
   }
   accept = DEFAULT_MEDIA_UPLOAD_TYPES
+  basePath = ''
+
+  protected fetchWithBasePath(path: string, init?: RequestInit) {
+    const fullPath = this.getFullPath(path)
+    return this.fetchFunction(fullPath, init)
+  }
+
+  protected getFullPath(path: string): string {
+    return `${this.basePath}${path}`
+  }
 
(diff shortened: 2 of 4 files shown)


packages/next-tinacms-s3/src/s3-tina-cloud-media-store.ts
```diff
@@ -10,6 +10,10 @@ export class TinaCloudS3MediaStore extends S3MediaStore {
   constructor(client: Client) {
     super()
     this.client = client
+
+    const basePath = this.client.schema.config.config.build?.basePath
+    this.basePath = basePath || ''
```
JackDevAU, line 15:
```suggestion
    this.basePath = `/${basePath}` || ''
```

packages/next-tinacms-s3/src/s3-tina-cloud-media-store.ts
```diff
@@ -10,6 +10,10 @@ export class TinaCloudS3MediaStore extends S3MediaStore {
   constructor(client: Client) {
     super()
     this.client = client
+
+    const basePath = this.client.schema.config.config.build?.basePath
+    this.basePath = `/${basePath}` || ''
```
JackDevAU, line 15:
```suggestion
    this.basePath = basePath || ''
```

packages/next-tinacms-s3/src/s3-media-store.ts
```diff
@@ -19,6 +19,17 @@ export class S3MediaStore implements MediaStore {
...
+
+  protected fetchWithBasePath(path: string, init?: RequestInit) {
+    const fullPath = this.getFullPath(path)
+    const normalizedPath = fullPath.startsWith('/') ? fullPath : `/${fullPath}`
+    return this.fetchFunction(normalizedPath, init)
+  }
+
+  protected getFullPath(path: string): string {
+    return `${this.basePath}${path}`
+  }
```
brookjeynes-ssw, line 32: Does this need to be an independent function?

packages/next-tinacms-s3/src/handlers.ts
```diff
@@ -65,7 +65,7 @@ export const createMediaHandler = (config: S3Config, options?: S3Options) => {
     }
     switch (req.method) {
       case 'GET':
-        if (req.url.startsWith('/api/s3/media/upload_url')) {
+        if (req.query.key) {
```
brookjeynes-ssw, line 68: Could you use includes instead of `startWith` to avoid the fact that now you're appending things to the base.

packages/next-tinacms-s3/src/s3-media-store.ts
JackDevAU: Not really - but I think its fine?

packages/next-tinacms-s3/src/s3-media-store.ts
```diff
@@ -19,6 +19,17 @@ export class S3MediaStore implements MediaStore {
     return fetch(input, init)
   }
   accept = DEFAULT_MEDIA_UPLOAD_TYPES
+  basePath = ''
+
+  protected fetchWithBasePath(path: string, init?: RequestInit) {
```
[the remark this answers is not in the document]
JackDevAU: Yea, unfortunately :(

packages/next-tinacms-s3/src/handlers.ts
JackDevAU: Yeah, we could use `includes` instead of `startsWith`, but I think the current setup works fine. Since we’re already requiring a valid key and an authorized user, and this package is the one handling the request, the check feels solid as is. Plus, it’s a nice bonus that we don’t have to worry about updating the URL manually if it changes on the backend.

ncn-ssw (approved):
🦙 Seen to work correctly

PR #5538, opened 2025-02-28
fix: #5534 to use correct collection when updating docs with references

### LINKED ISSUE #5534: 🐛 Bug - Multiple collections listed in reference list after deletion
@wicksipedia @JackDevAU @kldavis4 @Ben0189 @brookjeynes-ssw 

After deleting an item, any request for a list of references to the item's collection will return more than just the items in that collection.

To replicate:

1. Clone https://github.com/ncn-ssw/ref-integrity-bug
2. Link source code repository to latest commit on main for tinacms and [at]tinacms/graphql packages
3. Run `pnpm dev`
4. Navigate a web browser to http://localhost:3000/admin/index.html#/~/posts/learning-about-markdown
5. Click the 'author' drop down to note the listed items - see screenshot below
6. Navigate to http://localhost:3000/admin/index.html#/collections/author/~
7. Delete author Pedro
8. Navigate to http://localhost:3000/admin/index.html#/~/posts/learning-about-markdown
9. Click the 'author' drop down - note that posts are now also being included in the author list, as per the screenshot below.

Selecting any post item will cause the data store to lose integrity and cause reloads of the post item to fail.

Based on testing with different git commits, this issue appears to have been introduced with https://github.com/tinacms/tinacms/pull/5497

The expected behavior is that authors will only ever be included in the list.

## Tasks

- [ ] Fix

Image  
**Figure: Author list prior to deleting an author**

Image  
**Figure: Deleting Pedro as per step 7 above**

Image  
**Figure: Note that non-author items are now being included in the author list**

kldavis4:
Fix #5534  Root cause was the wrong collection was being used on the update call when updating documents containing references. 

- Also added logic to only perform updates when reference changes were made to avoid unnecessary updates.
- Fixed bug where null values were being inserted due to [null] values returned by JSONPath

### DIFF
--- a/packages/@tinacms/graphql/src/resolver/index.ts
+++ b/packages/@tinacms/graphql/src/resolver/index.ts
@@ -295,12 +295,14 @@ export const updateObjectWithJsonPath = (
   oldValue: any,
   newValue: any
 ) => {
+  let updated = false
   // Handle the case where path is a simple top-level property
   if (!path.includes('.') && !path.includes('[')) {
     if (path in obj && obj[path] === oldValue) {
       obj[path] = newValue
+      updated = true
     }
-    return obj
+    return { object: obj, updated }
   }
 
   // Extract the parent path (everything except the last segment)
@@ -317,12 +319,13 @@ export const updateObjectWithJsonPath = (
         if (parent[keyToUpdate] === oldValue) {
           // Update the property with the new value
           parent[keyToUpdate] = newValue
+          updated = true
         }
       }
     })
   }
 
-  return obj
+  return { object: obj, updated }
 }
 
 /**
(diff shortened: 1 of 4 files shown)


packages/@tinacms/graphql/src/resolver/index.ts
```diff
@@ -835,13 +838,34 @@ export class Resolver {
...
+                  path,
+                  realPath,
+                  null
+                )
+                refDoc = object
+                hasUpdate = updated || hasUpdate
               }
 
-              // save the updated doc
-              await this.database.put(pathToDocWithRef, refDoc, collection)
```
kldavis4, line 844: this was causing the issue (adding the doc with references to the referenced collection)

JackDevAU (approved):
🚀 Nice find! 
Tested and this has fixed the issue ✅

PR #5542, opened 2025-02-28
fix format using biome

Ben0189:


Ben0189:
close pr - this is just code format and too old to merge (too much conflict)

Issue #5545, opened 2025-03-02
Media folders containing non-ANSI characters are displayed as empty


Nuc1eoN:
### Describe the issue that you're seeing. Any videos (eg. YouTube or Loom) or screenshots help!

When a folders name in the media manager contains non-ansi characters (d.g. German umlauts like ä,ü,ö), its content will be shown as empty.

There does not seem to be any error in the console which would indicate that something is wrong.

### Reproduction

https://github.com/Nuc1eoN/tina-bug

### Steps to reproduce

1. Go to Media Manager
2. Create a folder with a non-ANSI character, e.g. "földer" :)
3. Enter the folder and upload an image
5. Go back, then re-enter the folder. The contents will be shown as empty, even though the image is "physically" there.

### System Info

```shell
System:
    OS: Linux 6.4 Arch Linux
    CPU: (12) x64 AMD Ryzen 5 5500U with Radeon Graphics
    Memory: 2.64 GB / 7.15 GB
    Container: Yes
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 20.3.1 - /usr/bin/node
    Yarn: 1.22.19 - ~/Desktop/tinaBug/my-tina-app/node_modules/.bin/yarn
    npm: 8.19.2 - /usr/bin/npm
    pnpm: 8.6.6 - /usr/bin/pnpm
  Browsers:
    Brave Browser: 114.1.52.130
    Chromium: 114.0.5735.198
  npmPackages:
    @tinacms/cli: 1.5.13 => 1.5.13 
    tinacms: 1.5.6 => 1.5.6
```

### Validations

- [x] Follow our [Code of Conduct](https://github.com/tinacms/tinacms/blob/main/CODE_OF_CONDUCT.md)
- [x] Read the [docs](https://tina.io/docs).
- [x] Check that there isn't [already an issue](https://github.com/tinacms/tinacms/issues) that reports the same bug.

ncn-ssw:
@kldavis4 - Has this been completed by the merging of https://github.com/tinacms/tinacms/pull/5555 ?

kldavis4:
@ncn-ssw yes, I think this can be closed

PR #5555, opened 2025-03-05
fix: local media server: uriDecode file paths in dev media server to …

kldavis4:
…fix issue with non-ANSI characters breaking file operations

Related issue #5545


kldavis4:
I tried to reproduce this issue with tina cloud and was not able to, so I think this is just a problem with the dev media server

ncn-ssw (approved):
🦙 Tested correct

PR #5572, opened 2025-03-15
fix: branch switcher bug when branch has been deleted

### LINKED ISSUE #5557: 🐛 Login - Branch Not Found Error Does Not Automatically Switch to Main Branch
Requested by: @wicksipedia via [YakShaver.ai](https://YakShaver.ai) 🦬
cc: @wicksipedia, @ncn-ssw, @adamcogan, @JackDevAU, @kldavis4 

Hi Team!

[Watch the video](https://www.youtube.com/watch?v=Ys3-UdP0eR4)

Url: [https://tina.io](https://tina.io)

### Pain
When logging into the admin interface of TinaCMS, if a user's last used branch is not found, the system no longer automatically switches to the main branch. This behavior needs to be reverted to ensure a seamless user experience.

### Acceptance Criteria
1. When a user's branch is not found, the system should automatically switch to the main branch.
2. The user should receive a notification indicating that they have been switched to the main branch.
3. The user should not need to manually clear local session variables to switch branches.

### Reproduce Steps
1. Log into the admin interface of TinaCMS.
2. Attempt to access a branch that does not exist.
3. Observe that the system does not automatically switch to the main branch and no notification is provided.

### Screenshot
Screenshot
**Figure: Switch to main branch broken**

kldavis4:
Fixes #5557 

When using editorial workflow and accessing the admin and the last accessed branch no longer exists, branch switcher should reset instead of trying to load the branch that doesn't exist

### DIFF
--- a/.changeset/healthy-elephants-impress.md
+++ b/.changeset/healthy-elephants-impress.md
@@ -1,0 +1,5 @@
+---
+"tinacms": patch
+---
+
+Fix issue with editorial workflow when last branch does not exist resulting in an error
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -62,6 +62,22 @@ jobs:
       - name: Build Types
         run: pnpm types
 
+      - name: Test
+        run: pnpm test
+
+      - name: Lint
+        if: matrix.os == 'ubuntu-latest'
+        run: pnpm lint
+
+      - name: Format
+        if: matrix.os == 'ubuntu-latest'
+        run: |
+          pnpm format
+          git diff --exit-code
+
+      - name: Check Tina lock changes
+        run: pnpm diff-tina-lock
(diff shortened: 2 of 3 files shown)


kldavis4:
@BrookJeynes it looks like the formatter is making different changes than what I get when I run `pnpm run format`

ncn-ssw (dismissed):
🦙 Tested correct

ncn-ssw (approved):
🦙 Tested correct

As per my conversation with @JackDevAU , the workflow has been modified to run the testing, linting, formatting and tina-lock tests before attempting a tagged release. This is to prevent the package publishing step causing modifications that may interfere with these checks. We accept that this will require the code to pass these steps in order to be published, which may impact the debugging loop when tagged releases are required.

PR #5577, opened 2025-03-18
fix: consider the remote schema to be the old and local to be new.

brookjeynes-ssw:
This pr fixes the comparison during the schema diff check. Previously, the remote was considered the new schema and local the old. This has been reversed in this pr.


ncn-ssw (approved):
🦙 Tested correct.

PR #5588, opened 2025-03-26
✨ feat: Allow new pages content to be queried on the server 

JackDevAU:
Previously, when using the auto-generated Tina client to query content, it would only query the branch specified at build time. As a result, server-side queries would fail if the user switched to a different branch (using the branch switcher), since the server had no information about the user's current branch.

This issue has been resolved by introducing a cookie-based approach. Now, whenever a new branch is set, a cookie is stored. When performing server-side queries with the Tina client, you can pass the branch via the `x-branch` header. TinaCMS will detect this header and correctly query the corresponding content index.

- Playground `/admin#/graphql` now uses currently selected Branch


JackDevAU:
Tested this on the SSW.Website and it is working as expected!

packages/tinacms/src/internalClient/index.ts
```diff
@@ -151,6 +151,9 @@ export class Client {
 
   setBranch(branchName: string) {
     const encodedBranch = encodeURIComponent(branchName);
+    //? When we change our branch, we add the 'x-branch' cookie. This is used when you change branches and want to use content from the new branch.
+    //? This is then used in the TinaClient to fetch the correct content from the correct branch. Instead of defaulting to the 'main' branch which is generated at build time.
```
JackDevAU, line 155:
```suggestion
    // When we change our branch, we add the 'x-branch' cookie. This is used when you change branches and want to use content from the new branch.
    // This is then used in the TinaClient to fetch the correct content from the correct branch. Instead of defaulting to the 'main' branch which is generated at build time.
```

packages/tinacms/src/unifiedClient/index.ts
```diff
@@ -102,6 +117,18 @@ export class TinaClient<GenQueries> {
       ...providedFetchOptions,
     };
 
+    //? Look for the header and change to use this branch instead of the build time generated branch.
+    //? This comes from the clients fetch options:
```
JackDevAU, line 121:
```suggestion
    // Look for the header and change to use this branch instead of the build time generated branch.
    // This comes from the clients fetch options:
```

JackDevAU (commented):
As per my conversation with Nick, These are distracting and need to be removed.

ncn-ssw (approved):
🦙 LGTM

PR #5624, opened 2025-04-14
 Add TinaDocs to NPX CLI templates

joshbermanssw:
‼️ PLEASE DONT MERGE UNTIL TINADOCS IS PRODUCTION READY

update: we're going live

**Changes**
- [ ] Added TinaDocs repo to NPX CLI starter templates


.changeset/plenty-mugs-remember.md
```diff
@@ -0,0 +1,5 @@
+---
+"create-tina-app": patch
+---
+
+🚧 WIP | Add TinaDocs to NPX CLI templates
```
isaaclombardssw: maybe update this message, TinaCMS users will see this

I believe there's a pnpm changeset command but check with the others

packages/create-tina-app/src/themes.ts
```diff
@@ -0,0 +1,34 @@
...
+    title: 'Pine',
+    value: 'pine',
+    description: 'A Pine theme for your project',
+  },
+  {
+    title: 'Indigo',
+    value: 'indigo',
+    description: 'An Indigo theme for your project',
+  },
+];
```
isaaclombardssw: Is this correct? now that the default theme is monochrome

.changeset/plenty-mugs-remember.md
joshbermanssw: ✅

packages/create-tina-app/src/themes.ts
joshbermanssw: Resolved

packages/create-tina-app/src/themes.ts
```diff
@@ -0,0 +1,34 @@
+import { BaseExample } from './templates';
+
+export const THEMES: BaseExample[] = [
```
JackDevAU, line 3: I think this should be a new type instead of using the `BaseExample` type

packages/create-tina-app/src/templates.ts
```diff
@@ -4,11 +4,11 @@ import path from 'path';
...
 import { Ora } from 'ora';
 
-type BaseExample = {
+export type BaseExample = {
   title: string;
   description?: string;
   value: string;
-  devUrl: string;
+  devUrl?: string;
 };
```
JackDevAU, line 12: future: undo and instead of exporting this. create a new theme type

PR #5640, opened 2025-04-26
fix: only delete existing items when partial indexing

kldavis4:
# problem

https://github.com/tinacms/tinacms/pull/5481 introduced an performance issue when performing full reindex due to unnecessary querying of every item.

# solution

Querying only needs to happen for a partial reindex so add a parameter to limit deletion to only partial reindexing

### DIFF
--- a/packages/@tinacms/graphql/src/database/index.ts
+++ b/packages/@tinacms/graphql/src/database/index.ts
@@ -1358,15 +1358,14 @@ export class Database {
         documentPaths,
         async (collection, documentPaths) => {
           if (collection && !collection.isDetached) {
-            await _indexContent(
-              this,
-              this.contentLevel,
+            await _indexContent({
+              database: this,
+              level: this.contentLevel,
               documentPaths,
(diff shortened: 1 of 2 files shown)


packages/@tinacms/graphql/src/database/index.ts
```diff
@@ -1611,7 +1613,8 @@ const _indexContent = async (
   documentPaths: string[],
   enqueueOps: (ops: BatchOp[]) => Promise<void>,
   collection?: Collection<true>,
-  passwordFields?: string[][]
+  passwordFields?: string[][],
+  isPartialReindex?: boolean
```
ncn-ssw, line 1617: **question**: Given that there are now 7 parameters for this function, some of which are not always needed, could `_indexContent` please be refactored to use named parameters?

kldavis4: updated

ncn-ssw (approved):
🦙 Tested correct

PR #5647, opened 2025-04-30
🔥 Deletion - Remove residual file from a previously existing integration test

ncn-ssw:
Based on https://github.com/tinacms/tinacms/commit/3603428b4c4d54aa65cd2dbfde7ed76edffcc0e3 , the integration tests for `@tinacms/graphql` previously resided in the `primitives/specs` directory. The `primitives` folder was removed at some point in the past, leaving only this residual file.

This change removes the file.


ncn-ssw:

kldavis4 (approved):
I'm not familiar with this file, but if tests are not failing with it remove it seems fine

PR #5648, opened 2025-04-30
🔥 Deletion - Remove 'pnpm serve' from @tinacms/graphql package

ncn-ssw:
While documenting the @tinacms/graphql package, I found that the `pnpm serve` command no longer works.

This change:

- Removes the `pnpm serve` command
- Removes the dev dependency on `nodemon` within the @tinacms/graphql package


ncn-ssw:

kldavis4 (approved):

PR #5653, opened 2025-05-02
🐛 Bug - Spurious error messages by FilesystemBridge on some relative paths

ncn-ssw:
When creating the test program included in the README.md of https://github.com/tinacms/tinacms/pull/5652 , strange error messages were seen when trying to use relative paths that included '.'.

These error messages were being raised by `FilesystemBridge` methods. For instance:

- Using '.' as the `rootPath` resulted in errors where an example file 'in.md' could not be found as it was trying to load 'inmd'
- Using './content' as the `rootPath` resulted in errors where FilesystemBridge was being asked to load `./content/content/in.md`

This change resolves the issue by doing the following:

- Paths that are received by the `FilesystemBridge` constructor are resolved.
- A replacement of the `outputPath` prefix is replaced with a substring operation.

This change also:

- Improves the consistency of typing and asynchronous operations within `FilesystemBridge`
- Adds basic tests for `FilesystemBridge.glob` and `FilesystemBridge.get`. These tests have been run successfully under Microsoft Windows.

Tests have not been added for `put` or `delete` operations as:

- Delays in the underlying filesystem updating are likely to result in flakey tests
- I assessed that the value of such tests was low


packages/@tinacms/graphql/tests/filesystem-bridge/index.test.ts
```diff
@@ -0,0 +1,27 @@
...
+          expect.arrayContaining(['file1.md', 'file2.md'])
+        );
+      });
+    }
+  );
+
+  test('can get file content based on filename', async () => {
+    const bridge = new FilesystemBridge(ABSOLUTE_CONTENT_PATH);
+    const itemContent = await bridge.get('file1.md');
+    expect(file1).equals(itemContent);
```
kldavis4, line 25:
```suggestion
    expect(file1).toEqual(itemContent);
```

kldavis4 (dismissed):
one minor recommendation

brookjeynes-ssw (approved):

PR #5654, opened 2025-05-02
🚚 Move - @tinacms/graphql/schema related tests

ncn-ssw:
This change aims to improve the positioning of tests related to the `packages/@tinacms/graphql/src/schema` directory.

- The tests that were in `@tinacms/graphql/src/spec/build-schema` are moved to `@tinacms/graphql/src/schema` so that they are co-located with the methods that they are testing. I have opted for co-locating the tests rather than moving them to `@tinacms/graphql/test` as these are unit tests and they don't have associated auxiliary files.
- The test that was in `@tinacms/graphql/src/schema/index.test.ts` has been moved to the `@tinacms/schema-tools` package, as the method it was testing originates from that package.

This change also removes unnecessary empty comments.


ncn-ssw:

kldavis4 (approved):

PR #5691, opened 2025-05-22
🐛 Bug - Editorial workflow fails to save from preview when path ends in slash

ncn-ssw:
This change rectifies a bug where saving will fail:

- when editorial workflow protections are enabled,
- the collection path ends in '/', and
- the `ui.router` is set on the collection.

The root cause was that, during indexing, path separators were not being processed correctly.

This change improves the path processing to ensure that the collection path is correctly removed from the `fullPath` of the file being modified. It does this by preprocessing the paths to ensure that there are no consecutive slashes and ensuring that both provided paths are non-absolute.

### DIFF
--- a/packages/tinacms/src/admin/pages/IndexingPage.tsx
+++ b/packages/tinacms/src/admin/pages/IndexingPage.tsx
@@ -5,7 +5,6 @@ import { useSearchParams } from 'react-router-dom';
 import { Client } from '../../internalClient';
 import { TinaAdminApi } from '../api';
 
-// TODO: Scott B styles
 type IndexingState =
   | 'starting'
   | 'indexing'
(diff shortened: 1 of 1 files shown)


packages/@tinacms/schema-tools/src/schema/TinaSchema.ts
```diff
@@ -116,37 +116,41 @@ export class TinaSchema {
...
         }
       }
-      // add / to the end of the path if it is not "''"
-      const path = collection.path ? collection.path.replace(/\/?$/, '/') : '';
-      return normalizedPath.startsWith(path);
+      const collectionPath = canonicalPath(collection.path);
+      return (
+        collectionPath == '' ||
+        canonicalFilepath.startsWith(`${collectionPath}/`)
+      );
```
JackDevAU, line 153: [nitpick] Use `===` over `==`

```suggestion
if (collectionPath === '') {
  return true;
}
return canonicalFilepath.startsWith(`${collectionPath}/`);
```

PR #5699, opened 2025-05-26
test coverage for data layer

kldavis4:

### DIFF
--- a/packages/@tinacms/graphql/src/database/index.ts
+++ b/packages/@tinacms/graphql/src/database/index.ts
@@ -1,8 +1,8 @@
-import path from "node:path";
-import fs from "fs-extra";
-import type { DocumentNode } from "graphql";
-import { GraphQLError } from "graphql";
-import micromatch from "micromatch";
+import path from 'node:path';
+import fs from 'fs-extra';
+import type { DocumentNode } from 'graphql';
+import { GraphQLError } from 'graphql';
+import micromatch from 'micromatch';
 
 import type {
   Collection,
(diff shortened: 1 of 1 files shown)


packages/@tinacms/graphql/src/database/index.ts
```diff
@@ -649,9 +650,22 @@ export class Database {
             : [];
         }
 
+        // TODO do we need to handle deletes
```
kldavis4, line 653: bug fix - without this, folders entries aren't created as expected

packages/@tinacms/graphql/src/database/index.ts
```diff
@@ -931,7 +945,7 @@ export class Database {
                   }
                 }
               }
-              for (const conflictedField in conflictedFields) {
+              for (const conflictedField of conflictedFields) {
```
kldavis4, line 948: this was not actually deleting the conflicted field

packages/@tinacms/graphql/src/database/index.ts
```diff
@@ -649,9 +651,21 @@ export class Database {
...
+        let folderOps: BatchOp[];
+        if (collection) {
+          folderOps = makeFolderOpsForCollection(
+            folderTreeBuilder.tree,
+            collection,
+            collectionIndexDefinitions,
+            'put',
+            level
+          );
+        }
```
kldavis4, line 663: without this folders were not getting created when specifying the a folder in the name of the file

packages/@tinacms/graphql/src/database/datalayer.ts
```diff
@@ -535,6 +535,9 @@ export const makeFilterSuffixes = (
   filterChain: (BinaryFilter | TernaryFilter)[],
   index: IndexDefinition
 ): { left?: string; right?: string } | undefined => {
+  if (index.fields.some((field) => field.list)) {
```
kldavis4, line 538: I don't think list values are currently filtering correctly so detect and return

packages/@tinacms/graphql/src/database/index.ts
```diff
@@ -290,10 +290,12 @@ export class Database {
...
         ? this.rootLevel
-            .sublevel('_content')
+            .sublevel('_content', SUBLEVEL_OPTIONS)
             .sublevel(this.contentNamespace, SUBLEVEL_OPTIONS)
             .sublevel(version, SUBLEVEL_OPTIONS)
-        : this.rootLevel.sublevel(version, SUBLEVEL_OPTIONS);
+        : this.rootLevel
+            .sublevel('_content', SUBLEVEL_OPTIONS)
+            .sublevel(version, SUBLEVEL_OPTIONS);
     }
```
kldavis4, line 299: I think this change is technically breaking but should not impact tina cloud since versioning is used in that case

kulesy:
Unsure what this is for or what PBI this is solving

Closing for now.

PR #5715, opened 2025-05-27
refactor: deprecate git media store

brookjeynes-ssw:
The git-client folder contains very old cold that is not used and is
no longer needed. These items will be removed in the next major TinaCMS
version.

Resolves https://github.com/tinacms/tinacloud/issues/2186


packages/tinacms/src/toolkit/git-client/git-client.ts
```diff
@@ -1,5 +1,6 @@
 /**
- * @deprecated
+ * @deprecated as the API is clunky and hard to use. Mutations should now be
+ * done via Graphql. This will be removed within the next month.
```
ncn-ssw: *issue*: When is "next month"?

ncn-ssw (approved):
🦙 Approved

Issue #5686, opened 2025-05-18
🐛 Bug - Error on Save Operation in ⭐ TinaCMS


uly1:
Requested by: @uly1 via [YakShaver.ai](https://YakShaver.ai) 🦬
cc: @wicksipedia, @uly1, @ncn-ssw, @adamcogan, @JackDevAU

Hi Team!

&nbsp;&nbsp;[🟥&nbsp;&nbsp;Watch the video (1 min 15 sec)](https://sswcom.sharepoint.com/:v:/s/SSWYakShaver/Ee9pQ3n8J4tEr4VQ8mMeFesB7staRHODMEY4QZhbulSwDA)

&nbsp;&nbsp;Url from screen share: [https://yakshaver.ai/admin/branch/new?branch=test%2Fst25652](https://yakshaver.ai/admin/branch/new?branch=test%2Fst25652)

### Pain
Users encounter an error when attempting to save changes in ⭐ TinaCMS. The error appears in the bottom right corner of the interface after trying to create a branch and save changes.

### Acceptance Criteria
1. The save operation completes successfully without errors.
2. The error message no longer appears in the bottom right corner after saving.
3. Users can create a branch and save changes without encountering issues.

### Reproduce Steps
1. Open ⭐ TinaCMS and navigate to the desired section of the website.
2. Make a change, such as capitalizing a letter.
3. Click 'Save' and provide a name for the branch.
4. Attempt to create the branch and save the project.
5. Observe the error message that appears in the bottom right corner.

### Screenshot
![Screenshot](https://sayakshaverproduction.blob.core.windows.net/images/Images/4b3766f0-783e-4698-9acb-9f09ede2cb67_frame_54.png?sv=2024-08-04&se=2075-05-18T23%3A51%3A41Z&sr=b&sp=rd&sig=QJTwLmKfyRBDuWSGk5DTGK%2F81smp60j%2B2VD5tUEw3bA%3D)
**Figure: Error fetching data in bottom right corner**

ncn-ssw:
@wicksipedia @uly1 @Calinator444 

Based on testing with SSW.Products, the trigger for the bug is including an ending slash in collection path names. When I remove the slashes, I am able to save without issue.

Based on TinaCloud logs, the record being retrieved has its collection prefix added twice before retrieval. This behavior is similar to that seen by the issue resolved by https://github.com/tinacms/tinacms/pull/5653 . However, I don't believe that the `FilesystemBridge` is involved in this case.

[...]

PR #5740, opened 2025-06-03
🚧 WIP - ♻️ Refactor - Use Form.id instead of Form.relativePath

ncn-ssw:
This change:

- Marks `Form.relativePath` and `FormOptions.relativePath` as deprecated.
- Refactors TinaCMS to use `Form.id` instead of `Form.relativePath`.

This purpose of this change is to clarify the usage of 'relativePath', as described in the comment at https://github.com/tinacms/tinacms/issues/5686#issuecomment-2899840518

### DIFF
--- a/.changeset/khaki-toes-tell.md
+++ b/.changeset/khaki-toes-tell.md
@@ -1,0 +1,5 @@
+---
+"tinacms": patch
+---
+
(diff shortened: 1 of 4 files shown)


ncn-ssw:
The refactoring of the `CollectionCreatePage` page is causing the editing page to fail completely.

This PR will be closed for now awaiting a better solution.

PR #5744, opened 2025-06-03
Plate v48 beta

Ben0189:
upgrade plate from v36 and code refactor


packages/tinacms/src/toolkit/fields/plugins/mdx-field-plugin/index.tsx
```diff
@@ -24,6 +24,7 @@ export type RichTextType = React.PropsWithChildren<
   >
 >;
 
+//(Plate Upgrade) Ask Jeff: Is this in used?
```
jeffsee55: I think so, yeah. It's effectively overwritten from `@tinacms/app`

jeffsee55:
This all works great for me. 

I noticed this in the build output, but not sure if that was already there:

Screenshot 2025-06-28 at 9 54 28 AM

---

Only other comment that I'm pretty sure we're all aware of, this rich text editor essentially no longer allows blocks inside blocks of any kind as far as I can tell

```
> # Hello
```

Becomes

```
> hello
```

We already didn't have great support for this, and from my testing this looks it handles it gracefully. 🚢

jeffsee55 (dismissed):

Ben0189:
@jeffsee55 

> I noticed this in the build output, but not sure if that was already there:

We will turn the `sourcemap` off and double check this error still exist

packages/tinacms/src/toolkit/fields/plugins/mdx-field-plugin/index.tsx
Ben0189: ok thanks, comment resolved

PR #5755, opened 2025-06-09
♻️ Refactor - Use Form.path instead of Form.relativePath

ncn-ssw:
This change:

- Marks `Form.relativePath` and `FormOptions.relativePath` as deprecated.
- Refactors TinaCMS to use `Form.path` instead of `Form.relativePath`.

This purpose of this change is to clarify the usage of 'relativePath', as described in the comment at https://github.com/tinacms/tinacms/issues/5686#issuecomment-2899840518

Note that the comment states to use `id` - however, during testing it was found that the `id` needed to remain constant during editing. A separate field, `path`, has been added in its place.

### DIFF
--- a/.changeset/khaki-toes-tell.md
+++ b/.changeset/khaki-toes-tell.md
@@ -2,4 +2,4 @@
 "tinacms": patch
 ---
 
-♻️ Refactor - Use Form.id instead of Form.relativePath, mark FormOptions.relativePath and Form.relativePath as deprecated
+♻️ Refactor - Use Form.path instead of Form.relativePath, mark FormOptions.relativePath and Form.relativePath as deprecated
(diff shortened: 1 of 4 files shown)


JackDevAU (changes requested):
Can't create a new document. Guessing there is a reference to `relativePath` somewhere that is throwing?

Using the `next-2024` example

https://github.com/user-attachments/assets/8596b96f-2a69-4b9c-b79f-85f0a9186206

ncn-ssw:
@JackDevAU - Which example are you using in your video above?

JackDevAU:
> @JackDevAU - Which example are you using in your video above?

Updated comment

JackDevAU (approved):
Unsure why I was getting the error I was getting last time I tested this... but all seems to be working fine after Re-cloning tinacms ✅

PR #5764, opened 2025-06-10
Fix: adjust blocks plugin to append instead of insert

mashakos:

PR #5766, opened 2025-06-10
Fix: adjust blocks plugin to append instead of insert

brookjeynes-ssw:
This PR updates the behaviour of block plugin inserts to append rather than insert at the beginning. It also introduces a new `addItemBehavior` prop to make the behavior of list and group field plugins in `tinacms` configurable.

It allows users to specify whether new items should be added at the start (`prepend`) or end (`append`) of a list, with the default behavior set to `append`.

This work is lifted from #5764


wicksipedia (approved):
lgtm

PR #5783, opened 2025-06-17
✨ Add a `StaticTinaMarkdown` component 

JackDevAU:
Can be used for server rendering content.

usage example: 

```tsx
import { StaticTinaMarkdown } from 'tinacms/dist/rich-text/static';

...

{
...
const data = await client.queries.post({
  relativePath: `${params.filename}.json`,
});
return (
  <StaticTinaMarkdown content={data.data.post.body} />
  )
}
```

### DIFF
--- a/.changeset/tall-turkeys-impress.md
+++ b/.changeset/tall-turkeys-impress.md
@@ -1,0 +1,5 @@
+---
+"tinacms": patch
+---
+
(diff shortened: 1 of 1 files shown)


wicksipedia (approved):
lgtm, refactoring can be done in another pr

PR #5784, opened 2025-06-18
🔧 Fix next-2024 example

JackDevAU:
Due to some weird merge issues - the `next-2024` example has become a bit broken. This fixes it.


wicksipedia (approved):
lgtm

PR #5823, opened 2025-07-01
🐛 Fix reference field "Edit" button when in visual editing on nested forms

JackDevAU:
- We now go through rich text templates to find reference fields. 
- Show path to usage in document 

image
**Figure: `apple.md` is used multiple times in various places**

### DIFF
--- a/.changeset/nervous-crabs-shake.md
+++ b/.changeset/nervous-crabs-shake.md
@@ -1,0 +1,6 @@
+---
+"tinacms": patch
+"@tinacms/app": patch
+---
+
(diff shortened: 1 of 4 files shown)


wicksipedia:
As per my conversation with @JackDevAU the current state of the PR has introduced a regression into visual editing.

We are thinking that instead of editing the `graphql-reducer` - we should be attacking this a different way.... unsure about the actual approach right now

JackDevAU:
closing as this has gone stale and this screen has now been redesigned

PR #5837, opened 2025-07-07
🚧 WIP - 🧑‍💻 DX - Cease overriding error stacks

ncn-ssw:
In the following sub-classes of Error, the stack member is overridden:

- `TinaFetchError`
- `TinaQueryError`

As a result, the normal stack trace produced by the error unwinding the stack becomes inaccessible, making bug hunting within Tina difficult.

This PR is WIP until a proper testing plan can be devised.

image
**Figure: Error output as seen in the TinaCloud logs - not only is this a duplicate of the `originalError.stack` value, it is not clear how the error was processed within Tina**

### DIFF
--- a/packages/@tinacms/graphql/src/database/index.ts
+++ b/packages/@tinacms/graphql/src/database/index.ts
@@ -674,7 +674,6 @@ export class Database {
         originalError: error,
         file: filepath,
         collection: collectionName,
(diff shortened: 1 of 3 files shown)


brookjeynes-ssw (approved):
Looks good, thanks!

PR #5842, opened 2025-07-07
VersionInfo - fix hardcoded version info

wicksipedia:
This pull request makes a small change to the `VersionInfo.tsx` component, replacing a hardcoded `currentVersion` value with an imported version from the `package.json` file.

### DIFF
--- a/.changeset/sour-readers-nail.md
+++ b/.changeset/sour-readers-nail.md
@@ -1,0 +1,5 @@
+---
+"tinacms": patch
+---
+
+Sidebar - fixed version number being reported in UI


packages/tinacms/src/toolkit/react-sidebar/components/VersionInfo.tsx
```diff
@@ -2,11 +2,9 @@ import { useCMS } from '@toolkit/react-tinacms';
 import formatDistanceToNow from 'date-fns/formatDistanceToNow';
 import { Check, LoaderCircle, TriangleAlert } from 'lucide-react';
 import React from 'react';
-// import { version as currentVersion } from '../../../../package.json';
+import { version as currentVersion } from '../../../../package.json';
```
JackDevAU, line 5: will this find the users package.json correctly?

wicksipedia: It was how we were displaying the version number before, its a reference to the tinacms package.json

PR #5872, opened 2025-07-16
🧪 GraphQL Test - Multi-field sorting query

ncn-ssw:
This change adds an automated test for sorting query output on multiple fields.


kulesy (approved):
lgtm

PR #5873, opened 2025-07-16
🧪 GraphQL Test - Numeric filtering query

ncn-ssw:
This change adds an automated test for filtering upon a numeric value.


kulesy (approved):
lgtm

PR #5874, opened 2025-07-16
🧪 GraphQL Test - Pagination query

ncn-ssw:
This change adds an automated test for paginating through query results.


ncn-ssw:
@kulesy - As per our conversation, as we cannot guarantee that the cursor values will be constant across all systems, I will rewrite the tests so that they are not reliant on the cursor value.

brookjeynes-ssw (approved):

PR #5876, opened 2025-07-16
🧪 GraphQL Test - Single-field sorting query

ncn-ssw:
This change adds an automated test for sorting with all types.


kulesy (approved):
lgtm

PR #5877, opened 2025-07-16
🧪 GraphQL Test - String filtering query

ncn-ssw:
This change adds an automated test for filtering on string types.


kulesy (approved):
lgtm

PR #5879, opened 2025-07-16
🧪 GraphQL Test - Polymorphic query test

ncn-ssw:
This change adds an automated test for filtering on attributes that differ between templates.


kulesy (approved):
lgtm

PR #5886, opened 2025-07-18
⚡ Performance - Improved query engine performance with slow databases, adds abort signal

ncn-ssw:
This change:

- **Modifies the query engine to execute the required "hydration" callbacks in parallel.** The hydration callback is used to read individual documents, which is sometimes done over a network as seen in TinaCloud. Network requests are slow, and do not require much involvement from the processor, so this is expected to improve TinaCloud query performance.

- **Adds the ability for the query engine to abort a query.** In future, this will be used to quicken queries that have already been shown to fail. The plumbing for the abort signal is included in this request as this change amends the type signature of `Database.query(...)` to explicitly list the arguments to be accepted by the hydrator.

- **Adds a performance test to ensure that requests are being executed in parallel.** This test consistently fails when executed with main, as the sequential document accesses (when the schema reading is included) total to a duration longer than the test timeout.

### DIFF
--- a/packages/@tinacms/graphql/src/database/index.ts
+++ b/packages/@tinacms/graphql/src/database/index.ts
@@ -1012,7 +1012,8 @@ export class Database {
 
   public query = async <T>(
     queryOptions: QueryOptions,
-    hydrator: (path: string, value?: Record<string, unknown>) => T
+    hydrator: (path: string, value?: Record<string, unknown>, abortSignal?: AbortSignal) => T,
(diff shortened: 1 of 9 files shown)


ncn-ssw:
❌ Closed - results of experiments with PR documented at https://github.com/tinacms/tinacloud/issues/1931#issuecomment-3142438679

This PR is being closed as too much merge debt would accumulate if it were left open.

PR #5901, opened 2025-07-25
🐛 Fix: content path trailing slash breaks indexer

ArkadiuszIrlik:
### On the surface
Leaving a trailing slash inside “localContentPath” in Tina config (e.g. "../../portfolio-website-content/") will result in “Unable to find collection for file at (…)” error whenever editing documents.

unable-to-find-collection-error

This restriction (no trailing slash) isn't documented anywhere, and the error doesn't do a good job of indicating the real source of the problem. Personally, I think using trailing slashes inside a directory path isn't uncommon and I'm sure other users will run into this issue.

### Reproduce Steps
1. Inside tina/config.ts, add a trailing slash at the end of “localContentPath” property.
2. Make a change to any document inside Tina admin console in browser and hit save.

### What actually happens
The issue originates in @tinacms/cli's [config-manager.ts#L343-L350](https://github.com/tinacms/tinacms/blob/main/packages/%40tinacms/cli/src/next/config-manager.ts#L343-L350):
```javascript
  printContentRelativePath(filename: string) {
    if (filename) {
      return filename
        .replace(/\\/g, '/')
        .replace(`${this.contentRootPath}/`, '');
    }
    throw `No path provided to print`;
  }
```
The replace() call hardcodes a “/”, assuming contentRootPath will never end in a trailing slash. 

Moving up the call stack to DevCommand's [watchContentFiles#L351-L362](https://github.com/tinacms/tinacms/blob/main/packages/%40tinacms/cli/src/next/commands/dev-command/index.ts#L351-L362):
```typescript
        await databaseLock(async () => {
          const pathFromRoot =
            configManager.printContentRelativePath(addedFile);
          await database
            .indexContentByPaths([pathFromRoot])
            .catch(console.error);
          if (searchIndexer) {
            await searchIndexer
              .indexContentByPaths([pathFromRoot])
              .catch(console.error);
          }
        });
```

Since printContentRelativePath() didn’t replace anything, “pathFromRoot” is instead the full, absolute path to the document. This value then gets passed to “database” and “searchIndexer” indexContentByPaths(), before ultimately ending up as “canonicalFilepath” in TinaSchema's [getCollectionByFullPath()#L150-L153](https://github.com/tinacms/tinacms/blob/main/packages/%40tinacms/schema-tools/src/schema/TinaSchema.ts#L150-L153), where it keeps failing this check:
```typescript
      return (
        collectionPath === '' ||
        canonicalFilepath.startsWith(`${collectionPath}/`)
      );
```
Causing the function to report zero possible collections, which throws the error in question.

## Solution
I wrote a small function which strips any trailing slashes from fullLocalContentPath inside [config-manager.ts](https://github.com/tinacms/tinacms/commit/66a251cadb500b90a08dadb253f04e3520fe2e2d#diff-379ac807caa172512c33c982fb1373fd1d533e2cb744fa4ad6b9444a3665b39eL237-R239). This leads to consistent, predictable behavior, whether the user provides a path with a trailing slash or not.

### Alternative solution
A less invasive solution would be to change the regex inside [printContentRelative()](https://github.com/tinacms/tinacms/blob/main/packages/%40tinacms/cli/src/next/config-manager.ts#L343-L350) function, so that it only **optionally** matches the "/". I'm not entirely familiar with Tina's codebase, but at a glance, there doesn't seem to be any other code which breaks when a trailing slash is provided to configManager.contentRootPath, so perhaps this more "naive" approach is good enough. 

However, I still think that's more of a plaster fix and normalizing the way ConfigManager's "contentRootPath" is stored is a more robust solution. Please, let me know what you think and I will adjust my PR as necessary.


JackDevAU:
Hey @ArkadiuszIrlik 

This looks great! 

Can you run pnpm install and add the lockfile 🙏

ArkadiuszIrlik:
Hi there, thanks. I pushed the updated lockfile, sorry for the delay.

brookjeynes-ssw (approved):
Looks good, thanks!

PR #5909, opened 2025-07-29
feat: add opt-out option for update checks in CMS configuration

wicksipedia:
Adds a new prop allowing users to opt out of the CMS's update version checks


packages/@tinacms/schema-tools/src/types/index.ts
```diff
@@ -625,6 +625,13 @@ export interface Config<
      * [more info](https://vercel.com/docs/concepts/deployments/generated-urls#url-with-git-branch)
      */
     previewUrl: (context: { branch: string }) => { url: string };
```
ncn-ssw: **question**: A `ui` setting cannot be defined without also setting the previewUrl - is this intentional?

wicksipedia: done - made it optional

ncn-ssw (approved):
🦙 Tested correct

PR #5917, opened 2025-07-31
🧪 GraphQL Test - addPendingDocument mutation with template

ncn-ssw:
This change adds a basic automated test to ensure that `addPendingDocument` can be successfully called with a template selection.


kulesy (approved):
lgtm

PR #5937, opened 2025-08-05
added missing devUrl

joshbermanssw:
cc: @JackDevAU adding a missing devUrl

### DIFF
--- a/packages/create-tina-app/src/themes.ts
+++ b/packages/create-tina-app/src/themes.ts
@@ -1,12 +1,12 @@
 import { BaseExample } from './templates';
 
-type TINADOCS_THEME = {
+type TinaDocsTheme = {
   title: string;
   value: string;
   description: string;
 };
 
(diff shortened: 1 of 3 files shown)


packages/create-tina-app/src/themes.ts
```diff
@@ -1,6 +1,12 @@
 import { BaseExample } from './templates';
 
-export const THEMES: BaseExample[] = [
+type TINADOCS_THEME = {
```
JackDevAU:
```suggestion
type TinaDocsTheme = {
```

packages/create-tina-app/src/themes.ts
```diff
@@ -1,6 +1,12 @@
...
+  title: string;
+  value: string;
+  description: string;
+};
+
+export const THEMES: TINADOCS_THEME[] = [
```
JackDevAU:
```suggestion
export const THEMES: TinaDocsTheme[] = [
```

PR #5963, opened 2025-08-29
add install dependencies step

joshbermanssw:
As per https://github.com/tinacms/tina-docs/issues/257

Screenshot 2025-08-29 at 3 16 01 pm

**Figure: Added Extra prompts for someone with error msg**

### DIFF
--- a/packages/create-tina-app/src/index.ts
+++ b/packages/create-tina-app/src/index.ts
@@ -22,7 +22,6 @@ import * as ascii from './util/asciiArt';
 import { THEMES } from './themes';
 import { useState } from 'react';
 
-
 export async function run() {
   // Dynamic import for ora to handle ES module compatibility
   const ora = (await import('ora')).default;
@@ -188,7 +187,9 @@ export async function run() {
     spinner.succeed();
   } catch (err) {
     spinner.fail(`Failed to install packages: ${(err as Error).message}`);
-    setPackageManagerErrorMsg(`Failed to install packages: ${(err as Error).message}`);
+    setPackageManagerErrorMsg(
+      `Failed to install packages: ${(err as Error).message}`
+    );
     exit(1);
   }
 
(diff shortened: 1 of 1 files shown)


packages/create-tina-app/src/index.ts
```diff
@@ -212,6 +212,7 @@ export async function run() {
   spinner.info(`${TextStyles.bold('To get started:')}
 
   ${padCommand(`cd ${appName}`)}# move into your project directory
+  ${padCommand(`${pkgManager} install`)}# install dependencies
```
wicksipedia: I think this should only be shown if there was an issue in running it on their behalf

joshbermanssw: 👍

packages/create-tina-app/src/index.ts
```diff
@@ -20,10 +20,12 @@ import { PackageManager, PKG_MANAGERS } from './util/packageManagers';
 import validate from 'validate-npm-package-name';
 import * as ascii from './util/asciiArt';
 import { THEMES } from './themes';
+import { useState } from 'react';
```
wicksipedia: I don't think we need react for this cli, seems overkill for tracking 1 piece of state

packages/create-tina-app/src/index.ts
```diff
@@ -185,6 +187,9 @@ export async function run() {
     spinner.succeed();
   } catch (err) {
     spinner.fail(`Failed to install packages: ${(err as Error).message}`);
+    setPackageManagerErrorMsg(
+      `Failed to install packages: ${(err as Error).message}`
```
wicksipedia: spinner.fail should show the error message in the terminal in place

packages/create-tina-app/src/index.ts
```diff
@@ -212,6 +217,7 @@ export async function run() {
   spinner.info(`${TextStyles.bold('To get started:')}
 
   ${padCommand(`cd ${appName}`)}# move into your project directory
+  ${packageManagerErrorMsg ? `${padCommand(`${pkgManager} install`)}# install dependencies` : ''}
```
wicksipedia: given the `exit(1)` call earlier after the message is set, this ternary will never evaluate to true

joshbermanssw: As per our convo - resolved

wicksipedia (approved):
lgtm

PR #5969, opened 2025-09-05
🐛 Bug Fix: Make UTC the default for the DateField

kulesy:
## 📝 Summary

This PR enhances the DateField component to ensures consistency between field display and saved values through improved UTC handling.

- Simplified UTC handling logic for better consistency

### DIFF
--- a/.changeset/poor-poets-argue.md
+++ b/.changeset/poor-poets-argue.md
@@ -2,4 +2,4 @@
 "tinacms": patch
 ---
 
-🗓️ feat: Make UTC the default for the DateField
+✨ Feature: Make UTC the default for the DateField
--- a/.changeset/poor-poets-argue.md
+++ b/.changeset/poor-poets-argue.md
@@ -2,4 +2,4 @@
 "tinacms": patch
 ---
 
-🗓️ feat: Enhance DateField locale support and UTC handling consistency
+🗓️ feat: Make UTC the default for the DateField
--- a/.changeset/poor-poets-argue.md
+++ b/.changeset/poor-poets-argue.md
@@ -1,0 +1,5 @@
+---
+"tinacms": patch
+---
+
+🗓️ feat: Enhance DateField locale support and UTC handling consistency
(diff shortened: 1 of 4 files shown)


packages/tinacms/src/toolkit/fields/plugins/date-format.ts
```diff
@@ -4,6 +4,39 @@ import type { DatetimepickerProps } from 'react-datetime';
 export const DEFAULT_DATE_DISPLAY_FORMAT = 'MMM DD, YYYY';
 export const DEFAULT_TIME_DISPLAY_FORMAT = 'h:mm A';
 
+/**
+ * Gets the user's locale-specific date format using abbreviated months
+ * Returns formats like "MMM DD, YYYY" for US or "DD MMM, YYYY" for UK, etc.
+ * Falls back to the hardcoded default format if locale detection fails
+ */
+export const getLocaleDateFormat = (): string => {
```
JackDevAU: Since this component uses `moment`. 

hmm instead of this can we use something like the following:

```tsx
import moment from 'moment';
...
export const getLocaleDateFormat = (): string => {
  try {

    const locale =
      Intl.DateTimeFormat().resolvedOptions().locale;

    moment.locale(locale);

    return moment.localeData().longDateFormat('ll');
 	} catch (error) {
    return DEFAULT_DATE_DISPLAY_FORMAT;
  }
```

kulesy: moment will always return the dates back as en-US so this will not work.

.changeset/poor-poets-argue.md
```diff
@@ -0,0 +1,5 @@
+---
+"tinacms": patch
+---
+
+🗓️ feat: Enhance DateField locale support and UTC handling consistency
```
JackDevAU:
```suggestion
🗓️ feat: Make utc the default for the DateField
```

JackDevAU (approved):

PR #5980, opened 2025-09-15
feat(cli): add Bun support to `tinacms init` package manager options

Felixdiamond:
## Overview
This PR adds support for the Bun package manager to the `tinacms init` command. 
The `create-tina-app` package already supports Bun, so this update brings parity between the two tools.

## Changes
- Added "Bun" option to the interactive package manager prompt (`prompts/index.ts`).
- Extended TypeScript union type in `prompts/types.ts` to include `'bun'`.
- Added Bun command mappings in `apply.ts`:
  - Regular dependencies: `bun add <deps>`
  - Dev dependencies: `bun add -D <devDeps>`
  - Next.js projects: `bun run dev`
  - Other frameworks: `bunx tinacms dev -c "<your dev command>"`

## Backward Compatibility
- No breaking changes.
- Existing npm, yarn, and pnpm flows remain unchanged.
- Bun support follows the same conventions used for other package managers.

## Testing
- Verified locally:
  - Bun commands generate correctly for dependencies and devDependencies.
  - Next.js projects run with `bun run dev`.
  - Other frameworks run with `bunx tinacms dev -c "<cmd>"`.
- Confirmed type safety with updated union types.

## Why
Fixes inconsistency between `create-tina-app` and `tinacms init`, giving users full Bun support across the TinaCMS toolchain.

### DIFF
--- a/packages/@tinacms/cli/src/cmds/init/apply.ts
+++ b/packages/@tinacms/cli/src/cmds/init/apply.ts
@@ -346,6 +346,7 @@ const addDependencies = async (
       : `pnpm add ${deps.join(' ')}`,
     npm: `npm install ${deps.join(' ')}`,
     yarn: `yarn add ${deps.join(' ')}`,
+    bun: `bun add ${deps.join(' ')}`,
   };
 
   if (packageManagers[packageManager] && deps.length > 0) {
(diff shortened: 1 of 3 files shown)


brookjeynes-ssw (approved):
Looks good, thanks for this!

PR #6013, opened 2025-09-29
Improve position of title overlay

joshbermanssw:
Overlay position was a child of the image, so it looked weird 

Screenshot 2025-09-29 at 12 46 25 pm

**Figure: Before**

Screenshot 2025-09-29 at 12 50 51 pm

**Figure: After**

### DIFF
--- a/packages/tinacms/src/toolkit/components/media/media-item.tsx
+++ b/packages/tinacms/src/toolkit/components/media/media-item.tsx
@@ -127,15 +127,6 @@ export function GridMediaItem({ item, active, onClick }: MediaItemProps) {
                 src={thumbnail}
                 alt={item.filename}
               />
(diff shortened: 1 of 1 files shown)


packages/tinacms/src/toolkit/components/media/media-item.tsx
```diff
@@ -118,15 +127,15 @@ export function GridMediaItem({ item, active, onClick }: MediaItemProps) {
                 src={thumbnail}
                 alt={item.filename}
               />
-              <span
+              {/* <span
```
kulesy: Can this just be deleted?

joshbermanssw: done

PR #6082, opened 2025-10-31
fix: update rich text editor to take up the full width

JackDevAU:
Before:

CleanShot 2025-10-31 at 14 50 03

After:

CleanShot 2025-10-31 at 14 49 23


Pat-Stuart:
LGTM - Tested

image

Figure: RichText editor

Issue #5926, opened 2025-08-01
♻️ Refactor - GraphQL Resolver class mega-method `resolveDocument`


ncn-ssw:
The [`Resolver.resolveDocument`](https://github.com/tinacms/tinacms/blob/31eca8481e834b1fb790de7d9c3311a1c8b4a400/packages/%40tinacms/graphql/src/resolver/index.ts#L740) method attempts to achieve multiple unrelated objectives.

For instance, it accepts the following parameters:

- `isMutation`
- `isCreation`
- `isDeletion`
- `isFolderCreation`

These are mutually incompatible operations, and all calls to this method are from the field resolver in https://github.com/tinacms/tinacms/blob/main/packages/%40tinacms/graphql/src/resolve.ts

To make it easier to reason about the field resolver, this method should be removed and replaced with individual methods handling individual operations such as:

- creating documents,
- deleting documents,
- creating folders.

The `Resolver` class does not appear to be exported from the @tinacms/graphql package, so changing its method signature should not affect backwards-compatibility.

## Tasks

- [ ] Replace `Resolver.resolveDocument` with methods dedicated to its individual tasks, such as creating documents, deleting documents and creating folders.
- [ ] Ensure all newly-created methods have tests.

kulesy:
Closing — implemented in PR #6084 (merged 2025-12-22). `Resolver.resolveDocument` was split into `resolveAddPendingDocument`, `resolveRetrievedDocument`, `resolveCreateFolder`, `resolveCreateDocument`, `resolveUpdateDocument`, and `resolveDeleteDocument` — matching the breakdown in the original task list.

Issue #5925, opened 2025-08-01
♻️ Refactor - GraphQL field resolver in @tinacms/graphql/src/resolve.ts


ncn-ssw:
The `resolve` function in https://github.com/tinacms/tinacms/blob/31eca8481e834b1fb790de7d9c3311a1c8b4a400/packages/%40tinacms/graphql/src/resolve.ts#L20 is difficult to reason about and modify.

Specifically, the field resolver has the following issues:

- Several parameters to the field resolver are prefixed with an underscore, yet those parameters are in use in defiance of convention.
- The logic for determining how to resolve a field is dependent on a combination of: GraphQL return type, lookup type attribute, and field name. This makes it likely that there are subtle bugs in the resolver, especially given that some field name checks (e.g. 'getOptimizedQuery', 'authenticate') do not check the path to that field.
- The 'getOptimizedQuery' resolver is deprecated and only echos the queryString, yet takes 35 lines including comments and a try-catch statement that is likely not in use.\
- The self-hosted account management functionality is directly included within the field resolver, rather than being broken logically out into separate functions.

The aim of this PBI is to make the field resolver easier to reason about.

## Tasks

- [ ] Rename the parameters starting with underscore where they are used within the function.
- [ ] Simplify the 'getOptimizedQuery' resolver code to just return `args.queryString` or '', removing the excessive comments and the try-catch logic.
- [ ] Split the 'authenticate', 'authorize' and 'updatePassword' logic out into their own functions.
- [ ] Determine if the field name guards (e.g. for 'authenticate') should be further guarded with the `GraphQLResolveInfo` instance `info`'s path value.
- [ ] Improve the consistency of the field determination logic, so that it is clearer why branches are being taken.

kulesy:
Closing — implemented in PR #6084 (merged 2025-12-22), which explicitly lists this issue under "Issues resolved". The PR splits auth functions, simplifies `getOptimizedQuery`, renames underscore params, and improves resolver clarity per the original task list.

PR #6084, opened 2025-11-02
Resolver - Refactor to improve clarity

n-crrn:
CC @brookjeynes-ssw 

The main purpose of this change is to improve the readability of the GraphQL resolver, thereby making it easier to resolve bugs and fix inefficiencies in TinaCMS' GraphQL query handling.

The PR attempts to do the following:

- Clarify the logic within the field resolver defined in `resolve.ts` so that the code-path executed on the `Resolver` object is obvious.
    - To support this object, the functionality of the `Resolver.resolveDocument` method is split out into new methods: `resolveAddPendingDocument`, `resolveRetrievedDocument`, `resolveCreateFolder`, `resolveCreateDocument`, `resolveUpdateDocument` and `resolveDeleteDocument`.
    - Previously existing public methods of the `Resolver` that are no longer required are marked as deprecated in preparation for removal at the next major version. They are not removed now as they are currently part of the public interface of the package, though they shouldn't be used outside of the `@tinacms/graphql` package.
- Improves type and shape checking throughout the resolution process, to make it clearer what types are expected at each point in the process.
- Simplifies the logic of the `getOptimizedQuery` query.
- Removes leading underscores from parameter names where the parameters are used.
- Replaces occurrences of `@ts-ignore` with `@ts-expect-error`, to catch future situations where suppression is no longer required for a line due to either improvements in TypeScript or changes to the line.

This change aims to maintain the existing behaviour of the resolver, with the exception of one bug affecting "create{{ COLLECTION-NAME }}" queries where the top-level entry in a collection does not have templates. This is because additional logic would need to be added to reinstate the bug. A test is added to confirm the correction and prevent regression. Note that the bug still exists for "create{{ COLLECTION-NAME }}" queries where the top-level entry does have templates.

## Issues resolved

- https://github.com/tinacms/tinacms/issues/5925
- https://github.com/tinacms/tinacms/issues/5926

## Future directions for consideration

- The `Resolver` could be reduced to two public methods that could be called from `resolve`: `typeResolver` and `fieldResolver`.  That way, all other methods could be made private to minimise the public interface of the class. Note that this would require that the functions in `auth-fields.ts` are folded into `Resolver.
- Further bug fixes and improvements to the Resolver will require changes to the generated database Lookup, but any changes to the generated Lookup will affect the schema checks conducted during `tinacms build`.

## Testing

I have not been able to test this change as thoroughly as I would like. The reviewer must test this change with sophisticated code bases such as https://github.com/tinacms/tina.io .


Pat-Stuart:
LGTM

Conducted testing with next.js starter and Tina.io.
Tested TinaCloud via tarball.

n-crrn:
CC: @kulesy 

Hi @Pat-Stuart,

Is there an intention to merge this change? I'm asking as the testing will need to be repeated if too much merge debt builds up.

PR #6114, opened 2025-11-12
fix: Adds in resolver for file path to handle windows

Pat-Stuart:
This is a small fix to ensure correct file path resolution on Windows systems.


JackDevAU (approved):
✅ lgtm

Issue #4642, opened 2024-07-15
Types - uploadDir doesn't exist


bradystroud:
### Describe the issue that you're seeing. Any Loom videos or screenshots usually help a lot!

Tina has a great feature that lets you specify where images go. you define this in your schema but there is always a type error.

Workaroud: use `@ts-expect-error`

```ts
    {
      type: "image",
      label: "Thumbnail",
      name: "eventThumbnail",
      // @ts-expect-error tinacms types are wrong
      uploadDir: () => "events",
    },
```

### Reproduction

https://github.com/sswconsulting/ssw.website

### System Info

```shell
n/a
```

### Validations

- [X] Follow our [Code of Conduct](https://github.com/tinacms/tinacms/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [docs](https://tina.io/docs).
- [X] Check that there isn't [already an issue](https://github.com/tinacms/tinacms/issues) that reports the same bug.

Pat-Stuart:
✅ Done - The type has been added correctly to TinaCMS and resolves this TS error now. Will be included in next release.

PR #6138, opened 2025-11-18
Add uploadDir to Image field type definition

Pat-Stuart:
[Relevant PBI](https://github.com/tinacms/tinacms/issues/4642)

This PR adds uploadDir as an optional property to the Image field type. Validation was also updated to accommodate this.
The use of this field is present in [image-field-plugin.tsx](https://github.com/tinacms/tinacms/blob/main/packages/tinacms/src/toolkit/fields/plugins/image-field-plugin.tsx)
This resolves a longstanding build-time TS error.

### DIFF
--- a/packages/@tinacms/schema-tools/src/types/index.ts
+++ b/packages/@tinacms/schema-tools/src/types/index.ts
@@ -243,26 +243,23 @@ export type DateTimeField = (
     type: 'datetime';
   };
 
-type ImageFieldOptions = {
-  /**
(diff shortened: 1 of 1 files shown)


Pat-Stuart:
Hey @JackDevAU 

I made a small change to add the definition to the root of the field as I see it being used. 
I have tested this using the tagged NPM release and this seems to resolve the TS error. I think using NPM release is the only way to correctly test this change.

Issue #4757, opened 2024-08-30
♻️TinaCMS - Remove lodash dependency


Ben0189:
[Lodash ](https://www.npmjs.com/package/lodash)packages is no longer updated (latest published 4 years ago) (it was introduced to provide some utility function for JavaScript project but Javascript has improved alot and we dont need lodash anymore, see this for some useful [resources](https://github.com/you-dont-need-x/you-dont-need-lodash)). These are just extra dependency to maintain.

Currently there are couple of packages in our monorepos that still using lodash, you might need to remove/replace all of them (graphql packages might be the tricky part because the code is hard to read)

- [x] Review the usage of lodash in each packages
- [ ] Replace lodash usage with something else (custom function if possible)
- [ ] Make sure the original function still working as it was before
- [ ] Remove lodash packages 

Something interesting to look at https://github.com/radashi-org/radashi

nicko-winner:
https://es-toolkit.slash.page/intro.html offers a compatibility layer to lodash, which might be worth looking at: https://es-toolkit.slash.page/compatibility.html

18-th:
Direct `lodash` usage has been removed in #6145. Usages have been replaced with either native or `es-toolkit` alternatives.

`yup` and `jest-dom` dependency on lodash removed in #6187.

Transitive dependencies investigation:

# Lodash Dependencies Analysis

This document outlines all lodash dependencies in the TinaCMS project and the path to eliminating them.

## Dependency Graph

```mermaid
graph TD
    subgraph "Your Packages"
        tinacms["tinacms"]
        next-cloudinary["next-tinacms-cloudinary"]
        tinacms-clerk["tinacms-clerk"]
        tinacms-mdx["@tinacms/mdx"]
        schema-tools["@tinacms/schema-tools"]
    end

    subgraph "Lodash Variants"
        lodash["lodash"]
        lodash-es["lodash-es"]
        lodash-startcase["lodash.startcase"]
        lodash-memoize["lodash.memoize"]
        lodash-mapvalues["lodash.mapvalues"]
        lodash-sortby["lodash.sortby"]
        lodash-truncate["lodash.truncate"]
    end

    subgraph "Intermediate Dependencies"
        react-color["✅ react-color"]
        reactcss["reactcss"]
        cloudinary["🔇 cloudinary"]
        cloudinary-core["cloudinary-core"]
        mermaid["⚠️ mermaid"]
        dagre-d3-es["dagre-d3-es"]
[...]

PR #6145, opened 2025-11-21
Remove Lodash

18-th:
The issue addressed is: #4757 

This PR removes the remaining usages of Lodash in the codebase and replaces them with equivalent native implementations or implementations from supported, maintained modern libraries.

Usages that could not be practically replaced with native implementations were substituted with the [es-toolkit](https://es-toolkit.dev) alternatives

Removed the following lodash usages:
* debounce - was not used, removed the reference
* camelcase - unused, removed the reference
* upperfirst - unused, removed the reference
* flatten - replaced by native .flat()
* get - replaced with an existing implementation from the GraphQL package
* cloneDeep - replaced with es-toolkit equivalent
* set - replaced with es-toolkit compat version. That implementation is identical to the one used by lodash
* uniqBy - replace with es-toolkit version. That implementation is identical to the one used by lodash

Fixed minimal type casting issues that arose from switching libraries.

### DIFF
--- a/.changeset/fifty-cameras-give.md
+++ b/.changeset/fifty-cameras-give.md
@@ -1,0 +1,7 @@
+---
+"@tinacms/schema-tools": patch
+"@tinacms/cli": patch
+"tinacms": patch
+---
+
+Fix: ui.max not applying to primitive type list:true
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -39,19 +39,21 @@ jobs:
           package_json_file: package.json
           run_install: false
 
-      - name: Get pnpm store directory
-        id: pnpm-cache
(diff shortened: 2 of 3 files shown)


.changeset/forty-roses-knock.md
```diff
@@ -0,0 +1,19 @@
+---
+"tinacms": patch
+"@tinacms/cli": patch
+"@tinacms/datalayer": patch
+"@tinacms/graphql": patch
+"@tinacms/mdx": patch
+"playwright-testing": patch
```
JackDevAU:
```suggestion
```

Don't need as its not user facing

18-th: Good point. Removed.

Issue #4412, opened 2024-01-04
String field with `list: true` does not respect `ui.min` and `ui.max` values


tyteen4a03:
### Describe the issue that you're seeing. Any Loom videos or screenshots usually help a lot!

When defining a `string` field with `list: true`, `ui.min` and `ui.max` is ignored and is not used in validation.

### Reproduction

N/A

### System Info

```shell
System:
    OS: macOS 14.2.1
    CPU: (10) arm64 Apple M1 Pro
    Memory: 22.52 MB / 32.00 GB
    Shell: 5.9 - /opt/homebrew/bin/zsh
  Binaries:
    Node: 18.17.1 - /private/var/folders/cg/srfblk2d7bq2dw2tqjt94qkh0000gn/T/xfs-cfcbfb04/node
    Yarn: 4.0.2 - /private/var/folders/cg/srfblk2d7bq2dw2tqjt94qkh0000gn/T/xfs-cfcbfb04/yarn
    npm: 9.6.7 - ~/Library/Caches/fnm_multishells/25880_1703958751436/bin/npm
    pnpm: 8.11.0 - ~/Library/pnpm/pnpm
    bun: 1.0.15 - ~/.bun/bin/bun
  Browsers:
    Brave Browser: 120.1.61.109
    Chrome: 120.0.6099.129
    Safari: 17.2.1
  npmPackages:
    @tinacms/cli: ^1.5.39 => 1.5.39
    tinacms: ^1.5.28 => 1.5.28
```

### Validations

- [X] Follow our [Code of Conduct](https://github.com/tinacms/tinacms/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [docs](https://tina.io/docs).
- [X] Check that there isn't [already an issue](https://github.com/tinacms/tinacms/issues) that reports the same bug.

tyteen4a03:
Still an issue.

JackDevAU:
Still an issue.

isaaclombardssw:
Still an issue, objects have the same limitation.

tyteen4a03:
Still an issue.

joshbermanssw:
✅ Done 

Prior to these changes the logic for removing the add button was `(!fixedLength || (fixedLength && !isMax))` - however this is unecessary as whether or not a list is fixedLength is we only need to disable the addButton on the maximum threshold 

**Testing Config**

```
{
						name: 'textListWithFixedLength',
						type: 'string',
[...]

PR #6147, opened 2025-11-23
Fix conditional logic for max items

joshbermanssw:
For #4412 

- Fixed conditional logic for add button being disabled on `isMax` 
- Removed @ts-ignore's 

Prior to these changes the logic for removing the add button was `(!fixedLength || (fixedLength && !isMax))` - however this is unecessary as whether or not a list is fixedLength is we only need to disable the addButton on the maximum threshold 

**Testing Config**

```
{
						name: 'textListWithFixedLength',
						type: 'string',
						label: 'Text List With Fixed Length',
						list: true,
						ui: {
							min: 3,
							max: 3,
						},
					},
					{
						name: 'someTextList',
						label: 'SomeText',
						type: 'string',
						list: true,
						ui: {
							min: 1,
							max: 3,
						},
					},
					{
						name: 'someOtherText',
						label: 'someOtherText',
						type: 'image',
						list: true,
						ui: {
							min: 1,
							max: 3,
						},
					},
```

Screenshot 2025-11-24 at 10 06 57 am

**Figure: Disabled Add Buttons after max is hit**

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -49,7 +49,7 @@ jobs:
         name: Setup pnpm cache
         with:
           path: ${{ steps.pnpm-cache.outputs.STORE_PATH }}
-          key: ${{ runner.os }}-pnpm-store-${{ hashFiles('pnpm-lock.yaml') }}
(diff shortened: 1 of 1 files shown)


kulesy:
LGTM - Approved

PR #6182, opened 2025-11-26
💄 Nav menu UI bugfix

Pat-Stuart:
[Relevant PBI](https://github.com/orgs/tinacms/projects/38?pane=issue&itemId=139252958&issue=tinacms%7Ctinacms%7C6118)

This PR implements a context and trigger for the nav menu. 
This allows us to fix a UI bug in which the menu icon took up an entire column of the screen pushing all content to the side.

This is a stripped back, different approach to the refactor I did.

From:

image

To:

image

From:

image

To:

image

image


Pat-Stuart:
As discussed, these screens use a different wrapper. I have added the trigger into that header and standardized the styling to our current 'header'. It looks like the only time this specific wrapper is used is for the media manager, select branch, and users (if self-hosting), all appear as expected now.

Issue #6235, opened 2025-12-08
🐛 Bug - Toast Blocks Interaction


joshbermanssw:
Requested by: @joshbermanssw via [YakShaver.ai](https://YakShaver.ai) 🦬
cc: @joshbermanssw, @kulesy, @adamcogan, @JackDevAU

Hi Team!

&nbsp;&nbsp;[🟥&nbsp;&nbsp;Watch the video (1 min 0 sec)](https://www.youtube.com/watch?v=IigyVEM71go)

&nbsp;&nbsp;Url from screen share: [https://localhost:3000/admin/index.html#/collections/tina-cloud/simples#i/apple](https://localhost:3000/admin/index.html#/collections/tina-cloud/simples#i/apple)

### Pain
The toast notification in TinaCMS is blocking interaction with elements underneath it due to excessive padding and high Z-index.

### Acceptance Criteria
1. Toast notification should not block interaction with elements underneath it.
2. Padding around the toast should be minimized to prevent blocking.
3. Z-index of the toast should be adjusted to allow interaction with underlying elements.

### Reproduce Steps
1. Open TinaCMS and perform an action that triggers a toast notification.
2. Attempt to click on elements underneath the toast notification.
3. Observe that clicks are blocked by the toast.

### Screenshot
![Screenshot](https://sayakshaverproduction.blob.core.windows.net/images/images/e21eadd2-2fa2-42b7-9e4a-845cba4290e6_frame_32.png?sv=2025-05-05&se=2045-12-08T04%3A33%3A26Z&sr=b&sp=rd&sig=m8qgi9X4VBKNeiTdPg5x8COfD462XVAN8h40p69ZWi0%3D)
**Figure: Toast overlay blocks interaction with underlying elements**

joshbermanssw:
✅ Done and merged

joshbermanssw:
<img width="532" height="257" alt="Screenshot 2025-12-08 at 3 25 47 pm" src="https://github.com/user-attachments/assets/db53c6d5-b31a-4a57-a343-868fe3ebaebb" />

**Figure: Before, padding blocking**

<img width="533" height="477" alt="Screenshot 2025-12-08 at 3 35 16 pm" src="https://github.com/user-attachments/assets/e9ccf47e-7690-438b-88a0-a23243cfc6f8" />

**Figure: After, position by top/left**

PR #6236, opened 2025-12-08
🐛 Fix padding block on toasts

joshbermanssw:
As per #6235 

- [ ] Removed `p-6` which positions toast but blocks content underneath - replace with `top-6 left-6`

Screenshot 2025-12-08 at 3 25 47 pm

**Figure: Before, padding blocking**

Screenshot 2025-12-08 at 3 35 16 pm

**Figure: After, position by top/left**


Pat-Stuart (approved):
LGTM

PR #6254, opened 2025-12-11
Update dependencies React 19 and Next.js 15 for tinacms-authjs package

MartinFiti:
**Discussion:** https://github.com/tinacms/tinacms/discussions/6248

This PR updates the packages for the tinacms-authjs package to React 19 and Next.js 15, along with required peer dependency changes.

**Note on Testing:**

I wasn't able to fully test the upgrade. All example apps rely on a localAuth provider in the test environment, and I couldn't find a simple way to run them without it. Because of this, I'm not sure whether the update works as expected.

**Request:**

If someone can point me to the correct way to test this setup or test the changes themselves. I'd really appreciate it.

### DIFF
--- a/.changeset/gentle-cycles-chew.md
+++ b/.changeset/gentle-cycles-chew.md
@@ -1,0 +1,5 @@
+---
+"tinacms-authjs": mayor
+---
+
+Update packages NextJs and React to be compatible with version 15 and 19 for tinacms-authjs package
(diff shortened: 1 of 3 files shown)


18-th:
Hey @MartinFiti, we appreciate your contribution! Unfortunately, the versions of React and Next that you used were affected by the React CVEs, so we couldn't accept this. (The same reason I am replying to you that late as we've been flat out for the past few weeks patching all of our projects and templates 😆). The work was completed and tested as part of #6354 with patched versions.

PR #6255, opened 2025-12-11
🐖 Add Posthog to create-tina-app

joshbermanssw:
As per https://github.com/tinacms/tinacloud/issues/2500 

- [ ] Added posthog-node to the create-tina-app 
- [ ] Added two events `create-tina-app-started` & `create-tina-app-finished` 
- [ ] Allow users to opt out of telemetry via `--noTelemetry` flag

Screenshot 2025-12-12 at 10 24 15 am

**Figure: Events showing in PostHog**

Screenshot 2025-12-12 at 11 26 50 am

**Figure: Telemetry Notice if Telemetry is displayed**

### DIFF
--- a/packages/create-tina-app/package.json
+++ b/packages/create-tina-app/package.json
@@ -44,7 +44,6 @@
         "typescript": "^5.7.3"
     },
     "dependencies": {
-        "@tinacms/metrics": "workspace:*",
         "chalk": "catalog:",
         "commander": "^12.1.0",
         "cross-spawn": "catalog:",
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -5,7 +5,7 @@
     // },
     "editor.defaultFormatter": "biomejs.biome",
     "[typescript]": {
-        "editor.defaultFormatter": "biomejs.biome"
+        "editor.defaultFormatter": "fronterior.biome-monorepo"
     },
     "[typescriptreact]": {
         "editor.defaultFormatter": "biomejs.biome"
--- a/packages/create-tina-app/src/index.ts
+++ b/packages/create-tina-app/src/index.ts
@@ -1,4 +1,3 @@
-import { Telemetry } from '@tinacms/metrics';
 import prompts from 'prompts';
 import path from 'node:path';
 import { createRequire } from 'node:module';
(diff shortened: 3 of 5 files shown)


.changeset/curvy-waves-add.md
```diff
@@ -0,0 +1,5 @@
+---
+"create-tina-app": patch
+---
+
+🐗 Posthog - add create-tina-app start + finish telemetry
```
wicksipedia:
```suggestion
📊 Telemetry - helping us build a better Tina
We’ve added anonymous telemetry to `⁠create-tina-app`. This helps our team understand which environments (Node version, OS, Package Manager) and starter templates are most popular so we can prioritize the right features and fixes.
Privacy First: We never collect personal data or project-specific code.
Opt-out: Prefer to stay off the grid? You can disable this at any time by passing the `⁠--noTelemetry` flag.
```

joshbermanssw: Good changes

packages/create-tina-app/.env.example
wicksipedia, line 1: is this required? it looks like it gets the api key from an api call

packages/create-tina-app/package.json
```diff
@@ -48,8 +48,10 @@
         "chalk": "catalog:",
         "commander": "^12.1.0",
         "cross-spawn": "catalog:",
+        "dotenv": "^17.2.3",
```
wicksipedia: if the `.env` is no longer required, then I would guess that this should be able to be dropped as well

wicksipedia (approved):
test pass with changes

packages/create-tina-app/package.json
joshbermanssw: ✅ Removed

wicksipedia:
PostHog received telemetry but I couldn't see OS info.

Can we capture that, node info, and available package managers on start?

This is in case the user doesn't select a template.

CleanShot 2025-12-16 at 15 49 24@2x

joshbermanssw:
> PostHog received telemetry but I couldn't see OS info.
> 
> Can we capture that, node info, and available package managers on start?
> 
> This is in case the user doesn't select a template.
> 
> CleanShot 2025-12-16 at 15 49 24@2x

✅ Added OS platform and OS version 

Screenshot 2025-12-16 at 4 53 06 pm

joshbermanssw:
Screenshot 2025-12-17 at 9 26 15 am

**Figure: Change to just use os.distro and os.release**

wicksipedia (approved):
lgtm

Issue #6285, opened 2025-12-21
✨ UI Enhancement - Half-width field specifier for string field


joshbermanssw:
Requested by: @joshbermanssw via [YakShaver.ai](https://YakShaver.ai) 🦬
cc: @joshbermanssw, @kulesy, @adamcogan, @JackDevAU

Hi Team!

&nbsp;&nbsp;[🟥&nbsp;&nbsp;Watch the video (1 min 22 sec)](https://www.youtube.com/watch?v=rfiU-wRzYKg)

&nbsp;&nbsp;Url from screen share: [https://localhost:3000/admin/index.html](https://localhost:3000/admin/index.html)

### Pain
Implement a boolean option for users to choose between full-width and half-width display for publication and updated date fields in the UI. Default to full-width, but allow half-width to enable two fields on one line.

### Acceptance Criteria
1. Add a boolean option for date fields to toggle between full-width and half-width.
2. Ensure the default setting is full-width.
3. Allow two half-width fields to be displayed on the same line.
4. Consider editor size and ensure usability with small input boxes.

### Screenshot
![Screenshot](https://sayakshaverproduction.blob.core.windows.net/images/images/3dc5935f-f302-4cf9-9296-b69c904afbed_frame_48.png?sv=2025-05-05&se=2045-12-21T22%3A37%3A30Z&sr=b&sp=rd&sig=vhrjBwV1dcnmL8XHxfLZmhYUlr4PMHKqKb8nM3xAICg%3D)
**Figure: Title string field set to half width**

<img width="401" height="342" alt="Image" src="https://github.com/user-attachments/assets/61df1cdf-fc0a-4a22-accb-f261263f5d3d" />

**Figure: Example**

kulesy:
✅ Done 

![Dec-30-2025 14-46-07](https://github.com/user-attachments/assets/c938cb8d-cc7b-4985-8f19-56ee7461ed2e)
**Figure: Half width fields**

PR #6286, opened 2025-12-22
✨ Feature - Add half width prop to text

joshbermanssw:
As per #6285 

- [ ] Added a property to the string type that allows users to define the `width` 
- [ ] Restructured the form builder to use flex-col instead of block to account for wrapping numerous elements onto a line

```
  fields: [
    {
      label: 'Title',
      name: 'title',
      type: 'string',
      width: 'half',
    },
    {
      label: 'Link',
      name: 'link',
      type: 'string',
      width: 'half',
    },
```

**Example Code Block**

Screenshot 2025-12-22 at 12 16 08 pm

**Figure: Example of two consecutive strings with width:'half'**

Screenshot 2025-12-22 at 12 18 38 pm

**Figure: Example of one field with width: 'half'**

### DIFF
--- a/packages/tinacms/src/toolkit/fields/plugins/wrap-field-with-meta.tsx
+++ b/packages/tinacms/src/toolkit/fields/plugins/wrap-field-with-meta.tsx
@@ -214,7 +214,7 @@ export const FieldWrapper = ({
 
   return (
     <div
-      className={`relative px-2 ${margin ? `mb-5 last:mb-0` : ``} ${field?.width === 'half' ? 'w-1/2' : 'w-full'} ${getFieldStateClasses()}`}
+      className={`relative w-full px-2 ${margin ? 'mb-5 last:mb-0' : ''} ${field?.width === 'half' ? '@sm:w-1/2' : ''} ${getFieldStateClasses()}`}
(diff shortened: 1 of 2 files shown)


packages/@tinacms/schema-tools/src/types/index.ts
```diff
@@ -224,6 +224,7 @@ export type StringField = (
     isTitle?: boolean;
     isBody?: boolean;
     options?: Option[];
+    editorWidth?: 'full' | 'half';
```
kulesy: why not just call this width?

kulesy:
On mobile should we revert back to full width?

image

**Figure: Form widths quite small**

packages/@tinacms/schema-tools/src/types/index.ts
kulesy: ✅ Renamed

kulesy:
> On mobile should we revert back to full width?
> 
> image
> **Figure: Form widths quite small**



✅ Done - when the form goes below <384px (sm) it will return to full width

Dec-30-2025 14-46-07

**Figure: Responsive form fields**

Issue #6297, opened 2025-12-30
Define a set of PostHog events required for the user facing website aspects of TinaCMS


brookjeynes-ssw:
This will include sending PostHog events on actions such as branch creation, branch switching, content saving, etc.

Once these events are confirmed they will split up into seperate PBIs to be implemented. Linking to the parent issue.

kulesy:
✅ Done

PR #6298, opened 2025-12-30
feat(posthog): add posthog event constants

brookjeynes-ssw:
This PR adds posthog event constants that are planned to be implemented

Closes https://github.com/tinacms/tinacms/issues/6297

### DIFF
--- a/.changeset/giant-beans-exercise.md
+++ b/.changeset/giant-beans-exercise.md
@@ -1,5 +1,1 @@
----
-"tinacms": patch
----
-
 🐖 PostHog - Define Events


packages/tinacms/src/lib/posthog/posthog.ts
```diff
@@ -1,12 +1,39 @@
+// User begins creating branch
 export const CreateBranchStartedEvent: string = 'create-branch-started';
+
+// The user finishes creating a branch
 export const CreateBranchFinishedEvent: string = 'create-branch-finished';
```
brookjeynes-ssw: Are these two events better suited to be logged from TinaCloud?

packages/tinacms/src/lib/posthog/posthog.ts
```diff
@@ -1,12 +1,39 @@
...
+
+// When a user clicks 'save' in the TinaCMS Editor 
 export const SavedContentEvent: string = 'saved-content';
+
+// When a user adds an item to the media manager 
 export const MediaManagerContentUploadedEvent: string =
   'media-manager-content-uploaded';
+
+// When a user deletes an item from the media manager 
 export const MediaManagerContentDeletedEvent: string =
```
brookjeynes-ssw, line 46: I wonder if we should also define the event payloads beside the event?

packages/tinacms/src/lib/posthog/posthog.ts
```diff
@@ -1,12 +1,39 @@
...
   'media-manager-content-deleted';
+
+// The format of the content uploaded to the media manager (.webm, .png, etc)
+export const MediaManagerContentFormat: string =
```
brookjeynes-ssw: Should this be an event or a property on `MediaManagerContentUploadedEvent`?

packages/tinacms/src/lib/posthog/posthog.ts
```diff
@@ -1,12 +1,39 @@
...
+export const MediaManagerContentFormat: string = 
+  'media-manager-content-type';
+
+// When a rich-text field is switched to Markdown mode (from Rich-text mode)
 export const EditorSwitchedToMarkdownEvent: string =
   'editor-switched-to-markdown';
+
+// When a rich-text field is switched to Rich-text mode (from Markdown mode)
 export const EditorSwitchedToRichtextEvent: string =
   'editor-switched-to-richtext';
```
brookjeynes-ssw: Should these be two separate events or a single "switched" event with a property

packages/tinacms/src/lib/posthog/posthog.ts
joshbermanssw: Property on the event.

packages/tinacms/src/lib/posthog/posthog.ts
joshbermanssw: Definitely could be a single event with a property indicating which mode it is on (rich-text or markdown)

packages/tinacms/src/lib/posthog/posthog.ts
joshbermanssw: 👍

.changeset/giant-beans-exercise.md
```diff
@@ -0,0 +1,5 @@
+---
+"tinacms": patch
+---
+
+🐖 PostHog - Define Events
```
wicksipedia, line 1: nitpick: if these are just definitions and not in use, I don't think that we need to rev the version number.

wicksipedia (dismissed):

joshbermanssw (approved):

PR #6303, opened 2026-01-01
Enable Dynamic Slash Commands for Tina MDX Templates

nojomyth-dev:
This PR removes the hardcoded slash menu items and wires the / menu up to Tina’s MDX templates instead. Any block defined in .tina/config.ts (like Callout) now shows up automatically and can be inserted from the editor. Inserting a block also cleans up the slash paragraph and keeps the cursor in a sensible place, so you can keep typing without extra empty lines.

The goal of this PR was to implement something similar to notion commands. While there were some hardcoded commands, for people like me who like to type rather than click, having the commands inside the slash menu is QoL.
I am not familiar with the code base, so please leave feedback / change requests. I currently have limited time, so might take a while (anyone can take over).

How it looks like for the user / editor:
image
image

How the code could look like to add custom MDX:
image

### DIFF
--- a/packages/tinacms/src/toolkit/fields/plugins/mdx-field-plugin/plate/components/plate-ui/slash-input-element.tsx
+++ b/packages/tinacms/src/toolkit/fields/plugins/mdx-field-plugin/plate/components/plate-ui/slash-input-element.tsx
@@ -84,12 +84,12 @@ export const SlashInputElement = withRef<typeof PlateElement>((props, ref) => {
 
   return (
     <PlateElement
-      as="span"
-      ref={ref}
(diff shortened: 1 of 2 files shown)


JackDevAU:
Hey @nojomyth-dev this is a super cool feature! thanks for adding this! 
I'm noticing a lot of `any` types around the place and double casting of types. Can we clean these up a bit!

nojomyth-dev:
Hi, thanks!

Yes, i'll get on it in a couple of hours!

nojomyth-dev:
@JackDevAU It should be typesafe now without the double type casting.

PR #6309, opened 2026-01-04
Bug - Runaway field resolver promises

n-crrn:
CC: @brookjeynes-ssw @kulesy 

This change fixes an issue where a leg of a query may continue execution after another leg has already failed and the resolver has returned the error message. The significance is that the continuing query can then fail during a separate later query, causing a spurious error that has nothing to do with the later query.

The fix works by tracking all field resolver promises that are created, and then ensuring that they are all settled before exiting the `resolve` function. The GitHub formatting of the change obscures the form of the fix within the `resolve` function of `resolve.ts`, but it can be summarised as follows:

```ts
// Track all field resolver promises to prevent runaway promises.
const resolverPromises: Promise<unknown>[] = [];
...
resolver: async (...) => {
    const fieldPromise = (async () => { /* Old resolver */ })();
    resolverPromises.push(fieldPromise);
    return fieldPromise;
}
...
// Ensure all field resolvers have completed before continuing, as errors can cause
// a premature return.
await Promise.allSettled(resolverPromises);
```

Regarding alternative approaches:

- An [AbortController](https://developer.mozilla.org/en-US/docs/Web/API/AbortSignal#examples) could be used as a cancellation token with the `Database.query` method to cease a query early. However, this would require more complicated API and logic changes.
- No attempt is made to remove completed promises from the `resolverPromises` array, as it would complicate the logic and risk race conditions when modifying the array.

In addition to the fix, a test is included that demonstrates the bug.

- The test works by delaying the return of calls to `Database.query`. When the bug is present, the test assertions are evaluated before the query completes.
- I have not been able to devise a reliable and easily understood way to coordinate the sequence without timeouts. To help ensure the integrity of the test during a run, a timeline of events is maintained within the test.
- The test does not use the existing test `setup*` functions in `tests/util.ts`, as the methods of the `Database` object need to be overridden rather than those of the `Bridge`.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -31,6 +31,7 @@ jobs:
         uses: actions/setup-node@v4
         with:
           node-version-file: ".nvmrc"
+          registry-url: "https://registry.npmjs.org"
 
       - name: Upgrade npm for OIDC support
         run: npm install -g npm@latest


kulesy:
For some reason, the publish tagged PR is not working - Can't figure out exactly why.

kulesy:
The tagged PR isn't publishing as our OIDC policy for creating tagged releases doesn't allow forked repositories.

I have tested this locally and it all looks good to me.

✅ Approved

PR #6354, opened 2026-01-15
Bump next and react for the auth package

18-th:
This PR addresses a part of #4720 to get the examples up to date and optimise them

* Bumped next to 15.5.9 and react to 19.2.3 for the auth package to allow the examples to use the latest and greatest
* Updated the self-hosted example to use ESM properly and run

### DIFF
--- a/.changeset/chilled-sheep-promise.md
+++ b/.changeset/chilled-sheep-promise.md
@@ -1,7 +1,0 @@
----
-"tinacms": patch
-"@tinacms/app": patch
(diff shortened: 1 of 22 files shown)


.changeset/early-rats-float.md
```diff
@@ -0,0 +1,9 @@
+---
+"tinacms-authjs": patch
+"@examples/self-hosted-starter": patch
```
JackDevAU:
```suggestion
```

don't need changesets for examples

PR #6379, opened 2026-01-28
💄 Improve branch validation error handling with structured error codes

kulesy:
## Summary
- Replaces fragile string matching with structured error codes for branch validation errors
- Adds proper handling for branch hierarchy conflicts (e.g., can't create `tina/foo` when `tina/foo/bar` exists)
- Maintains backwards compatibility with older API versions via fallback string matching

### DIFF
--- a/.changeset/real-suns-double.md
+++ b/.changeset/real-suns-double.md
@@ -1,0 +1,5 @@
+---
+"tinacms": minor
+---
+
+Added parse field for filename field
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -4,6 +4,18 @@ on:
     branches: [main]
   pull_request:
     types: [opened, synchronize, reopened, labeled]
+  workflow_dispatch:
+    inputs:
+      test_graphql_update:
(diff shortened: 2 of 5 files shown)


packages/tinacms/src/internalClient/index.ts
```diff
@@ -643,9 +643,16 @@ mutation addPendingDocumentMutation(
 
       if (!res.ok) {
         console.error('There was an error starting editorial workflow.');
-        throw new Error(
+        const error = new Error(
           responseBody?.message || 'Failed to start editorial workflow'
-        );
+        ) as Error & { errorCode?: string; conflictingBranch?: string };
```
brookjeynes-ssw: This type of cast is used in another place as well. should we have a named type for this instead?

kulesy: Good catch, @copilot fix this clanker
https://github.com/tinacms/tinacms/pull/6379/changes#diff-2e1be6dd095804911f3a3c2f89ee766b49975b84a9b8858291f355469520feafR218

PR #6380, opened 2026-01-28
Improve publish workflow comment with actual package versions

kulesy:
## Description

Updates the GitHub Actions publish workflow to generate more accurate and helpful installation instructions in PR comments. Instead of using a generic tag placeholder, which was causing issues when trying to install as it thought I was pointing to a sub packages like `package@foo/bar`.


.github/workflows/publish.yml
```diff
@@ -175,19 +175,20 @@ jobs:
             
             console.log('📝 Generated packages section:', packagesSection);
             
+            // Generate install commands using actual versions
+            let installCommands = packages.map(pkg => `pnpm add ${pkg.name}@${pkg.version}`).join('\n');
```
JackDevAU, line 179: do we expect this to run?

kulesy: No this is just for the user to copy to run it themselves easily

JackDevAU (approved):
:shipit:

Issue #6366, opened 2026-01-21
Remote GraphQL schema does not exist for dependabot branch


vorotech:
### Describe the issue that you're seeing. Any videos (eg. YouTube or Loom) or screenshots help!

Build failed when building from dependabot branch.
```
> tinacms build --partial-reindex && next build

Starting Tina build

WARN: Detected bot branch. Using schema/content from default branch 'main' instead of 'dependabot/github_actions/google/osv-scanner-action/dot-github/workflows/osv-scanner-reusable.yml-2.3.2'.
Checking indexing process in TinaCloud...
The remote GraphQL schema does not exist. Check indexing for this branch.

Additional info: Branch: dependabot/github_actions/google/osv-scanner-action/dot-github/workflows/osv-scanner-reusable.yml-2.3.2
```

Dependabot branch are not appearig in the Tine Cloud App

### Reproduction

https://github.com/vorotech/www-vorote-ch/actions/runs/21204384689/job/60997235668?pr=104

### Steps to reproduce

Any dependabot bracnh is currently failing

### System Info

```shell
System:
    OS: macOS 15.6.1
    CPU: (12) arm64 Apple M4 Pro
    Memory: 114.47 MB / 24.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 22.21.1 - /opt/homebrew/opt/node@22/bin/node
    npm: 10.9.4 - /opt/homebrew/opt/node@22/bin/npm
    pnpm: 9.10.0 - /opt/homebrew/bin/pnpm
  Browsers:
    Chrome: 143.0.7499.193
    Firefox: 147.0.1
    Safari: 18.6
  npmPackages:
    @tinacms/cli: ^2.0.7 => 2.0.7 
    tinacms: ^3.2.0 => 3.2.0
```

### Validations

- [x] Follow our [Code of Conduct](https://github.com/tinacms/tinacms/blob/main/CODE_OF_CONDUCT.md)
- [x] Read the [docs](https://tina.io/docs).
- [x] Check that there isn't [already an issue](https://github.com/tinacms/tinacms/issues) that reports the same bug.

vorotech:
The workaround I use.
Checkout the branch locally, create a new branch with custom name, push to remote. Tina Cloud doesn't detect it as a bot branch and everything work correctly.

[...]

PR #6381, opened 2026-01-29
fix: use default branch instead of dependabot branch when indexing

brookjeynes-ssw:
image
Figure: defaulting to main branch

Resolves https://github.com/tinacms/tinacms/issues/6366


brookjeynes-ssw:
Closing as this is not a TinaCMS related change and should be done in TinaCloud instead.

PR #6258, opened 2025-12-12
fixing vulnerabilities in tinacms

thethunderturner:
Should fix some of the major vulnerabilities found in my project
```
[0.03ms] ".env"
bun audit v1.3.4 (5eb2145b)
dompurify  <2.5.4
  tinacms › mermaid › dompurify
  high: DOMPurify allows tampering by prototype pollution - https://github.com/advisories/GHSA-mmhx-hmjr-r674
  moderate: DOMPurify allows Cross-site Scripting (XSS) - https://github.com/advisories/GHSA-vhxf-7vqr-mrjg
  critical: DOMPurify vulnerable to tampering by prototype polution - https://github.com/advisories/GHSA-p3vf-v8qc-cwcr
  high: DOMpurify has a nesting-based mXSS - https://github.com/advisories/GHSA-gx9m-whjm-85jf

jsonpath-plus  <10.2.0
  @tinacms/cli › @tinacms/graphql › jsonpath-plus
  tinacms › @tinacms/search › @tinacms/graphql › jsonpath-plus
  critical: JSONPath Plus Remote Code Execution (RCE) Vulnerability - https://github.com/advisories/GHSA-pppg-cpfq-h7wr
  high: JSONPath Plus allows Remote Code Execution - https://github.com/advisories/GHSA-hw8r-x6gr-5gjp

vite  <=5.4.19
  @astrojs/react › vite
  @tailwindcss/vite › vite
  @tinacms/cli › vite
  astro › vite
  low: Vite middleware may serve files starting with the same name with the public directory - https://github.com/advisories/GHSA-g4jq-h2w9-997c
  low: Vite's `server.fs` settings were not applied to HTML files - https://github.com/advisories/GHSA-jqfw-vq24-v9c3
  moderate: vite allows server.fs.deny bypass via backslash on Windows - https://github.com/advisories/GHSA-93m4-6634-74q7

mermaid  <=10.9.2
  tinacms › mermaid
  high: Prototype pollution vulnerability found in Mermaid's bundled version of DOMPurify - https://github.com/advisories/GHSA-m4gq-x24j-jpmf

esbuild  <=0.24.2
  @tinacms/cli › esbuild
  astro › esbuild
  @astrojs/react › vite › esbuild
  @tailwindcss/vite › vite › esbuild
  moderate: esbuild enables any website to send any requests to the development server and read the response - https://github.com/advisories/GHSA-67mh-4wv8-2f99
  ```
  
  tinacms seems to still be building locally
  ```
[...]

PR #6404, opened 2026-02-06
Dependencies: Upgrade tinacms

thethunderturner:
Upgrades some dependencies as there are quite a few vulnerabilities.
Before the PR
```
42 vulnerabilities found
Severity: 8 low | 13 moderate | 19 high | 2 critical
```

After the PR
```
35 vulnerabilities found
Severity: 8 low | 11 moderate | 16 high
```

In my original PR [here](https://github.com/tinacms/tinacms/pull/6258) there were far more vulnerabilities, but it seems a lot were automatically fixed

### DIFF
--- a/package.json
+++ b/package.json
@@ -42,7 +42,7 @@
 		"typescript": "^5.7.3",
 		"vite": "^4.5.9"
 	},
-	"packageManager": "pnpm@9.15.5",
+	"packageManager": "pnpm@10.28.2",
 	"dependencies": {
 		"highlight.js": "catalog:",
 		"react-colorful": "catalog:"
--- a/packages/tinacms/package.json
+++ b/packages/tinacms/package.json
@@ -101,7 +101,7 @@
     "graphql-tag": "catalog:",
     "is-hotkey": "catalog:",
     "lucide-react": "catalog:",
-    "mermaid": "9.3.0",
+    "mermaid": "11.12.2",
     "moment": "catalog:",
     "moment-timezone": "^0.6.0",
     "monaco-editor": "catalog:",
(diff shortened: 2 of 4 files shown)


JackDevAU:
Hey @thethunderturner

Can we simplify this down even more and just do 1 major change per pr so we can see a bit more clearly what is breaking and why. There is also another pr up - https://github.com/tinacms/tinacms/pull/6410 that is upgrading mermaid so that will handle that dep.  

I'm back from internal work so I will be a lot more active here to get these over the line.

Cheers

thethunderturner:
> Hey @thethunderturner
> 
> Can we simplify this down even more and just do 1 major change per pr so we can see a bit more clearly what is breaking and why. There is also another pr up - https://github.com/tinacms/tinacms/pull/6475 that is upgrading mermaid so that will handle that dep.  
> 
> I'm back from internal work so I will be a lot more active here to get these over the line.
> 
> Cheers

Hi, I dont understand how much more we can simplify 😅. I thought we agreed on one package per PR, and this one only changes dependencies on tinacms package. Also, the PR you referenced seems unrelated to mermaid?

JackDevAU:
> > Hey @thethunderturner
> > Can we simplify this down even more and just do 1 major change per pr so we can see a bit more clearly what is breaking and why. There is also another pr up - #6475 that is upgrading mermaid so that will handle that dep.
> > I'm back from internal work so I will be a lot more active here to get these over the line.
> > Cheers
> 
> Hi, I dont understand how much more we can simplify 😅. I thought we agreed on one package per PR, and this one only changes dependencies on tinacms package. Also, the PR you referenced seems unrelated to mermaid?

Hey, sorry about the incorrect PR link earlier; I’ve fixed that.

Ideally, we keep it to one package update per PR. For example, a Vite upgrade from 4 → 6 would be its own PR. That way, we can clearly see what breaks with each major update and address issues faster. I expect some of these major bumps will surface several problems that we’ll need to fix along the way.

Sorry for the runaround - I think this is the quickest and easiest path forward.

Issue #6430, opened 2026-02-16
🐖 PostHog - Opt in, out or anonymise Telemetry


joshbermanssw:
Add a field in config to allow users to either opt out completely or anonymise Telemetry. 

Field is opt in by default

joshbermanssw:
✅ Done - users can opt out of telemetry in the config of their tinacms project. By default we anonymize all projects. 

I have created playwright tests around this to ensure the logic is tested

We also signify to the user that we have telemetry in the build output 

<img width="999" height="579" alt="Screenshot 2026-02-17 at 4 56 25 pm" src="https://github.com/user-attachments/assets/1bd8f206-cf63-47df-94dc-1dac2399f879" />

**Figure: Notice of Telemetry in the build output**

Issue #6299, opened 2025-12-30
Implement PostHog logs for branch creation


brookjeynes-ssw:
We would like to track the time it takes for branches to be created and the amount of branches being created.

joshbermanssw:
✅ Done 

<img width="737" height="102" alt="Screenshot 2026-02-17 at 10 47 15 am" src="https://github.com/user-attachments/assets/068c2e60-0f41-49fe-8d50-cf0e10122dc4" />

**Figure: Event payload, local**

<img width="915" height="142" alt="Screenshot 2026-02-17 at 2 28 10 pm" src="https://github.com/user-attachments/assets/7922add4-8bba-4db3-b102-c6ddd7f86c97" />

PR #6438, opened 2026-02-16
Setup opt-in/opt-out, initialize on TinaAdmin

joshbermanssw:
As per 
- https://github.com/tinacms/tinacms/issues/6299 
- https://github.com/tinacms/tinacms/issues/6430

Adding PostHog telemetry to the TinaCMS Started Event. This event will allow us to identify: how often users are going to the /admin bar, and what versions (past 3.4.1) are out in the wild for TinaCMS

We will also get to see the amount of people self-hosted versus using TinaCloud

- [ ] If a user runs `npx create-tina-app@latest --noTelemetry`, we will create their tina project w telemetry disabled

Try it for yourself at https://tina-a07rujjqm-tinacms.vercel.app/admin#/~ , open the logs too

Screenshot 2026-02-17 at 10 47 15 am

**Figure: Event payload, local**

Screenshot 2026-02-17 at 2 28 10 pm

**Figure: When testing on staging for tina.io using a tagged release**

Screenshot 2026-02-17 at 4 56 25 pm

**Figure: Notice of Telemetry in the build output**

### DIFF
--- a/packages/@tinacms/cli/src/next/commands/dev-command/index.ts
+++ b/packages/@tinacms/cli/src/next/commands/dev-command/index.ts
@@ -325,11 +325,13 @@ export class DevCommand extends BaseCommand {
         // },
       ],
     });
-if (configManager?.config?.telemetry === 'anonymous') {
-    logger.info(`\n📊 Note: TinaCMS now collects anonymous telemetry regarding usage. More information on TinaCMS Telemetry: https://tina.io/telemetry\n`);
-}
+    if (configManager?.config?.telemetry === 'anonymous') {
+      logger.info(
+        `\n📊 Note: TinaCMS now collects anonymous telemetry regarding usage. More information on TinaCMS Telemetry: https://tina.io/telemetry\n`
+      );
+    }
 
-await this.startSubCommand();
+    await this.startSubCommand();
   }
 
   watchContentFiles(


packages/tinacms/src/lib/posthog/posthogProvider.ts
```diff
@@ -0,0 +1,184 @@
+import posthog, { PostHog } from 'posthog-js';
+
+/**
+ * Telemetry mode configured by the user in their tina config.
+ * - "enabled": Full telemetry with user/project identification
+ * - "anonymized": Telemetry without user/project identification
+ * - "disabled": No telemetry collected
+ */
+export type TelemetryMode = 'enabled' | 'disabled' | 'anonymized';
```
wicksipedia: it looks like the prop in [‎packages/@tinacms/schema-tools/src/types/index.ts](https://github.com/tinacms/tinacms/pull/6438/changes#diff-88b24591b0c36f508270e9ae261e94ea2770198c387d66a95b2fe1620810bde3) has the same string values, should it instead reference this prop?

packages/tinacms/src/admin/index.tsx
```diff
@@ -29,8 +29,81 @@ import CollectionUpdatePage from './pages/CollectionUpdatePage';
...
+    // Get telemetry mode from user config, default to 'enabled'
+    const telemetryMode: TelemetryMode =
+      client?.schema?.config?.config?.telemetry || 'enabled';
+
+    initializePostHog(telemetryMode).then((posthog) => {
+      if (posthog) {
+        const eventProperties: Record<string, string> = {
+          tinaCMSVersion: pkg.version,
+          system: 'tinacms/tinacms',
+          authType: authType,
```
wicksipedia: As per teams, whilst this is the auth, we really want to capture if their backend is TinaCloud or Self-hosted.

It may not be warranted to workout which auth platform they are using.

```suggestion
          backendType: authType,
```

wicksipedia:
I'm not seeing any tests for this, would like if there were checks to make sure that the different telemetry settings worked as expected e.g. when in anonymous mode, the client id isn't sent to post hog.

I think this should be possible with playwright and mocks

packages/tinacms/src/lib/posthog/posthog.ts
18-th: Do you mind fixing the typo please?

From

`CollectionListPage`**Itme**`ClickedEvent`

To

`CollectionListPage`**Item**`ClickedEvent`

packages/tinacms/src/lib/posthog/posthogProvider.ts
joshbermanssw: 👍

packages/tinacms/src/admin/index.tsx
```diff
@@ -29,8 +29,75 @@ import CollectionUpdatePage from './pages/CollectionUpdatePage';
...
+    initializePostHog(telemetryMode).then((posthog) => {
+      if (posthog) {
+        const eventProperties: Record<string, string> = {
+          tinaCMSVersion: pkg.version,
+          system: 'tinacms/tinacms',
+          backendType: backendType,
+        };
+
+        // Include clientId for TinaCloud users when telemetry is enabled
+        if (backendType === 'tinacloud' && client?.clientId) {
```
18-th, line 89: This does not seem to check if the telemetry is enabled

packages/tinacms/src/lib/posthog/posthogProvider.ts
```diff
@@ -0,0 +1,180 @@
+import posthog, { PostHog } from 'posthog-js';
+import type { TelemetryMode } from '@tinacms/schema-tools';
+
+export type { TelemetryMode };
+
+let posthogClient: PostHog | null = null;
+let isInitialized = false;
+let isDisabled = false;
+let defaultTelemetryMode: TelemetryMode = 'anonymous';
```
18-th: This should be just `telemetryMode`. `default` implies a constant.

playwright/tina-playwright/tests/posthog/posthog.spec.ts
```diff
@@ -0,0 +1,47 @@
...
+  const telemetryMode = getTelemetryMode();
+  const isDisabled = telemetryMode === 'disabled';
+
+  console.log(
+    `Telemetry mode from config: ${telemetryMode ?? 'undefined (defaults to anonymous)'}`
+  );
+
+  await page.goto('http://localhost:3000/admin/index.html#/~');
+
+  if (isDisabled) {
```
18-th: We should not have if blocks inside the test. This makes it non-deterministic. You could use before and after hooks to set the config to what you need it to be and clean up, respectively.

joshbermanssw: As per our conversation, we have changed the test

packages/@tinacms/cli/src/next/commands/dev-command/index.ts
```diff
@@ -326,6 +326,14 @@ export class DevCommand extends BaseCommand {
...
     });
     await this.startSubCommand();
+
+    {
+      if (configManager?.config?.telemetry === 'anonymous') {
+        logger.info(
+          `\n📊 Note: TinaCMS now collects anonymous telemetry regarding usage. More information on TinaCMS Telemetry: https://tina.io/telemetry\n`
+        );
+      }
+    }
```
wicksipedia: I noticed that there were extra curlies around the `if` statement.

Should the telemetry notice be printed before the sub command so that it shows up during dev?

```suggestion
if (configManager?.config?.telemetry === 'anonymous') {
    logger.info(`\n📊 Note: TinaCMS now collects anonymous telemetry regarding usage. More information on TinaCMS Telemetry: https://tina.io/telemetry\n`);
}

await this.startSubCommand();
```

packages/@tinacms/schema-tools/src/types/index.ts
```diff
@@ -774,6 +781,18 @@ export interface Config<
...
+   * Configuration for telemetry.
+   *
+   * This allows you to configure how telemetry is collected and sent to the TinaCMS team.
+   * Default is "anonymous"
+   *
+   * @example
+   * telemetry: "anonymous" // Default value - anonymous telemetry
+   * telemetry: "disabled" // Disable telemetry completely
+   */
+  telemetry?: TelemetryMode;
```
wicksipedia, line 789: It looks like the definition of the modes is duplicated. Should this use `/** @inheritdoc */`? or just drop the examples since the `TelemetryMode` type describes what each mode should do

wicksipedia (approved):
couple of nitpicks but otherwise lgtm

PR #6440, opened 2026-02-18
Add path traversal protection

18-th:
Add path traversal guards across CLI and GraphQL packages to prevent directory escape attacks via media and file system bridge operations.

- Add assertPathWithinBase utility to validate resolved paths stay within their intended base directory (cli/src/utils/path.ts)

- Add path traversal checks to media list, delete, and upload routes in both Express and Vite dev server handlers

- Add path validation to filesystem bridge get, put, delete, and glob operations (isomorphic.ts, filesystem.ts)

- Add comprehensive test coverage for all security-critical paths including percent-encoded traversal payloads

https://github.com/tinacms/tinacms/security/advisories/GHSA-5hxf-c7j4-279c
https://github.com/tinacms/tinacms/security/advisories/GHSA-2f24-mg4x-534q

### DIFF
--- a/packages/@tinacms/cli/src/next/commands/dev-command/server/media.ts
+++ b/packages/@tinacms/cli/src/next/commands/dev-command/server/media.ts
@@ -17,6 +17,8 @@ export const createMediaRouter = (config: PathConfig) => {
   const handleList = async (req, res) => {
     try {
       const requestURL = new URL(req.url, config.apiURL);
+      // @security User-controlled path — decoded here, validated inside
+      // mediaModel.listMedia via resolveWithinBase.
       const folder = decodeURIComponent(
         requestURL.pathname.replace('/media/list/', '')
       );
@@ -40,6 +42,8 @@ export const createMediaRouter = (config: PathConfig) => {
 
   const handleDelete = async (req: Connect.IncomingMessage, res) => {
     try {
+      // @security User-controlled path — decoded here, validated inside
+      // mediaModel.deleteMedia via resolveStrictlyWithinBase.
       const file = decodeURIComponent(req.url.slice('/media/'.length));
       const didDelete = await mediaModel.deleteMedia({ searchPath: file });
       res.end(JSON.stringify(didDelete));
@@ -61,6 +65,8 @@ export const createMediaRouter = (config: PathConfig) => {
     let responded = false;
 
     bb.on('file', async (_name, file, _info) => {
+      // @security User-controlled path — decoded here, validated immediately
+      // below via resolveStrictlyWithinBase.
       const fullPath = decodeURIComponent(
         req.url?.slice('/media/upload/'.length)
       );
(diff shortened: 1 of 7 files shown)


wicksipedia (dismissed):
lgtm, can we get this tagged and tested on some of the internal sites before publishing live

packages/@tinacms/cli/src/next/commands/dev-command/server/media.ts
```diff
@@ -105,6 +144,66 @@ export interface PathConfig {
...
+ *
+ * Matches (case-insensitive):
+ *   %2e%2e → ..  (double-dot – directory traversal)
+ *   %2f    → /   (forward slash)
+ *   %5c    → \   (backslash – Windows separator)
+ *
+ * A single %2e (encoded dot) is NOT matched — it is harmless and may
+ * appear in legitimate filenames or dotfile paths.
+ */
+const ENCODED_TRAVERSAL_RE = /%2e%2e|%2f|%5c/i;
```
JackDevAU, line 172: I noticed this is added twice - can we share this logic?

packages/@tinacms/cli/src/utils/path.ts
```diff
@@ -13,3 +13,60 @@ export function stripNativeTrailingSlash(p: string) {
...
+ *
+ * Matches (case-insensitive):
+ *   %2e%2e → ..  (double-dot – directory traversal)
+ *   %2f    → /   (forward slash)
+ *   %5c    → \   (backslash – Windows separator)
+ *
+ * A single %2e (encoded dot) is NOT matched — it is harmless and may
+ * appear in legitimate filenames or dotfile paths.
+ */
+const ENCODED_TRAVERSAL_RE = /%2e%2e|%2f|%5c/i;
```
JackDevAU, line 84: again do we need this again.

~~Also i thought the server is only used in dev - are we ensuring this is working on prod?~~
disregard that

packages/@tinacms/cli/src/next/commands/dev-command/server/media.ts
18-th: The duplication is intentional due to the way CodeQL does the analysis. It can't crawl onto the other package, so I had to inline some of those calls.

packages/@tinacms/cli/src/utils/path.ts
18-th: Same as above

packages/@tinacms/cli/src/next/commands/dev-command/server/media.ts
JackDevAU: Should we make a note of this somewhere in the code or docs?

18-th: Done

Issue #6421, opened 2026-02-13
🐖 PostHog Event Capture - branch-switched


joshbermanssw:
Capturing events on the branch selector modal 

Capture events: 
- BranchSwitchedEvent
- BranchSwitcherOpenedEvent
- BranchSwitcherSearchEvent
- BranchSwitcherDropDownEvent
- BranchSwitcherPREvent

joshbermanssw:
✅ Done - 

Created events for:
- BranchSwitchedEvent
- BranchSwitcherOpenedEvent
- BranchSwitcherSearchEvent
- BranchSwitcherDropDownEvent
- BranchSwitcherPREvent

<img width="1536" height="185" alt="Screenshot 2026-02-20 at 4 48 52 pm" src="https://github.com/user-attachments/assets/5be7c309-322a-4c43-b187-c29b82b7834a" />

**Figure: Events firing on PostHog**

Also added a series of unit tests with @18-th

PR #6446, opened 2026-02-20
🐖 PostHog - Add Telemetry around Branch Selector Modal 

joshbermanssw:
As per #6421 

Created events for:
- BranchSwitchedEvent
- BranchSwitcherOpenedEvent
- BranchSwitcherSearchEvent
- BranchSwitcherDropDownEvent
- BranchSwitcherPREvent

Screenshot 2026-02-20 at 4 48 52 pm

**Figure: Events firing on PostHog**

### DIFF
--- a/packages/tinacms/src/toolkit/plugin-branch-switcher/branch-button.test.tsx
+++ b/packages/tinacms/src/toolkit/plugin-branch-switcher/branch-button.test.tsx
@@ -3,60 +3,61 @@ import userEvent from '@testing-library/user-event';
 import * as React from 'react';
 import { beforeEach, describe, expect, it, vi } from 'vitest';
 
-
 const mockCaptureEvent = vi.fn();
 vi.mock('../../lib/posthog/posthogProvider', () => ({
-    captureEvent: (...args: any[]) => mockCaptureEvent(...args),
+  captureEvent: (...args: any[]) => mockCaptureEvent(...args),
 }));
 
 vi.mock('@utils/cn', () => ({
-    cn: (...args: any[]) => args.join(' '),
+  cn: (...args: any[]) => args.join(' '),
 }));
 
 vi.mock('../../lib/posthog/posthog', async () => {
-    return {
-      BranchSwitcherOpenedEvent: 'branch-switcher-opened',
-    };
+  return {
+    BranchSwitcherOpenedEvent: 'branch-switcher-opened',
+  };
 });
 
 vi.mock('./branch-data', () => ({
-    useBranchData: () => ({
-        currentBranch: 'main',
-        setCurrentBranch: vi.fn(),
-    }),
+  useBranchData: () => ({
+    currentBranch: 'main',
+    setCurrentBranch: vi.fn(),
+  }),
 }));
 
 vi.mock('@toolkit/react-tinacms', () => ({
-    useCMS: () => ({
-      flags: { get: (key: string) => key === 'branch-switcher' },
-      api: { tina: { usingProtectedBranch: () => false } },
-    }),
-  }));
+  useCMS: () => ({
+    flags: { get: (key: string) => key === 'branch-switcher' },
+    api: { tina: { usingProtectedBranch: () => false } },
+  }),
+}));
 
-  vi.mock('./branch-modal', () => ({
-    BranchModal: ({ close }: { close: () => void }) => <div data-testid="branch-modal" />,
-  }));
+vi.mock('./branch-modal', () => ({
+  BranchModal: ({ close }: { close: () => void }) => (
+    <div data-testid='branch-modal' />
+  ),
+}));
 
 import { BranchButton } from './branch-button';
 
 function renderBranchButton() {
-    const user = userEvent.setup();
-    return {
-      user,
-      ...render(<BranchButton />),
-    };
-  }
+  const user = userEvent.setup();
+  return {
+    user,
+    ...render(<BranchButton />),
+  };
+}
 
-  describe('BranchButton telemetry', () => {
-    beforeEach(() => {
-      mockCaptureEvent.mockClear();
-    });
-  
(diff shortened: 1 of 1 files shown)


packages/tinacms/src/toolkit/plugin-branch-switcher/branch-selector-table.tsx
```diff
@@ -39,6 +39,14 @@ import {
 import { ArrowUp, ArrowDown, ArrowUpDown } from 'lucide-react';
 import { cn } from '@utils/cn';
 import { Badge } from '@toolkit/react-sidebar/components/badge';
+import { captureEvent } from '../../lib/posthog/posthogProvider';
+import { BranchSwitcherSearchEvent } from '../../lib/posthog/posthog';
```
18-th: Could you move all PostHog imports into the same import block, please?

packages/tinacms/src/toolkit/plugin-branch-switcher/branch-selector-table.tsx
```diff
@@ -269,7 +277,15 @@ export default function BranchSelectorTable({
             <BaseTextField
               placeholder='Branch name or PR #'
               value={search}
-              onChange={(e) => setSearch(e.target.value)}
+              onChange={(e) => {
+                if (e.target.value && !searchEventFired) {
+                  searchEventFired = true;
+                  captureEvent(BranchSwitcherSearchEvent, {
+                    searchQuery: e.target.value,
```
18-th: This effectively captures just one letter. If the only thing you want to capture is that the search was interacted with, you can safely drop that part of the payload.

packages/tinacms/src/toolkit/plugin-branch-switcher/branch-selector-table.tsx
```diff
@@ -39,6 +39,14 @@ import {
 import { ArrowUp, ArrowDown, ArrowUpDown } from 'lucide-react';
 import { cn } from '@utils/cn';
 import { Badge } from '@toolkit/react-sidebar/components/badge';
+import { captureEvent } from '../../lib/posthog/posthogProvider';
+import { BranchSwitcherSearchEvent } from '../../lib/posthog/posthog';
+
+let searchEventFired = false;
```
18-th: Please put the variable declaration below the imports.

Also, in the way this is declared and never reset, consecutive opening of the modal won't trigger the search event again. So it is effectively disabled until the page reloads.

Could you double-check this behaviour for me, please?

packages/tinacms/src/auth/TinaCloudProvider.tsx
```diff
@@ -347,6 +349,9 @@ export const TinaCloudProvider = (
     cms.registerApi('tina', createClient({ ...props, branch: currentBranch }));
   } else {
     cms.api.tina.setBranch(currentBranch);
+    captureEvent(BranchSwitchedEvent, {
```
18-th: This will fire on every re-render. This will also fire when the page loads initially, which would give us false telemetry as the branch hasn't actually changed.

You could use something like this instead:

```js
const previousBranchRef = React.useRef(currentBranch);
  useEffect(() => {
    if (previousBranchRef.current !== currentBranch) {
      captureEvent(BranchSwitchedEvent, {
        branchSwitchedTo: currentBranch,
      });
      previousBranchRef.current = currentBranch;
    }
  }, [currentBranch]);
```

packages/tinacms/src/toolkit/plugin-branch-switcher/branch-selector-table.tsx
```diff
@@ -298,7 +314,10 @@ export default function BranchSelectorTable({
               id: 'branch-type',
               name: 'branch-type',
               value: filter,
-              onChange: (e: any) => setFilter(e.target.value),
+              onChange: (e: any) => {
+                setFilter(e.target.value);
+                captureEvent(BranchSwitcherDropDownEvent, {});
```
18-th: We might as well capture what was set in the dropdown as part of this

packages/tinacms/src/toolkit/plugin-branch-switcher/branch-selector-table.tsx
joshbermanssw: You're right. Resolved

packages/tinacms/src/toolkit/plugin-branch-switcher/branch-selector-table.tsx
joshbermanssw: You're correct. Resolved

packages/tinacms/src/auth/TinaCloudProvider.tsx
joshbermanssw: Applied

packages/tinacms/src/toolkit/plugin-branch-switcher/branch-selector-table.tsx
joshbermanssw: ✅

packages/tinacms/src/toolkit/plugin-branch-switcher/branch-selector-table.tsx
joshbermanssw: There are only 2 options. "All" and "Content" - I dont see much value in this

18-th: The value is that we can see how often users change from the default option to the alternative, and whether they change back to the default after that. If we see a bunch of users switching to all branches often, we can, for example, make it remember the last selected value to reduce friction for them

joshbermanssw: ✅ Done - added 

Screenshot 2026-02-23 at 11 16 46 am

**Figure: Option showing up in PostHog**

18-th (approved):
LGTM

PR #6480, opened 2026-03-04
Add TinaCMS Default Favicon

joshbermanssw:
As per https://github.com/tinacms/tinacms/issues/6481

Added TinaCMS default FavIcon instead of none 

Screenshot 2026-03-04 at 2 10 47 pm


18-th (approved):
LGTM

Issue #6486, opened 2026-03-04
🐛 Bug - Floating toolbar shows despite showFloatingToolbar config


joshbermanssw:
Requested by: @joshbermanssw via [YakShaver.ai](https://YakShaver.ai) 🦬
cc: @joshbermanssw, @kulesy, @JackDevAU

Hi Team!

&nbsp;&nbsp;[🟥&nbsp;&nbsp;Watch the video (49 sec)](https://www.youtube.com/watch?v=B0UHEihJC6s)

### Pain
A user reports that the TinaCMS floating toolbar remains visible even when the `showFloatingToolbar` configuration is set to disable it. The user has shared their actual configuration and a screenshot from the React Developer Tools showing that the `ToolbarProvider` still renders the floating toolbar. This indicates the configuration option may be ignored, overridden, or not correctly wired.

### Acceptance Criteria
1. Given `showFloatingToolbar` is set to false, the floating toolbar is not rendered in the UI.
2. The `ToolbarProvider` respects the `showFloatingToolbar` configuration consistently.
3. The behavior is verified using the provided user configuration or an equivalent setup.
4. Documentation or examples are updated if a specific setup or additional config is required.

### Reproduce Steps
1. Set up a TinaCMS project using the user's provided configuration (or equivalent).
2. Configure TinaCMS with `showFloatingToolbar: false`.
3. Run the application in development mode.
4. Observe the UI and confirm the floating toolbar is still visible.
5. Inspect React Developer Tools to verify the `ToolbarProvider` is rendering the floating toolbar.

### Screenshot
![Screenshot](https://sayakshaverproduction.blob.core.windows.net/images/images/9be28b56-f23e-4774-8f9e-3a87ce80eded_frame_33.png?sv=2025-05-05&se=2046-03-04T22%3A14%3A30Z&sr=b&sp=rd&sig=JGgmYO5BotFGZWqaZ%2B7lVhvnLb65X%2FnXauxO1nNzCM4%3D)
**Figure: Floating toolbar still appears despite configuration**

joshbermanssw:
Previously we were conditionally rendering the toolbar obj via conditional jsx AND the plate plugin (FloatingToolbarPlugin), however we were only applying the condition on the jsx obj and not the plugin itself. 

Even when I commented out the JSX showing the floatingtoolbar, we still see it rendered 

[...]

PR #6487, opened 2026-03-04
🐛 Fix FloatingToolbarPlugin config 

joshbermanssw:
As per #6486 and https://github.com/tinacms/tinacms/discussions/6484

Previously we were conditionally rendering the toolbar obj via conditional jsx AND the plate plugin (FloatingToolbarPlugin), however we were only applying the condition on the jsx obj and not the plugin itself. 

Even when I commented out the JSX showing the floatingtoolbar, we still see it rendered 

Screenshot 2026-03-05 at 9 05 26 am

This PR pops the FloatingToolbarPlugin correctly from the plate plugin array 

https://github.com/user-attachments/assets/f86caed5-fb09-46cb-9e79-3d9702a4886e

### DIFF
--- a/.changeset/soft-points-travel.md
+++ b/.changeset/soft-points-travel.md
@@ -1,0 +1,5 @@
+---
+"tinacms": patch
+---
+
+🐛 Fix showFloatingToolbar config not working by removing plugin correctly


Taiwonator:
Thanks for fixing this! Had just made a local patch for this, silly me 🙃

ifsimicoded:
@joshbermanssw @kulesy Any idea when the next release is scheduled? Would love to have this fix!

joshbermanssw:
Hey @ifsimicoded - we plan to have one today 🤞

PR #6509, opened 2026-03-12
Poc   build and bundle testing

Ben0189:
This PR improve package validation by cleaning up stale export references, retaining build artifact verification, add publint, and adding the combined package validation step to CI.


Ben0189:
As per conversation with @18-th here are some suggestion before we merge:

- [x] Check the git history to confirm whether css-transform and edit-slate were left there intentionally

image

**Figure: The git history shows no tracked css-transform file, so this looks like stale build metadata rather than an intentionally preserved source file.**

image

image

**Figure: This shows edit-state had already been removed from source, so deleting ./dist/edit-state from exports is metadata cleanup for that source file removed**


- [x] Look into the build script to see if it can actually catch all the export function to avoid any miss export function problem happen to the user

**So the build pipeline protects:**
✅ internal references inside this monorepo

**It does not automatically protect:**
❌ external consumer usage that lives outside this repo

**Recommendation:** We probably do not need that test for this PR/branch. The current change is about invalid file-level exports, and the existing verification tests are the right level for that. If later we wants stronger public API stability guarantees, that can be another pr for that.

- [x] Investigate if there is any tools exist which can replace the written test in this PR

As per investigation we can replaces the custom publish verification test with `publint`, while keeping the existing build verification test for local build artifact checks.

scripts/run-publint.mjs
```diff
@@ -0,0 +1,36 @@
+import fs from 'node:fs'
```
18-th: I strongly suspect that this entire script can just be an action in a package.json like so

```
"lintpackage": "pnpm --filter './packages/**' --no-bail exec publint --strict"
```

Ben0189: Done, good suggestion

18-th (approved):
LGTM

PR #6529, opened 2026-03-21
Fix containerized dev by adding configurable dev server URL for Tina admin

travisp:
## Summary

Adds a dev-only server.url config option so Tina admin can load correctly when tinacms dev is accessed through a non-localhost domain.

This updates the dev admin asset URLs and local GraphQL URL generation to use the configured dev server URL instead of hardcoded localhost

## Why

Tina currently hardcodes http://localhost:${port} when generating dev admin asset URLs, which breaks containerized, proxied, remote, and custom-hostname dev setups. This PR removes that localhost-only assumption so admin assets load from the actual host serving the admin page.

The current behavior causes the admin HTML to reference:

 - http://localhost:${port}/@react-refresh
 - http://localhost:${port}/@vite/client
 - http://localhost:${port}/src/main.tsx

If the admin is opened from a host other than localhost, those asset URLs are wrong and the admin fails to load.

## Changes

 - add server.url to the config types
 - use server.url when generating the dev admin HTML
 - use server.url for the local GraphQL apiURL
 - update the init template comments to show the new option
 - add tests for:
     - dev admin asset URL generation
     - local API URL generation
     
Example:

 ```ts
   server: {
     url: 'https://mycontainer.test',
     allowedOrigins: ['https://mycontainer.test'],
   }
 ```

 ## Notes

 - this is scoped to tinacms dev
 - default behavior is unchanged when server.url is not set
 - this does not change production behavior

This code was supported with the use of GPT 5.4, but manually edited, tested, and reviewed.


kulesy:
Hey Travis, sorry this sat for so long. I have been through it properly now and it is a keep. The approach is right, it merges onto main cleanly, and the whole CLI suite passes. The host-check test that boots a real Vite server is nicer than what usually lands here.

One thing blocks it. If you set `server.url` on its own, the admin throws "Invalid setup" and never mounts, so the option breaks the thing it is meant to fix. The admin decides whether it is in local mode by looking for the string "localhost" in the API URL it gets handed, so any other hostname reads as self-hosted and it bails out. A hostname that happens to contain "localhost" still works, which is probably why it looked fine when you tested it. The fix is in `packages/tinacms` rather than in this diff, and it would need a changeset entry too.

Second one: the `server.url` validation needs tightening. A value written without the protocol but with a port gets past it and produces a nonsense origin downstream, and there is a security angle to where that origin ends up, so I would rather not spell it out here. Happy to send you the detail.

Do you want to pick this back up, or would you rather we finish it off? Either is fine by us.

Issue #6515, opened 2026-03-17
🤖♻️ Kitchen-Sink Example (Astro)


KahaMason:
CC @wicksipedia @Ben0189 @18-th @JackDevAU @joshbermanssw 

Hey Team,

As per my conversation during the Sprint 97 review, the new kitchen-sink app has become a large body of work and has warranted it's own sub-issue to help improve the progress visibility in future sprints.

The criteria of this issue is as follows:
- [x] Recreate the Kitchen Sink app in Astro 5.x
- [x] Scaffold - Astro Project + Tina Config
- [x] Author Collection + Author Pages
- [x] Global Collection + Base Layout (Header, Footer, Theme)
- [x] Post Collection + Post Pages
- [x] Blog Collection + Blog Pages
- [x] Page Collection + Base Block Rendering
- [x] TinaCMS Visual Editing + Polish
- [x] Update the README.md and any relating docs.
- [x] Reviewed, approved and merged

KahaMason:
✅ Done

<img width="1919" height="1033" alt="Image" src="https://github.com/user-attachments/assets/34e0e352-1c6c-49b6-8cc4-13748a0dab70" />

**Figure: Astro Kitchen Sink - Landing Page**

<img width="1919" height="1032" alt="Image" src="https://github.com/user-attachments/assets/70088347-a819-452c-8410-52b5febb151a" />


**Figure: Astro Kitchen Sink - Posts Page**

<img width="1919" height="1029" alt="Image" src="https://github.com/user-attachments/assets/718f3ce6-df37-4447-a713-24e452487236" />


**Figure: Astro Kitchen Sink - Blogs Page**

<img width="1919" height="1030" alt="Image" src="https://github.com/user-attachments/assets/0bdac1bf-79d4-4867-91bc-4df1c29d9e0c" />


**Figure: Astro Kitchen Sink - TinaCMS Admin Editor**

PR #6534, opened 2026-03-23
🤖 Astro Kitchen Sink - TinaCMS config + Tag collection

KahaMason:
This PR partially addresses #6515

Stage 2: Tina Integration

- ✅ Added TinaCMS dependencies and config
- ✅ Created Tag collection with sample content
- ✅ Set up Biome linting (extends root config)
- ✅ Added cross-env for Windows dev support
- ✅ Updated Claude rules with data fetching pattern

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -743,9 +743,6 @@ importers:
       graphql:
         specifier: ^15.10.1
         version: 15.10.1
-      mongodb-level:
-        specifier: ^0.0.4
-        version: 0.0.4(mongodb@4.17.2)
       react:
         specifier: ^18.3.1
         version: 18.3.1


examples/astro/kitchen-sink/package.json
```diff
@@ -5,20 +5,35 @@
...
   },
   "dependencies": {
-    "astro": "^5.16.0",
-    "@astrojs/react": "^4.4.0",
     "@astrojs/mdx": "^4.3.0",
+    "@astrojs/react": "^4.4.0",
+    "@tinacms/datalayer": "workspace:*",
+    "astro": "^5.16.0",
+    "graphql": "^15.10.1",
+    "mongodb-level": "^0.0.4",
```
18-th: We only need the level packages for the self-hosted setups. Could we remove that?

KahaMason: Removed. 🙂

PR #6545, opened 2026-03-25
Fix path traversal bypass via backslash sequences

wicksipedia:
Fixes [GHSA-v9p7-gf3q-h779](https://github.com/tinacms/tinacms/security/advisories/GHSA-v9p7-gf3q-h779)

On POSIX systems, path.normalize() treats backslashes as literal characters, not directory separators. This allowed attackers to bypass the existing path traversal checks using paths like x\..\..\package.json — the validation saw no traversal, but downstream path.join()/fs operations could resolve the backslashes as separators.

### DIFF
--- a/.changeset/empty-teachers-itch.md
+++ b/.changeset/empty-teachers-itch.md
@@ -1,0 +1,5 @@
+---
+"tinacms": patch
+---
+
+Flipping the referenced files checkbox to off by default, and updating the label.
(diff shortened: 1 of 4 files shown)


18-th:
Tests missing if you want to be thorough:

Filesystem bridge:
* Get backslash traversal
* Get null byte
* Put backslash traversal
* Put null byte
* Delete backslash traversal
* Delete null byte
* Glob backslash traversal
* Glob null byte

Isomorphic bridge:
* Glob backslash traversal
* Glob null byte test
* Delete null byte test

Issue #6540, opened 2026-03-24
🐛 Bug - Rich text slash command menu cannot be dismissed with Escape


joshbermanssw:
Requested by: @joshbermanssw via [YakShaver.ai](https://YakShaver.ai) 🦬
cc: @joshbermanssw, @kulesy, @JackDevAU

Hi Team!

&nbsp;&nbsp;[🟥&nbsp;&nbsp;Watch the video (1 min 0 sec)](https://www.youtube.com/watch?v=XXLW0PDdAzM)

&nbsp;&nbsp;Url from screen share: [https://localhost](https://localhost)

### Pain
In the TinaCMS rich text editor, the slash command menu cannot be fully dismissed when a slash ("/") is present in the editor. Pressing the Escape key briefly closes the menu, but it immediately reopens, leaving the user with no way to exit the menu unless the slash character is removed. This results in a poor editing experience and prevents normal interaction with the editor.

### Acceptance Criteria
1. Pressing the Escape key closes the slash command menu and it remains closed.
2. After closing the slash command menu, it does not automatically reopen unless the user explicitly types a new slash.
3. The user can click outside the slash command menu to dismiss it.
4. The editor correctly tracks the open/closed state of the slash command menu.
5. No regression to existing slash command functionality.

### Reproduce Steps
1. Open a page in TinaCMS that uses the rich text editor.
2. Place the cursor inside the editor.
3. Type a forward slash ("/") to open the slash command menu.
4. Press the Escape key.
5. Observe that the menu briefly disappears and then immediately reopens.

### Screenshot
![Screenshot](https://sayakshaverproduction.blob.core.windows.net/images/images/c9c0138b-97f9-481f-95d9-570780a00abf_frame_27.png?sv=2025-05-05&se=2046-03-24T21%3A26%3A58Z&sr=b&sp=rd&sig=us0RpUzP6j6kYJ6T2nf5zYoB5hTc4n4oNpj4%2BkTcXXA%3D)
**Figure: Slash command menu reopens after Escape instead of closing**

joshbermanssw:
✅ Done

PR #6546, opened 2026-03-25
🐛 Fix slash command not being closable 

joshbermanssw:
As per #6540

The Inline Combo Box was closable via hacky methods or deleting the forward slash entirely. 

The / char is stored in a special node (SlashInputElement) in the editor. The insertPoint was being tracked via a state, so when the callback ran to add a pos it was reading an old state. Changing to useRef means callback reads the pos at the moment it runs, not the moment the component was last rendered. 

`useComboboxInput` was already trying to handle these cases for us instead of manually doing it in the onCancelInput 

The first method I tried (when we were keeping useState) was to add 

```
onCancelInput: (cause) => {
	...
      if (cause === 'escape') {
		replace node at position with a node {type: 'p', children: { trigger + value } } 
        });
      }
    },
```
Whilst this worked, it introduced console errors due to the stale position being iterated over

### DIFF
--- a/.changeset/small-items-dance.md
+++ b/.changeset/small-items-dance.md
@@ -1,0 +1,5 @@
+---
+"tinacms": patch
+---
+
+🐛 Fix slash command popup from not being escapable


isaaclombardssw (approved):
✅ Niiiice

PR #6552, opened 2026-03-25
Fix symlink path traversal bypass in media endpoints and FilesystemBridge

wicksipedia:
## Summary

- All path validation in TinaCMS used lexical string checks (`path.resolve` + `startsWith`) to prevent directory traversal. These checks don't follow symlinks — a symlink inside an allowed directory pointing outside it would pass validation while the actual filesystem operation targets a location outside the root.
- After the existing lexical check passes, path validation now resolves symlinks via `fs.realpathSync` and re-validates that the real filesystem target stays within the allowed directory. This applies to both media operations (list, upload, delete) and content operations via `FilesystemBridge` (get, put, delete, glob).
- For write operations where the target file doesn't exist yet, the check walks up to the nearest existing ancestor directory, resolves its real path, then verifies containment — so symlink escapes are caught even before the file is created.
- The dev server now warns at startup when the host is configured to listen on a non-localhost address, since it has no authentication and is not intended to be network-exposed.

## Security Advisories

- [GHSA-g87c-r2jp-293w](https://github.com/tinacms/tinacms/security/advisories/GHSA-g87c-r2jp-293w) — Media endpoints can escape the media root via symlinks or junctions
- [GHSA-g9c2-gf25-3x67](https://github.com/tinacms/tinacms/security/advisories/GHSA-g9c2-gf25-3x67) — FilesystemBridge path validation can be bypassed via symlinks or junctions

## Test plan

- [x] New symlink traversal tests across all validation sites (path utils, Vite media model, Express media model, FilesystemBridge) — create real symlinks in temp dirs pointing outside the base, assert operations are rejected
- [x] Positive test: symlinks that point within the base directory are still allowed
- [x] All existing path traversal and media tests continue to pass
- [x] `isHostExposed` utility tested for localhost, 0.0.0.0, LAN IPs, and default values
- [x] Manual: run dev server with `build.host: true` in tina config and verify the network exposure warning appears

### DIFF
--- a/packages/@tinacms/graphql/src/database/bridge/filesystem.ts
+++ b/packages/@tinacms/graphql/src/database/bridge/filesystem.ts
@@ -5,21 +5,6 @@ import normalize from 'normalize-path';
 import type { Bridge } from './index';
 
 /**
- * Defense-in-depth: validates that a filepath stays within a base directory.
- * This protects against CWE-22 (Path Traversal) even if callers fail to
- * sanitize user input before calling bridge methods.
- *
- * @security This is a local copy of the validation pattern. The canonical
- * reference is in `@tinacms/cli/src/utils/path.ts`. Unlike the media-model
- * copies, this one does NOT include the URL-encoded safety net because
- * bridge paths come from GraphQL, not raw HTTP URLs.
- *
- * @param filepath - The path to validate (relative to baseDir).
- * @param baseDir  - The trusted root directory.
- * @returns The resolved absolute path.
- * @throws {Error} If the path escapes the base directory.
- */
-/**
  * Follows symlinks to determine where a path actually points on disk.
  *
  * If the full path exists, returns its `fs.realpathSync` result. If it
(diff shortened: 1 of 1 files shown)


packages/@tinacms/graphql/src/database/bridge/filesystem.ts
```diff
@@ -19,6 +19,29 @@ import type { Bridge } from './index';
  * @returns The resolved absolute path.
  * @throws {Error} If the path escapes the base directory.
  */
+/**
```
18-th, line 7: Two docs are now glued together. Could you please move the doc above this to its correct place?

packages/@tinacms/cli/src/server/models/media.test.ts
```diff
@@ -58,6 +58,44 @@ describe('MediaModel (Express server)', () => {
...
+      await fs.writeFile(path.join(outsideDir, 'secret.txt'), 'sensitive');
+      const mediaDir = path.join(tmpDir, 'public', 'uploads');
+      await fs.symlink(outsideDir, path.join(mediaDir, 'escape'));
+
+      const model = new MediaModel(config);
+      await expect(model.listMedia({ searchPath: 'escape' })).rejects.toThrow(
+        PathTraversalError
+      );
+
+      await fs.remove(outsideDir);
```
18-th: Could we move the cleanup to the after hook, so that if this fails, the directory and the file are still cleaned up?

packages/@tinacms/cli/src/server/models/media.test.ts
```diff
@@ -58,6 +58,44 @@ describe('MediaModel (Express server)', () => {
...
+      await fs.writeFile(path.join(outsideDir, 'target.txt'), 'data');
+      const mediaDir = path.join(tmpDir, 'public', 'uploads');
+      await fs.symlink(outsideDir, path.join(mediaDir, 'escape'));
+
+      const model = new MediaModel(config);
+      await expect(
+        model.deleteMedia({ searchPath: 'escape/target.txt' })
+      ).rejects.toThrow(PathTraversalError);
+
+      await fs.remove(outsideDir);
```
18-th: Same with this one

packages/@tinacms/cli/src/server/models/media.test.ts
wicksipedia: ✅ done

packages/@tinacms/cli/src/server/models/media.test.ts
wicksipedia: ✅ done

PR #6604, opened 2026-03-30
Feat/create tina app theme flag

Ben0189:
The `tina-docs` template prompts the user to select a theme interactively.  In CI there is no interactive terminal, so the prompt hangs with no answer,  `create-tina-app` exits early, and the `template-repo` folder is never created — 
causing the build step to fail with `cd: template-repo: No such file or directory`.

### Changes

- `options.ts` — register `--theme` as a valid CLI flag
- `index.ts` — use `--theme` value if provided, otherwise fall back to interactive prompt
- `build-starter-templates.yml` — pass `--theme default` when testing `tina-docs` in CI

### DIFF
--- a/.github/workflows/build-starter-templates.yml
+++ b/.github/workflows/build-starter-templates.yml
@@ -72,7 +72,7 @@ jobs:
 
       - name: Run create-tina-app
         id: create-tina-app
-        run: npx create-tina-app@latest template-repo --pkg-manager ${{ matrix.package-manager }} --template ${{ matrix.template }} --noTelemetry${{ matrix.template == 'tina-docs' && ' --theme default' || '' }}
+        run: npx create-tina-app@latest template-repo --pkg-manager ${{ matrix.package-manager }} --template ${{ matrix.template }} --noTelemetry ${{ matrix.template == 'tina-docs' && '--theme default' || '' }}
 
       - name: Run build
         id: build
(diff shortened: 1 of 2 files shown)


packages/create-tina-app/src/util/options.ts
```diff
@@ -11,6 +11,7 @@ export interface CreateOptions {
   noTelemetry: boolean;
   projectName: string;
   verbose: boolean;
+  theme: string;
```
joshbermanssw:
```suggestion
  theme?: string;
```

.github/workflows/build-starter-templates.yml
```diff
@@ -72,7 +72,7 @@ jobs:
 
       - name: Run create-tina-app
         id: create-tina-app
-        run: npx create-tina-app@latest template-repo --pkg-manager ${{ matrix.package-manager }} --template ${{ matrix.template }} --noTelemetry
+        run: npx create-tina-app@latest template-repo --pkg-manager ${{ matrix.package-manager }} --template ${{ matrix.template }} --noTelemetry${{ matrix.template == 'tina-docs' && ' --theme default' || '' }}
```
18-th: Can we add a space after --noTelemetry to be consistent with other arguments here?

Ben0189: Done add added space

packages/create-tina-app/src/util/options.ts
Ben0189: Done

joshbermanssw (approved):

Issue #6557, opened 2026-03-26
🐖 PostHog - Add Telemetry around Media Dashboard Usage


joshbermanssw:
🩸 Pain - we just implemented a new UI via the Media dashboard, how are we going to know if users are using it? 

🥅 Goal - Add telemetry around the usage of the media dashboard so we can understand how users are interacting with it 

**Tasks**
- [ ] Define necessary events needed to track 
- [ ] Implement 
- [ ] Create relevant dashboard in PostHog

thomas-sepanosian:
> Define necessary events needed to track

## Proposed events (🤖)

### `media-usage-dashboard-opened`
- `backendType`: The backend mode, such as `local`, `tinacloud`, `self-hosted`, or `other`. For now this will effectively be `local`, since the feature is only supported in local mode.

### `media-usage-scan-started`
- `trigger`: What initiated the scan, such as `initial` or `refresh`.
- `backendType`: The backend mode, such as `local`, `tinacloud`, `self-hosted`, or `other`.

### `media-usage-scan-completed`
- `trigger`: What initiated the scan, such as `initial` or `refresh`.
- `backendType`: The backend mode, such as `local`, `tinacloud`, `self-hosted`, or `other`.
- `scanDurationMs`: Total scan time in milliseconds.
- `totalMediaFilesCount`: Number of media files discovered.
- `usedCount`: Number of files referenced by at least one document.
- `unusedCount`: Number of files with no references.
- `collectionCount`: Number of collections scanned.
- `documentCount`: Number of documents scanned.

### `media-usage-scan-failed`
- `trigger`: What initiated the scan, such as `initial` or `refresh`.
- `backendType`: The backend mode, such as `local`, `tinacloud`, `self-hosted`, or `other`.
- `errorType`: A normalized error category rather than a raw message.
- `errorPhase`: The stage of the scan that failed, such as `list-media`, `list-documents`, `parse-document`, or `unknown`.

### `media-usage-filter-changed`
- `filterName`: Which filter changed, such as `type` or `usage`.
- `filterValue`: The selected value, such as `all`, `image`, `video`, `other`, `used`, or `unused`.
- `resultCount`: Number of rows matching after the filter change.

### `media-usage-sort-changed`
- `sortKey`: Which column was sorted, such as `filename`, `directory`, or `usage`.
- `direction`: Sort direction, `asc`, `desc`, or `none`.
- `resultCount`: Number of rows visible under the current filtered state.

[...]

PR #6607, opened 2026-03-30
🐖 Add Telemetry Around the Media Usage Dashboard

joshbermanssw:
As per #6557

Adding a series of posthog events around the media usage dashboard so we can determine and track how users are interacting with it 

Screenshot 2026-03-31 at 10 08 42 am

**Figure: ✅ Events showing on PostHog**


joshbermanssw:
Tagged and tested on https://tina-io-git-jb-upd-tina-lock-22-tinacms.vercel.app/admin#/~

PR #6639, opened 2026-04-07
implement test for auth packages

Ben0189:
Add test for tinacms auth packages


packages/@tinacms/auth/src/index.test.ts
```diff
@@ -0,0 +1,130 @@
...
+  id: 'user-123',
+  email: 'test@example.com',
+  verified: true,
+  role: 'admin' as const,
+  enabled: true,
+  fullName: 'Test User',
+};
+
+// Replace the global fetch with a Jest mock before each test
+let fetchMock: jest.Mock;
```
18-th: We could use a fetchSpy instead, like so

`let fetchSpy: jest.SpyInstance`

Ben0189: Done - refactor using the spy instance

PR #6647, opened 2026-04-09
Remove node 20 from starter tests

joshbermanssw:
Node 20 is coming to EoL this month, we've seen a lot dependencies already start removing it's support. 

Due to this, some of our starter package testing is failing. Instead of bandage solutions; since node 20 is EoL lets remove


.github/workflows/build-starter-templates.yml
```diff
@@ -20,7 +20,7 @@ jobs:
           - tina-docs
           - tinasaurus
           - basic
-        node-version: [20, 22, 24]
+        node-version: [22, 24]
```
JackDevAU:
```suggestion
        node-version: [22, 24, 25]
```

18-th (approved):
LGTM

PR #6652, opened 2026-04-10
Fix HTML code block language tag from 'htmlbars' to 'html'

kulesy:
## Summary

Fixes incorrect language identifier for HTML code blocks in the rich-text editor. The correct identifier for HTML syntax highlighting is `html`, not `htmlbars`.

## Changes

- Changed line 78 in `code-block-combobox.tsx` from `{ label: 'HTML', value: 'htmlbars' }` to `{ label: 'HTML', value: 'html' }`

## Context

This addresses one of the issues reported in discussion #6651.

**Test with reproduction repository:** https://github.com/nkuehn/reproduce-tina-incompatible-mdx

## Related Issues

- Fixes issue #2 from discussion #6651


packages/tinacms/src/toolkit/fields/plugins/mdx-field-plugin/plate/components/plate-ui/code-block-combobox.tsx
```diff
@@ -75,7 +75,7 @@ const languages: { label: string; value: string }[] = [
   { label: 'Groovy', value: 'groovy' },
   { label: 'Handlebars', value: 'handlebars' },
   { label: 'Haskell', value: 'haskell' },
-  { label: 'HTML', value: 'htmlbars' },
+  { label: 'HTML', value: 'html' },
```
18-th, line 78: Do we want to add another entry here to keep supporting HTML Bars in case the existing user base relies on that?

Switching this to plain HTML with no templating might break stuff for people relying on it.

We could make this a minor version bump and note the syntax change in the release notes.

kulesy: Great point about backwards compatibility! I've updated the PR to support both:

- `{ label: 'HTML', value: 'html' }` - The correct identifier for standard HTML syntax highlighting
- `{ label: 'HTMLBars', value: 'htmlbars' }` - Kept for backwards compatibility with existing content

This way:
- Existing content with `htmlbars` continues to work without breaking
- New users can select the correct `html` option
- Both will appear as separate options in the dropdown

This keeps it as a patch-level change since we're not breaking existing functionality, just adding the correct option alongside the legacy one.

18-th (approved):
LGTM

Issue #6662, opened 2026-04-13
GraphQL API accepts whitespace relativePath in mutations, creating unrecoverable documents


18-th:
The validatePath method in the GraphQL resolver does not reject relativePath values that are empty, whitespace-only, or contain leading/trailing whitespace. The existing guard (if (relativePath)) only catches null/undefined/"" - whitespace strings like `" "` and `" .mdx"` are truthy and pass through.

  This allows mutations like:

```
  mutation {
    createPost(relativePath: " .md", params: { title: "test" }) {
      __typename
    }
  }
```

to succeed and write files with whitespace names (e.g. content/posts/ .md) to the filesystem. These documents then fail on retrieval with Invalid file extension: expected '.md' but got '.', making them inaccessible through the admin UI and irremovable without direct filesystem access.

Affected operations: createDocument, create<Collection>, updateDocument, createFolder, and their collection-specific variants.

18-th:
Done ✅. PRs #6663 and #6664 have been merged with fixes to the GraphQL and the collection creation page.

PR #6664, opened 2026-04-13
Validate relativePath to reject whitespace and invalid characters

18-th:
Addresses #6662

Relevant docos:
https://tina.io/docs/extending-tina/filename-customization
https://tina.io/docs/guides/naming-conventions

* Extend validateRelativePath to enforce the documented character allowlist (a-zA-Z0-9, -, _, ., /) and add the missing ^ anchor to the admin UI filename regex.

* Update tests to cover invalid characters, positive cases, and UI-level validation.


18-th:
Relevant changes to the regex:

Before:

image

**Figure: Leading whitespace is not rejected by the regex**

After:

image

**Figure: All whitespace is rejected by the regex**

kulesy (approved):

PR #6661, opened 2026-04-13
♻️ Cleanup Old Example Projects

KahaMason:
## Summary

Removes all the now obsolete example projects that are now being replaced by the next, react, astro, hugo kitchen sink apps.

## Changes:

* Removed `basic-iframe`, `empty`, `hugo-quickstart`, `old kitchen-sink`, `next-2024` example projects
* Moved `tina-self-hosted` => `next/tina-self-hosted`
* Biome formatting of all example projects

PR #6665, opened 2026-04-13
🛠️ Fix CI Pipeline after removing obsolete example projects

KahaMason:
## Summary
Fixed changesets CI ValidationError caused by the legacy workspaces field in root package.json not matching the new nested examples structure (`examples/*/*` instead of `examples/*`)

## Context
After #6661 moved examples from a flat structure (`examples/basic-iframe`, etc.) to nested (`examples/next/kitchen-sink`, etc.), the workspaces field in package.json still had `examples/*` which no longer matched any packages. Changesets uses `@manypkg/get-packages` which reads this field instead of `pnpm-workspace.yaml`, so it couldn't find any `@examples/*` packages to validate against the ignore config.


18-th (approved):
LGTM

PR #6669, opened 2026-04-14
Playwright test for filtering, pagination and search

Ben0189:
Adds end-to-end Playwright API tests that exercise real LevelDB query behaviour and the /searchIndex HTTP route — replacing the mocked-layer coverage that existed before.

#### What's tested:

filtering.spec.ts — postConnection filter by field value against real LevelDB
sorting.spec.ts — LevelDB index lifecycle (create/update/delete), ascending and descending sort direction
pagination.spec.ts — cursor-based navigation with first and after, hasNextPage accuracy
search.spec.ts — /searchIndex builds from real content, returns precise results

✅ All tests verified with red-green cycle against the actual dist code.


18-th (approved):
LGTM

Issue #31, opened 2019-08-05
Add "Toggle" field to the xeditor


ncphillips:
Some text data represent links to an image.

Here's an example of some data:

```yaml
draft: true
```

The field definition might look something like

```javascript
let fields = [
  { 
    name: "draft", 
    label: "Draft", 
    component: "toggle"
  },
]
```

PR #6682, opened 2026-04-16
Fix React Error #31 when inserting image from Media Library with custom MediaStore

### LINKED ISSUE #6679: React Error #31 when inserting image from Media Library with custom MediaStore
## Bug Description

When using a custom `MediaStore` (e.g., a `GitHubMediaStore` for self-hosted setups), selecting an image from the Media Library and clicking "Insert" causes the TinaCMS admin to go blank with a white screen. The browser console shows:

```
React error #31: Objects are not valid as React children
```

The error shows a Media object with keys `{id, src, filename, directory, type, thumbnails}` being passed directly as a React child.

## Environment

- TinaCMS v3.7.2
- Self-hosted TinaCMS with custom MediaStore
- Netlify deployment with serverless functions

## Steps to Reproduce

1. Set up TinaCMS with a custom `MediaStore` that does **not** implement a `parse()` method
2. Open TinaCMS admin
3. Navigate to any page with an image field (e.g., a Hero Banner block)
4. Click the image field → "Media Library" tab
5. Select any image from the media library
6. Click "Insert"

## Expected Behavior

The selected image is inserted into the field and the src string is saved.

## Actual Behavior

- TinaCMS admin goes blank/white screen
- Console shows: `React error #31: Objects are not valid as React children`
- The full Media object `{id, src, filename, directory, type, thumbnails}` is passed as a React child

## Root Cause

In `packages/tinacms/src/toolkit/fields/plugins/image-field-plugin.tsx` (lines 34-44), the `onChange` callback tries to parse the Media object via `cms.media.store.parse(media)`. If the custom MediaStore doesn't implement `parse`, the fallback passes the **entire Media object** as the field value instead of extracting `media.src`:

```typescript
async function onChange(media?: Media | Media[]) {
  if (media) {
    const parsedValue =
      typeof cms?.media?.store?.parse === 'function'
        ? cms.media.store.parse(media)
        : media;  // <-- BUG: passes full Media object when parse() is missing
    props.input.onChange(parsedValue);
  }
}
```

The same pattern exists in the MDX editor's image toolbar button (`image-toolbar-button.tsx`), which calls `insertImg(editor, media)` without running the media through `parse()` first.

## Fix

The fallback should extract `media.src` instead of passing the raw object:

```typescript
const item = Array.isArray(media) ? media[0] : media;
const parsedValue =
  typeof cms?.media?.store?.parse === 'function'
    ? cms.media.store.parse(media)
    : item.src || item;
```

## Additional Notes

- The media library itself displays thumbnails correctly
- Uploading new images works fine
- The issue only occurs when **inserting** a selected image into a form field
- The default `MediaStore` in `media-store.default.ts` does implement `parse()` (returns `img.src`), so this only affects custom stores

kulesy:
## Summary

- Fixes a crash (React Error #31: "Objects are not valid as React children") when inserting an image from the Media Library into an image field or MDX rich-text editor, when using a custom `MediaStore` that doesn't implement a `parse()` method (e.g., `GitHubMediaStore` for self-hosted setups)
- The image field plugin's `onChange` fallback now extracts `media.src` instead of passing the full Media object as the field value
- The MDX image toolbar button now normalizes `media.src` via `parse()` or falls back to `media.src` before calling `insertImg()`

## Root Cause

The `parse()` method is **not part of the `MediaStore` interface** (`packages/tinacms/src/toolkit/core/media.ts:65-100`), so custom stores don't implement it. When `cms.media.store.parse` is `undefined`, the fallback in `image-field-plugin.tsx` was passing the entire `Media` object `{id, src, filename, directory, type, thumbnails}` as the field value. React then tried to render this object as a child, causing Error #31.

## Test plan

- [x] Unit tests added covering both buggy (regression) and fixed code paths (12 vitest tests)
- [x] Integration test script verifying all scenarios (16 assertions)
- [x] Manual browser test: unpatched = white screen crash, patched = image inserts correctly
- [ ] Verify existing tests still pass (`pnpm test` in `packages/tinacms`)
- [ ] Test with default `TinaMediaStore` (should work as before since `parse()` is still called when available)

Closes #6679


packages/tinacms/src/toolkit/fields/plugins/mdx-field-plugin/plate/components/plate-ui/image-toolbar-button.tsx
```diff
@@ -30,7 +30,14 @@ const useImageToolbarButton = (state) => {
...
       directory: '',
       onSelect: (media) => {
-        insertImg(editor, media);
+        // Ensure media.src is a string, falling back to parse if available
+        const src =
+          // @ts-ignore
+          typeof cms?.media?.store?.parse === 'function'
+            ? // @ts-ignore
+              cms.media.store.parse(media)
+            : media.src;
```
kulesy, line 37: I don't like we are ignoring typing on this one.

packages/tinacms/src/toolkit/fields/plugins/image-field-plugin.tsx
```diff
@@ -33,12 +33,13 @@ export const ImageField = wrapFieldsWithMeta<InputProps, ImageProps>(
 
     async function onChange(media?: Media | Media[]) {
       if (media) {
+        const item = Array.isArray(media) ? media[0] : media;
         const parsedValue =
           // @ts-ignore
           typeof cms?.media?.store?.parse === 'function'
             ? // @ts-ignore
               cms.media.store.parse(media)
```
kulesy: Same for here

JackDevAU (approved):
Looks good! Ship it 🛳️

Issue #6675, opened 2026-04-15
Build fails in v3.7.2 with ESM: ReferenceError: __filename is not defined (sqlite-level compatibility)


cxr470235739:
### Describe the issue that you're seeing. Any videos (eg. YouTube or Loom) or screenshots help!

When I build the project using TinaCMS 3.7.2, an error occurs: "_filename is not defined in ES module scope".
database：sqlit3

### Reproduction

https://github.com/tinacms/tinacms/tree/main/examples/next/tina-self-hosted-demo

### Steps to reproduce

pnpm add sqlite-level

go to /tina/database.ts 
import { createDatabase, createLocalDatabase } from '@tinacms/datalayer';
import { GitHubProvider } from 'tinacms-gitprovider-github';
import path from 'path';
import { SqliteLevel } from 'sqlite-level';
const isLocal = process.env.TINA_PUBLIC_IS_LOCAL === 'true';

export default isLocal
  ? createLocalDatabase()
  : createDatabase({
      gitProvider: new GitHubProvider({
        branch: process.env.GITHUB_BRANCH,
        owner: process.env.GITHUB_OWNER,
        repo: process.env.GITHUB_REPO,
        token: process.env.GITHUB_PERSONAL_ACCESS_TOKEN,
      }),
      databaseAdapter: new SqliteLevel({
          filename: path.join(process.cwd(), "sqliteDB", 'tinacms.db'),
        }),
      namespace: process.env.GITHUB_BRANCH,
    });

pnpm build

### System Info

```shell
System:
    OS: Ubuntu 24.04.4 LTS (WLS2)
    CPU: x86_64
    Shell: /bin/bash
  Binaries:
    Node: 22.14.0
    npm: 10.9.2
    pnpm: 10.32.1
  Browsers:
    Chrome: 140.0.7339.186
    Safari: 17.4
```

### Validations

- [x] Follow our [Code of Conduct](https://github.com/tinacms/tinacms/blob/main/CODE_OF_CONDUCT.md)
- [x] Read the [docs](https://tina.io/docs).
- [x] Check that there isn't [already an issue](https://github.com/tinacms/tinacms/issues) that reports the same bug.

18-th:
Hi @cxr470235739, We've got an ESM migration PR ready for that. There are currently issues with npm access for that package that we need to resolve before we can merge it.

https://github.com/tinacms/sqlite-level/pull/24.

link89:
[...]

PR #6683, opened 2026-04-16
chore(cli): support TINA_EXTERNAL_PKGS for esbuild

link89:
As tina v3 migrate to ESM, some dependencies may be broken due to depend on CommonJS features, for exmaple https://github.com/tinacms/tinacms/issues/6675

This PR add a new environment variable `TINA_EXTERNAL_PKGS` to allow developers to exclude packages like `better-sqlite3` , 
so that such problem won't block the adoption of tina v3 before all packages are migrated to ESM.

### DIFF
--- a/packages/@tinacms/cli/src/next/config-manager.test.ts
+++ b/packages/@tinacms/cli/src/next/config-manager.test.ts
@@ -1,4 +1,4 @@
-import { getExternalPackagesFromEnv } from './config-manager';
+import { getExternalPackagesFromEnv } from './external-packages';
 
 describe('getExternalPackagesFromEnv', () => {
   it('returns an empty array when env var is undefined', () => {
--- a/packages/@tinacms/cli/src/next/config-manager.test.ts
+++ b/packages/@tinacms/cli/src/next/config-manager.test.ts
@@ -1,0 +1,24 @@
+import { getExternalPackagesFromEnv } from './config-manager';
+
+describe('getExternalPackagesFromEnv', () => {
+  it('returns an empty array when env var is undefined', () => {
+    expect(getExternalPackagesFromEnv(undefined)).toEqual([]);
+  });
+
+  it('returns an empty array when env var is empty', () => {
+    expect(getExternalPackagesFromEnv('')).toEqual([]);
+  });
+
+  it('parses comma-separated package names', () => {
+    expect(getExternalPackagesFromEnv('better-sqlite3,node-bindings')).toEqual([
+      'better-sqlite3',
+      'node-bindings',
+    ]);
+  });
+
+  it('trims package names and ignores empty entries', () => {
+    expect(
+      getExternalPackagesFromEnv(' better-sqlite3, node-bindings , ,')
+    ).toEqual(['better-sqlite3', 'node-bindings']);
+  });
+});
--- a/packages/@tinacms/cli/src/next/config-manager.ts
+++ b/packages/@tinacms/cli/src/next/config-manager.ts
@@ -1,6 +1,5 @@
 import fs from 'fs-extra';
 import path from 'path';
-import os from 'os';
 import { pathToFileURL } from 'url';
 import * as esbuild from 'esbuild';
 import type { Loader } from 'esbuild';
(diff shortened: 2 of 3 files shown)


18-th:
Hi @link89, this feels like a hack. If you are blocked on a particular set of packages, do you mind compiling a list, and we'll see if we can address the root cause instead?

link89:
Sure, you may check https://github.com/tinacms/tinacms/issues/6675 this issue. I have the same issue when running the example, and it is blocked by `better-sqlite3`, whose dependency `node-bindings` is make use of `__filename` to locate .node file.

link89:
Hi @18-th I'd be interested to hear your thoughts on this. I think we could either hard-code externals in esbuild to exclude affected packages like better-sqlite3, or go with the approach proposed in this PR and leave the decision up to developers.

18-th:
Hi @link89 

The long and short of it is that we need to show a bit of love to the self-hosted. I believe the answer here is to prioritise fixing the root cause instead of applying the band-aid solution.

At this point, TinaCMS v3 does not support the self-hosted scenario, and the recommended approach is to use v2 until the migration activities are done.

We've got a PR coming with the update to the documentation about this issue:

https://github.com/tinacms/tina.io/pull/4465

I understand this is likely not the answer you were expecting, but that is the best I can give you. I will raise it with the team to prioritise the self-hosted in the near future.

link89:
@18-th Thanks for the response. I’d like to clarify the technical root cause one last time in hopes of making a final case for this PR.

The issue stems from TinaCMS attempting to bundle database-related code into a single database.build.mjs file via esbuild. While this works fine for pure JS dependencies (like mongodb), it fails for libraries like better-sqlite3 that contain native .node modules. To ensure SQLite can pass next build, the only viable solution is to externalize these non-bundlable modules. This can be achieved either by letting developers decide (which is exactly what this PR does) or by hard-coding known modules for exclusion. This is an opt-in feature that won't break any existing workflows and can be easily removed if a more elegant solution surfaces later.

I completely understand the team's prioritization regarding self-hosting. TinaCMS is a piece of software with exceptional technical taste and deserves commercial success. That said, perhaps the community can shoulder more of the self-hosting burden. Community can help. We have been using the v2 self-hosted version for a long time, and while we've hit some roadblocks upgrading to v3, we are more than happy to explore, solve, and contribute these fixes back to the community.

kulesy:
For context on the @18-th's "band-aid" concern, I dug into this and the architecture is wedged between two incompatible constraints after the ESM migration:

- **Bundling** native modules fails (`__filename` not in ESM scope) what #6675 reports.
- **Externalizing** them (this PR) makes the bundling step succeed but creates a runtime resolution failure: the build output lives in `os.tmpdir()/<timestamp>/build.mjs`, and Node cannot resolve bare imports from `/tmp/` because `node_modules` lookup walks up from the importing file's directory. Verified empirically — both `import 'better-sqlite3'` and `createRequire(import.meta.url); require('better-sqlite3')` fail with `ERR_MODULE_NOT_FOUND` / `MODULE_NOT_FOUND` from a `/tmp/` file even when the project has the package installed.

So this PR likely converts the visible bundle-time error into a different runtime error rather than fixing it end-to-end. @link89 - could you share the exact reproduction where this PR makes `pnpm build` succeed? I suspect environment-specific factors (pnpm symlinks, NODE_PATH, or the failure manifesting later in a way that wasn't attributed to this) are involved.

I've opened #6750 with a proposed root-cause fix: move the build output to `<project>/.tina/__generated__/.cache/<timestamp>/build.mjs` so `node_modules` is reachable, then default to `packages: 'external'`. Native modules work without special-casing, and `TINA_EXTERNAL_PKGS` becomes either unnecessary or a narrow override.

link89:
> I suspect environment-specific factors (pnpm symlinks, NODE_PATH, or the failure manifesting later in a way that wasn't attributed to this) are involved.

Yes, `NODE_PATH=$(pwd)` has been used in the build script. 

> I've opened https://github.com/tinacms/tinacms/issues/6750 with a proposed root-cause fix: move the build output to <project>/.tina/__generated__/.cache/<timestamp>/build.mjs so node_modules is reachable, then default to packages: 'external'. Native modules work without special-casing, and TINA_EXTERNAL_PKGS becomes either unnecessary or a narrow override.

Moving build directory to project root and skip bundling external modules make more sense than the current implementation. I will looking forward to the fix.

kulesy:
@link89 — wanted to follow up: the #6750 fix shipped in #6790 (just merged). 

We actually ended up having to add an externalize option for users, as we couldn't completely externalize all packages, as some were CJS only and weren't compatible with our ESM-only CLI package.

Net result: your reproduction case (self-hosted with sqlite-level) now builds clean out of the box, no env var or workspace-level workaround required.

Thanks for the original investigation + PR — surfacing the externalize-vs-bundle question is what got us to the right architecture. This PR can be closed now.

PR #6678, opened 2026-04-16
fix: display editorial workflow modal on deleted branches

brookjeynes-ssw:
image  

**Figure: Branch list identifies deleted branch**

image  

**Figure: Editorial workflow for deleted branch content push**

Resolves #6344

PR #6687, opened 2026-04-17
feat: simplify errors shown to users

brookjeynes-ssw:
While the work done in this PR and #6678 won't solve the underlying issue of #6660, it aims to make the error message less scary to the user and gives them a way forward as after a refresh the issue should resolve itself. An investigation into the underlying issue will be conducted with the work required for TinaV4

image

### DIFF
--- a/packages/@tinacms/app/src/lib/errors.tsx
+++ b/packages/@tinacms/app/src/lib/errors.tsx
@@ -10,7 +10,7 @@ const ErrorModalContent = (props: { title: string; message: string }) => {
       <div>{title}</div>
       <p>{message}</p>
       <a href={TROUBLESHOOTING_URL} target='_blank' rel='noopener noreferrer'>
-        Learn more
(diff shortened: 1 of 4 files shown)


packages/@tinacms/app/src/lib/errors.tsx
```diff
@@ -1,24 +1,27 @@
...
     <>
       <div>{title}</div>
-      <ul>
-        {errors.map((error, i) => (
-          <li key={i}>{error}</li>
-        ))}
-      </ul>
+      <p>{message}</p>
+      <a href={TROUBLESHOOTING_URL} target='_blank' rel='noopener noreferrer'>
+        Learn more
```
kulesy: Make this more descriptive eg.

```suggestion
        Try these troubleshooting tips
```

18-th (dismissed):
LGTM

18-th (approved):
LGTM

PR #6698, opened 2026-04-20
fix(scripts): externalise peer dep sub-paths in rollup build

### LINKED ISSUE #6688: 🔨 tinacms-authjs — externalise React to support both React 18 and 19
## Description

`tinacms-authjs@21.0.2` bundles its own copy of `react/jsx-runtime` compiled against **React 19** internals. When the host project runs **React 18** — as the [`tina-self-hosted-demo`](https://github.com/tinacms/tina-self-hosted-demo) template does — the admin fails with:

```
Failed loading TinaCMS assets
```

…because the bundled runtime references symbols and properties that don't exist in R18.

## Root cause

A library shipping React components should leave React to the host — bundling `react/jsx-runtime` is what creates this version-straddling problem. The fix is to externalise `react`, `react-dom`, and `react/jsx-runtime` in `tinacms-authjs`'s bundle config so the host's React is used.

Specific React API differences the bundled runtime assumes (R19) vs breaks on (R18):

| Area | React 19 (what the bundle assumes) | React 18 (what the host provides) |
| --- | --- | --- |
| `REACT_ELEMENT_TYPE` | `Symbol.for("react.transitional.element")` | `Symbol.for("react.element")` |
| Internals export | `__CLIENT_INTERNALS_DO_NOT_USE_OR_WARN_USERS_THEY_CANNOT_UPGRADE` | `__SECRET_INTERNALS_DO_NOT_USE_OR_YOU_WILL_BE_FIRED` |
| Owner lookup | `ReactSharedInternals.A.getOwner()` | `ReactSharedInternals.ReactCurrentOwner.current` |
| `recentlyCreatedOwnerStacks` | present | absent (causes `undefined++` → NaN) |

## Reproduction

1. Deploy [`tina-self-hosted-demo`](https://github.com/tinacms/tina-self-hosted-demo) via the Vercel template button.
2. Upgrade `tinacms` to `^3.7.2`, `tinacms-authjs` to `21.0.2`, leave `react` at `^18.3.1`.
3. Visit `/admin` — "Failed loading TinaCMS assets".

Matching user report with same symptom: #6674. Community-authored patch demonstrating the workaround: [`stout-source/tina-self-hosted-react-fix`](https://github.com/stout-source/tina-self-hosted-react-fix) ([PATCH.md](https://github.com/stout-source/tina-self-hosted-react-fix/blob/main/PATCH.md)) — useful as a reference, but not safe to ship (pins R18 internals that can change in any patch release).

## Current workaround (docs-only)

[tina.io#4465](https://github.com/tinacms/tina.io/pull/4465) adds a troubleshooting guide telling self-hosted users to stay on the v2.x line:

```json
{
  "@tinacms/cli": "^1.12.6",
  "@tinacms/datalayer": "^1.4.3",
  "tinacms": "^2.10.1",
  "tinacms-authjs": "^5.0.9",
  "tinacms-gitprovider-github": "^2.0.19",
  "react": "^18.3.1",
  "react-dom": "^18.3.1",
  "next": "14.2.35"
}
```

That unblocks users but effectively writes off v3 for self-hosted indefinitely — this PBI is the real fix.

## Acceptance criteria

- [ ] `tinacms-authjs` bundle config marks `react`, `react-dom`, and `react/jsx-runtime` as externals (peer-only, not bundled).
- [ ] `tinacms-authjs` works against both `react@^18.3` and `react@^19` in a consuming app, without patching `dist/*`.
- [ ] Verify `next-auth@4.x` interop under the current ESM bundle config across the supported Next.js versions; if any config produces a `{ default: fn }` shape, add a `.default` unwrap at the `TinaCredentialsProvider` and `AuthJsBackendAuthProvider` call sites.
- [ ] [`tina-self-hosted-demo`](https://github.com/tinacms/tina-self-hosted-demo) is updated to a working version set and its Vercel deploy button re-verified end-to-end (login + save content).
- [ ] Release notes / upgrade guide explicitly call out the fix version.
- [ ] Docs cleanup: revert the v3-blocking warnings added in [tina.io#4465](https://github.com/tinacms/tina.io/pull/4465) (`upgrade-to-latest-version.mdx` callout + `troubleshooting.mdx` "Version Compatibility Issues" section) once the fix ships.

## Notes

- This is the `tinacms-authjs` counterpart to the already-closed #5549 ("Bundle React in TinaCMS"), which covered the main `tinacms` package.
- If the team decides to drop React 18 support entirely in v3+, that's a valid resolution — but the "stay on v2.x" docs guidance still needs a hard expiry date and the self-hosted template needs to be brought up to React 19 + Next 15.

kulesy:
## Summary

Fixes [#6688](https://github.com/tinacms/tinacms/issues/6688). Closes the "Failed loading TinaCMS assets" class of error for self-hosted consumers on a different React major than the monorepo.

Rollup's string-array `external` option only matches **exact** module IDs. That meant listing `react` / `next-auth` in a package's peer deps did **not** externalise sub-paths like `react/jsx-runtime` or `next-auth/react` — they got bundled into the package output.

Most visibly, `tinacms-authjs@21.0.2` inlined its own copy of `react/jsx-runtime` compiled against whichever React version the monorepo was on (currently 19). Consumers still on React 18 — notably the published `tina-self-hosted-demo` template — then hit the bundled runtime referencing R19 internals (`REACT_ELEMENT_TYPE` symbol, `__CLIENT_INTERNALS_...`, `recentlyCreatedOwnerStacks`) that don't exist on R18, producing "Failed loading TinaCMS assets". Full analysis in [#6688](https://github.com/tinacms/tinacms/issues/6688).

## Fix

In `packages/@tinacms/scripts/src/index.ts`, replace the string-array `external` in the Vite/Rollup build path with a predicate that matches both exact dep names AND any sub-path:

```ts
const isExternal = (id: string) =>
  external.some((dep) => id === dep || id.startsWith(`${dep}/`));
```

esbuild already treats a package name as externalising its sub-paths, so only the Vite/Rollup path needed changing. The node-target builds (e.g. `tinacms-authjs/dist/index.js`) were already correct.

## Effect

`tinacms-authjs/dist/tinacms.js`:

```
before (v21.0.2 published): 98 KB — inlined react/jsx-runtime, next-auth/react, ~dozen Babel helpers
after  (this PR):           4.4 KB — only authjs's own code
```

Imports in the post-fix bundle:
```
from "next-auth/react"  ← now external (was bundled)
from "tinacms"          ← already external
```

Other packages using `@tinacms/scripts` (`tinacms-clerk`, `@tinacms/vercel-previews`, `next-tinacms-*`, etc.) also get tighter bundles where they previously inlined sub-paths of peer deps — latent bugs that the same dialect-mismatch could have triggered in future React/tinacms bumps.

## Verification

- `pnpm build --force` from repo root: **20/20 packages build cleanly**.
- `grep 'jsx-runtime\|REACT_ELEMENT_TYPE\|ReactSharedInternals' packages/tinacms-authjs/dist/tinacms.js` → 0 matches.
- Remaining imports in `tinacms-authjs/dist/tinacms.js` are all peer-dep-provided (`next-auth/react`, `tinacms`).

## Changeset

`.changeset/externalise-react-subpaths.md`:
- `@tinacms/scripts`: patch
- `tinacms-authjs`: patch

Other affected packages (clerk, vercel-previews, cloudinary, s3, azure, dos) also get slightly smaller bundles from this change; happy to expand the changeset if the team wants those bumped too — I kept it narrow to the two packages with the clearest user-visible effect.

## Follow-ups not in this PR

- [`tina-self-hosted-demo`](https://github.com/tinacms/tina-self-hosted-demo) template still pins to the pre-v3 version matrix — worth revisiting once this ships, so users can deploy from the button straight onto v3.
- The related `next-auth@4.x` CJS/ESM `.default` interop risk called out in [#6688](https://github.com/tinacms/tinacms/issues/6688) is unconfirmed as user-visible; not addressed here. If it shows up, a defensive unwrap at the call sites in `tinacms-authjs/src/index.ts` is cheap insurance.
- Docs PR [tina.io#4465](https://github.com/tinacms/tina.io/pull/4465) tells self-hosted users to stay on v2.x; once this lands and ships, the v3-blocking warnings in that PR should be reverted.

## Test plan

- [x] `pnpm install` clean
- [x] `pnpm build --force` — all 20 packages build
- [x] `tinacms-authjs/dist/tinacms.js` no longer contains inlined `react/jsx-runtime`
- [ ] Reviewer: verify at least one consuming example still runs end-to-end (`examples/next/tina-self-hosted-demo` or similar)


18-th (approved):
LGTM

PR #6721, opened 2026-04-22
feat(schema-tools, cli): validate localContentPath + extract content-root resolution

joshbermanssw:
Addresses [tinacms/tinacloud#3294](https://github.com/tinacms/tinacloud/issues/3294) — "P4: Add contentPath configuration option to Tina config".

## Context

`localContentPath` already exists on the Tina config type at `packages/@tinacms/schema-tools/src/types/index.ts:713-725` with a docstring, and is fully wired into the CLI (`config-manager.ts` resolution + `contentRootPath` + `hasSeparateContentRoot()` + downstream consumers in `codegen`, `database`, `dev-command`, `build-command`, `searchindex-command`, `vite`). The multi-repo tinacloud work in [tinacms/tinacloud#3436](https://github.com/tinacms/tinacloud/issues/3436) depends on it.

Two gaps against the PBI's DoD:

1. **No runtime validation.** TypeScript caught the type at compile time; a non-string or empty-string value at runtime would fall through silently.
2. **No unit tests for path resolution.** The 22-line resolution block in `ConfigManager.processConfig` was inline and untestable without running the full `processConfig` pipeline (which loads a real config file via esbuild).

### `packages/@tinacms/cli`
- Extract the content-root resolution block from `ConfigManager.processConfig` into a new `resolveContentRootPath` function in `packages/@tinacms/cli/src/next/resolve-content-root.ts`.
- Validate `localContentPath` inside `resolveContentRootPath` using a local Zod schema (`z.string().min(1).optional()`). Rejects non-string and empty-string values at config-load time.
- `processConfig` now calls it with the four required inputs; behaviour is preserved exactly:
  - Single-repo (no `localContentPath`): returns `rootPath`.
  - Multi-repo + directory exists: returns joined path, info log fires.
  - Multi-repo + directory missing: warn log fires (citing the config file for user remediation), falls back to `rootPath`.
- Jest test (`resolve-content-root.test.ts`) covers all four behavioural branches plus trailing-slash stripping plus the two Zod rejection cases (empty string, non-string).
- Extracted rather than in-place so the Jest unit test does not have to import `ConfigManager` itself (which uses `import.meta.url` at module scope and is not loadable under the repo's current ts-jest CJS config).

### `packages/@tinacms/schema-tools`
- `localContentPath` is **deliberately excluded** from `tinaConfigZod`. Including it caused it to survive the Zod parse in `validateTinaCloudSchemaConfig` and leak into the hashed `_schema.json`. Any project that sets `localContentPath` then fails the "local Tina Schema matches server" check on upgrade until its remote schema is re-indexed. Empirically verified against the SSW.Rules generator repo.
- Validation lives in `resolveContentRootPath` (the only consumer), keeping this PR's intent without contaminating the server-bound schema.
- New regression test in `tinaCloudSchemaConfig.spec.ts` pins the hash invariant: `localContentPath` is stripped from `validateTinaCloudSchemaConfig` output.

## Net effect on existing users

- Single-repo projects: no change.
- Multi-repo projects with a valid `localContentPath: '../path'`: no change.
- Any project that (unusually) set `localContentPath: ''` — previously silently fell back to `rootPath`, now throws loudly on startup with a Zod error. Arguably a bug fix, but technically a behavior change.
- Projects that already set `localContentPath` and were on a prior version: the field will no longer leak into `_schema.json`, so the local/server hash match is preserved across upgrades.


kulesy (dismissed):

joshbermanssw:
✅ Tested to confirm no regression on building SSW Rules

kulesy (approved):

PR #6726, opened 2026-04-23
Use server-provided src URL for uploaded media assets

kulesy:
## Summary
- When uploading media, use the `src` URL returned by the server instead of always constructing it client-side as `https://assets.tina.io/{clientId}/{path}`
- Falls back to the existing constructed URL when the server doesn't provide one, maintaining backward compatibility

## Test plan
- [ ] Upload a media asset and verify the returned `src` URL is used correctly
- [ ] Verify backward compatibility when the server response doesn't include a `src` field


kulesy:
AI created the wrong PR

PR #6729, opened 2026-04-24
Implement test for mdx parse/stringify edge cases

Ben0189:
- Added src/parse/index.test.ts with 3 unit tests covering error recovery in the MDX parser (type: 'mdx' path):
- JSX child expression inside a registered component (<Cta>{name}</Cta>) produces invalid_markdown
- Malformed/truncated tag (<Cta) produces invalid_markdown
- Serializing an invalid_markdown node roundtrips back to the original string (no data loss on save)


Ben0189:
Three test fixtures had been prefixed FIXME- since April 2023 when the underlying parser logic for HTML inside shortcodes was reverted, leaving them as known broken cases. The parser has since been fixed, so all three now pass without any code changes needed.

instead just remove the fixme prefix

packages/@tinacms/mdx/src/parse/index.test.ts
```diff
@@ -0,0 +1,63 @@
+import { describe, expect, it } from 'vitest';
+import type { RichTextField } from '@tinacms/schema-tools';
+import { parseMDX } from './index';
+import { serializeMDX } from '../stringify';
+
+// parseMDX and serializeMDX require an image URL callback to transform image src values.
+// In tests we don't worry about image transformation, so we pass this no-op that returns the URL unchanged.
+const passthrough = (value: string) => value;
+
+const field: RichTextField = {
```
JackDevAU, line 9: should we make these tests match the other markdown/field tests (field.ts file, in.md file, etc) ?

packages/@tinacms/mdx/src/next/parse/index.ts
```diff
@@ -1,3 +1,5 @@
+// This is the newer MDX parser implementation, introduced in commit 651b6b53b ("Add next module for mdx behavior").
```
18-th: This comment does not seem correct. In the referenced file, only the markdown is passed to here, not the mdx. The mdx case there just falls through

Image

packages/@tinacms/mdx/src/parse/index.test.ts
Ben0189: Done

18-th (approved):
LGTM

Issue #6730, opened 2026-04-24
create-tina-app is failing intermittently in the Build TinaCMS Starter Templates workflow


KahaMason:
## Pain

The `Build TinaCMS Starter Templates` workflow is failing on a different template every weekday on its scheduled run. Each failure opens a new `bug`-labelled issue via the workflow's own error-reporting step, so the issue tracker has accumulated **8 separate auto-filed bugs in 2026 alone** for what appears to be one underlying cause.

### Failure signature

Every matching CI failure shares the same log shape:

```
Create Tina App v2.1.3
- Running pre-run checks...
✔ Node v22.22.2 is supported.
- Downloading template...
✖ Failed to download template: Repository information not found.
Error: Process completed with exit code 1.
```

The failure is before any install or build runs. The `Run create-tina-app` step exits non-zero, and `continue-on-error: true` in the matrix lets the rest of the workflow finish normally.

### Root cause

**The likely cause**: unauthenticated calls to `api.github.com` hitting rate-limit 403s under CI concurrency, combined with error handling that discards the real status before anyone sees it.

**How it currently functions**: `create-tina-app` resolves every template by calling `GET api.github.com/repos/<owner>/<name>` with no `Authorization` header. When the matrix fans out 63 concurrent jobs at the same scheduled cron tick, some of those calls hit GitHub's unauthenticated rate limit (60 requests / per hour).

**Why the logs are unhelpful**: the CLI treats any non-200 identically — discards the status code, response body, and rate-limit headers — then throws a generic `"Repository information not found."`

The failures span four different templates, three package managers, and three Node versions, with no correlating source changes — which is why the evidence points at the shared unauthenticated fetch rather than any individual template.

### Suggested Fixes

1. **Primary** — authenticate the GitHub API call using `GITHUB_TOKEN` / `GH_TOKEN` when present. Raises the rate limit from 60/hour-per-IP (unauthenticated) to 1,000/hour-per-repository (GitHub Actions token) — roughly 16× headroom for the 63-cell matrix.
2. **Supporting** — surface the real HTTP status and rate-limit headers in the thrown error, so future failures are self-diagnosing.
3. **Workflow wiring** — expose `secrets.GITHUB_TOKEN` to the `Run create-tina-app` step so the authenticated CLI path has a token to use. `@latest` stays — this workflow's job is to catch regressions in new CLI releases, and pinning would defeat that purpose.

## Related Issues

- #6625
- #6643
- #6676
- #6677
- #6685
[...]

PR #6735, opened 2026-04-24
create-tina-app: add diagnostic for Github API rate-limit errors

KahaMason:
## Summary

partially addresses #6730

- Adds detection for GitHub API primary (403 with `X-RateLimit-Remaining: 0`) and secondary (403/429 with `Retry-After`) rate-limit responses in `getRepoInfo`.
- Throws a specific `"GitHub API rate limit exceeded"` error instead of falling through to the generic `"Repository information not found."` message.
- Non-rate-limit failures are byte-for-byte unchanged — this is additive observability, not a behaviour change to the failure path.

## Why

The scheduled `Build TinaCMS Starter Templates` workflow has been intermittently failing across different templates, package managers, and Node versions, all surfacing as the same generic error. The shape of those failures — spread across unrelated templates with no corresponding source change since `v2.1.0` — points at GitHub's unauthenticated API rate limit (60 requests/hour per IP) as the probable cause. But the current code discards the HTTP status before the error is constructed, so the log never actually says "rate limit".

This PR makes rate-limit failures identifiable in the log so the theory can be confirmed from CI output.

## Scope — Phase 1 of 2

This is the **diagnostic** phase:

- ✅ Adds observability — rate-limit failures get a specific error message
- ❌ Does NOT fix the rate limit itself

Once CI confirms the cause by surfacing the new error, a follow-up PR will authenticate the `fetch` call with `GITHUB_TOKEN` / `GH_TOKEN` (raising the limit from 60/hour-per-IP to 1,000/hour-per-repository under `secrets.GITHUB_TOKEN`) and wire the token into the `Run create-tina-app` workflow step.

## Before / After

**Before — rate-limit failure in CI:**

```
✖ Failed to download template: Repository information not found.
```

**After:**

```
✖ Failed to download template: GitHub API rate limit exceeded
```

Other failures (missing repo, transient outage, etc.) are unchanged and continue to surface as the original generic message.

## Risk

Low. The change is additive — it inserts a new branch that throws a specific error only for responses matching GitHub's documented rate-limit signatures (primary: `X-RateLimit-Remaining: 0`; secondary: `Retry-After` header on 403 or 429). Every other non-200 falls through to the existing `return undefined` → generic error path unchanged. False-positive risk is very low (the header combinations we check for are GitHub's documented rate-limit fingerprints); false-negative risk falls back to current behaviour (not a regression).

## Test plan

- [x] `pnpm --filter create-tina-app build` — passes
- [x] `biome check ./packages/create-tina-app/src/util/examples.ts` — clean
- [ ] After release: wait for next scheduled cron of `Build TinaCMS Starter Templates` (or trigger via `workflow_dispatch`); verify a failing cell now logs `GitHub API rate limit exceeded` rather than the generic message
- [ ] If the new error appears on a failing cell, proceed with Phase 2 (the authentication fix)

## Files changed

- `packages/create-tina-app/src/util/examples.ts` — rate-limit detection + specific error throw
- `.changeset/ten-pumas-lick.md` — changeset entry for the patch release


KahaMason:
We have been able to identify that the root cause of the issue is that we're hitting Github's API Rate Limits. As such this PR is no longer needed.

image

**Figure: Build TinaCMS Starter Templates workflow log with diagnostic**

PR #6738, opened 2026-04-24
feat(cli): tinacms dev & build support for localContentPath (no duplicate content-repo writes)

joshbermanssw:
Closes [tinacms/tinacloud#3295](https://github.com/tinacms/tinacloud/issues/3295).

## Before

When a project sets `localContentPath` to point at a sibling content repo, every `tinacms dev` and `tinacms build` writes the generated artifacts (`_schema.json`, `_graphql.json`, `_lookup.json`, `tina-lock.json`) **twice**: once to the generator's `tina/__generated__/`, and again to the content repo. Editors then see a `tina/` folder full of build output land in their content repo on every commit. The content repo also has to *already contain* a `tina/` folder or the CLI errors out — so editors can't keep their content repo clean even if they wanted to.

## After

Generated artifacts go only to the generator's `tina/__generated__/`. The content repo is pure content; no `tina/` folder required, no build output committed, no duplicate `tina-lock.json`.

Two pieces make that work:
- The duplicate-write branches in `Codegen` and `DevCommand` are gone.
- `FilesystemBridge` routes any path starting with `tina/__generated__/` or `.tina/__generated__/` to `rootPath` (the generator); everything else still uses `outputPath` (the content root). Without this, schema/graphql/lookup reads after the duplicate-write removal would `ENOENT` on the content side.

## Testing

I had my whole regular setup locally; 
- [ ] tinacloud running locally on main branch 
- [ ] created a seperate content repo project pointing to SSW.Rules.Josh and SSW.Rules.Content.Test.JB
- [ ] SSW.Rules.Josh had tina config file ovverides pointing to the local instance of TinaCloud + .env set up 

- [ ] ran pnpm dev and pnpm build locally to ensure that tina files are only written to generator repo 
- [ ] deployed to vercel and tested editing 

- [ ] SSW.Rules.Josh is using the tagged tina versions (check out its repo if u want)

Screenshot 2026-05-06 at 4 30 33 pm

**Figure: my deployed site url**

---

## A bit more

**Public-API changes** (covered in the changeset migration notes):
- \`ConfigManager.generatedFolderPathContentRepo\` removed — use \`generatedFolderPath\`.
- \`ConfigManager.getTinaFolderPath\`'s \`isContentRoot\` option removed.
- \`FilesystemBridge\` routing for \`tina/__generated__/\` resolves to \`rootPath\` in multi-repo setups (custom bridge subclasses may need updates).
- Existing multi-repo users will have a stale \`tina/\` folder in their content repo after upgrade — safe to delete.

**Bonus fix rolled in.** Generated \`client.ts\` / \`database-client.ts\` now import \`./types\` (extensionless) for TS projects instead of \`./types.ts\` — avoids requiring \`allowImportingTsExtensions: true\` in consumer tsconfigs (broke under Next.js 15.5+ defaults). JS projects still import \`./types.js\` (Node ESM requires it).


joshbermanssw:
## Self-review

### Architecture / approach
The read/write split is the right pattern: route paths under `tina/__generated__/` (and legacy `.tina/__generated__/`) to `rootPath` at the bridge layer, leave content paths on `outputPath`. The new `baseFor()` helper centralises the rule so future readers don't have to puzzle out the split per call-site.

### What's solid
- **Bridge tests pin the four cases** (put/get to generated → rootPath, legacy `.tina/`, content → outputPath). The behavior split is locked in.
- **Codegen + DevCommand simplifications are pure deletions** of the duplicate-write blocks — no behavioural risk on single-repo apps.
- **Removed `generatedFolderPathContentRepo` field + `isContentRoot` option** — both were leaks of the duplicate-write design into the public surface.
- **Changeset migration notes are comprehensive**: stale `tina/` folder cleanup guidance, removed field, removed option, and the bridge behaviour change for custom subclasses.
- **Security preserved.** `assertWithinBase` still runs after `baseFor()` picks the base — `tina/__generated__/../../escape.txt` is still caught, just against `rootPath`.

### Things worth pushing on

1. **`isGeneratedPath` boundary edge.** `startsWith('tina/__generated__/')` (with trailing slash) means a caller passing the directory itself (`'tina/__generated__'`, no trailing slash) routes to `outputPath`. Realistically all bridge callers pass file paths, but a one-line test pinning that edge would close it.

2. **Telemetry checkbox in the PBI DOD is unaddressed.** PBI [#3295](https://github.com/tinacms/tinacloud/issues/3295)'s DOD says *"Telemetry added to track usage of `localContentPath` in dev and build commands"*. This PR doesn't add any — either drop the checkbox in a follow-up, or push a 5-line addition here.

3. **Higher-level integration tests deferred.** DOD lists `tinacms dev` and `tinacms build` integration tests with a separate content directory. Bridge unit tests cover the I/O routing but not command-level orchestration. Acceptable to defer to manual E2E on `berman-josh-v2`, but worth being honest about the gap.

4. **Backwards-compat E2E is manual.** DOD mandates *"stable pre-#3295 tinacms continues to build cleanly against post-#3403 tinacloud."* Confirm on a real stack before promoting to `@latest`.

5. **`baseFor` could use a one-liner comment** noting that the security check (`assertWithinBase`) still runs against whichever base is chosen — future readers shouldn't have to re-derive the safety reasoning.

6. **Codegen "exactly once" test** uses a flat fixture without modeling multi-repo. Since the duplicate-write code path was removed unconditionally (rather than gated by `hasSeparateContentRoot`), the simple test is sufficient — but a parallel test with `hasSeparateContentRoot: () => true` would strengthen the regression pin.

7. **Dead code grep before merge:** confirm whether `ConfigManager.hasSeparateContentRoot()` still has any callers post-this-PR. If not, follow-up to remove it.

### Verdict
Mergeable. Items 1, 5, 6, 7 are minor and could be follow-ups. Item 2 is the one I'd want resolved (or moved to a follow-up issue) so the PBI can close cleanly. Items 3 and 4 are pre-`@latest`-promotion manual gates.

kulesy:
**Minor — no end-to-end orchestration test for `localContentPath`**

Bridge routing is well unit-tested in `filesystem-bridge/index.test.ts`. CLI single-write is pinned in `codegen/index.test.ts`. But there's no test that exercises the full pipeline — `localContentPath` set → `processConfig` → bridge constructed with `(rootPath, contentRootPath)` → content reads hit `outputPath`, generated reads hit `rootPath`.

The PR description's `## Testing` checklist is all `[ ]` (manual flow), so a regression that broke the routing in a future refactor would only get caught by a human running through the same setup.

I just verified the full flow manually against `tina-multi-repo-generator` (with `localContentPath: '../../tina-multi-repo-content'`) → `tinacms build` and `tinacms dev` both wrote artifacts only to the generator, content reads via GraphQL came from the content repo. So the behavior is right; the gap is just the regression-pinning.

**Suggested test** (sketch — would live alongside the new bridge tests or in `cli/src/next/`):

```ts
// 1. Create temp generator dir + sibling content dir on disk
//    generator/tina/config.{ts,js} with localContentPath: '../content'
//    generator/tina/__generated__/_lookup.json + _schema.json + _graphql.json
//    content/posts/example.md (single collection content file)
// 2. const cm = new ConfigManager({ rootPath: generator })
//    await cm.processConfig()
// 3. expect(cm.contentRootPath).toBe(<content abs>)
//    expect(cm.rootPath).toBe(<generator abs>)
// 4. const bridge = new FilesystemBridge(cm.rootPath, cm.contentRootPath)
//    await bridge.glob('posts', 'md') → ['posts/example.md'] (from content)
//    await bridge.get('tina/__generated__/_lookup.json') → contents from generator
//    await bridge.put('posts/new.md', '...') → file lands in content, NOT generator
// 5. await fs.pathExists(<content>/tina/__generated__/_lookup.json) === false
```

Catches: bridge constructed with swapped args, `localContentPath` accidentally feeding `rootPath`, future refactor that drops the `baseFor` routing in one of the bridge methods, anyone re-introducing the duplicate-write code path. ~30 lines, no mocks, runs fast.

Given the rollout-sensitivity (the changeset's `⚠️ Rollout gate` notes that pre-#3403 TinaCloud breaks if this ships wrong), pinning the orchestration end-to-end is worth the small upfront cost.

kulesy:
**Minor — `@tinacms/metrics` needs a `minor` bump in the changeset**

```md
---
"@tinacms/cli": minor
"@tinacms/graphql": minor
"@tinacms/metrics": minor
---
```

joshbermanssw:
Done in b020e5b. Added `packages/@tinacms/cli/src/next/localcontentpath-orchestration.test.ts` — real fs, real bridge, five cases:

- `localContentPath` set + sibling exists → generated lands in generator, content lands in content repo
- stale `tina/__generated__/` in content repo → generator wins on reads
- `localContentPath` set + sibling missing → falls back to single-repo
- `localContentPath` omitted → single-repo
- swapped bridge args → routes to wrong repos (sentinel)

Couldn't drive `ConfigManager.processConfig()` directly — `import.meta.url` at module scope blocks it under jest CJS (same reason `resolveContentRootPath` was extracted). Test wires resolver + bridge exactly as the four callsites do.

joshbermanssw:
Done in b020e5b. New `TinaCMSDevInvoke` + `TinaCMSBuildInvoke` event types in `@tinacms/metrics/src/interfaces/index.ts` warrant the bump.

kulesy (approved):

PR #6740, opened 2026-04-28
feat(@tinacms/graphql): branch-aware media URL generation

### LINKED ISSUE #6504: GraphQL Resolver — Branch-Aware Media URL Generation
## Background

**Media URL resolution:** When TinaCMS GraphQL returns a document containing an image field, the raw value stored in git is a relative path (e.g. `/uploads/hero.jpg`). The resolver converts this to a full CDN URL (`https://assets.tina.io/{clientId}/uploads/hero.jpg`) via `resolveMediaRelativeToCloud()`. On mutation/save, `resolveMediaCloudToRelative()` strips the CDN prefix back to the relative path before writing to git.

**`GraphQLConfig`:** a plain object constructed once per request and passed into the resolver. Already contains `assetsHost`, `clientId`, and `useRelativeMedia` fields. TinaCloud constructs this object in `graphql-content.ts` (covered in a separate TinaCloud issue).

**`useRelativeMedia: true`:** When TinaCMS is self-hosted (not using TinaCloud), image paths stay relative and are never expanded to CDN URLs. This code path is completely unaffected by this change.

**CloudFront fallback:** Branch CDN URLs point to staging S3. If the file was not modified on the branch (inherited from main), CloudFront automatically falls back to the production S3 origin. The resolver does not need to check whether a file exists on the branch.

**V1 limitation:** If a file is deleted on a branch, its image field in a document will still resolve to the staging CDN URL, which CloudFront serves from production via fallback. The file appears to still exist in content previews. Deleted-file detection in the resolver is deferred to a future version.

## Pain

When content is queried on a feature branch, image fields in documents resolve to production CDN URLs (`https://assets.tina.io/{clientId}/image.jpg`). This means that even if an editor uploaded a new version of an image on the branch, the old production version is shown in content previews — the branch's staging file is never referenced. Vercel PR preview deployments cannot show correct branch media because all image URLs point to production.

## Description

Update the GraphQL resolver to generate staging-prefixed CDN URLs for image fields when the request is on a non-main branch. The host (TinaCloud) passes `branch` and `mediaBranch` into `GraphQLConfig`; the resolver uses them to decide which URL form to produce.

**`GraphQLConfig` type extension (`packages/@tinacms/graphql/src/types.ts`):**
```typescript
// Add to the non-relative variant:
branch?: string
mediaBranch?: string
```

**`resolveMediaRelativeToCloud()` (`resolver/media-utils.ts`):** when `branch` is set and `branch !== mediaBranch`, prepend `/__staging/{encodeURIComponent(branch)}` to the CDN path. Main branch and self-hosted behaviour unchanged.

**`resolveMediaCloudToRelative()` (`resolver/media-utils.ts`):** strip the `/__staging/{branch}/` segment when converting CDN URLs back to relative paths (used on mutation/save to store the canonical relative path in the document).

## Affected Files

**TinaCMS:**
- `packages/@tinacms/graphql/src/types.ts` — add `branch?`, `mediaBranch?` to `GraphQLConfig`
- `packages/@tinacms/graphql/src/resolver/media-utils.ts` — staging URL generation, staging URL stripping

## Acceptance Criteria

- [ ] `GraphQLConfig` type includes optional `branch` and `mediaBranch` fields
- [ ] `resolveMediaRelativeToCloud()` returns `https://assets.tina.io/{clientId}/__staging/{encodedBranch}/{path}` when `branch` is set and `branch !== mediaBranch`
- [ ] `resolveMediaRelativeToCloud()` returns the production URL when `branch === mediaBranch` or `branch` is unset
- [ ] Self-hosted (`useRelativeMedia: true`) behaviour is completely unchanged
- [ ] `resolveMediaCloudToRelative()` correctly strips `__staging/{branch}/` when saving a document on a branch
- [ ] `resolveMediaCloudToRelative()` is unchanged for production URLs (no `__staging` segment)

## Depends On

_No dependencies. This is a pure package change; it takes effect once the host passes `branch`/`mediaBranch` into `GraphQLConfig`._

kulesy:
## Summary
- Add optional `branch` / `mediaBranch` to `GraphQLConfig` (non-relative variant).
- `resolveMediaRelativeToCloud()` prefixes `/__staging/{encodeURIComponent(branch)}` when `branch` is set and differs from `mediaBranch`; production URL otherwise.
- `resolveMediaCloudToRelative()` strips a leading `/__staging/{branch}/` segment on save so documents always store the canonical relative path.
- Self-hosted (`useRelativeMedia: true`) and main-branch behaviour are unchanged.

Closes #6504. The host (TinaCloud) wiring that supplies `branch`/`mediaBranch` is tracked separately.

## Test plan
- [x] `pnpm vitest run` in `packages/@tinacms/graphql` — 403/403 pass, including 6 new `media-utils.test.ts` cases:
  - production URL when `branch === mediaBranch`
  - staging-prefixed URL when `branch !== mediaBranch`
  - production URL when `branch` is unset
  - staging prefix applied to every entry of an array value
  - round-trip strip of staging URL back to relative
  - production cloud URL untouched by stripper
- [ ] Integration verification once TinaCloud passes `branch`/`mediaBranch` into `GraphQLConfig`.


packages/@tinacms/graphql/src/resolver/media-utils.ts
```diff
@@ -92,6 +89,21 @@ export const resolveMediaRelativeToCloud = (
   }
 };
 
+const stagingPrefix = (config: {
+  branch?: string;
+  mediaBranch?: string;
+}): string =>
+  config.branch && config.branch !== config.mediaBranch
```
18-th, line 98: What's the expected behavior here if the mediaBranch is not set? Because currently if it is `undefined` and config.branch is not empty - we get staging

e.g. `some-feature && some-feature !== undefined` => true

Probably a minor edge case though

TinaInfo: @kulesy - Yes, that is the expected behaviour. If the user has set the branch and it doesn't correspond to the media branch (source of truth), it will get created under a staging branch.

We could add a validation check to ensure that if a branch is passed through (branch-aware enabled), the mediaBranch should always be set, but this is a very minor edge case.

18-th (approved):
LGTM

PR #6742, opened 2026-04-28
fix(create-tina-app): avoid GitHub API rate limits when scaffolding starter templates

KahaMason:
## Summary

partially addresses #6730

The `Build TinaCMS Starter Templates` workflow has been intermittently failing because `create-tina-app` was hitting GitHub's unauthenticated API rate limit (60 requests/hour). Each scaffold made one call to `GET api.github.com/repos/<owner>/<name>` solely to discover the starter's default branch, and when the workflow's matrix fanned out at the scheduled cron tick, those calls collided with the rate limit and surfaced as a generic *"Repository information not found"* error. To avoid hitting the limit, this PR removes the `getRepoInfo()` call (and its supporting helpers `isUrlOk`, `hasRepo`, `existsInRepo`) from the install path and replaces it with a `branch` field declared directly on each entry in the `TEMPLATES` config. The scaffold now goes straight to `codeload.github.com` for the tarball with zero `api.github.com` traffic, resolving the recurring workflow failures.

## Files

- **`packages/create-tina-app/src/templates.ts`** — declared `branch` on each starter and rewrote `downloadTemplate()` to use it instead of calling the GitHub API.
- **`packages/create-tina-app/src/util/examples.ts`** — removed `getRepoInfo`, `isUrlOk`, `hasRepo`, and `existsInRepo` now that the API lookup is gone.
- **`.changeset/many-bears-clap.md`** — patch-level changeset for `create-tina-app`.

## Trade-offs

- **Renaming a default branch on any starter repo** (`tina-nextjs-starter`, `tina-docs`, `tina-astro-starter`, `tina-hugo-starter`, `tina-remix-starter`, `tinasaurus`, `tina-barebones-starter`) now requires a matching update to `templates.ts` and a patch release of `create-tina-app`. The scheduled workflow will fail on that specific starter the next morning if it's missed, so detection is automatic — but the fix isn't.
- **Adding a new starter template** requires setting the `branch` field on the new entry. TypeScript will refuse to compile without it, so this is hard to forget, but worth flagging in any contributor docs covering "how to add a starter".
- **The `Template` type now diverges from the upstream `create-next-app` lineage** that `examples.ts` was originally ported from. Future cherry-picks of changes from Vercel's `helpers/examples.ts` will need manual reconciliation.
- **The job matrix in `build-starter-templates.yml`** still fans out to 63 cells (3 package managers × 7 templates × 3 Node versions) at the same cron tick. The rate-limit pressure is gone, but the matrix shape is worth a separate look for CI runtime cost — dropping the non-LTS Node version, capping concurrency with `max-parallel`, or curating via `include:` are all options.
- **`codeload.github.com` is now the only GitHub surface in the install path.** It has its own (much more permissive) anti-abuse limits. If GitHub ever tightens those or has an outage, failures will manifest as tarball download errors rather than the previous *"Repository information not found"* signature — worth knowing for triage.

## Verification

- [x] `pnpm --filter create-tina-app build` passes
- [x] No remaining references to `api.github.com` in the package source
- [x] Local smoke test: `npx create-tina-app --template basic` against the built CLI


18-th (approved):
LGTM

PR #6745, opened 2026-04-28
fix(cli): add absWorkingDir to esbuild calls to prevent Yarn PnP interference

coldbrow:
## Problem

When using TinaCMS in a monorepo environment, the CLI may fail to bundle
`tina/config.ts` with module-not-found errors even though all packages are
correctly installed.

The root cause: `esbuild.build()` calls in `loadConfigFile()` and
`loadDatabaseFile()` do not set `absWorkingDir`. Without it, esbuild walks up
the directory tree looking for configuration files. If a `.pnp.cjs` file from
a Yarn PnP workspace exists in a parent directory — which is common in
monorepos — esbuild detects it automatically and routes all module resolution
through PnP, breaking projects that are not themselves using Yarn PnP.

## Fix

Add `absWorkingDir: process.cwd()` to all `esbuild.build()` calls in
`loadConfigFile()` and `loadDatabaseFile()`. This pins esbuild's working
directory to the project root and prevents ancestor-directory traversal.

The double-bundle step in `loadConfigFile()` uses `path.dirname(outfile)`
since its entry point is a temp file written to `os.tmpdir()`.

## Changes
- loadDatabaseFile() — 1 call
- loadConfigFile() — 3 calls (prebuild, node bundle, double-bundle)

## How to reproduce

1. In a monorepo, have a Yarn PnP workspace in a parent directory (`.pnp.cjs`
   exists at or above the repo root)
2. Run `tinacms dev` or `tinacms build` in a non-PnP TinaCMS project within
   that monorepo
3. Module-not-found errors appear during esbuild bundling
4. After this fix: works correctly

## Why this is safe
- Keeps esbuild resolution scoped to the project root
- Prevents unintended interference from parent directories
- Does not affect projects explicitly using Yarn PnP


packages/@tinacms/cli/src/next/config-manager.ts
```diff
@@ -379,6 +379,7 @@ export class ConfigManager {
       format: 'esm',
       outfile: outfile,
       loader: loaders,
+      absWorkingDir: process.cwd(),
```
JackDevAU: could we use `this.rootPath` ? instead of process.cwd?

coldbrow: @JackDevAU Replaced all `process.cwd()` with `this.rootPath` so it also honours the --rootPath option.

packages/@tinacms/cli/src/next/config-manager.ts
```diff
@@ -393,6 +393,7 @@ export class ConfigManager {
       format: 'esm',
       outfile: outfile,
       loader: loaders,
+      absWorkingDir: this.rootPath,
```
kulesy, line 396: `this.rootPath` is not guaranteed absolute: `--rootPath` is a raw `Option.String` in `baseCommands.ts` and the constructor only passes it through `normalizePath`, which converts backslashes but leaves `./apps/web` relative. I reproduced the consequence against tina-cloud-starter with the published CLI: `tinacms build --rootPath ./tina-cloud-starter` builds fine today, and with this PR's line applied to that same CLI it fails with `ERR_CONFIG_LOAD_FAILED / The working directory "./tina-cloud-starter" is not an absolute path`, while an absolute `--rootPath` still passes. Worth resolving the path before handing it to esbuild; main already computes `path.resolve(this.rootPath)` in both loaders, so that value is there to reuse after a rebase.

kulesy, line 396: I do not think this can fix the reported bug, because esbuild anchors its PnP manifest lookup on `absWorkingDir` and walks *up* from there rather than treating it as a floor. Reproducing your scenario with esbuild 0.25.12 (a `.pnp.cjs` in the parent, a non-PnP project below it with its own `node_modules`, absolute entry point): leaving `absWorkingDir` unset with the process cwd outside the tree succeeds, setting it to the project root **fails** with `Could not resolve "foo"`, and setting it to `os.tmpdir()` succeeds. The middle case is what this PR does. Also worth noting that `rootPath` defaults to `process.cwd()` and esbuild's own default is the cwd its service was spawned with, so for anyone not passing `--rootPath` this is a no-op, and the original `process.cwd()` version was that default exactly. The direction that did work in my test is the opposite one, pointing esbuild at a working directory outside the PnP tree, which is safe because the bundle's imports still resolve from the importing file's own directory.

packages/@tinacms/cli/src/next/config-manager.ts
```diff
@@ -440,6 +441,7 @@ export class ConfigManager {
       outfile: preBuildConfigPath,
       loader: loaders,
       metafile: true,
+      absWorkingDir: this.rootPath,
```
kulesy, line 444: Flagging that this needs more than a mechanical rebase: the merge base is 6 May and a local test-merge shows this file is the one conflict. `loadDatabaseFile` no longer has an inline options object on main, it calls `buildDatabaseEsbuildConfig()` from `next/build-database-esbuild-config.ts`, whose externalize and output-path contract is deliberately locked down by unit tests (#6785), so adding `absWorkingDir` there means changing the helper and its test rather than a call site.

kulesy:
Hey @coldbrow, thanks for the report and the first pass at this.

I am closing this one because the approach cannot fix the bug as it stands: in my test, pinning absWorkingDir to the project root still let esbuild find the parent .pnp.cjs, and what did work was pointing esbuild outside the PnP tree. The file has also drifted a long way from main.

The bug is now tracked in https://github.com/tinacms/tinacms/issues/7617 with your write-up linked from it, so nothing here is lost.

Issue #6214, opened 2025-12-03
Update ColorPicker Swatch from a div to a button


18-th:
### Describe the issue that you're seeing. Any videos (eg. YouTube or Loom) or screenshots help!

Currently, the Color Picker Swatch uses a div for styling but behaves as a button. It needs to become a semantic button and match the current styling.

<img width="926" height="413" alt="Image" src="https://github.com/user-attachments/assets/0ca08cf2-538d-47a0-bec4-c4b9309172d9" />

### Reproduction

N/A

### Steps to reproduce

N/A

### System Info

```shell
N/A
```

### Validations

- [x] Follow our [Code of Conduct](https://github.com/tinacms/tinacms/blob/main/CODE_OF_CONDUCT.md)
- [x] Read the [docs](https://tina.io/docs).
- [x] Check that there isn't [already an issue](https://github.com/tinacms/tinacms/issues) that reports the same bug.

joshbermanssw:
✅ Done and merged 

<img width="1543" height="292" alt="Image" src="https://github.com/user-attachments/assets/8e7b2aef-78ed-4541-9630-e32f5abb02ac" />

**Figure: Colour swatch**

PR #6746, opened 2026-04-28
TinaCMS Colour Picker Swatch - change html element and type

joshbermanssw:
As per #6214 

- [ ] Changes html element from div to button 
- [ ] Makes type an extension of button


packages/tinacms/src/toolkit/fields/components/color-picker/color-picker.tsx
```diff
@@ -34,7 +34,8 @@ const getTextColorForBackground = function (backgroundColor?: ColorRGBA) {
 };
 
 const Swatch = ({ colorRGBA, colorFormat, width, ...props }: SwatchProps) => (
-  <div
+  <button
+    type='button'
     className='bg-gray-100 rounded shadow-[0_2px_3px_rgba(0,0,0,0.12)] cursor-pointer m-0'
```
18-th: Since this is the button now, do we still need the `cursor-pointer`?

packages/tinacms/src/toolkit/fields/components/color-picker/color-picker.tsx
```diff
@@ -34,7 +34,8 @@ const getTextColorForBackground = function (backgroundColor?: ColorRGBA) {
 };
 
 const Swatch = ({ colorRGBA, colorFormat, width, ...props }: SwatchProps) => (
-  <div
+  <button
```
18-th, line 37: div's display is block by default, button is inline-block. Could we check that it still respects the width correctly in all cases?

packages/tinacms/src/toolkit/fields/components/color-picker/color-picker.tsx
joshbermanssw: resolved

packages/tinacms/src/toolkit/fields/components/color-picker/color-picker.tsx
joshbermanssw: good catch; resolved 

Screenshot 2026-04-29 at 9 14 27 am

18-th (approved):
LGTM

Issue #6718, opened 2026-04-22
Add tinacms doctor CLI command to check for outdated Tina packages


joshbermanssw:
Part of #6719

## Problem

A `/versions` page on tina.io answers "what's current?" — but not "am *I* current?" To answer the second question today, a user has to eyeball the web page with their `package.json` open, or run `npm outdated` and mentally filter to Tina packages.

## Proposed solution

Add a `tinacms doctor` command (or a flag on an existing CLI command) that:

1. Reads the project's `package.json` and lockfile.
2. Filters to the TinaCMS family: `tinacms`, `@tinacms/*`, `tinacms-*`, `next-tinacms-*`, `create-tina-app`.
3. Fetches `dist-tags.latest` from `registry.npmjs.org` for each installed package in parallel.
4. Prints a table showing **installed version** vs. **latest**, with outdated rows highlighted.
5. Exits non-zero if anything is outdated, so it's CI-friendly.

## Prior art

- `storybook upgrade` / `storybook doctor` — reads `package.json`, filters to `@storybook/*`, diffs against registry.
- `nx migrate latest` — same pattern for `@nx/*`.
- `npm-check-updates` with `ncu -f 'tinacms,@tinacms/*,tinacms-*,next-tinacms-*'` — the generic version people use today ad-hoc.

## Relationship to `/versions` page

Independent and parallel — this is the CLI counterpart to the sibling `/versions` web page. Web page answers "what exists"; doctor answers "am I current."

joshbermanssw:
✅ Done - Example with `tina.io`

<img width="851" height="161" alt="Screenshot 2026-04-29 at 4 03 34 pm" src="https://github.com/user-attachments/assets/73a1c608-17f5-4f5a-8706-a22146ed6d86" />

**Figure: ✅ Currently highlights all the out-dated packages**

PR #6758, opened 2026-04-29
tinacms doctor package command

joshbermanssw:
As per #6718 

Adds `tinacms doctor` to check a project's direct TinaCMS package dependencies against the npm `latest` dist-tag.

The command now:

- reads Tina package declarations from `dependencies`, `devDependencies`, `optionalDependencies`, and `peerDependencies`
- filters package names matching `tinacms`, `@tinacms/*`, `tinacms-*`, `next-tinacms-*`, and `create-tina-app`
- resolves installed versions from `node_modules` first, then `package-lock.json`, `pnpm-lock.yaml`, or `yarn.lock`
- fetches npm `dist-tags.latest` for each package in parallel
- prints a table, supports `--json`, and exits non-zero when a package is outdated or cannot be checked
- documents the command in the CLI README

Example with `tina.io`

Screenshot 2026-04-29 at 4 03 34 pm

**Figure: ✅ Currently highlights all the out-dated packages**


18-th (approved):
LGTM

PR #6763, opened 2026-04-30
fix(tinacms): normalise slashes in editorial-workflow branch name

### LINKED ISSUE #6762: 🐛 Bug - Branch creation fails when saving new article
Requested by: @StefStar via [YakShaver.ai](https://YakShaver.ai) 🦬
cc: @StefStar, @kulesy, @JackDevAU

Hi Team!

&nbsp;&nbsp;[🟥&nbsp;&nbsp;Watch the video (1 min 5 sec)](https://www.youtube.com/watch?v=ZJjPd12YdIk)

&nbsp;&nbsp;Url from screen share: [https://ssw.com.au/admin/index.html#/collections/new/articles/-/](https://ssw.com.au/admin/index.html#/collections/new/articles/-/)

### Pain
Users are unable to save a new article to a new branch in ⭐ TinaCMS. After clicking "Save to new branch", the operation fails after several seconds with a "Branch operation failed" message. Browser console logs show an error related to failing to check the editorial workflow status, indicating a possible issue with async workflow execution or GitHub/TinaCloud integration.

### Acceptance Criteria
1. User can save a new article to a new branch without encountering a branch operation failure.
2. No "Branch operation failed" error is shown during normal save-to-new-branch operations.
3. Editorial workflow status check completes successfully without console errors.
4. If a failure occurs, the error message is actionable and includes relevant diagnostic information.
5. The issue is reproducible neither on TinaCloud-hosted nor self-hosted environments after the fix.

### Reproduce Steps
1. Open ⭐ TinaCMS admin UI.
2. Create or edit a new article with content filled in.
3. Click "Save" and choose "Save to new branch".
4. Wait approximately 5–10 seconds.
5. Observe the error message: "Branch operation failed. Please try again."
6. Open browser developer console and observe error: "Failed to check workflow status" related to async editorial workflow execution.

### Screenshot
Screenshot
**Figure: Saving an article to a new branch fails with a 'branch operation failed' error in TinaCMS**

kulesy:
Closes #6762

**TL;DR** Normalise branch names to valid Git refs at every branch creation site, and disable saving when the name normalises to empty, so "Save to new branch" no longer fails with a generic *Branch operation failed*.

**Pain:** A collection `path` with a trailing slash (e.g. `content/articles/` on ssw.com.au) makes the create form build file paths like `content/articles//foo.mdx`. The default branch name derived from that path becomes `tina/articles//foo`, an invalid Git ref, so the server-side workflow fails and the UI only surfaces *Branch operation failed* plus a `Failed to check workflow status` console error. User-typed names containing spaces or other Git-forbidden characters failed the same way, and the `formatBranchName` helper was duplicated with diverged behaviour (the legacy branch switcher deleted invalid characters instead of hyphenating) while collapsing slash runs nowhere.

**Solution:** Centralise the naming helpers in `@utils/branch-name` (`formatDefaultBranchName`, `normalizeBranchName`, and the previously duplicated `formatBranchName`) and normalise to `check-ref-format` rules at every creation site: the save-to-new-branch modal, both branch switchers, and the deleted-branch recovery modal. Normalisation collapses repeated and leading/trailing slashes, turns characters Git forbids in refs (whitespace, control characters, `~ ^ : ? * [ \` and the `@{` sequence) into hyphens, collapses `..` runs, and strips leading dots and trailing `.` / `.lock` per path component, using loop-based trims rather than the anchored regexes CodeQL flags. Save buttons are disabled while the name normalises to an empty string (e.g. `///`), and the helpers have unit tests (18 passing).


packages/tinacms/src/toolkit/form-builder/create-branch-modal.tsx
```diff
@@ -43,6 +45,9 @@ const formatDefaultBranchName = (
   return result;
 };
 
+const normalizeBranchSlashes = (name: string): string =>
```
JackDevAU: Can we reuse the other functions in the codebase that do this. If they are all over the place can we create a central util for this. I believe it is called `formatBranchName`.

packages/tinacms/src/toolkit/form-builder/create-branch-modal.tsx
```diff
@@ -111,7 +116,7 @@ export const CreateBranchModal = ({
     setIsBranchGuardChecking(false);
 
     const success = await executeWorkflow({
-      branchName: `tina/${newBranchName}`,
+      branchName: `tina/${normalizeBranchSlashes(newBranchName)}`,
```
JackDevAU: Would entering only slashes (///) normalize to '' ? If so... the button's newBranchName === '' guard stays enabled and builds the invalid ref `tina/` - See if we can guard on the normalized value instead

JackDevAU (changes requested):

kulesy:
**Test plan**
- [x] Unit tests: `vitest run src/toolkit/form-builder/branch-name.test.ts` (15 passing)
- [ ] Configure a collection whose `path` has a trailing slash (or use a filename starting with `/`)
- [ ] Edit/create a document, click **Save to a new branch**, and confirm the prefilled name has no doubled or leading/trailing slashes
- [ ] Confirm the editorial workflow completes and a PR opens on the resulting branch
- [ ] Type `//foo bar//` into the branch input and confirm the resulting branch is `tina/foo-bar`
- [ ] Type `///` and confirm the save button is disabled

packages/tinacms/src/toolkit/form-builder/create-branch-modal.tsx
kulesy: Done in 6de433c - good call, it existed as two diverged copies (the main switcher hyphenates invalid chars, legacy deleted them, neither collapsed slash runs). All helpers now live in @utils/branch-name, both switchers and both workflow modals share it, and every creation site funnels through normalizeBranchName so invalid refs can't reach Git.

packages/tinacms/src/toolkit/form-builder/create-branch-modal.tsx
kulesy: Fixed in 7513391 - the modal derives normalizedBranchName and both submits with and disables on it, so /// can't be submitted; 6de433c applies the same guard to the deleted-branch recovery modal.

packages/tinacms/src/toolkit/plugin-branch-switcher/branch-switcher.tsx
joshbermanssw: Any reason we didnt include the normalizedBranchName(...) check here?

packages/tinacms/src/toolkit/plugin-branch-switcher/branch-switcher-legacy.tsx
joshbermanssw: Any reason we didnt include the normalizedBranchName(...) check here?

packages/tinacms/src/toolkit/plugin-branch-switcher/branch-switcher.tsx
kulesy: Good catch, no reason - fixed in d71b47522. Both creators now derive `normalizedBranchName` once and use it for the disabled guard and the submitted value. Worth noting the editorial one was the worse of the two: `///` normalised to a bare `tina` ref, which would D/F-conflict with every `tina/*` branch. Added component tests covering both.

packages/tinacms/src/toolkit/plugin-branch-switcher/branch-switcher-legacy.tsx
kulesy: Good catch, no reason - fixed in d71b47522. Both creators now derive `normalizedBranchName` once and use it for the disabled guard and the submitted value. Worth noting the editorial one was the worse of the two: `///` normalised to a bare `tina` ref, which would D/F-conflict with every `tina/*` branch. Added component tests covering both.

packages/tinacms/src/toolkit/plugin-branch-switcher/branch-switcher.tsx
```diff
@@ -334,9 +343,9 @@ const BranchCreator = ({ setViewState, handleCreateBranch, currentBranch }) => {
           variant='primary'
           type='submit'
           style={{ flexGrow: 2 }}
-          disabled={branchName === ''}
+          disabled={normalizedBranchName === ''}
           onClick={() => {
```
joshbermanssw, line 349: This guard only holds for mouse clicks.

The toolkit `Button` destructures `disabled` out of `props` and uses it purely for styling (`pointer-events-none opacity-30 cursor-not-allowed`, `styles/button.tsx:30`) — it never reaches the DOM, so the button stays focusable and activatable. `pointer-events-none` blocks the pointer, not the keyboard.

Since this one is `type="submit"` inside a `<form>` with no `onSubmit`, typing `///` and pressing Enter triggers implicit form submission, fires this `onClick`, and creates a branch named `tina/`.

The new test picks this up indirectly (`isDisabled()` has to check for `cursor-not-allowed` rather than the attribute) but stops short of the consequence.

Cheapest fix in scope is an early return rather than touching `Button`:

```ts
onClick={() => {
  if (!normalizedBranchName) return;
  handleCreateBranch("tina/" + normalizedBranchName);
}}
```

Same applies to `CreateBranch` in `branch-switcher-legacy.tsx`.

joshbermanssw: Thanks for the quick turnaround — but I think this is fixed at the wrong layer.

The actual defect is `styles/button.tsx:30`: `disabled` is destructured out of `props` and only used to pick styling classes, so it never reaches the element. Guarding in the handler treats the symptom at two call sites and leaves the cause in place.

What that costs us here:

- the same condition now lives twice per button — `disabled={normalizedBranchName === ''}` for styling and `if (!normalizedBranchName) return;` for behaviour — with a verbatim-duplicated comment in two files explaining why
- the test has to assert on a Tailwind class (`className.includes('cursor-not-allowed')`); renaming it to `isStyledDisabled` documents the smell rather than removing it
- there are ~46 `disabled={` call sites across `src/toolkit`, and every one of them that renders a toolkit `Button` has this same hole, still unfixed

Forwarding it on the `<Tag>` in `Button`:

```diff
       )}
+      disabled={disabled}
       {...props}
```

fixes all of them, and lets both guards, both comments and `isStyledDisabled` be deleted — the test becomes `expect(createButton()).toBeDisabled()`.

Worth checking whether any existing caller relies on a "looks disabled but still clickable" button before flipping it, but I doubt it.

The `onSubmit={(e) => e.preventDefault()}` on the form is a good catch and should stay either way.

kulesy: Agreed on the root cause - the early return treats the symptom. I tested your diff and it works: the DOM attribute lands, both activation paths go dead, all 633 unit tests pass, and it makes both guards plus `isStyledDisabled` deletable. I would still rather it landed separately, because the bulk of its value is the ~46 unrelated call sites and I do not want the fix for a user-facing branch-naming failure reverted over a regression in the media manager or sidebar. It also needs more than the one-liner: `IconButton` has the identical bug, `Button` is polymorphic so the three `as="a"` sites want `Tag === "button" ? disabled : undefined` rather than a bare forward, and `busy` has the same keyboard hole (double-submit). Raising it as its own PBI with that evidence attached. The `onSubmit` preventDefault stays here either way.

joshbermanssw (approved):

PR #6765, opened 2026-04-30
feat(tinacms): pass branch on all TinaMediaStore cloud calls

### LINKED ISSUE #6503: CMS Client — Pass Branch on All Media Calls
## Background

**`TinaMediaStore`** (`packages/tinacms/src/toolkit/core/media-store.default.ts`): the default cloud media store used by the TinaCMS editor. It exposes `persist()` (upload), `list()`, and `delete()` methods that call the TinaCloud assets-api via HTTP.

**Branch encoding:** `TinaMediaStore` stores the current branch as `this.api.branch`, which is already URL-encoded via `encodeURIComponent(branchName)` (set in `setBranch()` in `internalClient/index.ts`). When appending it to a query parameter, decode it first with `decodeURIComponent` and re-encode with `encodeURIComponent` to avoid double-encoding (e.g. `feat%2Fmy-branch` → `feat/my-branch` → `feat%2Fmy-branch` remains correct, but naively appending the pre-encoded string to a URL that also encodes query params would produce `feat%252Fmy-branch`).

**`MediaManager`**: the TinaCMS class that coordinates UI interactions with the media store. The `rename` stub is reserved on the `MediaStore` interface but `MediaManager` does not need to be updated in this issue — no rename UI or routing is added.

## Pain

The TinaCMS media store (`TinaMediaStore`) always calls the assets-api without identifying which branch the editor is on. Even after the assets-api is fully branch-aware, the client continues sending branch-agnostic requests — uploads go to production, listings read from production, deletions hit production. The branch context is already available on the client (`this.api.branch`) but is never forwarded.

## Description

Add `?branch=` to all three cloud media API calls in `TinaMediaStore`. The branch is already stored as `encodeURIComponent(branchName)` on the client (set in `setBranch()`, `internalClient/index.ts:167`). Decode then re-encode when constructing the query param to avoid double-encoding.

**`persist_cloud()`:** append `?branch={encodedBranch}` to the upload URL.

**`list()`:** append `?branch={encodedBranch}` to the list URL.

**`delete()`:** append `?branch={encodedBranch}` to the delete URL.

**`MediaStore` interface extension (`media.ts`):** add an optional `rename?` hook stub — `rename?(from: string, to: string): Promise<Media>` — as a reserved extension point. No implementation is added to `TinaMediaStore` or `MediaManager` in this issue. When the assets-api rename endpoint is built, the implementation goes in `TinaMediaStore.rename()` calling `POST ${this.url}/rename` with `{ from, to, branch }`.

**Local mode:** no changes. `persist_local()`, `list()` for local, and `delete()` for local all use different URL patterns and are unaffected.

**Backward compatibility:** the `?branch=` parameter is ignored by assets-api versions that don't parse it, so this change is safe to deploy before the assets-api is updated.

## Affected Files

**TinaCMS:**
- `packages/tinacms/src/toolkit/core/media-store.default.ts` — add `?branch=` to `persist_cloud()`, `list()`, `delete()`
- `packages/tinacms/src/toolkit/core/media.ts` — add optional `rename?` stub to `MediaStore` interface

## Acceptance Criteria

- [ ] `persist_cloud()` includes `?branch={encodedBranch}` in the upload URL
- [ ] `list()` includes `?branch={encodedBranch}` in the list URL
- [ ] `delete()` includes `?branch={encodedBranch}` in the delete URL
- [ ] Branch names with special characters (e.g. `/`) are correctly encoded
- [ ] Local mode methods are not affected
- [ ] `MediaStore` interface has optional `rename?(from, to): Promise<Media>` stub
- [ ] When on main branch, assets-api receives the main branch name; the server's `isBranchRequest()` helper treats it as a production request (unchanged behaviour) because `branch === app.mediaBranch`

## Depends On

- [Assets-API Events and Routes](https://github.com/tinacms/tinacloud/issues/3330)

kulesy:
## Summary

Closes #6503.

This PR makes media operations branch-aware end-to-end:

1. **Forward the branch on every cloud media call.** `TinaMediaStore`'s `upload_url`, `list`, and `delete` requests now carry `?branch=<encodedBranch>`, so once an app opts into branch-aware media those operations are scoped to the editor's current branch instead of always hitting production.
2. **Fix staging URL handling for multi-segment branches.** `@tinacms/graphql`'s media URL resolver now formats staging URLs as `/__staging/<branch>/__file/<path>` instead of `/__staging/<encoded-branch>/<path>`. The previous form broke for branches containing `/` (e.g. `feat/my-branch`) because CloudFront decodes paths before downstream components see them — the S3 write key (with a literal `%2F`) wouldn't match the decoded read path. The `__file` delimiter lets the branch contribute its natural `/` segments while still marking where the file path begins.
3. **Refresh from the server after upload.** The media manager now drops its locally-constructed entries and re-fetches the list, since the server is the source of truth for the canonical `src` URL (including the staging-branch path when applicable). Constructing it on the client would mean mirroring the server's branch-aware path logic plus the per-stage CDN host, which is fragile to keep in sync.

### Branch-forwarding details

- `persist_cloud()` — appends `?branch=<encodedBranch>` to the `upload_url` request, then returns `[]` so the media manager's refresh path runs (see point 3 above)
- `list()` cloud branch — appends `&branch=<encodedBranch>` to the list URL
- `delete()` cloud branch — appends `?branch=<encodedBranch>` to the delete URL

`this.api.branch` is already URL-encoded by [`Client.setBranch()`](https://github.com/tinacms/tinacms/blob/main/packages/tinacms/src/internalClient/index.ts#L167). A new private `encodedBranchParam()` helper decodes then re-encodes at the use site so the value is always single-encoded, even if the URL containing it is later re-processed.

The local-mode code paths (`persist_local`, the local-mode list branch, the local-mode delete branch) are untouched.

Also reserves an optional `rename?(from: string, to: string): Promise<Media>` hook on the `MediaStore` interface as a future extension point — no implementation in `TinaMediaStore`, no surfacing in `MediaManager`.

## Backward compatibility

Older assets-api versions ignore unknown query parameters, so this can ship before [tinacms/tinacloud#3330](https://github.com/tinacms/tinacloud/issues/3330) lands. When the flag isn't enabled on an app, the server's [`isBranchRequest()`](https://github.com/tinacms/tinacloud/blob/main/js/assets-api/src/api/koa-admin.ts#L16) helper falls through to production behaviour — and that fall-through is regression-tested for all three endpoints by [tinacms/tinacloud#3472](https://github.com/tinacms/tinacloud/pull/3472).

## Test plan

- [x] `pnpm --filter tinacms build` — clean
- [x] `pnpm --filter tinacms test` — 321 passed, 5 skipped (no regressions)
- [x] Manual diff inspection: each of the three cloud URLs now ends with `?branch=…` or `&branch=…` exactly once; local-mode branches are untouched
- [x] Encoding round-trip: `feat/my-branch` (stored as `feat%2Fmy-branch`) decodes to `feat/my-branch` and re-encodes back to `feat%2Fmy-branch`


18-th (dismissed):
LGTM

18-th (approved):
LGTM

PR #6768, opened 2026-05-01
fix(@tinacms/graphql): await async writes in OutputBridge to fix flaky snapshot tests

kulesy:
## Summary

Fixes the flaky `tests/is-title` snapshot test (and the same latent bug in `tests/is-body`) seen in CI runs like [actions/runs/25143506962](https://github.com/tinacms/tinacms/actions/runs/25143506962/job/73698320188?pr=6763).

- **Root cause:** `OutputBridge.put()` in `tests/util.ts` called `super.put()` (a `fs.outputFile` wrapper) without `await`, leaving the disk write fire-and-forget. The subsequent `toMatchFileSnapshot('out.md')` could read the file before the write finished — most likely to flake on macOS CI under load.
- Added `await` to `super.put()` in `OutputBridge` (the actual fix).
- Also added `await` to all 63 `toMatchFileSnapshot` calls across the test suite. Vitest 2.x auto-awaits them at test end, but without explicit `await` there are no ordering guarantees, and Vitest 3 will reject un-awaited assertions outright.

## Test plan

- [x] `pnpm vitest run tests/is-title tests/is-body` passes
- [x] Full `pnpm vitest run tests/` passes (31 files, 99 tests)
- [ ] Re-run the previously flaky CI job to confirm


JackDevAU:
@copilot resolve the merge conflicts in this pull request

JackDevAU (approved):
:shipit:

PR #6771, opened 2026-05-04
feat(@tinacms/astro): vanilla-JS visual editing bridge for non-React frontends

### LINKED ISSUE #6774: 🤖 Visual editing for Astro + framework-agnostic bridge
## User story

> **As an** Astro developer using TinaCMS,
> **I want** to use TinaCMS's visual editing (click-to-focus, live preview as I type) on my Astro pages
> **without shipping React in the page tree,**
> **so that** I keep Astro's "ship as little JS as possible" guarantee while still getting full editing UX for content teams.

> **As a** TinaCMS maintainer,
> **I want** the editing protocol to be framework-agnostic
> **so that** future integrations (Hugo, Eleventy, plain HTML, anything else) reuse the same bridge instead of each rebuilding the postMessage / form-payload / refresh pipeline from scratch.

---

## Background

Visual editing has always required React. The existing flow runs through `useTina()` — a React hook that subscribes to admin postMessages and re-renders the page tree. For Astro this meant either accepting a React bundle on every editable page, or accepting that the CMS lived in a separate admin tab with no live preview.

This is the most-requested gap from the Astro community. Astro users consistently chose other CMSes when their alternative was "ship React for editing." Closing it unlocks Astro as a first-class TinaCMS target.

---

### 1. New package: `@tinacms/bridge`
- Vanilla TypeScript, ESM, zero runtime dependencies
- Speaks the existing TinaCMS admin postMessage protocol unchanged (`open` / `close` / `updateData` / `field:selected` / `quick-edit` / `isEditMode` / `quickEditEnabled`)
- ~2 KB gzipped, self-bails when `window.parent === window` (cheap dead code in production)
- Submodules: edit-mode handshake, form registration with retry, data store, click-to-focus, per-island refresh
- Subpath exports: `@tinacms/bridge`, `@tinacms/bridge/tina-field`, `@tinacms/bridge/preview`

### 2. Extracted helper: `tinacms/tina-field`
- Move `tinaField()` out of the React-coupled `tinacms/dist/react` to its own subpath
- Existing `tinacms/dist/react` keeps re-exporting → no breaking change
- Lets non-React frontends import the helper without pulling React (and Plate, dnd-kit, etc.) into their bundle

### 3. Reference example: `examples/astro/visual-editing/`
- Mirrors `examples/astro/kitchen-sink/` field-for-field — same six collections (Tag, Author, Global, Post, Blog, Page), same shared content via `localContentPath`, same eight routes
- Pure Astro components, no `client:*` directives in production
- Production parity with kitchen-sink visually; zero bridge JS shipped to non-admin visitors

### 4. Stateless overlay protocol
- POST + JSON body via `Content-Type: application/x-tina-preview+json`
- Works against any GraphQL backend — self-hosted Tina, TinaCloud, anything that exposes `client.queries.X()`

### 5. Documentation
- README in the example app with architecture diagram, integration play-by-play, hosting matrix, "adding a new editable region" walkthrough, "adding a custom MDX component" walkthrough, side-by-side feature table vs `astro/kitchen-sink`
- Blog post draft for tina.io (`_docs/blog-draft-astro-visual-editing.md`)

---

## Out of scope (follow-up PBIs)

| # | Follow-up | Notes |
|---|-----------|-------|
| 1 | **Alternative framework bridge** | Uses same protocol to enable another popular framework |
| 2 | **Standalone `@tinacms/rich-text-astro` package** | Extract the in-example rich-text renderer once the API stabilises. |
| 3 | **TinaCloud overlay channel** | Shouldn't be needed — the stateless POST protocol works against any backend. Open if/when CDN-cached production previews become a use case. |
| 4 | **Astro view transitions** | Use Astro 5's built-in view transitions for swap polish on island refresh. Nice-to-have. |
| 5 | **Eleventy / plain-HTML adapter** | The contract is framework-free; just needs an integration guide and a sample. Demand-driven. |
| 6 | **Per-island scope tracking** | Currently every island refetches on any updateData. Could route updateData by id and refetch only matching islands. Optimization, not correctness. |

---

### Bridge package
- [x] `pnpm test --filter=@tinacms/bridge` passes (18 tests covering UTF-8, large payloads, multi-form, malformed input, content-type matching, full bridge→server round-trip)
- [x] Bridge bundle ≤ 3 KB gzipped (currently ~2 KB)
- [x] Bridge has zero runtime dependencies (verified via `package.json`)
- [x] Bridge is a no-op outside iframes (`window.parent === window` check before any postMessage)
- [x] `@tinacms/bridge/preview` is consumable from Node, Bun, Cloudflare Workers (uses standard Web `Request` / `Response` only — no `Buffer`, `fs`, or other Node APIs)

### `tinacms/tina-field` extraction
- [x] `pnpm test --filter=tinacms` passes (321 tests)
- [x] `tinacms/dist/react` continues to re-export `tinaField` (no breaking change)
- [x] New subpath export `tinacms/tina-field` resolves to a standalone module with no React imports

### Astro example
- [x] All eight routes return 200 (`/`, `/:slug`, `/posts/`, `/posts/:slug`, `/blog/`, `/blog/:filename`, `/authors/`, `/authors/:filename`)
- [x] All six collections list in admin sidebar
- [x] Edit a hero headline → block re-renders within ~300ms (per-island scoped, verified in Network tab)
- [x] Click a marked element → admin sidebar focuses the matching form field
- [x] POST overlay round-trips UTF-8 (em-dashes, smart quotes, emoji) without 431 / encoding errors
- [x] `pnpm build:local` succeeds; production HTML ships ~2 KB gzipped of bridge JS, no other runtime
- [x] Adding a new editable region requires editing exactly one file (`src/lib/islands.ts`)
- [x] Custom MDX components (BlockQuote, NewsletterSignup, DateTime) render via Astro renderer when content uses MDX-template syntax

### Code quality
- [x] `pnpm format` (root) passes — no parse errors, no formatting drift
- [x] `pnpm biome check` passes for the bridge package and the example
- [x] Bridge submodules have file-level docstrings explaining why they exist + which React behaviour they mirror
- [x] No regressions: `astro/kitchen-sink`, `next/kitchen-sink`, `react/kitchen-sink` untouched and still passing

### Documentation
- [x] Example README documents architecture, marker contract, stateless overlay, hosting requirements, migration path
- [x] Changeset entry (minor bump on both `@tinacms/bridge` and `tinacms`) covers what shipped + the audience framing
- [ ] Update astro template to enable visual editing
- [ ] Blog post draft ready for tina.io
- [ ] Docs updated on tina.io/docs
- [ ] Discord / Twitter announcement
- [ ] Walkthrough recorded for the TinaCMS YouTube channel

---

## Risks & mitigations

| Risk | Likelihood | Impact | Mitigation |
|------|------------|--------|------------|
| Astro's hoisted-script bundling silently drops conditional `<script>` tags in prod build | Realised during dev | High (bridge wouldn't load) | Switched to always-load + runtime iframe check (~2 KB cost). Documented in Base.astro comment. |
| HTTP header size cap (431) when overlay carries large rich-text body | Realised during dev | High (live update broken for posts) | Switched transport from header to POST body. Tests cover >40 KB payloads. |
| Latin-1 header restriction trips UTF-8 content | Realised during dev | High (live update broken for any em-dash) | Same fix as above (POST body round-trips UTF-8 natively). |
| Admin's form-builder needs `... on Document { _sys, id }` selection to attach `_tina_metadata` | Realised during dev | Medium (click-to-focus broken) | Hand-rolled queries replaced with the auto-generated query string verbatim, extracted from `tina/__generated__/queries.gql` at build time. |
| GraphQL rejects unused fragments → admin error modal on form expand | Realised during dev | Medium (admin unusable) | Walk fragment dependencies starting from the query body and emit only transitively-referenced fragments. |
| Hand-authored `<Component>` syntax in markdown bodies parses as `html` (not `mdxJsxFlowElement`) | Known limitation | Low (cosmetic — same as React renderer) | Documented in README + changeset. Content authored via the Tina editor inserts proper template references. |
| Other framework integrations diverge from the protocol over time | Open | Medium (fragments the contract) | `@tinacms/bridge/preview` and the postMessage shape are the contract; future integrations should share these and add per-framework shims only on top. |

wicksipedia:
Closes the "you have to ship React for editing" gap that's been keeping Astro users off TinaCMS. Three packages land together: a framework-agnostic bridge, an Astro integration that wraps it, and a reference example.

## Why

Visual editing has always required React via `useTina()` - the hook subscribes to admin postMessages and re-renders the page tree. For Astro that meant either accepting a React bundle on every editable page or accepting that the CMS lived in a separate admin tab with no live preview. Astro grew because authors care about runtime cost; asking them to ship React is exactly the tradeoff they picked Astro to avoid.

This is the most-requested gap from the Astro community. Closing it unlocks Astro as a first-class TinaCMS target and gives every future framework integration (Hugo, Eleventy, plain HTML) a shared protocol to build against instead of each one rebuilding the postMessage / form-payload / refresh pipeline from scratch.

## What's shipping

**`@tinacms/bridge`** 

- ~2 KB gzipped, zero-dep ESM that speaks the existing admin postMessage protocol.
- Bails when not in an iframe so production visitors pay nothing.
- Protocol is **stateless** -  admin pushes already-resolved data to the bridge, bridge POSTs it to a per-island refresh endpoint, the endpoint re-renders the matching component with the unsaved overlay, the bridge swaps the HTML in place.

**`@tinacms/astro`**

- Default export: an Astro `TinaMarkdown` with the same `content`/`components` contract as the React version, emitting pure HTML
- `/bridge`, `/tina-field`, `/preview` re-export `@tinacms/bridge` so Astro consumers install one package
- `/types` + `/sanitize` for renderer types and URL safety helpers
- Ships `.astro` source uncompiled — the consumer's Astro pipeline compiles it. `@tinacms/bridge` stays publishable on its own for non-Astro adopters.

**`tinacms/tina-field`**

— the existing helper, now exported as a standalone subpath so non-React frontends can import it without pulling React, Plate, dnd-kit, etc. into their bundle.
- `tinacms/dist/react` keeps re-exporting it; no breaking change.

**`examples/astro/visual-editing`**

— reference example mirroring `examples/astro/kitchen-sink` but rendered with pure Astro components.
- Includes a single dynamic island endpoint backed by a registry, a `withOverlay()` helper that shares the load path between production and in-editor, and recipes for adding a new editable region or a custom MDX component.


joshbermanssw:
init() doesn't survive Astro view transitions — please expose a per-navigation re-scan

  Trying this out on a ClientRouter-enabled Astro site (the Astro view-transitions integration). When the admin iframe navigates between pages — e.g. /admin/index.html#/~/ →
  /admin/index.html#/~/about — the new page's <script type="application/tina+json"> payload is swapped into the DOM, but the admin sidebar never updates: it keeps showing the form
  for the previous doc.

  Root cause is in packages/@tinacms/bridge/src/index.ts:

  let initialized = false;
  function init(options = {}) {
    if (initialized) return;
    initialized = true;
    …
  }

  initForms runs once and captures payloads in closure, so subsequent navigations have no way to make the bridge re-read the new <script type="application/tina+json"> blocks.
  Calling init() a second time on astro:page-load is a no-op, and removing the guard would double-bind the click listener, double-subscribe to the data store, and start racing
  announce() loops.

  Workaround for now is to drop ClientRouter from the layout and accept full reloads inside the iframe, but view transitions are a first-class Astro feature people will expect to
  keep using on the public site.

  Suggested API: a separate exported refreshForms() (or setActiveForms()) that:

  1. Re-queries script[type="application/tina+json"]
  2. Diffs against the last active set, posts close for forms that disappeared and open for forms that appeared
  3. Re-seeds the store with the new payload data

  Caller wiring would be:

  import { init, refreshForms } from '@tinacms/astro/bridge';
  init();
  document.addEventListener('astro:page-load', refreshForms);

  This keeps init()'s one-time global listeners (click capture, store subscription, message handlers) untouched and only re-runs the form-scan/open-close cycle, which is the part
  that's actually per-page.

JackDevAU (dismissed):
:shipit:

JackDevAU (approved):
:shipit:

Issue #6789, opened 2026-05-08
Promote starter-template smoke subset to PR-blocking CI


kulesy:
## Context

`build-starter-templates.yml` runs the full `create-tina-app` → install → build flow across 7 templates × 3 package managers × 3 Node versions. Today it's **nightly cron only** with `continue-on-error: true` — failures auto-file an issue but don't block PRs. So a #6750-style regression in the user-facing init flow could ship to main and only be discovered the next morning.

This sub-issue adds a **slim, fast subset** of that matrix to `main.yml` so the canonical happy-path is verified on every PR. The full nightly matrix stays as the broader safety net.

## What

Add a `template-smoke` job to `.github/workflows/main.yml`:

```yaml
template-smoke:
  needs: build
  strategy:
    fail-fast: false
    matrix:
      template: [basic, tina-nextjs-starter]
      package-manager: [pnpm, npm]
  runs-on: ubuntu-latest
  steps:
    - setup node + pm
    - run: npx create-tina-app@latest test-repo --pkg-manager ${{ matrix.package-manager }} --template ${{ matrix.template }} --noTelemetry
    - run: cd test-repo && ${{ matrix.package-manager }} install && ${{ matrix.package-manager }} run build
```

= 4 jobs per PR. Total runtime ~5 min.

## Acceptance criteria

- [ ] New `template-smoke` job in `main.yml`
- [ ] Matrix: `basic` + `tina-nextjs-starter`
- [ ] Matrix: pnpm + npm (yarn stays nightly-only)
- [ ] Ubuntu only (cross-OS coverage handled by the self-hosted-sqlite job)
- [ ] Node version from `.nvmrc` (current default — not a multi-Node matrix)
- [ ] **Blocking** (no `continue-on-error`)
- [ ] Job completes in under 8 minutes wall-clock
- [ ] Existing nightly `build-starter-templates.yml` left untouched as the broader matrix
- [ ] Documented in repo README / contributing docs that PR CI now includes template smoke tests

## Catches

- Template `package.json` regressions that break `create-tina-app`
- New CLI bugs that only manifest in a fresh project layout (vs. the monorepo dev environment)
- Package-manager-specific resolution issues that the existing examples don't exercise (because they're inside the monorepo with hoisted deps)

## Notes

Why only 2 templates × 2 PMs on the critical path:
- `basic` and `tina-nextjs-starter` are by far the most-used starting points
- yarn classic is on the way out; pnpm + npm cover ~95% of users
[...]

Issue #6788, opened 2026-05-08
Add adapter compatibility test matrix (sqlite-level, mongodb-level, memory-level, TinaLevelClient)


kulesy:
Part of #6750.

## Context

The motivating bug is "user picks adapter X, build breaks." Externalizing changes how every CJS package enters the runtime — they go through Node's CJS loader instead of esbuild's interop wrapper. CJS default-export semantics differ subtly between the two, and we already see defensive coding for this in the codebase:

```ts
// packages/@tinacms/search/src/client/index.ts:11
import * as sqliteLevelModule from 'sqlite-level';
const SqliteLevel =
  (sqliteLevelModule as any).default?.SqliteLevel ??
  (sqliteLevelModule as any).SqliteLevel;
```

That `default?.X ?? X` pattern hints that some adapters work bundled but might surprise when externalized. We need explicit coverage per adapter so any regression in CJS interop is caught immediately and named.

## What

Add a parameterized test (in `config-manager.test.ts` or a sibling file) that runs `loadDatabaseFile` against a fixture per adapter, then asserts the loaded default export is a working adapter via a put/get round-trip.

Add a new `adapter-matrix` job to `main.yml`:

```yaml
adapter-matrix:
  needs: build
  strategy:
    fail-fast: false
    matrix:
      adapter: [sqlite-level, mongodb-level, memory-level, tina-level-client]
  runs-on: ubuntu-latest
  steps:
    - run: pnpm test:adapters --adapter=${{ matrix.adapter }}
```

## Acceptance criteria

- [ ] Fixture `tina/database.ts` per adapter:
  - [ ] `sqlite-level` (the motivating native-module case)
  - [ ] `mongodb-level` (current demo's adapter — pure-JS regression check)
  - [ ] `memory-level` (no-op baseline)
  - [ ] `TinaLevelClient` via `createLocalDatabase()` (local dev path)
- [ ] Each fixture builds via `loadDatabaseFile`, instantiates the default export, performs put + get of a known key, asserts the returned value matches
- [ ] `mongodb-level` test runs against an in-memory or test-container Mongo — must work in CI without external services or be cleanly mocked
- [ ] `pnpm test:adapters` script wired up in `packages/@tinacms/cli/package.json`
- [ ] New `adapter-matrix` job in `main.yml` runs on every PR
- [ ] Job is blocking (not `continue-on-error`)
- [ ] Job runs in under 5 minutes total (faster than the full e2e workflow)

## Catches

[...]

Issue #6787, opened 2026-05-08
Add unit tests for config-manager.ts (loadDatabaseFile + loadConfigFile)


kulesy:
Part of #6750.

## Context

`packages/@tinacms/cli/src/next/config-manager.ts` has **zero** test coverage today. It's the file being changed by #6750 (output relocation + `packages: 'external'`), so we're flying blind on the unit-level invariants. This sub-issue adds focused tests against real `esbuild` runs with fixture inputs.

## What

Create `packages/@tinacms/cli/src/next/config-manager.test.ts`. Use real esbuild (not mocked) against on-disk fixtures, then assert on the produced bundle and side effects.

## Acceptance criteria

Each bullet is one test case (or small group):

- [ ] **Output location** — assert build file lands inside `<project>/.tina/__generated__/.cache/<timestamp>/`, not `os.tmpdir()`
- [ ] **No node_modules inlined** — fixture importing `sqlite-level`; read the generated bundle and assert it contains a literal `from 'sqlite-level'` import statement, not the package's source
- [ ] **Local files still bundled** — fixture `tina/database.ts` importing `./helpers.ts`; assert helper code IS inlined (relative imports remain bundled)
- [ ] **Cache cleanup post-load** — after `loadDatabaseFile()` resolves, assert the timestamp dir is removed
- [ ] **Cache sweep on startup** — pre-create stale `.cache/<old-timestamp>/` dirs, instantiate the manager, assert they're removed
- [ ] **Crash residue tolerated** — leave a `.cache/<timestamp>/` dir from a "crashed" prior run (no lockfile, partial files), assert next load proceeds and sweeps it
- [ ] **Read-only project fallback** — chmod the project dir read-only, assert behavior matches the design decision (clean error OR documented fallback to `os.tmpdir()` with bundled deps)
- [ ] **Cache-busting** — call `loadDatabaseFile()` twice with the same source between calls; assert both resolve and the second returns a freshly-evaluated module (not a cached instance)
- [ ] **Watch list excludes node_modules** — fixture importing both local `./schema.ts` and external `sqlite-level`; assert returned `watchList` contains `./schema.ts` but not anything under `node_modules/`
- [ ] **`createRequire` banner still injected** — verify the ESM bundle still includes the `createRequire(import.meta.url)` banner for any remaining bundled CJS paths
- [ ] All tests pass on Linux + macOS + Windows in CI

## Notes

[...]

Issue #6786, opened 2026-05-08
Add self-hosted SQLite example + PR-blocking CI matrix


kulesy:
Part of #6750.

## Context

The current CI has no example using `sqlite-level` — the only self-hosted demo (`examples/next/tina-self-hosted-demo`) uses `mongodb-level`, which is pure JS. This means the entire bug class from #6675 / #6750 (native modules, esbuild + ESM) is structurally invisible to CI. Any regression in the externalize/relocate fix would slip through.

This sub-issue adds the missing fixture and wires it into PR-blocking CI as a matrix across OS × Node × package manager.

## What

1. Add `examples/next/tina-self-hosted-sqlite-demo/` mirroring the existing mongo demo but with:
   - `sqlite-level` dep
   - `databaseAdapter: new SqliteLevel({ filename: '...' })` in `tina/database.ts`
2. Add a `scripts/smoke-test-adapter.mjs` that dynamically imports the built bundle, instantiates the adapter, and does a put/get round-trip. (The bug fails before adapter use, but the round-trip catches resolution-from-output-location regressions.)
3. Add a new `self-hosted-sqlite` job to `.github/workflows/main.yml` (PR-blocking, not the nightly `build-starter-templates`):

```yaml
matrix:
  os: [ubuntu-latest, macOS-latest, windows-latest]
  node: ['20', '22', '24']
  pm: [npm, pnpm, yarn]
```

= 27 jobs per PR. Steps: install with `${{ matrix.pm }}`, build, run smoke test.

## Acceptance criteria

- [ ] `examples/next/tina-self-hosted-sqlite-demo/` committed with working `tina/database.ts` using `SqliteLevel`
- [ ] `scripts/smoke-test-adapter.mjs` performs put/get round-trip and exits non-zero on failure
- [ ] New `self-hosted-sqlite` job in `main.yml` runs on every PR
- [ ] Matrix covers ubuntu-latest, macOS-latest, windows-latest
- [ ] Matrix covers Node 20, 22, 24
- [ ] Matrix covers npm, pnpm, yarn
- [ ] Job is **blocking** (not `continue-on-error`)
- [ ] On a branch *without* the #6750 fix applied, the new job fails (proving it reproduces #6675)
- [ ] On a branch *with* the #6750 fix applied, the new job passes across all 27 combinations

## Catches

- The actual #6675 / #6750 regression on every PR
- Windows path-separator bugs in `pathToFileURL`
- pnpm symlink-farm resolution failures
[...]

PR #6790, opened 2026-05-08
Fix native SQLite crash in ESM database build

### LINKED ISSUE #6675: Build fails in v3.7.2 with ESM: ReferenceError: __filename is not defined (sqlite-level compatibility)

Ben0189:
Closes #6675
Part of #6750

## Problem

TinaCMS v3's CLI bundles `tina/database.ts` with esbuild and dynamically imports the result. Since the December 2025 ESM migration, this is wedged between two failure modes:

1. **Bundling** native modules (`better-sqlite3` via `sqlite-level`) crashes — `bindings` reads `__filename`/`__dirname`, which don't exist in ESM scope. See #6675.
2. **Externalizing** native modules from `os.tmpdir()` doesn't work either — Node's resolver walks up from the importing file's directory and never reaches the project's `node_modules`.

## Fix

Two architectural changes plus the API + ergonomics work that comes out of them:

### Build cache moved into the project tree

- `loadDatabaseFile` and `loadConfigFile` now write to `<project>/tina/__generated__/.cache/<timestamp>/` instead of `os.tmpdir()`, so Node can resolve `node_modules` from the bundle's location at runtime.
- Each load removes its own subdir after the dynamic-import resolves, and `ConfigManager.processConfig()` sweeps `.cache/` on startup to clean up residue from crashed prior runs (Ctrl+C mid-build, OOM kills).
- The empty `<timestamp>/` parent is also reaped after both loads complete (using `rmdirSync`, so concurrent in-flight loads don't conflict).
- Read-only project mounts (Docker `:ro`, AWS Lambda `/var/task`, sandboxed CI runners) now fail with an actionable error explaining the cause and how to resolve it, instead of a cryptic mid-build `EACCES` from esbuild.

### Externalize curated baseline + user-extensible

- `loadDatabaseFile` externalizes `better-sqlite3` so Node loads it as CJS where `__filename` exists. The `sqlite-level` wrapper stays bundled so esbuild handles its CJS named-export interop.
- New `defineConfig` field: `build.externalDependencies?: string[]`. Users with custom native adapters outside the baseline can extend the externalize list from their config — discoverable, type-safe, version-controlled. Supersedes the env-var approach proposed in #6683.
- The merge logic and baseline live in `external-resolver.ts` with 9 unit tests.

### `tina init` template

- New projects (and existing projects without it) get `tina/__generated__` added to `.gitignore`. Belt + suspenders for the cache cleanup.

## Verification

Manually verified end-to-end against `examples/next/tina-self-hosted-demo` temporarily reconfigured to use `SqliteLevel`:

- Bundle lands in `<project>/tina/__generated__/.cache/<timestamp>/database/database.build.mjs` (not `/tmp/`)
- Node loads `bindings.js`'s `__filename`-based path successfully
- `better-sqlite3` native binary loads
- SqliteLevel instantiates, opens the DB file, and gets through to content indexing
- Cache directory is empty after the build

The remaining indexing crash is a known SQL-injection bug in sqlite-level **v1**, fixed by [tinacms/sqlite-level#24](https://github.com/tinacms/sqlite-level/pull/24) — tracked as a follow-up in #6848. Not in scope for this PR.

The pre-existing mongo demo is unchanged so its CI coverage is preserved. A standalone SQLite example with full CI matrix is filed as #6786.

## Tests

- 9 new unit tests on the externalize merge logic (`external-resolver.test.ts`)
- All 180 CLI tests pass (was 171 pre-PR — 9 new)
- Build clean, type-check clean (via `pnpm types`), format clean

## Follow-ups (filed)

- #6786 — Add self-hosted SQLite example + PR-blocking CI matrix
- #6787 — Add unit tests for `config-manager.ts` (output location, cache cleanup, watch list, etc.)
- #6788 — Adapter compatibility test matrix
- #6789 — Promote starter-template smoke subset to PR-blocking
- #6847 — Migrate `mongodb-level` to ESM
- #6848 — Bump `sqlite-level` to v2.x in `@tinacms/search`
- #6849 — Expand the externalize baseline as adapters migrate


JackDevAU:
@copilot resolve the merge conflicts in this pull request

kulesy:
## `.cache/<timestamp>/` directories accumulate forever

Nice fix on the relocation + externalize. One thing missing from #6750's proposal that I think we should land in this PR rather than defer: **cache cleanup**.

The current code only removes the bundle file, not the timestamp directory:

https://github.com/tinacms/tinacms/blob/main/packages/%40tinacms/cli/src/next/config-manager.ts#L391-L392

```ts
const result = await import(pathToFileURL(outfile).href);
fs.removeSync(outfile);  // removes database.build.mjs but leaves the parent <timestamp>/ dir
return result.default;
```

So after this PR, every `tina dev` invocation, every rebuild, every `tina build` leaves an empty (or near-empty) `<timestamp>/` directory behind:

```
.tina/__generated__/.cache/
  1730000000001/
  1730000000002/
  1730000000003/
  ... grows indefinitely
```

`loadConfigFile` has the same shape — it cleans up `outfile` and `outfile2` but not the parent `buildDir`.

### Suggested fix

Two changes inside this PR:

1. **Per-load cleanup** — replace `fs.removeSync(outfile)` with `fs.removeSync(buildDir)` in both `loadDatabaseFile` and `loadConfigFile`, so the whole timestamped directory disappears once the dynamic-import resolves.

2. **Startup sweep** — in `ConfigManager`'s `init()` (or a new method called from it), `fs.removeSync(this.generatedFolderPath + '/.cache')` before creating the new timestamp dir. This handles residue from crashed prior runs (Ctrl+C during `tina dev`, OOM kills, etc.) which `try/finally` can't catch.

Both are a few lines. Without them, this PR introduces a new long-term housekeeping problem in every Tina project.

### Related — `.gitignore`

While we're here: `.tina/__generated__/.cache/` should be added to the `.gitignore` shipped with `tina init`, otherwise users will start `git add .`-ing these directories. Not strictly required for this PR but the deferral risk is high — once it ships without the gitignore entry, every existing project on upgrade will see the noise on first `tina dev`.

cc'd from the parent #6750 — these were called out in the proposal under "CLI sweeps `.cache/` on startup and after each successful load to prevent buildup" and the gitignore bullet.

kulesy:
## Replacing the Mongo demo loses coverage — should be additive, not destructive

The diff converts `examples/next/tina-self-hosted-demo/` from Mongo to SQLite:

```diff
-import { MongodbLevel } from 'mongodb-level';
-import { GitHubProvider } from 'tinacms-gitprovider-github';
+import { SqliteLevel } from 'sqlite-level';
```

Two concerns with this approach:

### 1. We lose the Mongo reference + its CI coverage

Anyone running self-hosted Tina with Mongo today (and that's most production self-hosted users — it's been the canonical adapter) just lost their reference example. More importantly, **the Mongo + esbuild interop path is no longer exercised by any example in the repo**. Mongo-level uses `scmp` which `require('crypto')` — that's exactly the kind of CJS pattern the existing `createRequire` banner exists to handle. If a future change breaks that path, no example will catch it.

### 2. The noop git provider drops the "self-hosted" demonstration

```ts
const noopGitProvider = {
  onPut: async () => {},
  onDelete: async () => {},
};
```

The whole point of `tina-self-hosted-demo` is to show how self-hosting works end-to-end, including the git push that persists content changes. Swapping `GitHubProvider` for a noop turns it into a "local-only demo with SQLite" — which is what `createLocalDatabase()` already provides. Users following this example to set up real self-hosting will hit a wall when their content edits don't persist anywhere.

### Suggested fix

Make this additive:

1. **Restore `examples/next/tina-self-hosted-demo/`** to its original Mongo + GitHubProvider state
2. **Add a new `examples/next/tina-self-hosted-sqlite-demo/`** that uses `SqliteLevel` + `GitHubProvider` (or, if a real git provider is too much setup for the demo, document that limitation explicitly in the demo's README rather than silently noop'ing it)

This was the intent of sub-issue #6786 — "mirroring the existing mongo demo." The word *mirroring* matters: side-by-side examples, not a replacement.

The added cost is small (one duplicated example directory), and it preserves both:
- Production-pattern documentation for both adapters
- CI regression coverage for both code paths

kulesy:
✅ Addressed.

- **Per-load subdir cleanup** — each `loadDatabaseFile` / `loadConfigFile` now removes its entire `<timestamp>/database/` or `<timestamp>/config/` subdir after the dynamic-import resolves (commit [4e153f842](https://github.com/tinacms/tinacms/pull/6790/commits/4e153f842)).
- **Empty timestamp parent reaped** — once both sibling loads finish, the empty `<timestamp>/` parent itself is removed via `rmdirSync` (commit [c735e6be0](https://github.com/tinacms/tinacms/pull/6790/commits/c735e6be0)). `ENOTEMPTY` is the expected concurrent-load case and is safely deferred to the next startup sweep; other errors re-throw (commit [c840cf9a0](https://github.com/tinacms/tinacms/pull/6790/commits/c840cf9a0)).
- **Startup sweep** — `ConfigManager.processConfig()` now nukes any `.cache/` residue from crashed prior runs before reserving its new timestamp dir (commit [4e153f842](https://github.com/tinacms/tinacms/pull/6790/commits/4e153f842)).
- **`.gitignore` template** — `tina init` now adds `tina/__generated__` for new projects (or projects missing it). Updates the `InitEnvironment` type, detection, apply logic, and the fresh-create path (commits [4e153f842](https://github.com/tinacms/tinacms/pull/6790/commits/4e153f842) + [c840cf9a0](https://github.com/tinacms/tinacms/pull/6790/commits/c840cf9a0)).
- **Tests** — extracted to `cache-manager.ts` with 13 unit tests covering startup sweep, crash residue, per-load reap, sibling-load coexistence, and the empty-parent reap invariant (commit [c8d8baf17](https://github.com/tinacms/tinacms/pull/6790/commits/c8d8baf17)).

Verified end-to-end against the demo: `.cache/` is empty after every successful build.

kulesy:
✅ Addressed in commit [4e153f842](https://github.com/tinacms/tinacms/pull/6790/commits/4e153f842).

- `examples/next/tina-self-hosted-demo/tina/database.ts` and `package.json` restored to their pre-PR state — `MongodbLevel` + `GitHubProvider`, no more `noopGitProvider` hack, `sqlite-level` dep removed.
- `pnpm-lock.yaml` synced to drop the unused `sqlite-level` entry from this example's importer.

A separate self-hosted SQLite example with PR-blocking CI matrix (the regression fixture) is filed as #6786 and will be additive — preserving both production-pattern documentation and CI coverage for both adapters.

End-to-end SQLite verification was done locally during PR development against a temporarily-modified copy of this demo; the build succeeded all the way through to content indexing, confirming the original `__filename` bug is gone.

JackDevAU (approved):
:shipit:

PR #6828, opened 2026-05-08
Strip cloud URL on save regardless of host

### LINKED ISSUE #6827: 🤖 🐛 Cloud→relative strip in `resolveMediaCloudToRelative` fails when content-api's `assetsHost` differs from dashboard-inserted URL
## Summary

`@tinacms/graphql`'s `resolveMediaCloudToRelative` only strips a value's leading cloud URL when it exactly matches `https://${config.assetsHost}/${config.clientId}` ([`media-utils.ts:32`](https://github.com/tinacms/tinacms/blob/main/packages/%40tinacms/graphql/src/resolver/media-utils.ts#L32)). On every TinaCloud setup, the dashboard's default `MediaStore` inserts upload URLs with a hardcoded host (`https://assets.tina.io`, [`media-store.default.ts:181`](https://github.com/tinacms/tinacms/blob/main/packages/tinacms/src/toolkit/core/media-store.default.ts#L181)). The two have historically agreed and the round-trip worked: insert absolute, strip on save, store relative.

The moment those two hosts disagree — typical of any non-prod multi-host setup — the strip fails silently and the absolute URL gets committed to the content repo. The next time anyone reads that field, the URL still resolves to the wrong host.

## Why it was dormant

For users running against prod TinaCloud (the vast majority), content-api returns `assetsHost: "assets.tina.io"`, which matches the dashboard's hardcoded URL. The strip succeeds and content saves as a relative path — exactly as designed.

For users running against a non-prod stage (PR / stage / personal-dev / `local-*`):

- Dashboard still inserts `https://assets.tina.io/<clientId>/<path>` (hardcoded).
- Content-api returns the stage-specific `assetsHost`, e.g. `"assets-local-<stage>.tinajs.dev"` (set by `TINA_CLOUD_MEDIA_STORE_ASSETS_HOST` env in TinaCloud's `ContentApiStack`).
- `resolveMediaCloudToRelative` looks for `https://assets-local-<stage>.tinajs.dev/<clientId>` in the value, doesn't find it, returns the value untouched.
- The committed MDX/JSON now contains `src=\"https://assets.tina.io/...\"`, which won't resolve to the actual file (which lives in the stage's bucket).

This was effectively invisible until someone exercised image upload + commit + read on a stage with a non-prod `assetsHost`.

## Repro

1. Configure a TinaCloud project against any non-prod stage:

   ```ts
   tinaioConfig: {
     contentApiUrlOverride: \"https://my-stage-content.tinajs.dev\",
     assetsApiUrlOverride: \"https://assets-api-my-stage.tinajs.dev\",
   }
   ```
2. Open the media manager on any document, upload and insert an image.
3. Inspect the saved MDX or the resulting PR diff.

**Expected:** image `src` is relative (`/uploads/…`), matching pre-existing content.
**Actual:** image `src` is absolute and points at `https://assets.tina.io/<clientId>/...`. Committed to repo as-is. Image renders broken because the file lives at the stage bucket, not on prod.

## Suggested fix — primary

**Make the strip permissive across cloud hosts.** Recognise any TinaCloud-style cloud URL pattern, not just the one matching the current `config.assetsHost`. e.g. match `^https://[^/]+/${config.clientId}/` — the host segment can vary, the client-id-prefixed path is the durable invariant. Strip when matched.

This:

- Fixes new uploads in multi-host setups (the dashboard's hardcoded URL still gets stripped because the client-id prefix matches).
- Also fixes content migrated between stages — old absolute URLs from a previous stage round-trip cleanly back to relative on the next save.
- Is one targeted change inside `media-utils.ts`, no callers affected, no API changes.
- Symmetric: `resolveMediaRelativeToCloud` is already permissive (it always emits the configured cloud URL); the strip should be too.

## Suggested fix — alternative (or in addition)

**Don't hardcode the upload host on the dashboard side.** [`media-store.default.ts:181`](https://github.com/tinacms/tinacms/blob/main/packages/tinacms/src/toolkit/core/media-store.default.ts#L181) hardcodes `https://assets.tina.io`. The cleanest version of this fix is to insert a relative path (`/${path}`) into the editor and let `resolveMediaRelativeToCloud` handle host resolution at read time — same shape as historical content. Less preferred on its own because it doesn't address content migrated from stages with mismatched hosts; but pairs well with the primary fix.

## Acceptance

- [ ] On a non-prod stage with `config.assetsHost !== "assets.tina.io"`, uploading and saving an image causes the committed value to be a relative path (`/<mediaRoot>/...`), matching pre-existing content's format.
- [ ] Pre-existing relative content keeps round-tripping unchanged.
- [ ] Loading content that contains an absolute cloud URL whose host does not match `config.assetsHost` (e.g. cross-stage content migration) still strips correctly on next save.
- [ ] New tests in `media-utils.test.ts` cover (a) a string value with a non-matching host, (b) an array value with mixed hosts.

## Context

- Symptom surfaced while validating multi-repo editorial-workflow on a personal-dev TinaCloud stage. Saved MDX showed `src=\"https://assets.tina.io/...\"` for new uploads while existing image embeds in the same doc had `src=\"/uploads/...\"` (relative).
- Historical commits: dashboard hardcoded URL added in [`4696e15ba`](https://github.com/tinacms/tinacms/commit/4696e15ba) (Jan 2023, \"rm previewSrc from images\"). `resolveMediaCloudToRelative` strip condition predates that — both have lived alongside each other harmlessly because the hosts agreed in practice.
- Independent of the TinaCloud-side wiring of `branch`/`mediaBranch` (#6504, #6740). This is purely a dashboard ↔ resolver-on-save mismatch.

joshbermanssw:
## TL;DR

`resolveMediaCloudToRelative` only stripped values whose host exactly matched `config.assetsHost`, so non-prod TinaCloud stages were silently committing absolute `https://assets.tina.io/...` URLs to content repos instead of relative paths. Match by `<clientId>/...` path prefix instead — host segment varies across stages, the client-id-prefixed path is the invariant.

## Before / After

| | Before | After |
|---|---|---|
| Upload on prod stage (`assetsHost = "assets.tina.io"`) | Strips → relative path saved | Strips → relative path saved (unchanged) |
| Upload on non-prod stage (`assetsHost = "assets-stage.tinajs.dev"`) | Doesn't strip → `https://assets.tina.io/...` saved as absolute | Strips → relative path saved |
| Cross-stage content migration | URLs from previous stage left as-is | URLs from any stage strip on next save |

## Why

`resolveMediaCloudToRelative` runs on every image-typed field on save and strips a leading cloud URL to leave a relative path in storage (`/uploads/...`). That round-trip works as long as the dashboard's default `MediaStore` ([`media-store.default.ts:181`](https://github.com/tinacms/tinacms/blob/main/packages/tinacms/src/toolkit/core/media-store.default.ts#L181) — hardcodes `https://assets.tina.io/<clientId>/<path>`) and content-api's configured `assetsHost` agree.

They agree on prod TinaCloud (both are `assets.tina.io`), so 99% of users have always seen relative paths in their content. They disagree on every non-prod stage — content-api's `TINA_CLOUD_MEDIA_STORE_ASSETS_HOST` env returns `"assets-{stage}.tinajs.dev"`, the dashboard still inserts `https://assets.tina.io/...`, and the strip's `value.includes(\`https://${config.assetsHost}/${config.clientId}\`)` check returns false → URL passes through to disk.

Surfaced while validating multi-repo editorial-workflow on a personal-dev TinaCloud stage. Saved MDX showed `src=\"https://assets.tina.io/...\"` for new uploads while existing image embeds in the same doc had `src=\"/uploads/...\"`.

## How

```ts
// Before — exact host match
const assetsURL = `https://${config.assetsHost}/${config.clientId}`;
if (typeof value === 'string' && value.includes(assetsURL)) { ... }

// After — match any host whose path begins with `<clientId>/`
const cloudUrl = cloudUrlPattern(config.clientId);
if (typeof value === 'string' && cloudUrl.test(value)) { ... }
```

`cloudUrlPattern` is a small helper at the bottom of `media-utils.ts` next to the existing `stagingPrefix` / `stripStagingPrefix` helpers. Uses an `escapeRegExp` for client-id safety. The array branch gets the same treatment, plus a guard so non-cloud entries pass through untouched (the old behavior would `.replace()` a non-matching URL and prepend `mediaRoot` to it — a latent bug surfaced by adding the regex test).

`resolveMediaRelativeToCloud` is unchanged — it was already host-agnostic on the way out (always emits with `config.assetsHost`).

## Reviewer notes

- **Symmetric with what's already there.** `resolveMediaRelativeToCloud` always emits a URL using the configured host. The strip on the way back was the asymmetric piece — it required an exact match. Now it accepts the same pattern shape regardless of which host was originally used.
- **Cross-stage migration falls out for free.** A user moving content between stages (or between prod and a fork) used to see absolute URLs persisting forever; they now strip on next save.
- **Non-TinaCloud URLs left alone.** A URL whose path doesn't start with `<clientId>/` (e.g. an external `` source pasted in by a user, or a non-TinaCloud asset URL) doesn't match the regex and passes through. Existing test for the `other-cloud-media-service.com` case still passes; new test covers the wrong-clientId case explicitly.
- **No API change.** `GraphQLConfig` shape unchanged. No callers of either resolver function need to change.

## Tests

`packages/@tinacms/graphql/src/resolver/media-utils.test.ts` — 4 new cases:

- string value with cloud host that doesn't match `config.assetsHost` (e.g. `assets.tina.io` while config says `assets.tinajs.dev`) → strips to relative.
- array of mixed-host cloud URLs → all strip.
- `__staging/{branch}` prefix on a cross-host URL → still strips correctly (preserves prior staging-prefix behavior).
- URL with a different clientId in the path → not stripped (negative case).

All 501 tests in `@tinacms/graphql` pass; `media-utils.test.ts` is now 15 cases (was 11).

## Out of scope

The dashboard's hardcoded `https://assets.tina.io` in [`media-store.default.ts:181`](https://github.com/tinacms/tinacms/blob/main/packages/tinacms/src/toolkit/core/media-store.default.ts#L181) is independently worth cleaning up (insert a relative `/${path}` and let `resolveMediaRelativeToCloud` handle it), but resolving it is no longer required to fix the bug. Tracking it under #6827's "alternative fix" section.

## Links

- Closes [#6827](https://github.com/tinacms/tinacms/issues/6827)


18-th (dismissed):
LGTM

18-th (approved):
LGTM

Issue #6840, opened 2026-05-11
Add PostHog telemetry to @tinacms/cli to track build command usage


joshbermanssw:
## Background

The CLI currently emits first-party telemetry via `@tinacms/metrics` (POSTs to `https://metrics.tina.io/record`). Today we send a single event per `tinacms build` / `tinacms dev` invocation with a small payload (e.g. `hasLocalContentPath`). See:

- `packages/@tinacms/cli/src/next/commands/build-command/index.ts:139`
- `packages/@tinacms/cli/src/next/commands/dev-command/index.ts:85`
- `packages/@tinacms/metrics/src/telemetry/telemetry.ts`

This is enough to count invocations but not enough to answer product questions like:
- Which CLI commands/flags are used in the wild?
- How does usage break down across `build`, `dev`, `audit`, `init`?
- What % of users opt into `localContentPath`, `--local`, `--noSDK`, `--skipCloudChecks`, etc.?
- Funnels (e.g. `init` → first `dev` → first `build`).

## Proposal

Add PostHog telemetry **directly in `@tinacms/cli`** (not in `@tinacms/metrics`) so we can use PostHog's dashboards/funnels/retention views for CLI usage. `@tinacms/metrics` stays as-is; PostHog is a CLI-only concern.

## Scope

- [ ] Add `posthog-node` as a dependency of `packages/@tinacms/cli` only.
- [ ] New CLI-local telemetry module (e.g. `packages/@tinacms/cli/src/telemetry/posthog.ts`) that wraps init / capture / shutdown.
- [ ] Capture an event for each CLI command entrypoint:
  - `cli.build.invoked`
  - `cli.dev.invoked`
  - `cli.audit.invoked`
  - `cli.init.invoked`
- [ ] Properties to include (non-PII):
  - CLI version, Node version, OS/arch
  - Command name + non-sensitive flags (`local`, `noSDK`, `skipCloudChecks`, `localContentPath` boolean, `port` boolean, `verbose`, etc.)
  - Anonymous install ID (reuse existing `getId()` from `@tinacms/metrics`)
  - Result (success/failure, duration) on command exit
- [ ] Respect existing opt-out:
  - `--noTelemetry` flag (`baseCommands.ts:51`)
  - `config.telemetry === 'anonymous'` / disabled config
  - Standard env vars (`DO_NOT_TRACK=1`, `CI=true` → consider sampling or skip)
[...]

PR #6841, opened 2026-05-11
📊 Add PostHog telemetry to tinacms build

joshbermanssw:
As per #6840 

## TL;DR

The CLI today fires one minimal event per `tinacms build` to `metrics.tina.io` carrying only `hasLocalContentPath` — useful for invocation counts, useless for flag-distribution or build-duration questions. This PR replaces the `tinacms build` telemetry with PostHog: an `invoke` event with a flag payload, a `finished` event with `success` + `durationMs`, and a per-run `distinctId` so the two events correlate into a funnel.

- [ ] Tested locally with SSW.Rules on tagged versions of CLI and TinaCMS - against prod PostHog (see sc below)

## Why

Today we can ask "how many builds ran?" but not "what % use `--content=local`?", "what's p50 build duration?", or "what % of `--skip-cloud-checks` builds also pass `--local`?". #6840 called for switching CLI telemetry to PostHog so dashboards / funnels / retention views become available.

```mermaid
graph LR
  A[tinacms build start] --> B[fetch posthog-token<br/>from identity-v2]
  B --> C[invoke event<br/>flag payload + buildRunId]
  C --> D[do the work]
  D --> E{success?}
  E -->|yes| F[finished event<br/>durationMs + buildRunId]
  E -->|throws| G[catch: finished event<br/>success:false + errorCode]
  F --> H[posthog.shutdown 2s flush]
  G --> H
  H --> I[process.exit]
```

Screenshot 2026-05-11 at 1 57 09 pm

**Figure: ✅ Start + End have the same ID**

Screenshot 2026-05-11 at 2 12 20 pm

**Figure: ✅ Start body has correct flag params for `"build": "tinacms build --content=local -c \"next build\"",`**

Screenshot 2026-05-11 at 2 12 36 pm

**Figure: ✅ Finish Body**


packages/@tinacms/cli/src/utils/posthog.ts
```diff
@@ -0,0 +1,279 @@
...
+  });
+}
+
+export function postHogCapture(
+  client: PostHog,
+  distinctId: string,
+  event: string,
+  properties: Record<string, any>
+): void {
+  if (process.env.TINA_DEV === 'true') return;
```
18-th: Does it make more sense to make that check in the init function? Otherwise, we do the init, make the network call and then don't use the client for anything

packages/@tinacms/cli/package.json
```diff
@@ -91,10 +92,12 @@
         "memory-level": "catalog:",
         "minimatch": "catalog:",
         "normalize-path": "catalog:",
+        "posthog-node": "^5.17.2",
         "prettier": "catalog:",
         "progress": "catalog:",
         "prompts": "catalog:",
         "readable-stream": "catalog:",
+        "systeminformation": "^5.27.13",
```
18-th: Why do we need this dep? It doesn't seem to be referenced anywhere

packages/@tinacms/cli/src/utils/fetchPostHogConfig.ts
```diff
@@ -0,0 +1,32 @@
+export interface PostHogConfig {
+  POSTHOG_API_KEY?: string;
+  POSTHOG_ENDPOINT?: string;
+}
+
+export default async function fetchPostHogConfig(
+  endpointUrl: string
+): Promise<PostHogConfig> {
+  try {
+    const response = await fetch(endpointUrl, {
```
18-th, line 10: Can we put a timeout on this? If someone is building locally with no internet connection or with a firewall up, every call to the CLI now has extra latency up to the maximum default timeout for no good reason

packages/@tinacms/cli/src/utils/posthog.ts
```diff
@@ -0,0 +1,279 @@
...
+  message: string;
+  sanitizedStack: string;
+  originalStackHash: string;
+}
+
+/**
+ * Sanitize a stack trace by removing sensitive local paths
+ * while preserving useful debugging information
+ */
+function sanitizeStackTrace(stack: string): string {
```
18-th: Can we get some tests around functions in this file?

packages/@tinacms/cli/src/utils/posthog.ts
```diff
@@ -0,0 +1,279 @@
...
+ *   - 'create-tina-app-user-cancelled' for user cancellations (Ctrl+C)
+ * - Includes error code, sanitized stack, step name, and telemetry data in properties
+ * - Non-fatal errors are tracked but allow the process to continue
+ *
+ * @example
+ * ```typescript
+ * try {
+ *   await downloadTemplate();
+ * } catch (err) {
+ *   postHogCaptureError(client, userId, sessionId, err as Error, {
```
18-th: The signature of the example doesn't seem to match the signature of the method it describes

packages/@tinacms/cli/src/next/commands/build-command/index.ts
```diff
@@ -82,7 +91,17 @@ export class BuildCommand extends BaseCommand {
     description: `Build the CMS and autogenerated modules for usage with TinaCloud`,
   });
 
+  private posthogClient: PostHog | null = null;
+  private buildStartedAt = 0;
+  private buildRunId = generateSessionId();
+
   async catch(error: any): Promise<void> {
+    postHogCapture(this.posthogClient, this.buildRunId, BuildFinishedEvent, {
```
18-th, line 100: Is it intentional that for all of the if blocks in the execute that exit with `process.exit(1)`, we don't log it as a failed build?

packages/@tinacms/cli/src/utils/posthog.ts
```diff
@@ -0,0 +1,279 @@
...
+    errorCategory,
+    step,
+    fatal = true,
+    additionalProperties = {},
+  } = context;
+
+  // Determine event name based on category type
+  let eventName: string;
+  if (errorCategory === 'user-cancellation') {
+    eventName = 'create-tina-app-user-cancelled';
```
18-th: This is a build command in the tinamcs CLI. Why are we using the create-tina-app prefix here?

packages/@tinacms/cli/src/utils/posthog.ts
```diff
@@ -0,0 +1,279 @@
...
+  const {
+    errorCode,
+    errorCategory,
+    step,
+    fatal = true,
+    additionalProperties = {},
+  } = context;
+
+  // Determine event name based on category type
+  let eventName: string;
```
18-th: We declare this and never use it again. Is it a leftover from something?

packages/@tinacms/cli/src/utils/posthog.ts
```diff
@@ -0,0 +1,279 @@
...
+ * Capture an error event in PostHog with categorized tracking and sanitized stack traces
+ *
+ * @param client - The PostHog client instance
+ * @param event - The name of the event to track
+ * @param error - The error object that was thrown
+ * @param context - Context about the error including code, category, step, and additional properties
+ *
+ * @remarks
+ * - Sanitizes stack traces to remove local file paths
+ * - Maps error categories to three event types:
```
18-th: Is this block true? Seems like a copy-paste issue

packages/@tinacms/cli/src/next/commands/build-command/index.ts
```diff
@@ -134,15 +153,41 @@ export class BuildCommand extends BaseCommand {
...
+      skipIndexing: Boolean(this.skipIndexing),
+      partialReindex: Boolean(this.partialReindex),
+      hasPreviewName: Boolean(this.previewName),
+      specifiesTinaGraphQLVersions: Boolean(this.tinaGraphQLVersion),
+      skipCloudChecks: Boolean(this.skipCloudChecks),
+      skipSearchIndex: Boolean(this.skipSearchIndex),
+    };
+    this.buildStartedAt = Date.now();
+    try {
+      postHogCapture(
```
18-th: postHogCapture already does a catch-all on lines 74-76 of `posthog.ts`, is there another way we can get to that catch block that I'm missing?

18-th (dismissed):
LGTM

18-th (approved):
LGTM

Issue #6829, opened 2026-05-08
Fix generated TinaCMS client `./types` import to satisfy both TypeScript strict mode and Node.js ESM


joshbermanssw:
## Background

The TinaCMS CLI auto-generates `tina/__generated__/client.ts` containing an import of the sibling generated types file:

```ts
import { queries } from "./types";
```

Choosing the right string for that import path has been a recurring source of bugs. Different consumers have incompatible rules:

| Consumer | Required form |
|---|---|
| TypeScript compiler (default) | `"./types"` (no extension) |
| TypeScript with `allowImportingTsExtensions` | `"./types.ts"` |
| Node.js native ESM | `"./types.js"` (extension required) |
| Bundlers (Next.js, Vite, esbuild) | All forms work |

There is no single string that satisfies all four. The CLI must pick one and arrange the on-disk filenames so every consumer can resolve it.

## History — what's been tried so far

1. **Original behaviour:** emit `"./types"` (extensionless).
   - Worked for TypeScript and bundlers.
   - Broke for users running the generated file under Node.js native ESM. Filed as #6062.

2. **PR #6739** (`fix(@tinacms/cli): add .js extension to ./types import in generated client (fix #6062)`, merged 2026-04-30, released in #6748 on 2026-05-08): switched the emission to be extension-aware:
   - TypeScript projects → `"./types.ts"`
   - Non-TypeScript projects → `"./types.js"`

   **Why it didn't work:** TypeScript's default rules forbid the explicit `.ts` extension in import paths unless `"allowImportingTsExtensions": true` is set in `tsconfig.json` — and that flag itself requires `"noEmit": true`. Next.js runs `tsc` for type-checking during `next build`, so every Next.js-based starter (`tina-nextjs-starter`, `tina-docs`) failed with:

   > Type error: An import path can only end with a '.ts' extension when 'allowImportingTsExtensions' is enabled.

   This took out the entire `Build TinaCMS Starter Templates` scheduled run on 2026-05-08 ([run 25548247838](https://github.com/tinacms/tinacms/actions/runs/25548247838)) and auto-filed issue #6806 plus duplicates across Node 22/24/25.

   The PR's tests passed because they only string-matched the generator output — they didn't actually run `tsc` over the generated file or `node` against it. So no real consumer was exercised in CI before merge.

[...]

Issue #6062, opened 2025-10-27
Generated client.ts import of ./types missing file extension


Daniel15:
### Describe the issue that you're seeing. Any videos (eg. YouTube or Loom) or screenshots help!

If I try to import the `tina/__generated__/client.ts` file in a Node.js script (e.g. to run an express server for the datalayer), I get an error like this:
```
Error [ERR_MODULE_NOT_FOUND]: Cannot find module '/home/daniel/src/tina-selfhost-test1/tina/__generated__/types' imported from /home/daniel/src/tina-selfhost-test1/tina/__generated__/client.ts
```

According to the Node.js documentation, [file extensions must be provided](https://nodejs.org/api/esm.html#mandatory-file-extensions) when using ES modules, so this import should actually have a `.ts` extension, like this:
```js
import { queries } from "./types.ts";
```

not:
```js
import { queries } from "./types";
```

### Reproduction

N/A

### Steps to reproduce

Create new TinaCMS site (I used the Astro template)
Create new `server.ts` file:
```js
import {client} from "../../tina/__generated__/client.ts";
```

Run it:
```
node server.ts
```

### System Info

```shell
System:
    OS: Linux 6.15 Fedora Linux 42 (KDE Plasma Desktop Edition)
    CPU: (24) x64 AMD Ryzen 9 5900X 12-Core Processor
    Memory: 16.04 GB / 31.24 GB
    Container: Yes
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 22.21.0 - /home/daniel/.volta/tools/image/node/22.21.0/bin/node
    Yarn: 4.10.3 - /home/daniel/.volta/tools/image/yarn/4.10.3/bin/yarn
    npm: 10.9.4 - /home/daniel/.volta/tools/image/node/22.21.0/bin/npm
  Browsers:
    Firefox: 141.0
    Firefox Developer Edition: 141.0
  npmPackages:
    @tinacms/cli: ^1.11.0 => 1.11.0 
    @tinacms/datalayer: ^1.4.1 => 1.4.1 
    tinacms: ^2.9.0 => 2.9.0
```

### Validations

- [x] Follow our [Code of Conduct](https://github.com/tinacms/tinacms/blob/main/CODE_OF_CONDUCT.md)
- [x] Read the [docs](https://tina.io/docs).
- [x] Check that there isn't [already an issue](https://github.com/tinacms/tinacms/issues) that reports the same bug.

PR #6850, opened 2026-05-13
Fixing up imports

18-th:
Addresses #6829 and #6062

The generated client.ts and databaseClient.ts now import "./types.js" unconditionally, and the CLI ships both types.ts and types.js side by side for TypeScript projects.

Different consumers disagree on the right shape for that import string: TypeScript wants ".ts" or extensionless, Node native ESM wants ".js", bundlers don't care. Prior attempts (#6739, #6738) flipped between the forms and each one broke a different consumer. This sidesteps the string-choice problem by shipping both files on disk:

* TypeScript (bundler / node16 / nodenext, and TS 4.7+ classic node) reads types.ts via the .js → .ts resolution rewrite.
* Node ESM reads types.js literally at runtime.
* Bundlers don't care which one they pick.

Added a fixture-driven regression test that spawns real `tsc --noEmit` and `node --input-type=module` against the generator output. This is the test that would have caught #6739 before merge. The prior unit tests only string-matched the generated source.

### DIFF
--- a/.changeset/fix-generated-client-types-import.md
+++ b/.changeset/fix-generated-client-types-import.md
@@ -1,0 +1,5 @@
+---
+"@tinacms/cli": patch
+---
+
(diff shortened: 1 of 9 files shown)


kulesy (approved):

18-th:
Tested by importing the tagged release into the Astro, Hugo, Next and Barebones starters. Build is passing. Merging the PR.

PR #6853, opened 2026-05-13
fix(graphql): preserve absolute external URLs in image-type resolvers

kulesy:
## Summary

- `resolveMediaRelativeToCloud` previously treated every `image`-typed value as a media-library path — wrapping it with the cloud assets prefix on read. That corrupts any value that's already an absolute URL (external link, `data:` URL, etc.), because the prefix gets concatenated directly in front of the original URL with no separator.
- Adds a guard for absolute URLs (any RFC 3986 scheme, plus protocol-relative `//`) so they round-trip identically in both directions.
- Adds a self-heal: if stripping the configured media root from a stored value reveals an absolute URL underneath (the shape of a previously-corrupted round-trip), return the URL directly instead of re-wrapping. Documents already on disk in the broken shape transparently recover on next save.
- The save-side resolver already gated rewriting on a cloud-URL pattern match, so no change there beyond inheriting cleaner inputs.

## Verification

Drove the full `parseMDX` → resolver → `serializeMDX` pipeline against a rich-text body containing both a media-library src and an external URL, on a feature branch with a staging prefix active:

| Phase | Local `src` | External `src` |
|---|---|---|
| Stored MDX | `/uploads/library/local.png` | `https://cdn.example.com/external/image.png` |
| After read (editor sees) | `https://<assetsHost>/<clientId>/__staging/<branch>/__file/library/local.png` | `https://cdn.example.com/external/image.png` |
| After write (re-serialized MDX) | `/uploads/library/local.png` | `https://cdn.example.com/external/image.png` |

Self-heal verified end-to-end: a corrupted stored src of the form `/uploads<absolute-url>` is read as the clean absolute URL and serialized back to disk in that form.

Confirmed the same demo fails against the unpatched resolver — the external URL gets clobbered with a leading `https://<assetsHost>/<clientId>/__staging/<branch>/__file` and then mangled further on save.

## Test plan

- [x] `pnpm --filter @tinacms/graphql test` — 516/516 passing across 50 files
- [x] New `media-utils.test.ts` cases:
  - [x] `it.each` over `https://`, `http://`, `//`, `data:`, `blob:`, `file:` — identity in both directions on a feature branch with active staging prefix
  - [x] Mixed array — relative entries get the staging prefix, absolute entries pass through untouched
  - [x] `it.each` self-heal cases — `/uploads<url>` shapes for `https://`, `http://`, `//`, `data:` recover to the underlying URL on read
  - [x] Array self-heal — mixed clean + corrupted entries are all normalized
- [x] New `media-rich-text.test.ts` end-to-end cases:
  - [x] Rich-text body with a mix of local + external `imageEmbed` srcs — full parse/serialize round-trip identity
  - [x] Protocol-relative and `http://` srcs — same identity guarantee
  - [x] Corrupted src — read self-heals, save writes clean form back to disk
- [ ] Smoke test in a branch-enabled site: open a document containing an `image`-typed field whose stored value is an external URL, save without touching it, confirm the stored value is unchanged

### DIFF
--- a/.changeset/fresh-tomatoes-look.md
+++ b/.changeset/fresh-tomatoes-look.md
@@ -1,0 +1,5 @@
+---
+"@tinacms/graphql": patch
+---
+
+fix(graphql): preserve absolute external URLs in image-type resolvers
--- a/packages/@tinacms/graphql/src/resolver/media-rich-text.test.ts
+++ b/packages/@tinacms/graphql/src/resolver/media-rich-text.test.ts
@@ -1,0 +1,150 @@
+/**
+ * End-to-end round-trip through `parseMDX` + `serializeMDX` for a rich-text
+ * field that contains a JSX template with an `image`-typed src prop. Exercises
+ * the same pipeline the GraphQL `rich-text` resolver uses, with the actual
+ * resolver callbacks plugged in.
+ *
+ * Guards against regressing the absolute-URL guard in `media-utils.ts`: an
+ * `image`-typed src that is already an external URL must survive both the
+ * relative→cloud (read) and cloud→relative (save) transforms unchanged, while
+ * a media-library path must still pick up the staging branch prefix on read
+ * and round-trip back to its original form on save.
+ */
+
+import { parseMDX, serializeMDX } from '@tinacms/mdx';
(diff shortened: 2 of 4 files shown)


18-th (approved):
LGTM

PR #6855, opened 2026-05-14
Improve build starter templates action

18-th:
Previously ran on a weekday cron with no version control over tina packages. Now triggers automatically after each publish, pins each tina package to its exact published version (packages carry different version numbers), and falls back to latest for packages not in the release. Lock files for non-tina dependencies are preserved. Manual dispatch supports both a flat version and a JSON package map for testing tagged/candidate builds.

Note: pnpm is pinned to version 10 for now. Version 11 required explicitly specifying which packages are allowed to run scripts, which requires an update to all starter templates. After that's done - we can bump it.

### DIFF
--- a/.github/workflows/build-starter-templates.yml
+++ b/.github/workflows/build-starter-templates.yml
@@ -2,8 +2,19 @@ name: Build TinaCMS Starter Templates
 
 on:
   workflow_dispatch:
+    inputs:
+      tina_version:
+        description: 'Version applied to any Tina package not listed in published_packages (semver, dist-tag, or snapshot tag, e.g. 2.5.0, latest, or my-branch). Defaults to latest.'
+        required: false
+        default: 'latest'
+        type: string
+      published_packages:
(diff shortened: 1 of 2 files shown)


18-th:
New options:

image

18-th:
All 63 jobs passing.

Run: https://github.com/tinacms/tinacms/actions/runs/25840183387

image

kulesy:
The jq filter only matches `tinacms` and `@tinacms/*`, but the publish workflow also releases `next-tinacms-s3`, `tinacms-authjs`, `tinacms-clerk`, and `tinacms-gitprovider-github`. A starter that depends on one of those won't get the just-published version pinned — it'll silently resolve to whatever the lockfile or `latest` had at install time, which weakens the release-validation purpose for those packages.

Worth either widening the regex (e.g. also match `^tinacms-`) or making the scope explicit in the PR description.

kulesy (dismissed):

kulesy (dismissed):

kulesy (approved):

PR #6864, opened 2026-05-14
test(@tinacms/cli): lock down loadDatabaseFile esbuild contract

### LINKED ISSUE #6785: 🧪 Tests for self-hosted - esbuild externalization in `loadDatabaseFile`
⚠️ Not Ready until #6750 is done

## Background

Issue #6750 — `better-sqlite3` (used transitively by `sqlite-level`) uses `require('bindings')` internally to locate its native `.node` binary. `bindings` relies on CJS globals `__filename`/`__dirname` to resolve the binary path. When `better-sqlite3` is bundled inline into the generated ESM output by esbuild, those globals are unavailable at runtime and the process crashes.

The fix (in `config-manager.ts` → `loadDatabaseFile()`) is:
- `external: ['better-sqlite3']` — esbuild emits an ESM import for it instead of inlining it; Node then loads it as CJS from the project's `node_modules` where `__filename` is available
- `sqlite-level` stays **bundled** — esbuild handles `{ SqliteLevel }` named-export interop from the CJS module

Config builds (`loadConfigFile`) are intentionally **not** changed — they import React/Next.js UI components which also break under Node ESM when externalized.

---

## What to test

That the `esbuild.build()` call in `loadDatabaseFile()`:
1. Includes `external: ['better-sqlite3']` so the native-binding chain is never bundled
2. Does **not** include `packages: 'external'` (which would also externalize `sqlite-level` and break named imports)

---

## Why we want this test

Without it, a future contributor could remove `external: ['better-sqlite3']` or replace it with the simpler `packages: 'external'`, silently re-introducing either:
- `ReferenceError: __filename is not defined` (better-sqlite3 bundled into ESM), or
- `Named export 'SqliteLevel' not found` (sqlite-level externalized, CJS named import fails)

Both failures are runtime-only and hard to catch without a real SQLite environment.

---

### Approach
`config-manager.ts` uses `import.meta.url` at module level, which breaks Jest's CJS runtime. Mock all heavy dependencies and capture the `esbuild.build()` options passed at call time.

### Test file location
`packages/@tinacms/cli/src/next/config-manager.test.ts`

### Required mocks
```ts
const mockBuild = jest.fn().mockResolvedValue({ metafile: { inputs: {} } });

jest.mock('esbuild', () => ({ build: mockBuild }));
jest.mock('fs-extra', () => ({
  outputFileSync: jest.fn(),
  removeSync: jest.fn(),
  existsSync: jest.fn(() => true),
}));
jest.mock('normalize-path', () => (p: string) => p);
jest.mock('../logger', () => ({ logger: { info: jest.fn(), warn: jest.fn(), error: jest.fn() } }));
jest.mock('./resolve-content-root', () => ({ resolveContentRootPath: jest.fn() }));
jest.mock('url', () => ({ pathToFileURL: (p: string) => ({ href: p }) }));
```

#### 1. `loadDatabaseFile` externalizes `better-sqlite3` but not `sqlite-level`
```ts
it('externalizes better-sqlite3 but keeps sqlite-level bundled', async () => {
  const manager = Object.create(ConfigManager.prototype);
  manager.selfHostedDatabaseFilePath = '/project/tina/database.ts';
  manager.generatedCachePath = '/project/tina/__generated__/.cache/12345';

  let captured: any;
  mockBuild.mockImplementationOnce((opts) => {
    captured = opts;
    return Promise.resolve({ metafile: { inputs: {} } });
  });

  await manager.loadDatabaseFile().catch(() => {}); // import of tmp file will fail — expected

  expect(captured.external).toContain('better-sqlite3');
  expect(captured.packages).toBeUndefined(); // must NOT be 'external'
});
```

#### 2. `loadDatabaseFile` output path is inside the project tree (not `/tmp`)
```ts
it('writes output inside generatedCachePath, not os.tmpdir()', async () => {
  const manager = Object.create(ConfigManager.prototype);
  manager.selfHostedDatabaseFilePath = '/project/tina/database.ts';
  manager.generatedCachePath = '/project/tina/__generated__/.cache/12345';

  let captured: any;
  mockBuild.mockImplementationOnce((opts) => {
    captured = opts;
    return Promise.resolve({ metafile: { inputs: {} } });
  });

  await manager.loadDatabaseFile().catch(() => {});

  expect(captured.outfile).toContain('/project/tina/__generated__/.cache/12345');
  expect(captured.outfile).not.toContain(require('os').tmpdir());
});
```

---

## Acceptance criteria
- [ ] Test 1 passes and fails when `external: ['better-sqlite3']` is removed or replaced with `packages: 'external'`
- [ ] Test 2 passes and fails when the output path is changed back to `os.tmpdir()`
- [ ] All existing CLI tests still pass (`pnpm test` in `packages/@tinacms/cli`)

kulesy:
Closes #6785

## Why

#6785 asks for regression tests that catch the two ways the #6790 fix could be silently undone:

1. Removing `external: ['better-sqlite3']` (re-introduces the `__filename is not defined` crash from #6675)
2. Replacing it with `packages: 'external'` (silently breaks user-side named imports of CJS UMD packages like `sqlite-level` v1 and `mongodb-level`)

The original issue proposed mocking esbuild and inspecting captured args via `Object.create(ConfigManager.prototype)`. That approach is blocked by `import.meta.url` in `config-manager.ts` — Jest's CJS runtime can't parse it. #6790 worked around this by extracting helpers (`external-resolver.ts`, `cache-manager.ts`); this PR applies the same pattern one more time.

## What

Extract the esbuild options out of `loadDatabaseFile()` into a pure `buildDatabaseEsbuildConfig()` helper. The helper is trivially testable — no module-load side effects, no `import.meta`, no fs access.

## Tests added (9)

Three describe blocks, all directly asserting the regressions #6785 calls out:

**`buildDatabaseEsbuildConfig — externalize contract`**
- ✅ external includes `'better-sqlite3'`
- ✅ caller-provided external list is forwarded unchanged
- ✅ **`packages: 'external'` is NEVER set** — the critical broad-externalize regression guard

**`buildDatabaseEsbuildConfig — output path contract`**
- ✅ `outfile` is forwarded unchanged from the caller (caller-side guarantee that the path is in the project tree is enforced by `prepareCacheLocation()` in `cache-manager.ts`, which has its own tests)
- ✅ no remapping happens to the path

**`buildDatabaseEsbuildConfig — fixed esbuild options`**
- ✅ `bundle: true` + `entryPoints` set
- ✅ `platform: 'node'` + `format: 'esm'`
- ✅ caller's loader map forwarded
- ✅ `createRequire` banner injected (so bundled CJS deps using `require()` for Node built-ins still work in ESM context)

## Verification

- `pnpm types` (turbo): 42/42 tasks pass
- `pnpm --filter @tinacms/cli test`: 206/206 tests pass (was 197 → +9)
- `pnpm format`: clean
- The single behavior change in `config-manager.ts` is a refactor: `loadDatabaseFile()` now constructs the esbuild config via the helper. The runtime behavior is identical.


18-th (approved):
LGTM

JackDevAU (approved):

PR #6878, opened 2026-05-14
Allow pnpm template builds to run dependency scripts

### LINKED ISSUE #6873: Starter template 'tinasaurus' failed during update-tina with pnpm using Node 24
The **tinasaurus** starter template failed during update-tina with **pnpm** using Node **24**.

Check the full workflow logs for more details: [View Workflow Run](https://github.com/tinacms/tinacms/actions/runs/25845159103)

mzl2233:
The pnpm template jobs are failing because pnpm ignores dependency build scripts by default, which leaves native packages unavailable for the subsequent build. This sets pnpm's build-script approval config before the install step in the starter-template workflow while leaving the yarn and npm paths unchanged. Closes #6873.

### DIFF
--- a/.github/workflows/build-starter-templates.yml
+++ b/.github/workflows/build-starter-templates.yml
@@ -78,6 +78,9 @@ jobs:
         id: build
         run: |
           cd template-repo
+          if [[ "${{ matrix.package-manager }}" == "pnpm" ]]; then
+            pnpm config set dangerouslyAllowAllBuilds true
+          fi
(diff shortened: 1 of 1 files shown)


18-th:
Hey @mzl2233, thanks for looking into this. I actually just messed up my regex while updating the action to run starters, which broke up a few runs. Now the pipeline is green again. I explicitly avoided approving all builds, as that exposes us to potential attacks if those packages are compromised and the pipeline runs automatically. Closing the PR.

PR #6902, opened 2026-05-18
Handle media workflow on protected branches

18-th:
This addresses https://github.com/tinacms/tinacloud/issues/3474

* Intercept Tina media uploads and deletes on protected branches when the Editorial Workflow is enabled. Create the workflow branch/PR first, switch the editor to the new branch, then run the original media operation against that branch.
* Share editorial workflow status handling with content saves, surface the existing workflow progress UI for media operations, and generate media-specific branch names from the target asset path.
* Add media-store coverage for protected-branch upload/delete interception, direct writes on unprotected branches, empty upload batches, branch switch acknowledgement, and branch-aware media requests.


packages/tinacms/src/toolkit/form-builder/editorial-workflow-utils.ts
```diff
@@ -0,0 +1,27 @@
+export const getEditorialWorkflowPrTitle = (branchName: string): string =>
+  `${branchName.replace('tina/', '').replaceAll('-', ' ')} (PR from TinaCMS)`;
```
kulesy, line 11: The previous inline version used `.replace('-', ' ')` (first dash only); this helper uses `replaceAll`. So `tina/page-updates-2026` now becomes `"page updates 2026"` instead of `"page updates-2026"`. Almost certainly the desired behaviour, but flagging because this changes content-save PR titles, not just media ones — worth a line in the changeset.

packages/tinacms/src/auth/TinaCloudProvider.tsx
```diff
@@ -509,7 +510,18 @@ export const TinaCloudProvider = (
...
+            try {
+              currentBranchExists = await client.branchExists(currentBranch);
+            } catch (err) {
+              console.error(
+                '[tina:branch-guard] setupEditorialWorkflow: branchExists threw, falling back to default branch:',
+                err
+              );
+            }
+            if (!currentBranchExists) {
+              setCurrentBranch(project.defaultBranch || 'main');
```
kulesy: Could we add (a) a one-line comment explaining the `branchExists` check — something like *"metadata can lag git after a fresh `createBranch`; trust git as the source of truth here"* — and (b) a small test for "branch missing from `project.metadata` but exists in git → user stays on it"? The content-save flow doesn't exercise this path in practice because its server-side waits hide the metadata lag, so without a test this guard is invisible against future refactors of either side.

packages/tinacms/src/toolkit/core/media-store.default.ts
```diff
@@ -11,6 +13,27 @@ import {
...
+const INDEX_MAX_UNKNOWN = 24;
+
+interface MediaWorkflowConfirmBranchEvent {
+  type: 'media:workflow:confirm-branch';
+  branchName: string;
+  baseBranch: string;
+  onConfirm: (branchName: string) => Promise<void>;
+  onCancel: () => void;
+  onSaveToProtectedBranch: () => void;
+}
```
kulesy: `MediaWorkflowConfirmBranchEvent` is declared in both places — here and in [`media-workflow-overlay.tsx`](packages/tinacms/src/toolkit/components/media/media-workflow-overlay.tsx#L31-L38) — and has to stay in sync by hand. Worth hoisting into a shared types file so the contract is single-sourced.

packages/tinacms/src/toolkit/core/media-store.default.ts
```diff
@@ -123,88 +150,335 @@ export class TinaMediaStore implements MediaStore {
...
+          } catch (error) {
+            reject(error);
+            throw error;
+          }
+        },
+        onCancel: () => resolve('cancelled'),
+        onSaveToProtectedBranch: () => resolve(undefined),
+      });
+    });
+  }
```
kulesy, line 277: If nothing is subscribed to `media:workflow:confirm-branch` when this dispatches, none of the resolve/reject callbacks ever fire and the Promise hangs forever — the upload appears to silently stall. The overlay is mounted in `TinaCloudProvider` so this only bites custom-provider setups, but it's a sharp edge. Defensively, we could either check that a listener exists after dispatch (and resolve to `undefined` to fall through to direct write, or reject with a clear "overlay not mounted" error), or document the requirement explicitly.

packages/tinacms/src/auth/TinaCloudProvider.tsx
kulesy: This is now irrelevant with the PR https://github.com/tinacms/tinacloud/pull/3563 which ensures the branch metadata is populated with the canonical branch on creation.

packages/tinacms/src/toolkit/plugin-branch-switcher/format-branch-name.ts
```diff
@@ -0,0 +1,10 @@
+/**
+ * Normalizes user-supplied strings into git-ref-safe branch segments.
+ * Kept in its own file (no React or workspace-internal imports) so
+ * non-React callers like `core/media-store.default.ts` can import it
+ * without dragging in the rest of `plugin-branch-switcher` and creating
+ * a cycle through `@toolkit/core`.
+ */
+export function formatBranchName(str: string): string {
+  return str.replace(/[^/\w-]+/g, '-').toLowerCase();
```
kulesy: CodeQL alerts [#127](https://github.com/tinacms/tinacms/security/code-scanning/127) and [#128](https://github.com/tinacms/tinacms/security/code-scanning/128) are still open against this regex. They look like false positives in practice — the input is realistically length-bounded — but either way it's worth resolving before merge. Two options:

1. **Dismiss with rationale** in the CodeQL UI (input length-bounded by browser file API; pattern is effectively linear).
2. **Drop the regex** in favor of an explicit character-by-character normalizer (5-line loop, no regex, alert goes away).

I'd lean toward option 2 — small function, easy swap, removes the alert without needing to justify it. Either way is fine, but leaving the alert open complicates the security tab going forward.

packages/tinacms/src/toolkit/form-builder/editorial-workflow-utils.ts
18-th: Changeset updated

packages/tinacms/src/toolkit/core/media-store.default.ts
18-th: Addressed

packages/tinacms/src/toolkit/core/media-store.default.ts
18-th: Addressed

packages/tinacms/src/toolkit/core/media-store.default.ts
```diff
@@ -123,88 +150,335 @@ export class TinaMediaStore implements MediaStore {
...
+        0
+      );
+      this.cms.events.dispatch({ type: 'media:workflow:finish' });
+      this.mediaWorkflowInProgress = false;
+    } catch (err) {
+      this.workflowBranchOverride = undefined;
+      this.mediaWorkflowInProgress = false;
+      this.dispatchMediaWorkflowError(err);
+    }
+  }
```
kulesy, line 371: **Major — orchestration shape diverges from the linked issue and from the content-save flow.**

[tinacloud#3474](https://github.com/tinacms/tinacloud/issues/3474) explicitly asks for this to go through `executeEditorialWorkflow({ branchName, baseBranch, prTitle, /* no graphQLContentOp */, onStatusUpdate })` — the same single server-orchestrated endpoint content saves use. What this PR ships is a client-driven chain: `createBranch` → client polls index → media op → client polls index → `createPullRequest`. Five roundtrips and two polling loops, where content saves do all of this in one awaitable promise with status streamed back from the server.

This matters beyond style:

- **Ordering & retries live on the client.** If the user closes the tab between branch creation and PR opening, the branch is orphaned. Content saves don't have this problem because the server owns the whole transaction.
- **Two indexing-wait loops are needed only because the client is orchestrating.** Server-side orchestration collapses both into one server-internal wait, no client polling required.
- **`finalizeMediaWorkflow` (this function) is essentially a hand-rolled state machine that duplicates the server-side workflow engine** — step 2/3/4 dispatches, alerts, branch override cleanup. `executeEditorialWorkflow` already streams these states via `onStatusUpdate`.

The wrinkle is that the file bytes can't live in the workflow payload — the browser has to PUT to S3 directly. But that's the only client-side hop that's actually inherent; everything else (branch creation, indexing wait, PR opening, status updates) can move into the workflow. The natural shape:

```ts
await tinaApi.executeEditorialWorkflow({
  branchName, baseBranch, prTitle,
  mediaOp: { type: 'upload' | 'delete', path, ... },
  onStatusUpdate, // surfaces signed-URL handle for upload, then 'media:ack' callback
})
```

Client awaits one promise, server orchestrates, browser only handles the byte transfer mid-flow when it gets the signed-URL status event.

I'd recommend reworking before merge. The current shape works and is tested, but it's architectural debt that bakes a parallel workflow engine into the client, and it ages badly the moment we want to add a third trigger (e.g. rename, batch-upload, anything else) — each one would re-implement this loop.

packages/tinacms/src/internalClient/index.ts
```diff
@@ -622,6 +637,114 @@ mutation addPendingDocumentMutation(
...
+          throw error;
+        }
+        console.warn(
+          `Editorial workflow status poll failed (attempt ${attempts}/${maxAttempts}), retrying...`,
+          error
+        );
+      }
+    }
+
+    throw new Error('Editorial workflow timed out after 5 minutes');
```
kulesy: **Minor — 5-minute polling cap can silently abandon a still-running workflow.** `maxAttempts × pollInterval = 60 × 5000ms = 5 minutes` hard cap. If the server-side workflow takes longer (slow indexer, large schema, retried operations, transient GitHub delays), this throws `"Editorial workflow timed out after 5 minutes"` while the server continues running. The user has no recovery path — re-triggering creates a *second* workflow branch (collision-suffixed), not a resume of the first, so they end up with two orphaned branches: one mid-flight on the server, one fresh from the retry.

Three options, any one would close the gap:

1. **Bump the cap** to something like 15 minutes that matches the realistic server-side worst case.
2. **Surface the `requestId` in the timeout error** so an operator can manually check workflow status (`GET /editorial-workflow/.../status/{requestId}`) and avoid the duplicate-branch problem.
3. **Prompt the user before giving up** — "this is taking longer than expected, continue waiting?" — and continue polling on yes.

(1) is the smallest fix. (2) is the most operationally useful. (3) is the cleanest UX. Up to you which.

Related: tinacloud#3594's server-side `waitForWorkflowIndexing` is already up to 120s × 2 + branch creation + commit + PR creation, so 5 minutes is plausibly tight even on a happy path. Worth confirming the typical p99 wall-clock before picking a number.

18-th: Updated to 15 minutes. We'll need to do some UI/UX work over that modal but that's a separate problem from what we are trying to do in this PR.

kulesy: Verified at `internalClient/index.ts:644` — `maxAttempts: 180` (15 minutes) with the new error message explicitly advising users not to retry while it may still be completing. That closes both the silent-timeout-on-slow-workflow concern and the duplicate-branch-on-retry problem (since the message gives a clear "wait, don't re-trigger" nudge). Generous headroom above the server-side worst case. 👍

kulesy (approved):
Reviewed end-to-end, all prior concerns addressed:

- Server-orchestrated rework adopted (Option 5 — client registers intent, polls status)
- All 5 of my prior line comments resolved (dedup type, Promise hang via `hasExplicitListenerFor`, CodeQL regex via char-by-char rewrite, `replaceAll` PR-title in changeset, `branchExists` guard no longer needed under server-side flow)
- New polling cap raised to 15 minutes with user-friendly error advising against premature retry
- CI green, 1009-line test coverage on `media-store.default.test.ts`

Companion PR: tinacms/tinacloud#3594. LGTM 🎉

PR #6904, opened 2026-05-18
🐛 Fix <mark> highlight serialization for MDX v2 / v3 compatibility

### LINKED ISSUE #6899: Hilight feature breaks Docusaurus 3.10 builds

The highlight feature introduced in v3.7 uses raw HTML in the form of a `<mark>` tag with parameters. This causes a problem under Docusaurus 3.10 because it forced MDXv2+ which does not permit HTML parameters in markdown. It expects parameters to be included in JSX format. One option is to just not use the feature, but if it's present, a user will probably use it. The workaround I'm using is to configure all rich text fields in the templates to specifically exclude hilights. I'm also using my own hilight solution (which predates 3.7) of repurposing the strikethrough style in markdown for hilights in the editor only (using CSS). I haven't implemented the fix in main yet, so you can see the problme.

### Reproduction

https://github.com/aowendev/docstatic

### Steps to reproduce

Add a hilight to the doc. Build the site.

### System Info

```shell
System:
    OS: macOS 26.4.1
    CPU: (8) arm64 Apple M1
    Memory: 85.97 MB / 8.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 24.15.0 - /opt/homebrew/opt/node@24/bin/node
    Yarn: 1.22.22 - /opt/homebrew/bin/yarn
    npm: 11.12.1 - /opt/homebrew/opt/node@24/bin/npm
    pnpm: 11.0.9 - /opt/homebrew/bin/pnpm
  Browsers:
    Firefox: 150.0.3
    Safari: 26.4
```

### Validations

joshbermanssw:
## TL;DR

`<mark style="background-color: #BFDBFE">` aborts Docusaurus 3.10+ builds because MDX v2/v3 parses the tag as JSX and React rejects string-valued `style` props with "Style prop value must be an object." This PR rewrites highlight serialization to emit `<mark style={{ backgroundColor: "#BFDBFE" }}>` instead, which is valid JSX and a valid React `style` prop.

## Why

The highlight feature serializes `<mark>` via `mdast-util-mdx-jsx`. Before this PR the `value` was set to a CSS string, producing `style="background-color: ..."`. In MDX v2+, every angle-bracketed tag is JSX, and React enforces an object-shaped `style` prop, so SSR builds abort.

## How

Both the legacy and `/next/` stringify paths now produce an `mdxJsxAttributeValueExpression` containing a camelCased object literal:

```ts
{
  type: 'mdxJsxAttribute',
  name: 'style',
  value: {
    type: 'mdxJsxAttributeValueExpression',
    value: `{ backgroundColor: "${color}" }`,
  },
}
```

The parser was widened to accept both the legacy HTML-string form and the new JSX-expression form, so existing files load without data loss. The file is rewritten to the new form on the next save through Tina.

### Tested for regressions locally 

Screenshot 2026-05-18 at 12 25 07 pm

**Figure: Confirm no regression on rich-text fields for nextJS project as example**

- Fixes #6899


packages/@tinacms/mdx/src/parse/mark.ts
```diff
@@ -14,15 +14,36 @@ export const getHighlightColorFromAttributes = (
...
+    return backgroundColorMatch?.[1]?.trim();
+  }
+
+  if (
+    styleAttribute.value &&
+    typeof styleAttribute.value === 'object' &&
+    styleAttribute.value.type === 'mdxJsxAttributeValueExpression'
+  ) {
+    const expression = styleAttribute.value.value;
+    const camelMatch = /backgroundColor\s*:\s*['"]([^'"]+)['"]/.exec(
```
18-th: Do you reckon we could support an optional quotation marks match (`['"]?`) on either side of the background color to align with the kebab matcher below, and in case someone hand-rolls their markdown?

18-th (approved):
LGTM

Issue #6848, opened 2026-05-13
Bump sqlite-level to v2.x and remove namespace-import workaround (tinacms + tinacloud)


kulesy:
Part of #6750.

## Context

[tinacms/sqlite-level#24](https://github.com/tinacms/sqlite-level/pull/24) migrated `sqlite-level` to pure ESM (major version bump to v2.x). TinaCMS still depends on v1.2.1 in:

- `pnpm-lock.yaml` (root): `sqlite-level: 1.2.1(sucrase@3.35.1)`
- `@tinacms/search/package.json`: catalog reference

Because v1 is CJS, [`@tinacms/search/src/client/index.ts`](https://github.com/tinacms/tinacms/blob/main/packages/%40tinacms/search/src/client/index.ts#L8-L14) currently uses a namespace-import workaround:

```ts
// TODO: Update to use named import once https://github.com/tinacms/sqlite-level/pull/24 is merged and released
// Use namespace import because `sqlite-level` is a CJS module and esbuild
// rewrites default imports in ways that break ESM named-export resolution.
import * as sqliteLevelModule from 'sqlite-level';
const SqliteLevel =
  (sqliteLevelModule as any).default?.SqliteLevel ??
  (sqliteLevelModule as any).SqliteLevel;
```

The TODO comment is now actionable. Bumping to v2 also unblocks #6750's externalize strategy from including `sqlite-level` in the curated baseline.

## What

1. Confirm `sqlite-level` v2.x is published to npm (PR #24 is merged but check the publish step ran)
2. Bump the dep in TinaCMS to `^2.0.0` (pnpm-workspace.yaml catalog + any direct references)
3. Remove the namespace-import workaround in `@tinacms/search/src/client/index.ts` — replace with clean named import:
   ```ts
   import { SqliteLevel } from 'sqlite-level';
   ```
4. Delete the TODO comment and the explanation block above it
5. Run the search package tests to confirm everything still works
6. Update `examples/next/tina-self-hosted-sqlite-demo/package.json` (when that example is added per #6786)

## Acceptance criteria

- [ ] `sqlite-level` v2.x verified published on npm
- [ ] Workspace catalog updated to `^2.0.0`
- [ ] `@tinacms/search/src/client/index.ts` uses clean named import; workaround code + TODO removed
- [ ] All existing search-related tests pass
- [ ] `@tinacms/search` builds without warnings about CJS interop
- [ ] (Optional, depends on order with #6786) New SQLite demo uses v2.x

[...]

PR #6905, opened 2026-05-18
chore: bump .nvmrc from v20 to v22 (Node 20 EOL)

### LINKED ISSUE #6686: Windows local install fails on Node 24 because TinaCMS pulls a better-sqlite3 line without matching prebuilt binaries
We are seeing a local installation failure on Windows with Node 24 in a project using TinaCMS.

The issue is not the React peer dependency warnings. The actual blocker is a native dependency in the TinaCMS package chain:

- `tinacms`
- `@tinacms/search`
- `sqlite-level`
- `better-sqlite3`

On Windows + Node 24, the currently resolved `better-sqlite3` version does not provide a matching prebuilt binary, so installation falls back to `node-gyp rebuild`. That then requires a local Python and MSVC toolchain and causes `npm install` to fail on machines that do not have that setup.

## What happens

`npm install` fails locally on Windows with Node 24.

The relevant error is:

```text
npm error path D:\...\node_modules\better-sqlite3
npm error command failed
npm error command C:\WINDOWS\system32\cmd.exe /d /s /c prebuild-install || node-gyp rebuild --release
npm error prebuild-install warn install No prebuilt binaries found (target=24.15.0 runtime=node arch=x64 libc= platform=win32)
...
npm error gyp ERR! find Python Could not find any Python installation to use
```

There are also peer dependency warnings from TinaCMS-related packages, but those are not the installation blocker in this case.

## Dependency chain

From our lockfile the relevant chain is:

- `tinacms`
- `@tinacms/search`
- `sqlite-level@1.2.1`
- `better-sqlite3@11.10.0`

The failure happens because `better-sqlite3@11.10.0` does not provide a matching prebuilt binary for our local target:

- `win32`
- `x64`
- Node ABI `137` (Node 24)

## Important context

This same project installs and builds successfully on Netlify with Node 24.

So the issue appears to be specifically about the local Windows installation path, not a general TinaCMS + Node 24 incompatibility.

## What we verified

We tested the latest `better-sqlite3` line separately on the same Windows + Node 24 machine.

`better-sqlite3@12.9.0` installs successfully in an isolated test and provides matching prebuilt binaries for this target, including:

- `node-v137-win32-x64`

That strongly suggests this issue could be resolved upstream by updating the TinaCMS dependency chain to a newer compatible `better-sqlite3` line, either directly or via an updated `sqlite-level` dependency.

## Requested change

Could TinaCMS please update the dependency chain used for local search and storage so that Node 24 local installs do not fall back to `node-gyp` for this case?

Concretely, this likely means one of the following:

- updating the relevant `sqlite-level` and `better-sqlite3` chain to a Node 24 compatible release line
- or otherwise switching to a dependency version set that provides prebuilt Windows x64 binaries for Node 24

## Why this matters

Right now, local Windows users on Node 24 are forced into native build tooling for a transitive dependency:

- Python
- Visual C++ Build Tools
- `node-gyp` toolchain setup

If TinaCMS moved to a compatible prebuilt binary line, local installation on Windows would become much smoother and would match the successful Node 24 experience we already see in CI and on Netlify.

Thanks for taking a look.

### Reproduction

 npx create-tina-app@latest

### Steps to reproduce

1. Use Windows x64.
2. Use Node 24
3. Run `npm install` in a project using TinaCMS 3.7.1.
4. Installation fails while building `better-sqlite3`.

### System Info

```shell
- OS: Windows 11
- Architecture: x64
- Node: 24.15.0
- npm: bundled with Node 24
- TinaCMS: 3.7.1
- Project type: plain JavaScript project using TinaCMS with Next.js
```

### Validations

kulesy:
## Context

Node 20 reached its official end-of-life on **2026-04-30** ([Node.js release schedule](https://endoflife.date/nodejs)). The contributor and CI default in this repo (\`.nvmrc\`) still points to Node 20, 18 days after EOL.

## What

Single one-line change: \`.nvmrc\` from \`v20\` to \`v22\`.

## Why now

Two reasons compounding:

1. **Node 20 is no longer receiving security patches.** Anyone running \`nvm use\` / \`fnm use\` in this repo today gets an EOL Node version. The repo's own end-user verification CI ([\`build-starter-templates.yml\`](.github/workflows/build-starter-templates.yml)) already runs against Node 22 / 24 / 25 — Node 20 is excluded from the user-facing matrix. The \`.nvmrc\` was just lagging.

2. **Coordinated with [tinacms/sqlite-level#34](https://github.com/tinacms/sqlite-level/pull/34).** That PR bumps \`better-sqlite3\` from \`^11.8.1\` to \`^12.9.0\` (resolves to 12.10.0) to fix [#6686](https://github.com/tinacms/tinacms/issues/6686) on Node 24. \`better-sqlite3@12.10.0\` [explicitly dropped Node 20 prebuilds](https://github.com/WiseLibs/better-sqlite3/releases/tag/v12.10.0): *"Add support for Node.js v26 prebuilds and remove EOL builds (Node.js v20, v23)"*. Once sqlite-level publishes from #34 and \`@tinacms/search\` consumes it via [#6848](https://github.com/tinacms/tinacms/issues/6848), the [\`publish.yml\`](.github/workflows/publish.yml) workflow — which uses \`node-version-file: .nvmrc\` — would hit a \`node-gyp rebuild\` attempt and fail without a local Python + C++ toolchain in the runner image.

This bump prevents that.

## Workflows affected

Workflows that use \`node-version-file: '.nvmrc'\` (and will now run on Node 22):

- \`.github/workflows/publish.yml\`
- \`.github/workflows/main.yml\`
- \`.github/workflows/e2e.yml\`
- \`.github/workflows/playwright-next-kitchen-sink.yml\`
- \`.github/workflows/playwright-react-kitchen-sink.yml\`
- \`.github/workflows/playwright-astro-kitchen-sink.yml\`
- \`.github/workflows/playwright-hugo-kitchen-sink.yml\`

Workflows with their own explicit \`node-version\` matrices are unaffected (\`build-starter-templates.yml\` already runs 22 / 24 / 25).

## Out of scope

- \`packages/create-tina-app/package.json\` declares \`engines.node: ">=18.18.0"\`. That's the *user-facing* floor for scaffolding new projects and is a separate decision — not changed here. Worth revisiting in a future PR.
- This does **not** change any production runtime targets. Tina packages themselves don't declare \`engines\` and are Node-version-agnostic at runtime.

## Test plan

- [ ] CI workflows pass on Node 22 (will run automatically when PR opens)
- [ ] No lockfile or package.json changes needed — this is purely the \`.nvmrc\` hint file


18-th (approved):
LGTM

PR #6906, opened 2026-05-18
fix(tinacms): hide highlight for markdown rich text fields

### LINKED ISSUE #6903: Bug Report 🐞 Rich text Markdown parser should filter/disable highlight to prevent broken .md output
@kulesy @JackDevAU

[🟥 Watch the video (1 min 1 sec)](https://www.youtube.com/watch?v=m3WslcqLlPY)

Rich text highlight in MD parser breaks output

In a collection using **MD** (not MDX) with a rich-text field configured to use the **Markdown parser**, Tina correctly filters out unsupported features like embeds, but it **does not filter out the "highlight" feature**.

When highlighted text is saved (e.g., colored red), the resulting edited Markdown file becomes **broken/invalid**, indicating that the highlight markup is being persisted even though it's not compatible with the Markdown parser output.

Expected: When a rich-text field is configured with the Markdown parser, editor features that cannot be represented safely in Markdown (like highlight) should be disabled or stripped from the output.
Actual: Highlight can be applied, saved, and it breaks the generated MD content.

### Reproduction
N/A (video demonstrates the behavior). A minimal repro repo can be added if needed.

### Steps to reproduce
1. Create/open a collection whose format is **MD**.
2. Add a rich text field and set its parser to **Markdown**.
3. In the Tina visual editor, highlight a portion of text.
4. Apply a highlight color (e.g., red).
5. Save the document.
6. Inspect the generated/edited `.md` file.
7. Observe the Markdown output is broken due to persisted highlight markup.

### System Info
N/A

### Validations
- [ ] Follow our [Code of Conduct](https://github.com/tinacms/tinacms/blob/main/CODE_OF_CONDUCT.md)
- [ ] Read the [docs](https://tina.io/docs).
- [ ] Check that there isn't [already an issue](https://github.com/tinacms/tinacms/issues) that reports the same bug.

---

## Investigation

1. **Parser is auto-derived correctly.** `packages/@tinacms/schema-tools/src/schema/TinaSchema.ts:51-60` assigns `parser: { type: 'markdown' }` to any rich-text field on a collection with `format: 'md'` (or anything other than `'mdx'`). Parsing works as expected.

2. **The toolbar does not consult `field.parser.type`.** `packages/tinacms/src/toolkit/fields/plugins/mdx-field-plugin/plate/components/fixed-toolbar-buttons.tsx` renders every entry in `toolbarItems` (or whatever `overrides.toolbar` lists), with no gate based on parser type. The only parser-aware code in the editor is a `slatejson` check in `plate/index.tsx:16`. (Note: embeds appear filtered only because no templates were registered for that field. If templates were registered, the embed button would show regardless of parser type, so embed is not actually format-aware today.)

3. **On save, the stringify path always emits JSX for highlights.** `serializeMDX` writes `<mark>...</mark>` via `mdxJsxTextElement` regardless of parser. The markdown parser does not recognize `<mark>` as a registered JSX pattern, so on reload the tag and its contents are dropped.

## Audit: which toolbar items round-trip through the markdown parser

| Button | Emits | Markdown-safe? |
|---|---|---|
| `heading` | `heading` | Yes |
| `link` | `link` | Yes |
| `image` | `image` | Yes |
| `hr` | `thematicBreak` | Yes |
| `quote` | `blockquote` | Yes |
| `ul` | `list` (unordered) | Yes |
| `ol` | `list` (ordered) | Yes |
| `bold` | `strong` | Yes |
| `italic` | `emphasis` | Yes |
| `strikethrough` | `delete` (GFM) | Yes |
| `code` | `inlineCode` | Yes |
| `codeBlock` | fenced `code` block | Yes |
| `mermaid` | `code` block with `lang: 'mermaid'` | Yes |
| `table` | `table` (GFM) | Yes |
| `raw` | UI-only toggle | N/A |
| `embed` | `mdxJsxFlowElement` or `mdxJsxTextElement` | No, unless the template defines a `match` shortcode pattern, in which case it survives via the shortcode mechanism |
| `highlight` | `<mark>` via `mdxJsxTextElement` | No |

## Proposed fix

When `field.parser?.type === 'markdown'`, filter the rendered toolbar to drop:

- **`highlight`** unconditionally. Pure JSX, cannot round-trip.
- **`embed`** only when the field has no templates with a `match` shortcode pattern. Templates with `match` survive in markdown.

The filter belongs in `fixed-toolbar-buttons.tsx`, alongside the existing `showEmbedButton` filtering on line 167. The `ToolbarProvider` context already exposes the field, so the parser type is accessible without prop drilling.

A first cut can land just the `highlight` removal since that is the actively reported case. The `embed` nuance can ship as a follow-up once we decide how to detect "has a match shortcode template" cleanly.

rotempasharel1:
Fixes #6903

## Summary

This PR hides the `highlight` toolbar item when a rich-text field uses the Markdown parser.

Highlight serializes to JSX markup, which does not safely round-trip through the Markdown parser and can result in broken `.md` output.

## Changes

- Pass the parser type into the toolbar context.
- Hide the `highlight` toolbar item when `parserType === 'markdown'`.
- Keep highlight available for non-Markdown rich-text fields.

## Notes

This PR intentionally keeps the scope small and does not change embed/shortcode behavior.

### DIFF
--- a/packages/tinacms/src/toolkit/fields/plugins/mdx-field-plugin/plate/components/fixed-toolbar-buttons.tsx
+++ b/packages/tinacms/src/toolkit/fields/plugins/mdx-field-plugin/plate/components/fixed-toolbar-buttons.tsx
@@ -143,7 +143,7 @@ const toolbarItems: { [key in ToolbarOverrideType]: ToolbarItem } = {
 export default function FixedToolbarButtons() {
   const toolbarRef = React.useRef(null);
   const [itemsShown, setItemsShown] = React.useState(11);
-  const { overrides, templates, field } = useToolbarContext();
+  const { overrides, templates, parserType } = useToolbarContext();
   const showEmbedButton = templates.length > 0;
 
   let items = [];
@@ -168,7 +168,7 @@ export default function FixedToolbarButtons() {
     items = items.filter((item) => item.label !== toolbarItems.embed.label);
   }
 
-  if (field?.parser?.type === 'markdown') {
+  if (parserType === 'markdown') {
     items = items.filter((item) => item.label !== toolbarItems.highlight.label);
   }
 
--- a/packages/tinacms/src/toolkit/fields/plugins/mdx-field-plugin/plate/components/fixed-toolbar-buttons.tsx
+++ b/packages/tinacms/src/toolkit/fields/plugins/mdx-field-plugin/plate/components/fixed-toolbar-buttons.tsx
@@ -143,7 +143,7 @@ const toolbarItems: { [key in ToolbarOverrideType]: ToolbarItem } = {
 export default function FixedToolbarButtons() {
   const toolbarRef = React.useRef(null);
   const [itemsShown, setItemsShown] = React.useState(11);
-  const { overrides, templates } = useToolbarContext();
+  const { overrides, templates, field } = useToolbarContext();
   const showEmbedButton = templates.length > 0;
 
   let items = [];
@@ -166,6 +166,10 @@ export default function FixedToolbarButtons() {
 
   if (!showEmbedButton) {
     items = items.filter((item) => item.label !== toolbarItems.embed.label);
+  }
+
+  if (field?.parser?.type === 'markdown') {
+    items = items.filter((item) => item.label !== toolbarItems.highlight.label);
   }
 
   const editorState = useEditorState();
--- a/packages/tinacms/src/toolkit/fields/plugins/mdx-field-plugin/plate/toolbar/toolbar-provider.tsx
+++ b/packages/tinacms/src/toolkit/fields/plugins/mdx-field-plugin/plate/toolbar/toolbar-provider.tsx
@@ -12,7 +12,7 @@ interface ToolbarContextProps {
   tinaForm: Form;
   templates: MdxTemplate[];
   overrides: ToolbarOverrideType[] | ToolbarOverrides;
-  field?: any;
+  parserType?: string;
 }
 
 interface ToolbarProviderProps extends ToolbarContextProps {
@@ -27,11 +27,13 @@ export const ToolbarProvider: React.FC<ToolbarProviderProps> = ({
   tinaForm,
   templates,
   overrides,
-  field,
+  parserType,
   children,
 }) => {
   return (
-    <ToolbarContext.Provider value={{ tinaForm, templates, overrides, field }}>
+    <ToolbarContext.Provider
+      value={{ tinaForm, templates, overrides, parserType }}
+    >
       {children}
     </ToolbarContext.Provider>
   );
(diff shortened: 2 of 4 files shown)


rotempasharel1:
Hi! I updated the PR title and description, signed the CLA, added a changeset, and changed the toolbar context to pass only `parserType` instead of the full field object.

The checks currently show `action_required`; could a maintainer please approve the workflow run?

joshbermanssw (changes requested):
Thanks for trying to tackle this @rotempasharel1 !

The highlighter functionality is still bypassable via a keyboard shortcut (On MacOS Cmd + Shift + H) 

Perhaps the ideal fix is removing/ disabling `HighlightPlugin` 

Additionally nested rich-text fields can silently keep highlight on markdown collections when using the filter `field.parser.type === 'markdowm'` check

rotempasharel1:
I fixed it, I'd be happy if you would approve the merge.

packages/tinacms/src/toolkit/fields/plugins/mdx-field-plugin/plate/toolbar/toolbar-provider.tsx
```diff
@@ -12,6 +12,7 @@ interface ToolbarContextProps {
   tinaForm: Form;
   templates: MdxTemplate[];
   overrides: ToolbarOverrideType[] | ToolbarOverrides;
+  parserType?: string;
```
JackDevAU, line 15: can we change this from a string to `parserType?: Parser['type']` ?

packages/tinacms/src/toolkit/fields/plugins/mdx-field-plugin/plate/components/fixed-toolbar-buttons.tsx
```diff
@@ -168,6 +168,10 @@ export default function FixedToolbarButtons() {
     items = items.filter((item) => item.label !== toolbarItems.embed.label);
   }
 
+  if (parserType === 'markdown') {
+    items = items.filter((item) => item.label !== toolbarItems.highlight.label);
```
JackDevAU, line 172: Can we instead filter by key, not label?
e.g. filter on the ToolbarOverrideType key (or HighlightPlugin.key) instead of item.label 
Just worried if we ever change the label here.. will this break. Same applies to the embed filter above

packages/tinacms/src/toolkit/fields/plugins/mdx-field-plugin/plate/components/fixed-toolbar-buttons.tsx
```diff
@@ -168,6 +168,10 @@ export default function FixedToolbarButtons() {
     items = items.filter((item) => item.label !== toolbarItems.embed.label);
   }
 
+  if (parserType === 'markdown') {
```
JackDevAU, line 171: can we also add some tests around this - test if highlight is filtered when parserType === 'markdown', and kept for 'mdx'/undefined etc. (would recommend vitetest and @testing-library/react should have examples in the codebase)

JackDevAU (changes requested):

kulesy:
Hey @rotempasharel1, thanks for this, and sorry for the slow reply.

Closing because the review showed hiding the toolbar button is not enough on its own: highlight is still reachable with Cmd+Shift+H, and a nested rich-text field can keep it because the parser type is read from the outer field.

None of that is wasted. I have written what your PR and the review established into https://github.com/tinacms/tinacms/issues/6903, so whoever picks it up starts from there rather than repeating the same approach. Very welcome to be that person if you are still keen.

Issue #6437, opened 2026-02-03
🐛Optional datetime fields autofill a value instead of allowing null/empty


matt-antone:
### 🐛 Describe the Bug

When a `datetime` field is marked optional (`required: false`)—e.g. an “Expiration” field where content may have no end date—the editor still auto-fills a date value (e.g. current date). Saving then writes that value to the content file, so we cannot persist “no date” (null/empty) for optional datetime fields. Required datetime fields (e.g. “Date”) behave correctly; the issue is only with optional ones. In older Tina versions we could leave optional datetime fields empty.

### 🧑‍💻 Reproduction Steps

1. Define a collection (e.g. blog) with a template that has a required datetime (“Date”) and an optional datetime (“Expiration” with `required: false`), e.g.:
   ```ts
   { type: "datetime", name: "date", label: "Date", description: "The date the page was created." },
   { type: "datetime", name: "expiration", label: "Expiration", required: false }
   ```
2. Create or open a document that has `date` set but **no** `expiration` in frontmatter (key missing or null).
3. Open the document in the Tina editor.
4. Observe: the Expiration field shows a value (e.g. today’s date) even though the content has none.
5. Save the document.
6. See that the content file now contains `expiration:` with an ISO string (the autofilled value) instead of remaining empty/null.

### 🎯 Expected Behaviour

- Optional `datetime` fields (e.g. “Expiration”) should allow **no value**.
- If the editor leaves the field empty, the saved document should have **no date** for that field (key omitted or value `null`).
- No date should be written for that field unless the user explicitly picks one. Required datetime fields (e.g. “Date”) should continue to require a value.

### 💼 Tasks

- [ ] For `datetime` fields with `required: false`, do not initialize the field with a default value (e.g. “today”) in the form/date picker.
- [ ] Ensure initial value for optional datetime is `undefined` / `null` so the UI shows empty and saving without user input does not write a date.
- [ ] (Alternatively) Document a way to disable the default value for optional datetime (e.g. `ui.defaultValue: undefined` or “no initial value”).

### ✅ Acceptance Criteria

- [ ] Opening a document with no value for an optional `datetime` field shows the field as empty in the editor (no autofilled date).
- [ ] Saving that document does not write a date value for the optional field (key remains absent or value is `null`).
- [ ] Required `datetime` fields are unchanged and still require a value.

### ➕ More Information

[...]

PR #6926, opened 2026-05-19
Fix optional datetime fields defaulting to current date

Aibono1225:
Relates to #6437 

This pull request addresses an issue where optional datetime fields in TinaCMS were automatically filled with the current date when no value was set. The changes ensure that only required fields default to the current date, while optional fields remain unset unless a value is provided.

**Bug fix for optional datetime fields:**

* Fixed the logic in `DateField` (in `date-field-plugin.tsx`) so that optional datetime fields (`required: false`) do not auto-fill with the current date; only required fields will default to the current date if no value is set.
* Updated the `Field` interface to include an optional `required` property, allowing fields to explicitly specify whether they are required.

image

**Figure: No autofilled date for optional field**


packages/tinacms/src/toolkit/fields/plugins/date-field-plugin.tsx
```diff
@@ -44,15 +44,25 @@ export const DateField = wrapFieldsWithMeta<InputProps, DatetimepickerProps>(
       return dateFormat;
     }, [dateFormat]);
 
-    const date = input.value ? new Date(input.value) : new Date();
+    const date = input.value
+      ? new Date(input.value)
+      : (rest as any).required === false
```
joshbermanssw: to avoid using `any` type and improve readability of this, I would:
1. add required as an optional field on the `Field` interface (see `src/toolkit/forms/field.ts`) 
2. destructure in `field: { .... }` 
3. use `required` instead of `(rest as any).required`

Aibono1225: ✅ Done

joshbermanssw (dismissed):
lgtm just add format pls

joshbermanssw (approved):

PR #6929, opened 2026-05-19
Fix dropdown options hidden behind fields

### LINKED ISSUE #5684: Template dropdown overlap issue

When using a list of object fields, using templates, the dropdown will underlay the form label and content boxes, making it impossible to add more.

You can get around this by adding a few of the first item to make the list long enough, which means you can access the further-down dropdown items.

Image

Can't access the 3rd onwards here

### Reproduction

https://github.com/JackBailey/tinaoverlap-demo

### Steps to reproduce

Create a object field with `list: true`, with at least 3 templates.
Try to use the 3rd item or further, you are unable to do so.

### System Info

```shell
System:
    OS: Linux 6.14 Fedora Linux 42 (KDE Plasma Desktop Edition)
    CPU: (12) x64 AMD Ryzen 5 7640U w/ Radeon 760M Graphics
    Memory: 7.95 GB / 30.65 GB
    Container: Yes
    Shell: 5.2.37 - /bin/bash
  Binaries:
    Node: 22.14.0 - ~/.nvm/versions/node/v22.14.0/bin/node
    npm: 10.9.2 - ~/.nvm/versions/node/v22.14.0/bin/npm
    pnpm: 10.8.0 - ~/.nvm/versions/node/v22.14.0/bin/pnpm
  npmPackages:
    @tinacms/cli: ^1.9.5 => 1.9.7 
    tinacms: ^2.7.5 => 2.7.7
```

### Validations

Aibono1225:
Relates to #5684 

This pull request addresses a UI bug where dropdown options in fields were being hidden behind other elements in the `tinacms` package. The main fix involves improving how the `zIndex` style is applied to field components, ensuring dropdowns display correctly.

**Bug Fixes (Dropdown Visibility):**

* Updated the `style` prop in both `ListFieldMeta` (`list-field-meta.tsx`) and `FieldMeta` (`wrap-field-with-meta.tsx`) components to set `zIndex` only when `index` is defined, preventing dropdowns from being hidden behind other fields.

**Before:**
image

**After:**
image

### DIFF
--- a/packages/tinacms/src/toolkit/fields/plugins/list-field-meta.tsx
+++ b/packages/tinacms/src/toolkit/fields/plugins/list-field-meta.tsx
@@ -50,7 +50,7 @@ export const ListFieldMeta = ({
       margin={margin}
       {...hoverEvents}
       onClick={() => setFocusedField({ id: tinaForm.id, fieldName: name })}
-      style={{ zIndex: index !== undefined ? 1000 - index : undefined }}
+      style={{ zIndex: index != null ? 1000 - index : undefined }}
       {...props}
     >
       <ListHeader>
--- a/packages/tinacms/src/toolkit/fields/plugins/list-field-meta.tsx
+++ b/packages/tinacms/src/toolkit/fields/plugins/list-field-meta.tsx
@@ -50,7 +50,7 @@ export const ListFieldMeta = ({
       margin={margin}
       {...hoverEvents}
       onClick={() => setFocusedField({ id: tinaForm.id, fieldName: name })}
-      style={{ zIndex: index ? 1000 - index : undefined }}
+      style={{ zIndex: index !== undefined ? 1000 - index : undefined }}
       {...props}
     >
       <ListHeader>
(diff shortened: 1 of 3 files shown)


packages/tinacms/src/toolkit/fields/plugins/list-field-meta.tsx
```diff
@@ -50,7 +50,7 @@ export const ListFieldMeta = ({
       margin={margin}
       {...hoverEvents}
       onClick={() => setFocusedField({ id: tinaForm.id, fieldName: name })}
-      style={{ zIndex: index ? 1000 - index : undefined }}
+      style={{ zIndex: index !== undefined ? 1000 - index : undefined }}
```
joshbermanssw: if index is `null` index !== undefined would return true... `index != null` would catch both

Aibono1225: ✅ Done

joshbermanssw (approved):

PR #6941, opened 2026-05-21
fix(@tinacms/graphql): pad datetime index keys to fix pre-2001 sorting

### LINKED ISSUE #4519: Sorting by datetime breaks ordering for dates before September 10th 2001

When you have collection items with the `datetime` field and you sort by that field in the UI and also using sort in the API client GraphQL connection, any collection item fields that have a date older than September 10th 2001, seem to get mistaken for being a more recent collection item in the sort order:

Posts sorted by releaseDate:
firefox_xMgz03SDev

Post test2 with releaseDate September 10th 2001:
firefox_AQs9EJLMd0

Post test3 with releaseDate September 9th 2001:
firefox_ANmUpjPfUf

### Reproduction

https://github.com/HarveyPeachey/bug-sort-date-sep-2001

### Steps to reproduce

Run `npm run dev`
Visit http://localhost:3000/admin/index.html#/collections/post/~
Click sort by dropdown and select `Release Date (Oldest First)`
See test3 being moved to the bottom of the sorted list when it should be at the top

### System Info

```shell
System:
    OS: Windows 10 10.0.19045
    CPU: (12) x64 AMD Ryzen 5 3600 6-Core Processor
    Memory: 15.74 GB / 31.95 GB
  Binaries:
    Node: 20.11.1 - C:\Program Files\nodejs\node.EXE
    Yarn: 1.22.22 - ~\Documents\Projects\bug-sort-date-sep-2001\node_modules\.bin\yarn.CMD
    npm: 8.5.2 - ~\AppData\Roaming\npm\npm.CMD
    pnpm: 9.0.6 - ~\AppData\Roaming\npm\pnpm.CMD
  npmPackages:
    @tinacms/cli: ^1.5.34 => 1.5.45
    tinacms: ^1.5.24 => 1.6.3
```

### Validations

isaaclombardssw:
## Summary
- Stores datetime index keys as ISO 8601 strings instead of Unix timestamps, fixing lexicographic sorting for all dates
- Normalizes timezone-naive datetime strings to UTC so index keys are deterministic across build environments
- Adds number padding to composite (multi-field) indexes, which previously had no padding at all

Closes #4519

## Root cause
LevelDB uses lexicographic (string) ordering. Datetime values were stored as Unix timestamps in milliseconds, which broke sorting in two ways:
1. **Digit-length boundary** — timestamps crossed from 12 to 13 digits on Sep 10, 2001, so `"999993600000"` (Sep 9) sorted after `"1000080000000"` (Sep 10)
2. **Pre-1970 dates** — negative timestamps (e.g. `"-86400000"` for Dec 31, 1969) don't sort correctly as strings at all

ISO 8601 strings (`YYYY-MM-DDTHH:mm:ss.sssZ`) are inherently lexicographically sortable across all dates, so no padding or timestamp conversion is needed.

## Timezone handling
Hand-edited frontmatter can contain datetime strings without a timezone indicator (e.g. `2025-01-15T10:30:00`). Per the JS spec, `new Date()` parses these as local time, which would produce different index keys depending on the build machine's timezone. All datetime strings are now normalized to UTC before indexing — if a value has a `T` but no `Z` or offset, `Z` is appended before parsing.

## Changes
- **`datalayer.ts`** — added `ensureUTC`/`parseDatetimeUTC` helpers; replaced `new Date(x).getTime()` with `parseDatetimeUTC(x)` in index key building, filter operand coercion, and in-memory filtering; removed `DEFAULT_DATETIME_LPAD`
- **`index.ts`** — removed datetime padding from index definitions; added number padding to composite indexes (was missing before)
- **`movie-zeta.md`** — test fixture with pre-1970 date (`1965-03-15`)
- **`index.test.ts`** — test verifying correct sort order for pre-1970 dates and across the former digit boundary
- Updated snapshots, cursors, and unit test expectations for ISO string format

## Test plan
- [x] All 517 `@tinacms/graphql` tests pass (6 pre-existing failures in `validate.test.ts`/`filter-utils.test.ts` unrelated to this change)
- [ ] Manual test: create posts with dates before and after Sep 10, 2001, verify sort order

**Index rebuild required:** This changes the index key format. Local dev/build users are unaffected (indices rebuild automatically in-memory). Self-hosted setups with persistent storage should run a full `tinacms build` to rebuild indices.


isaaclombardssw:
Screenshot 2026-05-21 at 5 23 57 pm

**Figure: before**

Screenshot 2026-05-21 at 5 25 32 pm

**Figure: after**

18-th (approved):
LGTM

PR #6947, opened 2026-05-22
fix: match rich text editor typography with frontend styles

### LINKED ISSUE #6700: Make rich text editor typography match frontend styles
@tiagov8 @Freego1783 @adamcogan @joshbermanssw 

## Pain

The rich text editor in Tina currently uses different typography from the actual website. For example, headings in the editor are shown with a serif font, while the frontend uses Inter.

This creates a mismatch between the editing experience and the final rendered content.

## Expected Behavior

The rich text editor should allow the same typography styles as the frontend, so the content in the editor looks closer to the final result.

## Suggestion

It would be great to have a way to define typography styles once and share them between:

- the website
- the Tina rich text editor

This would make the editing experience more truly WYSIWYG.

### Screenshots

Image

MirkaJuliet34:
## What changed

Updated the rich text editor typography to match the frontend typography styles.

Previously, the rich text editor used `Libre Baskerville`, while the frontend displayed content with a different typography style. This created a mismatch between the editing experience and the final rendered content.

The update changes the editor typography so content displayed inside the editor more closely reflects what users see on the actual site.

## Why

This improves the WYSIWYG experience and makes editing feel more consistent with the frontend.

Fixes #6700

## How to test

1. Run the TinaCMS kitchen sink example
2. Open the Hero section
3. Edit content using the Rich Text field
4. Create a Heading (H1/H2)
5. Compare the typography in the editor and rendered page
6. Verify both use the same typography style


joshbermanssw:
Hey @MirkaJuliet34  - please don't forget to add a changeset - `pnpm changeset`

packages/@tinacms/app/src/index.css
```diff
@@ -14,7 +14,7 @@
 }
 
 .font-libre-baskerville {
```
18-th: The handle doesn't match the font inside anymore. This needs to be aligned and all references need to be updated

MirkaJuliet34: Thanks for pointing that out! I renamed the class to match the actual font being used and updated all related references accordingly.

MirkaJuliet34:
@joshbermanssw Thanks for the reminder! Added a changeset for this update.

joshbermanssw:
Nice work, and the direction is right — the kitchen-sink frontend uses Inter (via `next/font/google`), so matching the editor to it makes sense. Two follow-ups before merge:

- [ ] **Remove Libre Baskerville from the font loader.** Nothing references it anymore after this change, but `packages/tinacms/src/toolkit/styles/font-loader.tsx:8` still downloads all four weights of `Libre+Baskerville` from Google Fonts on every editor load. Trim it to just Inter: `https://fonts.googleapis.com/css2?family=Inter:wght@400;600&display=swap`.
- [ ] **Rename `headerSerifStyle`.** The constant at `components.tsx:170` (applied to H5/H6) now sets `fontFamily: "'Inter', sans-serif"`, so the `Serif` name is misleading — suggest `headerFontStyle`.

MirkaJuliet34:
@joshbermanssw Thanks for the review! I updated the naming (`headerFontStyle`), verified the font loader, and confirmed that only Inter is being loaded now. The requested changes have been pushed.

18-th (approved):
LGTM

18-th:
@MirkaJuliet34 thanks again for your contribution, the PR is now merged 🙂

MirkaJuliet34:
Hi! Thanks again for merging my previous contribution.

I'd love to continue contributing to TinaCMS.

Is there any frontend/UI issue that would be a good next contribution while the v4 work is being scoped?

18-th:
Hey @MirkaJuliet34, have a look at the backlog. There are quite a few issues tagged with `UI` and not tagged with `v4`. You can have a go at any one of them 🙂

PR #6961, opened 2026-05-26
feat(create-tina-app): make Astro the default starter

kulesy:
## Summary

- Astro is now the pre-selected option in the interactive `What starter code would you like to use?` prompt (moved to the top of `TEMPLATES`, and the ⭐ "recommended" marker is shifted from `NextJS starter` / `TinaDocs` onto `⭐ Astro Starter`).
- Introduces a resolved `DEFAULT_TEMPLATE` export in `templates.ts` so the default lives in one place. The prompt's `initial` index is derived from it, so any future reordering won't desync the visual cursor, and a bad value throws at import instead of degrading silently.
- When `create-tina-app` is invoked **without a TTY** (e.g. piped, CI, scripted) and no `--template` flag is given, Astro is selected directly instead of falling into the prompt and exiting as `user cancelled`. Existing interactive behavior is unchanged for TTY users beyond the new default.

## Behavioural diff

| Invocation | Before | After |
|---|---|---|
| Interactive, no `--template` | Prompt opens on `⭐ NextJS starter` | Prompt opens on `⭐ Astro Starter` |
| `--template tina-nextjs-starter` (or any other) | Unchanged | Unchanged |
| Non-TTY **with `--pkg-manager` and a project name**, no `--template` | Prompt fires, gets empty input, exits with `ERR_CANCEL_TEMPLATE_PROMPT` | Astro selected, scaffolding proceeds |
| Non-TTY, bare (no `--pkg-manager`, no project name) | Stops at the package manager prompt | Unchanged, still stops there |
| Non-TTY (e.g. `< /dev/null`), with `--template foo` | Unchanged | Unchanged |

**Scope note:** this PR only gives the *template* prompt a non-TTY fallback. The package manager prompt (`index.ts:226-271`) and the project name prompt (`index.ts:274-307`) are still ungated, so a bare `npx create-tina-app` in CI exits 0 without scaffolding, same as before. Covering all three prompts behind an explicit `--yes` is tracked separately.

The existing `ERR_CANCEL_TEMPLATE_PROMPT` telemetry event now only fires when an interactive user dismisses the prompt, which is what its `errorCategory: 'user-cancellation'` already claimed. Non-TTY runs still record `telemetryData.template = 'tina-astro-starter'` via the existing line below the `if (!template)` block, so PostHog analytics still capture the choice.

## Test plan

- [x] `pnpm tsc` in `packages/create-tina-app` — passes
- [x] `pnpm format` (CI command) — no diff
- [x] `pnpm build` — produces dist/index.js containing the resolved default and the `!process.stdin.isTTY` branch
- [x] Non-TTY smoke test: `node bin/create-tina-app my-test-app --pkg-manager npm --noTelemetry < /dev/null` — output includes `✔ Downloading files from repo tinacms/tina-astro-starter` and `dev server http://localhost:4321`; resulting `package.json` has `"astro": "^6.4.4"`
- [x] Bare non-TTY: `node bin/create-tina-app my-test-app --noTelemetry < /dev/null` — stops at the package manager prompt, nothing scaffolded (documented above, not fixed here)
- [ ] Reviewer smoke (manual, requires a real terminal): run `node packages/create-tina-app/bin/create-tina-app` interactively and confirm the cursor lands on `⭐ Astro Starter`

### DIFF
--- a/packages/create-tina-app/src/index.ts
+++ b/packages/create-tina-app/src/index.ts
@@ -11,7 +11,12 @@ import {
 } from './util/fileUtil';
 import { install } from './util/install';
 import { initializeGit, makeFirstCommit } from './util/git';
-import { TEMPLATES, Template, downloadTemplate } from './templates';
+import {
+  DEFAULT_TEMPLATE_VALUE,
+  TEMPLATES,
+  Template,
+  downloadTemplate,
+} from './templates';
 import { preRunChecks } from './util/preRunChecks';
 import { checkPackageExists } from './util/checkPkgManagers';
 import { TextStyles, TextStylesBold, box } from './util/textstyles';
@@ -302,30 +307,44 @@ export async function run() {
   }
 
   if (!template) {
-    const res = await prompts({
-      name: 'template',
-      type: 'select',
-      message: 'What starter code would you like to use?',
-      choices: TEMPLATES.map(formatTemplateChoice),
-    });
-    if (!Object.hasOwn(res, 'template')) {
-      postHogCaptureError(
-        posthogClient,
-        userId,
-        sessionId,
-        new Error('User cancelled template selection'),
-        {
-          errorCode: ERROR_CODES.ERR_CANCEL_TEMPLATE_PROMPT,
-          errorCategory: 'user-cancellation',
-          step: TRACKING_STEPS.TEMPLATE_SELECT,
-          fatal: true,
-          additionalProperties: telemetryData,
-        }
+    const defaultTemplate = TEMPLATES.find(
+      (t) => t.value === DEFAULT_TEMPLATE_VALUE
+    );
+    const defaultTemplateIndex = defaultTemplate
+      ? TEMPLATES.indexOf(defaultTemplate)
+      : 0;
+
+    if (!process.stdin.isTTY && defaultTemplate) {
+      template = defaultTemplate;
+    } else {
+      const res = await prompts({
+        name: 'template',
+        type: 'select',
+        message: 'What starter code would you like to use?',
+        choices: TEMPLATES.map(formatTemplateChoice),
+        initial: defaultTemplateIndex,
+      });
+      if (!Object.hasOwn(res, 'template')) {
+        postHogCaptureError(
+          posthogClient,
+          userId,
+          sessionId,
+          new Error('User cancelled template selection'),
+          {
+            errorCode: ERROR_CODES.ERR_CANCEL_TEMPLATE_PROMPT,
+            errorCategory: 'user-cancellation',
+            step: TRACKING_STEPS.TEMPLATE_SELECT,
+            fatal: true,
+            additionalProperties: telemetryData,
+          }
+        );
+        if (posthogClient) await posthogClient.shutdown();
+        exit(1);
+      }
+      template = TEMPLATES.find(
+        (_template) => _template.value === res.template
       );
-      if (posthogClient) await posthogClient.shutdown();
-      exit(1);
     }
-    template = TEMPLATES.find((_template) => _template.value === res.template);
   }
   telemetryData['template'] = template.value;
 
--- a/packages/create-tina-app/src/index.ts
+++ b/packages/create-tina-app/src/index.ts
@@ -307,12 +307,12 @@ export async function run() {
   }
 
   if (!template) {
-    const defaultTemplate = TEMPLATES.find(
+    const foundDefaultIndex = TEMPLATES.findIndex(
       (t) => t.value === DEFAULT_TEMPLATE_VALUE
     );
-    const defaultTemplateIndex = defaultTemplate
-      ? TEMPLATES.indexOf(defaultTemplate)
-      : 0;
+    const defaultTemplate =
+      foundDefaultIndex >= 0 ? TEMPLATES[foundDefaultIndex] : undefined;
+    const defaultTemplateIndex = foundDefaultIndex >= 0 ? foundDefaultIndex : 0;
 
     if (!process.stdin.isTTY && defaultTemplate) {
       template = defaultTemplate;
--- a/packages/create-tina-app/src/index.ts
+++ b/packages/create-tina-app/src/index.ts
@@ -12,7 +12,7 @@ import {
 import { install } from './util/install';
 import { initializeGit, makeFirstCommit } from './util/git';
 import {
-  DEFAULT_TEMPLATE_VALUE,
+  DEFAULT_TEMPLATE,
   TEMPLATES,
   Template,
   downloadTemplate,
@@ -307,22 +307,15 @@ export async function run() {
   }
 
   if (!template) {
-    const foundDefaultIndex = TEMPLATES.findIndex(
-      (t) => t.value === DEFAULT_TEMPLATE_VALUE
-    );
-    const defaultTemplate =
-      foundDefaultIndex >= 0 ? TEMPLATES[foundDefaultIndex] : undefined;
-    const defaultTemplateIndex = foundDefaultIndex >= 0 ? foundDefaultIndex : 0;
-
-    if (!process.stdin.isTTY && defaultTemplate) {
-      template = defaultTemplate;
+    if (!process.stdin.isTTY) {
+      template = DEFAULT_TEMPLATE;
     } else {
       const res = await prompts({
         name: 'template',
         type: 'select',
         message: 'What starter code would you like to use?',
         choices: TEMPLATES.map(formatTemplateChoice),
-        initial: defaultTemplateIndex,
+        initial: TEMPLATES.indexOf(DEFAULT_TEMPLATE),
       });
       if (!Object.hasOwn(res, 'template')) {
         postHogCaptureError(
--- a/packages/create-tina-app/src/templates.ts
+++ b/packages/create-tina-app/src/templates.ts
@@ -27,11 +27,33 @@ export type ExternalTemplate = BaseExample & {
 };
 export type Template = InternalTemplate | ExternalTemplate;
 
+export const DEFAULT_TEMPLATE_VALUE = 'tina-astro-starter';
+
 export const TEMPLATES: Template[] = [
   {
-    title: '⭐ NextJS starter',
+    title: '⭐ Astro Starter',
     description:
-      'Kickstart your project with Next.js – our top recommendation for a seamless, performant, and versatile web experience.',
+      'Get started with Astro - a modern static site generator designed for fast, lightweight, and flexible web projects.',
+    value: 'tina-astro-starter',
+    isInternal: false,
+    features: [
+      {
+        name: 'Visual Editing',
+        description: '✅',
+      },
+      {
+        name: 'SSG',
+        description: '✅',
+      },
+    ],
+    gitURL: 'https://github.com/tinacms/tina-astro-starter',
+    branch: 'main',
+    devUrl: 'http://localhost:4321',
+  },
+  {
+    title: 'NextJS starter',
+    description:
+      'Kickstart your project with Next.js – a seamless, performant, and versatile web experience.',
     value: 'tina-nextjs-starter',
     isInternal: false,
     features: [
@@ -53,7 +75,7 @@ export const TEMPLATES: Template[] = [
     devUrl: 'http://localhost:3000',
   },
   {
-    title: '⭐️ TinaDocs',
+    title: 'TinaDocs',
     description:
       'Get your documentation site up and running with TinaCMS and Next.js in minutes.',
     value: 'tina-docs',
@@ -75,26 +97,6 @@ export const TEMPLATES: Template[] = [
     gitURL: 'https://github.com/tinacms/tina-docs',
     branch: 'main',
     devUrl: 'http://localhost:3000',
-  },
-  {
-    title: 'Astro Starter',
-    description:
-      'Get started with Astro - a modern static site generator designed for fast, lightweight, and flexible web projects.',
-    value: 'tina-astro-starter',
-    isInternal: false,
-    features: [
-      {
-        name: 'Visual Editing',
-        description: '✅',
-      },
-      {
-        name: 'SSG',
-        description: '✅',
-      },
-    ],
-    gitURL: 'https://github.com/tinacms/tina-astro-starter',
-    branch: 'main',
-    devUrl: 'http://localhost:4321',
   },
   {
     title: 'Hugo Starter',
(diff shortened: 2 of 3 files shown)


kulesy:
We are going to hold off on this until we have polished our astro-starter template.
https://github.com/tinacms/tina-astro-starter/issues/37

kulesy:
🛑 **Hold for polish before merge.**

This is mergeable as-is, but we've decided to wait until the Astro starter (`tinacms/tina-astro-starter`) has had another polish pass before flipping the default. The shape of the change is settled; the timing isn't. Converting to draft so it's clear this isn't imminent.

Heads-up: the Astro team has noticed this PR and would like to mention it in the May 31 *What's New in Astro* newsletter — see [withastro/docs#13936 (comment)](https://github.com/withastro/docs/pull/13936#issuecomment-4543947388). We've replied to flag that the May 31 window is unlikely, and we'll re-ping them when this lands.

kulesy:
**Canary testing evidence for defaulting to Astro** (PBI: #7179): before flipping the default, the real `tina-astro-starter` was validated end to end by 4 independent testers, each running the full clone, TinaCloud, and deploy flow and signing off in their own child issue. All four passed, and every blocking bug surfaced during testing is resolved:

- #7181 Tester 1 sign-off - Eddie K ✅
- #7182 Tester 2 sign-off - Calum S ✅
- #7183 Tester 3 sign-off - Baba ✅
- #7184 Tester 4 sign-off - Rob T ✅

Supplementary: the nightly [starter-template build canary](https://github.com/tinacms/tinacms/actions/workflows/build-starter-templates.yml) also has `tina-astro-starter` green in all 6 cells (yarn/npm/pnpm x Node 22/24) across the [last](https://github.com/tinacms/tinacms/actions/runs/29570435741) [three](https://github.com/tinacms/tinacms/actions/runs/29639277177) [runs](https://github.com/tinacms/tinacms/actions/runs/29681810304).

joshbermanssw (dismissed):

packages/create-tina-app/src/index.ts
```diff
@@ -302,30 +307,44 @@ export async function run() {
...
-          additionalProperties: telemetryData,
-        }
+    const foundDefaultIndex = TEMPLATES.findIndex(
+      (t) => t.value === DEFAULT_TEMPLATE_VALUE
+    );
+    const defaultTemplate =
+      foundDefaultIndex >= 0 ? TEMPLATES[foundDefaultIndex] : undefined;
+    const defaultTemplateIndex = foundDefaultIndex >= 0 ? foundDefaultIndex : 0;
+
+    if (!process.stdin.isTTY && defaultTemplate) {
```
joshbermanssw: **This branch is unreachable in the case the PR description advertises.**

The table says non-TTY with no `--template` now scaffolds. It only does that if the caller *also* passes `--pkg-manager` and a positional project name, because two ungated prompts run before this one:

- `index.ts:226-271` package manager
- `index.ts:274-307` project name

Repro on the current build:

```
$ node bin/create-tina-app my-test-app --noTelemetry < /dev/null
✔ Node v24.14.1 is supported.
? Which package manager would you like to use? › - Use arrow-keys. Return to submit.
❯   pnpm (recommended)
$ echo $?
0        # nothing scaffolded, no error, no telemetry
```

It dies at the package manager prompt, well before reaching this line. The PR's own smoke test passes `--pkg-manager npm` plus a positional name, which is what masks it.

Both earlier prompts already have defaults available (`initial: pnpm` on :245, `initial: 'my-tina-app'` on :280), so one helper covers all three call sites:

```ts
const interactive = Boolean(process.stdin.isTTY);
async function askOrDefault<T>(config, fallback: T, cancelCtx): Promise<T> { ... }
```

Otherwise the description needs scoping to "non-TTY **with `--pkg-manager` and a project name**", so this doesn't land on the expectation that bare `npx create-tina-app` works in CI.

packages/create-tina-app/src/index.ts
```diff
@@ -302,30 +307,44 @@ export async function run() {
...
+      template = defaultTemplate;
```
joshbermanssw: Two smaller things on this assignment.

**It's silent.** Everywhere else this CLI narrates itself. One line here ("No `--template` given and no interactive terminal detected, using Astro Starter") turns an opaque CI log into a readable one.

**`isTTY` is a proxy, not the signal.** It gets the piped case right but misses a container run with `-t`: TTY present, no human, so the prompt hangs forever instead of defaulting. The convention other scaffolders settled on is an explicit `--yes`, optionally with `process.env.CI` as a secondary hint. Follow-up rather than a blocker, but worth doing if the non-interactive path grows to cover the other prompts.

packages/create-tina-app/src/index.ts
```diff
@@ -302,30 +307,44 @@ export async function run() {
...
-          step: TRACKING_STEPS.TEMPLATE_SELECT,
-          fatal: true,
-          additionalProperties: telemetryData,
-        }
+    const foundDefaultIndex = TEMPLATES.findIndex(
+      (t) => t.value === DEFAULT_TEMPLATE_VALUE
+    );
+    const defaultTemplate =
+      foundDefaultIndex >= 0 ? TEMPLATES[foundDefaultIndex] : undefined;
+    const defaultTemplateIndex = foundDefaultIndex >= 0 ? foundDefaultIndex : 0;
```
joshbermanssw: This defensive branch can't execute. `DEFAULT_TEMPLATE_VALUE` is a literal in the same module as `TEMPLATES`, so `foundDefaultIndex` is provably `0`.

The cost isn't the dead code, it's the failure mode it sets up: if the constant is ever typo'd, the `&& defaultTemplate` guard on :317 silently falls through to the prompt, which in non-TTY means the exit-0 no-op described above. Quiet degradation, hard to trace.

Suggest moving the invariant into the module that owns `TEMPLATES` so it fails loudly instead (see the comment on `templates.ts:30`). That deletes all three of these lines.

packages/create-tina-app/src/index.ts
```diff
@@ -302,30 +307,44 @@ export async function run() {
...
+      }
+      template = TEMPLATES.find(
+        (_template) => _template.value === res.template
       );
-      if (posthogClient) await posthogClient.shutdown();
-      exit(1);
     }
-    template = TEMPLATES.find((_template) => _template.value === res.template);
   }
   telemetryData['template'] = template.value;
```
joshbermanssw, line 342: This now records `tina-astro-starter` identically whether the user selected it, passed `--template`, or never saw a prompt.

Since the point of this field is measuring which starter people pick, auto-selections will inflate Astro's share in exactly the window where you most want a clean read on whether promoting Astro worked.

```suggestion
  telemetryData['template'] = template.value;
  telemetryData['template-source'] = opts.template
    ? 'flag'
    : process.stdin.isTTY
      ? 'prompt'
      : 'default';
```

packages/create-tina-app/src/templates.ts
```diff
@@ -27,36 +27,34 @@ export type ExternalTemplate = BaseExample & {
 };
 export type Template = InternalTemplate | ExternalTemplate;
 
+export const DEFAULT_TEMPLATE_VALUE = 'tina-astro-starter';
```
joshbermanssw: Good to have this in one place. One tweak: export the *resolved* template rather than the raw string, so a typo fails at import instead of silently degrading in `index.ts`:

```ts
// after the TEMPLATES array
export const DEFAULT_TEMPLATE = TEMPLATES.find(
  (t) => t.value === 'tina-astro-starter'
);
if (!DEFAULT_TEMPLATE) {
  throw new Error('DEFAULT_TEMPLATE missing from TEMPLATES');
}
```

`index.ts` then reduces to `TEMPLATES.indexOf(DEFAULT_TEMPLATE)` and the three ternaries on :310-315 disappear.

Deriving the prompt's `initial` from a constant instead of a hardcoded `0` is the right call either way. That's the part of this PR still paying off in a year.

packages/create-tina-app/src/templates.ts
```diff
@@ -27,36 +27,34 @@ export type ExternalTemplate = BaseExample & {
...
       },
-      {
-        name: 'ISR',
-        description: '✅',
-      },
       {
         name: 'SSG',
         description: '✅',
       },
     ],
```
joshbermanssw, line 50: Astro is the only template that omits the ISR row entirely. Hugo (:113) and Remix (:137) both list `ISR: ❌` explicitly.

Rendered through `formatTemplateChoice`, the starred recommendation shows two checkmarks while the option directly below it shows three, with nothing explaining the gap.

```suggestion
    features: [
      {
        name: 'Visual Editing',
        description: '✅',
      },
      {
        name: 'ISR',
        description: '❌',
      },
      {
        name: 'SSG',
        description: '✅',
      },
    ],
```

Pre-existing, but this PR is what moves it to the top of the list.

packages/create-tina-app/src/templates.ts
```diff
@@ -27,36 +27,34 @@ export type ExternalTemplate = BaseExample & {
 };
 export type Template = InternalTemplate | ExternalTemplate;
 
+export const DEFAULT_TEMPLATE_VALUE = 'tina-astro-starter';
+
 export const TEMPLATES: Template[] = [
   {
-    title: '⭐ NextJS starter',
+    title: '⭐ Astro Starter',
```
joshbermanssw, line 32: Copy nits, now that this entry is the headline:

- Titles in the top three are inconsistent: `⭐ Astro Starter`, `NextJS starter` (lowercase s), `TinaDocs`.
- The description on :36 uses a hyphen where the Next.js one uses an en dash, and "static site generator" undersells Astro given it does SSR too.
- "our top recommendation" was stripped from Next.js but never added here, so the promoted option now carries the flatter blurb of the two.

joshbermanssw (commented):
Reviewed the diff, both file versions, and the surrounding code, and reproduced the non-TTY behaviour locally.

**First, a false alarm worth clearing:** the reorder is done as an in-place field swap rather than by moving object literals, so the diff reads as though the feature lists got scrambled. They didn't. Diffing the resulting file against `main`, Astro keeps `Visual Editing + SSG`, Next.js keeps `Visual Editing + ISR + SSG`, and TinaDocs keeps `Visual Editing + ISR + SSG`. All correct.

**One blocker**, inline on `index.ts:317`: the non-TTY branch is unreachable in the scenario the description advertises. Two ungated prompts (package manager, project name) run before it, so bare `npx create-tina-app` in CI still exits 0 having done nothing. The PR's own smoke test passes `--pkg-manager npm` plus a positional name, which is what hides it.

Relatedly, the "Before" column isn't accurate. With a closed stdin the `prompts` promise never resolves, the event loop drains, and node exits **0** with no error and no PostHog event, so `ERR_CANCEL_TEMPLATE_PROMPT` never fired on that path to begin with. Not a code defect, but it means the real prior failure was a silent success-coded no-op, and that failure still exists at the two earlier prompts.

The rest is small: a dead defensive branch, telemetry that can no longer tell a choice from a default, a silent selection, and the ISR row missing from the now-promoted Astro entry.

Good parts: deriving `initial` from a named constant instead of a hardcoded `0` is the right call and is the piece of this that keeps paying off. The changeset is correctly scoped. The canary evidence in the thread (four tester sign-offs plus 6/6 green nightly cells across three runs) is solid diligence for a change that redirects every new user's default path.

Housekeeping: `mergeStateStatus` is `BEHIND`, so this needs a refresh against `main`. The May 26 "hold for polish" looks cleared by the July 20 canary comment, but the PR body still doesn't reflect that. And the Astro team's newsletter interest (withastro/docs#13936) is an external commitment tied to this landing, so someone should re-ping them on merge as promised.

Leaving this as a comment rather than a formal block, but I'd want finding 1 resolved (either the fix extended to the other prompts, or the description scoped to what it actually delivers) before this goes in.

packages/create-tina-app/src/index.ts
joshbermanssw: Applied in 5cffacbd1. `templates.ts` now resolves the default next to `TEMPLATES` and throws at import if it is missing, so `index.ts` is down to `TEMPLATES.indexOf(DEFAULT_TEMPLATE)` and all three ternaries are gone.

packages/create-tina-app/src/templates.ts
joshbermanssw: Applied in 5cffacbd1, resolved export rather than the raw string. Agreed on the `initial`-from-a-constant part; that is the piece still paying off in a year.

packages/create-tina-app/src/templates.ts
joshbermanssw: Applied in b8b1b43f9. Astro now carries `ISR: ❌`, matching Hugo, Remix and Docusaurus.

packages/create-tina-app/src/templates.ts
joshbermanssw: Applied in b8b1b43f9: `Next.js Starter` for the casing, comma instead of the en dash, and the "our top recommendation" blurb moved onto Astro so the promoted option is not the flatter of the two. Left `TinaDocs` alone since it is a product name rather than a framework starter.

packages/create-tina-app/src/index.ts
joshbermanssw: Reproduced again on the current build, still exits 0 at the package manager prompt with nothing scaffolded. Went with the scoping option rather than growing this PR: the behavioural table now has an explicit bare-non-TTY row, and the changeset says so too (99b626c25), since that text ships in the release notes. The `askOrDefault` helper covering all three prompts is filed as #7330.

packages/create-tina-app/src/index.ts
joshbermanssw: Both points moved to #7330 rather than into this PR. Agreed `isTTY` is the weaker signal; the container-with-`-t` case still hangs, and that wants an explicit `--yes`. The missing narration line is in there too.

packages/create-tina-app/src/index.ts
joshbermanssw: Leaving this one to you, it is an analytics schema call rather than a cleanup. The concern stands though: for the window where you most want to read whether promoting Astro worked, flag-passed and never-prompted runs are currently indistinguishable from deliberate picks. Happy to push the four-line version if you want it in this PR.

joshbermanssw:
@kulesy took the liberty of pushing the three mechanical items from my review straight to the branch, since they were all decision-free. Shout if you'd rather I'd left them.

- b8b1b43f9 Astro's missing `ISR: ❌` row, plus the title casing and the "our top recommendation" blurb
- 5cffacbd1 resolved `DEFAULT_TEMPLATE` export, which deletes the unreachable lookup branch in `index.ts`
- 99b626c25 scopes the changeset's non-TTY claim

Verified: `tsc` clean on each of the three commits individually, `biome lint` and `biome format` clean, build emits the resolved default, and the non-TTY smoke test scaffolds Astro end to end.

Two left for you: the `template-source` telemetry field is your call, and the `--yes` flag is filed as #7330. Nothing here blocks the approval.

18-th (approved):
LGTM

PR #6995, opened 2026-05-29
fix(media): sanitize and NFC-normalize uploaded filenames

### LINKED ISSUE #6359: 🤖🐛 Media Manager - Sanitize/normalize asset filenames
Reported from Discord:

Users are experiencing issues with media assets not sanitizing filenames correctly when uploading files that contain special characters.

"Tina doesn't sanitize filenames, which can cause different systems to recognize the files differently (I've had a combination of macOS and git struggle with ß in filename, for instance. By default, git also can struggle with capitalization in filenames)"

https://discord.com/channels/835168149439643678/1325453715682889770

We would need to add the filename sanitization around here:
https://github.com/tinacms/tinacms/blob/ced9076d25d7177bd8d9ae3c6133d0560366511c/packages/tinacms/src/toolkit/core/media-store.default.ts#L111

Ben0189:
## What

Sanitize and normalize uploaded media filenames so they stay consistent across operating systems, git, and the URLs that end up in content.

Closes #6359

## Why

Filenames were passed through to the media stores verbatim. This caused two problems:

- Names exposed in decomposed (NFD) form — common on macOS — were percent-encoded into content fields as multi-codepoint sequences, producing image URLs that didn't resolve on the frontend.
- Special characters and inconsistent casing could break URLs or be recognized differently across filesystems and git.

## What changed

- Added a shared `sanitizeFilename` helper (`packages/tinacms/src/toolkit/components/media/utils.ts`) that:
  - normalizes to NFC so a single codepoint is encoded in URLs
  - replaces URL-breaking (`# % &`) and filesystem-reserved (`< > : " | ? *`) characters with hyphens, collapsing the resulting runs
  - strips control characters and any path components
  - caps the base-name length while preserving the extension
- Applied it in the default, S3, Azure, Cloudinary, and DigitalOcean Spaces media stores, so the upload key, the stored asset, and the value persisted into content all agree.
- Exported the helper from `tinacms` and added unit tests.

## Testing

- `vitest run` on the new `utils.test.ts` — 14/14 passing (covers NFC normalization, macOS-style NFD input, URL/filesystem-unsafe characters, separator collapsing, length cap, and fallbacks).

## Notes

Sanitization is applied client-side at upload time. Editors who have already uploaded files with affected names will need to re-upload them to pick up the normalized name.

### DIFF
--- a/packages/next-tinacms-azure/src/azure-media-store.ts
+++ b/packages/next-tinacms-azure/src/azure-media-store.ts
@@ -5,7 +5,7 @@ import type {
   MediaStore,
   MediaUploadOptions,
 } from 'tinacms';
-import { DEFAULT_MEDIA_UPLOAD_TYPES } from 'tinacms';
+import { DEFAULT_MEDIA_UPLOAD_TYPES, sanitizeFilename } from 'tinacms';
 import { E_UNAUTHORIZED, E_BAD_ROUTE, interpretErrorMessage } from './errors';
 
 export type AzureMediaStoreOptions = {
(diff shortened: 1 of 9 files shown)


isaaclombardssw (approved):
✅ This code makes sense, but possible filename duplication effect – should have a separate PR looking into de-duping behaviour

PR #7006, opened 2026-06-01
Harden Forestry migration code generation against untrusted input

18-th:
This closes [#3593](https://github.com/tinacms/tinacloud/issues/3598)

The Forestry migrator serializes field definitions and un-quotes generator-produced expressions via an internal marker. Make that marker unforgeable with a per-process random nonce so values originating from migration input cannot be emitted as executable code in the generated template file.

Also extract the pure naming helpers (stringifyLabel, stringifyLabelWithField) into a dependency-free util/naming module, breaking a circular import so the code generator is unit-testable in isolation. Adds regression tests for the marker round-trip and the hardening.

### DIFF
--- a/.changeset/perky-ears-serve.md
+++ b/.changeset/perky-ears-serve.md
@@ -1,0 +1,5 @@
+---
+"@tinacms/cli": patch
+---
+
+Harden Forestry migration code generation against untrusted input
--- a/packages/@tinacms/cli/src/cmds/forestry-migrate/index.ts
+++ b/packages/@tinacms/cli/src/cmds/forestry-migrate/index.ts
@@ -18,6 +18,7 @@ import {
   makeTemplateFile,
   makeFieldsWithInternalCode,
 } from './util/codeTransformer';
+import { stringifyLabel, stringifyLabelWithField } from './util/naming';
 
 const BODY_FIELD = {
   // This is the body field
@@ -28,13 +29,7 @@ const BODY_FIELD = {
   isBody: true,
 };
 
-export const stringifyLabel = (label: string) => {
-  return label.replace(/[^a-zA-Z0-9]/g, '_').toLowerCase();
-};
-export const stringifyLabelWithField = (label: string) => {
-  const labelString = stringifyLabel(label);
-  return `${labelString}Fields`;
-};
+export { stringifyLabel, stringifyLabelWithField };
 const transformForestryMatchToTinaMatch = (match: string) => {
   const newMatch = match
     // remove white space
--- a/packages/@tinacms/cli/src/cmds/forestry-migrate/util/codeTransformer.test.ts
+++ b/packages/@tinacms/cli/src/cmds/forestry-migrate/util/codeTransformer.test.ts
@@ -1,0 +1,111 @@
+import type { TinaField } from '@tinacms/schema-tools';
+import {
+  addVariablesToCode,
+  makeFieldsWithInternalCode,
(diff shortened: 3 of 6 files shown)


AnGrY-Althaf (commented):
Thanks @18-th — I took the patch for a spin and can confirm it closes the vulnerability. Verification details below.

## Verified: the nonce-marker approach fully neutralises the injection

I drove the real `generateCollections()` from the `fix-forestry-migrate-command` branch against the original PoC plus an extended battery, and ran the new regression suite.

**Original attack vector — all neutralised:**

| Payload | `ae1ab5d0f` (before) | This PR |
|---|---|---|
| IIFE break-out `1}] }; (function(){…})(); function _i(){ return [{x:1` | 🔴 RCE | ✅ inert quoted string |
| Comma-operator `(globalThis.X=1,'x')` | 🔴 RCE | ✅ inert quoted string |
| Bare legacy marker `__TINA_INTERNAL__:::…:::` | 🔴 RCE | ✅ inert quoted string |
| Forged nonce-style marker (`__TINA_INTERNAL_deadbeef…__`) | — | ✅ inert (wrong nonce) |
| Close-then-reopen a legitimate marker | — | ✅ inert |

**"Another way around" — every other attacker-controlled position is also inert:**

- `select` / `list` field `options[]` values → quoted
- nested `field_group` child `label` → quoted
- field `name` (sanitised + `nameOverride`) → quoted
- `section.label` in `settings.yml` (the collections path through `apply.ts`) → quoted

In every case the injected sentinel stays a quoted JSON string in the generated `tina/templates.ts` / collections output — never a bare executable token.

**Regression suite:** `codeTransformer.test.ts` → 6/6 passing (legitimate marker round-trip + the injection-hardening cases).

## Why it holds

- Both code-gen sinks (`codeTransformer.ts` templates and `apply.ts` collections) funnel through the single hardened `addVariablesToCode`, so there's one chokepoint, not two.
- No raw-concatenation path: every attacker string is either `JSON.stringify`'d (now nonce-gated) or identifier-sanitised via `stringifyLabel` (`[^a-zA-Z0-9] → _`) for function names / imports.
- The 128-bit per-process nonce is unguessable and has no oracle — migration runs entirely on the victim's host with no feedback channel — so it can't be brute-forced or leaked back to an attacker.
- `.` doesn't cross newlines and `JSON.stringify` escapes real newlines, so no multiline trick can bridge a legitimate marker's `OPEN…CLOSE` around attacker content.

LGTM. 👍

## Optional hardening (non-blocking)

The current fix is sufficient as-is. One belt-and-suspenders idea if you ever want to revisit: the scheme still relies on string-matching (`"OPEN(.*?)CLOSE"`) plus nonce secrecy. You could remove the string-matching surface entirely by **not routing user content and generated-code markers through the same `JSON.stringify` pass** — e.g.:

- Serialise the field array, substituting each *internally generated* expression with a unique placeholder token collected in a `Map<token, rawExpression>`.
- After `JSON.stringify`, replace each placeholder by looking it up in that map (exact-string `split`/`join` on the token, no regex, no `.*?`).

Because the placeholders are object-identity-keyed and only ever produced by the generator, user data can never collide with one regardless of its contents — it sidesteps the "what if user input reconstructs the delimiter" question by construction rather than by secrecy. Purely defence-in-depth; the nonce already covers the practical threat.


Happy to help review if you pick that up. Thanks for the quick turnaround on the fix!

JackDevAU (approved):
:shipit:

PR #7005, opened 2026-06-01
chore(v4): scaffold @tinacms/tinacms package structure

JackDevAU:
This addresses #6930 

First v4 commit. Adds an empty @tinacms/tinacms package under
packages/v4/ with the folder structure for the v4 architecture:

- adapters/{react,astro}/    framework integrations
- bridge/                    visual editing iframe infra
- core/                      types, config, plugin registry, lifecycle
- store/{slices,middleware}/ Zustand
- plugins/{fields,media,search,auth}/  built-in plugins
- rich-text/                 production renderer
- cli/{commands,server,graphql,codegen}/
- app/                       admin SPA
- ui/                        Radix-wrapped primitives
- utils/

No code yet — folders are tracked with .gitkeep so the structure is
visible. package.json is a minimal private marker; build/exports map
will be added when implementations start landing.

The unscoped v3 tinacms package stays untouched under packages/tinacms/
and continues shipping under tinacms@3.x during the v4 transition.

PR #7008, opened 2026-06-02
[⚠️ DO NOT MERGE] (v4): package deprecation notices

JackDevAU:
⚠️ Merge https://github.com/tinacms/tinacms/pull/7005 first as this branch is based off it

add deprecation notices to all packages that won't be transitioning over to v4 or adding notes to where they will be moved


JackDevAU:
As per my conversation with @18-th we are closing this as we are not yet ready to deprecate all these packages. Instead we will do this after v4 has launched.

PR #7018, opened 2026-06-03
🧹 Derive supported Node versions from a single source of truth

### LINKED ISSUE #4822: `create-tina-app` CLI - Don't hardcode supported Node versions.
Currently the `create-tina-app` hardcodes the supported Node versions. This leads to tech debt if not updated.

We support all supported supported LTS versions.

## Tasks
- [ ] Investigate into a way to retrieve all supported LTS versions
- [ ] Replace the static variable with a dynamic list

joshbermanssw:
## TL;DR

`create-tina-app` hardcoded its supported Node versions in a bounds constant that, due to an off-by-design bug, accepted v20–v41 instead of the intended v20–v22. This makes `engines.node` the source of truth for the CLI's runtime check instead.

## Why

Per #4822, supported versions were hardcoded and had drifted — `engines` said `>=18.18.0` (Node 18 is EOL) while the CLI bounds said 20–22, and the bounds math was wrong on top of that.

## How

- `create-tina-app`'s `engines.node` (`22.x || 24.x`) is the source of truth for the CLI.
- `preRunChecks` reads that `engines` value at runtime and parses the supported majors, replacing the buggy bounds constant.
- Root `package.json` gains the same `engines.node` for the dev environment.
- The starter-template CI matrix is set to `[22, 24]`.

## Reviewer notes

- **LTS-only.** Node 18 (EOL) and odd/non-LTS lines are no longer accepted; the CLI now allows 22 and 24.

## Follow-up

The supported versions are still edited by hand. A scheduled job that bumps `engines.node` when Node's LTS set changes (e.g. when 26 promotes to LTS) would make it fully dynamic — out of scope here.

## Links

- Closes #4822

### DIFF
--- a/.changeset/node-version-ssot.md
+++ b/.changeset/node-version-ssot.md
@@ -1,0 +1,5 @@
+---
+"create-tina-app": patch
+---
+
+Derive supported Node versions from `engines.node` instead of a hardcoded list.
--- a/.github/workflows/build-starter-templates.yml
+++ b/.github/workflows/build-starter-templates.yml
@@ -14,39 +14,14 @@ on:
         default: ''
         type: string
   schedule:
-    # Daily at 09:00 UTC. The workflow's full matrix (7 templates × 3 PMs ×
-    # the supported Node majors, derived from root engines.node) is the broader
-    # safety net for starter-template drift — daily cadence keeps worst-case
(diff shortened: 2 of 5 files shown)


18-th (approved):
LGTM

PR #7026, opened 2026-06-03
docs: align bug report template with Discord post guidelines

kulesy:
## What hurts

The current bug report template asks for *"describe the issue"* — a single free-text blob. The literal error string, expected-vs-actual split, and *"anything non-default"* (the field that most often points us straight at root cause) all live in the user's head, not in the form. Our team and our AI triage tools end up guessing.

## What this does

Replaces the blob with structured fields matching the same checklist now enforced in the Discord `#ask-for-help` post guidelines. The exact error gets its own copy-pasteable `render: shell` field. The custom-adapter detail is explicit in the environment prompt.

Same form, just prompts for the things that matter.


KahaMason (approved):
✅ LGTM

PR #7028, opened 2026-06-04
🐛 Skip filesystem cache on edge runtimes to prevent client hang

### LINKED ISSUE #7017: Docs: deploying the TinaCMS Astro starter to Cloudflare Workers
## Summary

We need a guide covering how to deploy the TinaCMS Astro starter template to Cloudflare Workers. Today there is no walkthrough for this path, so users adapting the Astro starter for a Workers deployment are left to piece it together themselves.

## Why

Cloudflare Workers (with the `@astrojs/cloudflare` adapter) is a common target for Astro sites, and the TinaCMS Astro starter is a frequent entry point for new users. A clear, end-to-end guide reduces setup friction and cuts down on repeat support questions.

## Scope

Documentation only. The guide should cover:

- Prerequisites (Cloudflare account, Wrangler, a TinaCloud project or self-hosted setup)
- Configuring the Astro Cloudflare adapter for the starter
- Building the Tina admin (`tinacms build`) and the Astro site, and where each output lands
- Environment variables needed at build and runtime (client ID, token, branch, search, etc.)
- Deploying via Wrangler and/or the Cloudflare dashboard / Git integration
- Verifying the deployed admin (`/admin`) and the content API
- Known gotchas (Node compatibility flags, edge runtime limits, asset/media handling)

## Acceptance criteria

- [ ] A new docs page walks through deploying the Astro starter to Cloudflare Workers from a clean checkout to a live site
- [ ] Required environment variables are listed with where to set them
- [ ] Steps are verified against the current Astro starter and a real Cloudflare Workers deploy
- [ ] The guide is linked from the relevant Astro / deployment docs index

## Notes

Reference the existing Astro kitchen-sink (`examples/astro/kitchen-sink/`) for the current starter setup. Confirm the Astro version and adapter currently shipped before writing steps.

JackDevAU:
✅ Tested and working!! https://astro.jack-082.workers.dev/

The generated Tina client enables an fs-backed NodeCache whenever a cacheDir is set and `window` is undefined. On Cloudflare Workers (workerd) the Node fs shim is present but read/write calls hang, so the per-key cache lock never releases and concurrent identical queries deadlock. Detect edge runtimes (Cloudflare Workers, Vercel Edge) in init() and skip the cache so requests take the lock-free path. Also add a `cache` client option to force-disable the cache.

### DIFF
--- a/.changeset/astro-cloudflare-import-meta-url.md
+++ b/.changeset/astro-cloudflare-import-meta-url.md
@@ -1,0 +1,5 @@
+---
+"@tinacms/astro": patch
+---
+
+Make the experimental island route work under the `@astrojs/cloudflare` adapter by supplying a valid `import.meta.url` to server bundles only. Adds a `cloudflareWorkers` option to force the workaround on or off.
--- a/.changeset/edge-cache-skip.md
+++ b/.changeset/edge-cache-skip.md
@@ -1,0 +1,5 @@
+---
+"tinacms": patch
+---
+
+Skip the filesystem-backed response cache on edge runtimes (Cloudflare Workers, Vercel Edge) where Node's `fs` API is present but unusable, which could otherwise hang concurrent identical queries. Adds a `cache` option to `createClient` to force-disable the cache.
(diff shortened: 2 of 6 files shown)


18-th (approved):
LGTM

PR #7037, opened 2026-06-05
🐛 Surface real error when create-tina-app dependency install fails

wicksipedia:
## What

When `create-tina-app`'s "install packages" step failed, the CLI printed:

```
✖ Failed to install packages: undefined
```

…with no detail, because:

- `src/util/install.ts` rejected with a plain object `{ command }` (not an `Error`), so `.message` was `undefined`.
- stderr was discarded (`stdio: 'ignore'`) outside `--verbose`, hiding the real cause.

## Changes

- **`install.ts`** — buffer stderr (even in non-verbose mode), reject with a real `Error` containing the command, exit code, and captured output. Handle the spawn `error` event. Detect pnpm's `ERR_PNPM_IGNORED_BUILDS` and append a hint to run `approve-builds`.
- **`index.ts`** — safe message extraction so a non-`Error` reject can never print `undefined` again.
- Changeset (patch).

Before → after on failure:

```
✖ Failed to install packages: "pnpm install" exited with code 1.

  <real pnpm stderr>

  Some dependencies' build scripts were blocked by pnpm. Run
  "pnpm approve-builds" (then "pnpm install") in your project directory.
```

## Notes

Addresses the empty-error report in #7031. The underlying blocked-build (better-sqlite3/sharp/esbuild/core-js under pnpm 10+) is fixed in the starter repo: tinacms/tina-astro-starter#57.

### DIFF
--- a/.changeset/create-tina-app-install-error.md
+++ b/.changeset/create-tina-app-install-error.md
@@ -1,0 +1,5 @@
+---
+"create-tina-app": patch
+---
+
+Surface the real cause when dependency installation fails. Previously a failed install showed `Failed to install packages: undefined` with no detail; the CLI now captures the package manager's output and reports the exit code and error, including a hint to run `approve-builds` when pnpm has blocked dependency build scripts.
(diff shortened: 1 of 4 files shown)


JackDevAU (approved):
:shipit:

PR #7055, opened 2026-06-12
fix(tinacms): prevent crash when reopening Global Configuration after navigating to a collection

### LINKED ISSUE #7038: 🟥 🐛 Bug - Navigation between collections and global configuration causes undefined error and breaks UI
Requested by: @JimmyChenSSW via [YakShaver.ai](https://YakShaver.ai) 🦬
cc: @JimmyChenSSW, @kulesy, @JackDevAU @isaaclombardssw 

Hi Team!

&nbsp;&nbsp;[🟥&nbsp;&nbsp;Watch the video (1 min 37 sec)](https://www.youtube.com/watch?v=V2jO3BMqT5E)

&nbsp;&nbsp;Url from screen share: [https://localhost:4321/admin/index.html#/screens/global_config](https://localhost:4321/admin/index.html#/screens/global_config)

### Pain
When using ⭐ TinaCMS with Astro, navigating between a collection (e.g. Blog) and the Global Configuration via the Hamburger menu causes a runtime error ("Cannot read properties of undefined"). After the error occurs, the UI becomes stuck and the user cannot navigate back to other pages or sections. This appears to happen consistently when a user clicks into a collection and then navigates away to Global Configuration or another section.

### Acceptance Criteria
1. Users can navigate between any collection (e.g. Blog) and Global Configuration without encountering runtime errors.
2. The error "Cannot read properties of undefined" no longer occurs during normal navigation.
3. After navigating between sections, the UI remains responsive and allows further navigation.
4. The fix applies consistently across all collection sections, not just Blog.
5. Navigation behavior works reliably for non-technical users who click between sections arbitrarily.

### Reproduce Steps
1. Open a project using ⭐ TinaCMS with Astro.
2. Open the TinaCMS admin UI.
3. Open the Hamburger menu.
4. Click on Global Configuration and wait for it to load.
5. Open the Hamburger menu again.
6. Click on a collection (e.g. Blog).
7. Open the Hamburger menu once more and click back to Global Configuration (or another section).
8. Observe the runtime error "Cannot read properties of undefined" and the UI becoming unusable.

### Screenshot
Screenshot
**Figure: Navigation bug when switching between collections and global config causes undefined error and blocks navigation**

Aibono1225:
Relates to #7038 

## Summary
Navigating Global Configuration → a collection → Global Configuration (via the hamburger menu) in the TinaCMS admin crashed with `Cannot read properties of undefined`, locking the UI on a "TinaCMS Render Error" card. This re-establishes the global form so the round trip works and the form stays editable.

## Root cause
The Global Configuration screen is registered as a persistent `GlobalFormPlugin` in `cms.plugins`. Visual-editing navigation cleanup (`cms.removeAllForms()` + `form-lists:clear`) empties `cms.state.forms`, but the plugin stays registered. On returning to the Global screen, `GlobalFormPlugin.Component` found no matching form and passed `undefined` to `FormBuilder`, which dereferences `form.tinaForm` → crash. Collection update pages avoid this because their remount path re-dispatches form state; the global screen had no equivalent.

## Fix
When the backing form is missing from `cms.state.forms`, `GlobalFormPlugin.Component` re-establishes it from the plugin's stored `Form` instance (captured at registration) via `forms:add`, and restores `activeFormId` via `forms:set-active-form-id` — mirroring how collection update pages re-add form state on mount. Both dispatches are idempotent (`forms:add` dedupes by id), so there are no duplicate forms and no render loop (effect deps `[cms, cmsForm]`); a transient `null` render avoids passing `undefined` to `FormBuilder` for the single frame before the re-add lands. Scoped to the global-screen lifecycle; collection behavior is untouched.

No `FormBuilder` guard was added — an early-return guard there would violate the repo's `useHookAtTopLevel` lint rule, and the plugin-level fix already guarantees `FormBuilder` never receives an undefined form.

## Tests run
- New `global-form-plugin.test.tsx` (drives the real reducer): asserts the missing form is re-added, `FormBuilder` never receives an undefined form, `activeFormId` is restored, and no duplication when the form already exists.
- `forms` + `plugin-screens` suites pass; Biome lint clean; TypeScript adds no new errors.

## Real Astro starter verification
Reproduced and verified against a real Astro starter using `@tinacms/astro` visual editing, linked to the local `tinacms` build. **Before:** Global → Blog → Global crashed with `Cannot read properties of undefined (reading 'tinaForm')` and the render-error card. **After:** repeated arbitrary navigation (Blog→Global, Pages→Global, Blog→Global) renders the full, editable Global Configuration form every time with no crash and no console error.

## Limitations
None functional. The `null` render is a single frame and only while the form is briefly absent; the global form's `final-form` state survives in the stored instance, so data is preserved on return.


kulesy:
**Local before/after verification** 🧪

Reproduced #7038 against `examples/astro/visual-editing` (uses `@tinacms/astro` visual editing, matching the report) with the local `tinacms` build — toggling only `packages/tinacms/src/toolkit/plugin-screens/index.tsx`.

Steps: load admin → **Site → Global** (open & close) → **Collections → Blogs** → **Site → Global** (`#/screens/global`).

**Before (un-fixed):**

before – TinaCMS Render Error card

Figure: `TinaCMS Render Error` — `Cannot read properties of undefined (reading 'tinaForm')`, thrown in `FormBuilder` under `GlobalFormPlugin.Component`.

**After (this PR):**

after – Global form renders and is editable

Figure: The Global form renders and stays editable (Name / Color / Nav Links), with no console errors.

kulesy (approved):
Approving — reviewed the code and verified the fix end-to-end.

**Code:** Correct and well-scoped. `GlobalFormPlugin.Component` re-establishes the stored form via `forms:add` (dedupes by id) + restores `activeFormId`, and returns `null` for the single frame before the re-add lands so `FormBuilder` never gets `undefined`. Effect deps `[cms, cmsForm]` stabilize after the re-add (no render loop), mirroring the `CollectionUpdatePage` mount pattern. Tests drive the real reducer; lint/types clean.

**Verified live** against a real `@tinacms/astro` visual-editing example linked to a local build: *Before* — Global → Blog → Global reproduced `Cannot read properties of undefined (reading 'tinaForm')` + the render-error card. *After* — the same navigation renders the editable Global form every time; repeated/arbitrary cycles stay clean (0 console errors), the recovered form edits live with correct iframe binding, and collection editing is unaffected.

Non-blocking nit: unlike `CollectionUpdatePage`, the effect has no unmount cleanup (never nulls `activeFormId`/removes the form). Confirmed benign in practice — the leftover active-form state is always followed by a navigation that resets it — so optional polish, not a correctness issue.

PR #7058, opened 2026-06-15
Fix Local Mode banner for absolute contentApiUrlOverride

### LINKED ISSUE #5520: ✨ Remove Local Mode Banner Without Requiring Relative URL
CC: @JackDevAU @ncn-ssw @kldavis4 @wicksipedia 

### Description

Update TinaCMS to allow the removal of the Local Mode banner without requiring `contentApiUrlOverride` to be a relative URL. Currently, the logic detects non-relative URLs and automatically enables the Local Mode banner, even when an authentication provider is configured. This issue aims to adjust the behavior so that simply setting `contentApiUrlOverride` is sufficient to remove the banner.

### Acceptance Criteria

- [ ] Modify the logic that determines when to show the Local Mode banner.
- [ ] Ensure `contentApiUrlOverride` does not need to be a relative URL to disable Local Mode.
- [ ] Verify that users with a configured authentication provider do not see the Local Mode banner.
- [ ] Add documentation on `contentApiUrlOverride`
- [ ] Update documentation to clarify how `contentApiUrlOverride` interacts with Local Mode detection.
- [ ] Add tests to confirm the expected behavior when using an absolute `contentApiUrlOverride`.

🤖

Aibono1225:
Closes #5520

## PR Summary

**Root cause** — `parseURL` classified any non-TinaCloud absolute URL as `isLocalClient: true`. As a result, an absolute `contentApiUrlOverride` routed the admin UI to `LocalClient` (`isLocalMode === true`), causing the Local Mode banner to appear even when an auth provider was configured.

**Fix** — Updated the non-TinaCloud branch of `parseURL` to return `isLocalClient: false`. A non-TinaCloud absolute override represents a custom/self-hosted content API, not the local dev server. True local dev server URLs containing `localhost` are still handled by the earlier local branch, so existing Local Mode behavior is preserved.

**Tests** — Added/updated coverage for relative and absolute `contentApiUrlOverride` values, including the auth provider case. Also verified true local mode still returns `LocalClient` / `isLocalMode: true`.

**Docs** — Updated the `contentApiUrlOverride` JSDoc to clarify that it may be relative or absolute, and that setting it opts the admin UI out of Local Mode detection.

## Testing

* `pnpm --filter @tinacms/schema-tools test -- parseURL`
* `pnpm --filter tinacms test -- src/utils/index.test.ts src/internalClient/index.test.ts`
* `pnpm --filter @tinacms/schema-tools types`

### DIFF
--- a/.changeset/hot-rice-pull.md
+++ b/.changeset/hot-rice-pull.md
@@ -1,0 +1,6 @@
+---
+"@tinacms/schema-tools": patch
+"tinacms": patch
+---
+
+Fix Local Mode banner for absolute contentApiUrlOverride
--- a/packages/@tinacms/schema-tools/src/types/index.ts
+++ b/packages/@tinacms/schema-tools/src/types/index.ts
@@ -654,6 +654,14 @@ export interface Config<
   Store = undefined,
   SearchClient = undefined,
 > {
+  /**
+   * Points the admin UI at a custom/self-hosted content API instead of TinaCloud.
+   *
+   * Can be a relative URL, such as `/api/tina/gql`, or an absolute URL, such as
+   * `https://example.com/api/content`. When set, the admin UI will not show the
+   * Local Mode banner; Local Mode is only detected when this option is unset and
+   * the content API resolves to `localhost`.
+   */
   contentApiUrlOverride?: string;
   authProvider?: AuthProvider;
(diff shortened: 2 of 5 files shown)


18-th (approved):
LGTM

Issue #5611, opened 2025-04-03
tina-astro-starter returns 404 when user clicks on newly created content entries


tinanoob:
### 🐛 Describe the Bug
When running the "tina-astro-starter" locally in development mode,
new entries added to a collection aren't rendered properly after **Save**.
The entry appears in the list of entries in the collection but clicking on the entry shows a 404 page instead.  
This condition persists until tina/astro is restarted manually.

### 🎯 Expected Behaviour 
The new content should render properly after it's created and the user clicks on the new content.
The user should not have to manually restart tina/astro after adding a new entry.

### Notes

I am new to Tina and Astro and cannot determine if this is a bug in Tina, Astro or a limitation of the Tina-Astro integration.  I also reported this as https://github.com/tinacms/tina-astro-starter/issues/17 since if this is a known limitation of the tina-astro integration the documentation on the website should state that.

### Reproduction

https://github.com/tinanoob/tinacms-astro-new-content-404

### Steps to reproduce

1. `cd /tmp`
2. `pnpx create-tina-app@latest --template tina-astro-starter`
     ✔ Which package manager would you like to use? › pnpm
     ✔ What is your project named? … my-tina-app
3. `cd my-tina-app`
4. `pnpm run dev`
5. open http://localhost:4321/admin/index.html#/collections/blog/~
6. press **Add Files**
7. Enter 'New' for Title, Description, Body and Filename.  Choose todays for Publication Date and Updated Date.
8. Press **Save**
9. Observe "New" appears in the Blogs list
10. Click on the "New" blog just created
11. Observe browser navigates to url http://localhost:4321/admin/index.html#/~/blog/New
12. Observe the astro logo and "404: Not found", "Path: /blog/New"
13. Observe error in the terminal 
 
```
12:22:46 [WARN] [router] A `getStaticPaths()` route pattern was matched, but no matching static path was found for requested path `/blog/New`.
Possible dynamic routes being matched: src/pages/blog/[...slug].astro, src/pages/[...slug].astro.
12:22:46 [404] /blog/New 3ms
```

14. In the terminal, press Ctrl-C
15. `pnpm run dev`
16. Refresh the page in the browser
17. Observe the http://localhost:4321/admin/index.html#/~/blog/New page renders without error

### System Info

```shell
System:
    OS: macOS 10.15.7
    CPU: (8) x64 Intel(R) Core(TM) i5-1038NG7 CPU @ 2.00GHz
[...]

PR #7069, opened 2026-06-18
fix(@tinacms/astro): re-run getStaticPaths in dev on Tina content writes

kulesy:
## The problem

Creating a new entry in the admin leaves it unreachable: it shows in the collection list and on index pages, but visiting its own route 404s until you restart `tinacms dev` by hand. Reported in #5611 (and the same symptom blocked the Astro starter's create-content flow).

## Why

Astro caches `getStaticPaths()` in dev and only re-runs it for a change it recognises: a module in the graph, or a file in a watched content-layer collection. When content is sourced through Tina's GraphQL client rather than Astro's content layer, a newly written file is neither, so the route's path list stays stale and the new page is unreachable.

Confirmed by hand: touching the route file (forcing `getStaticPaths` to re-run) flips the new page from 404 to 200 with nothing else changed.

## The fix

A dev-only Vite plugin in the `tina()` integration watches the configured Tina collection paths and, when a content file is added or removed, emits the same `astro:content-changed` HMR signal Astro fires for its own content layer. Astro clears the route cache, `getStaticPaths` re-runs, the new path resolves, and the open tab reloads.

- Content paths come from the generated schema (`tina/__generated__/_schema.json`), the same source `tinacms dev`'s own watcher uses, so content is matched wherever a collection's `path` points, not just under `content/`. A path heuristic (markdown anywhere; structured data under a `content` dir) is used as a fallback when the schema isn't on disk yet.
- Bursts (a multi-file save) are debounced into a single reload.
- Scoped to Astro 6+ via the installed Astro major: only Astro 6's dev runtime listens for `astro:content-changed` and clears the route cache. Astro 5 has no listener, so it is left a genuine no-op. Gating on the env API would not do this, since Astro 5 ships Vite 6 and has that API too.
- Known limitation: content kept in a separate repo (a `localContentPath` resolving outside the Astro project) isn't watched by Vite, so its new entries still need a manual dev restart. `tinacms dev` indexes that content fine; only the Astro route-cache refresh can't reach outside the project root. This is unchanged from before this PR, not a regression.

## Before / after

Same URL (`/blog/Hello-Tina-Demo/`) in `tinacms dev`, created through the admin; only the fix differs.

Before the fix: 404
Figure: Before the fix, a newly created entry 404s on its own page until the dev server is restarted.

After the fix: renders
Figure: After the fix, the same entry resolves immediately, no restart needed.

### DIFF
--- a/packages/@tinacms/astro/src/integration.ts
+++ b/packages/@tinacms/astro/src/integration.ts
@@ -255,6 +255,12 @@ function devContentInvalidationPlugin(
         if (timer) clearTimeout(timer);
         timer = setTimeout(flush, 50);
       };
+      // Vite only watches files inside the Astro project root, so content kept
+      // in a separate repo (a `localContentPath` resolving outside the project)
+      // isn't seen here and won't auto-refresh the route cache in dev. That
+      // content is still indexed by `tinacms dev`'s own watcher; only this
+      // Astro-side refresh can't reach outside the project root, which matches
+      // the behaviour before this fix existed.
       server.watcher.on('add', onChange);
       server.watcher.on('unlink', onChange);
     },
--- a/packages/@tinacms/astro/src/integration.ts
+++ b/packages/@tinacms/astro/src/integration.ts
@@ -48,11 +48,25 @@ export default function tina(
   return {
     name: '@tinacms/astro',
     hooks: {
-      'astro:config:setup': ({ addMiddleware, updateConfig }) => {
+      'astro:config:setup': ({
+        addMiddleware,
+        updateConfig,
+        command,
+        logger,
+      }) => {
         addMiddleware({
           entrypoint: '@tinacms/astro/middleware',
           order: middlewareOrder,
         });
+        const astroMajor = astroMajorVersion();
+        // Live dev content refresh needs Astro 6+ (see
+        // devContentInvalidationPlugin). On Astro 5 nothing handles the signal,
+        // so let editors know new entries won't appear without a manual restart.
+        if (command === 'dev' && astroMajor !== undefined && astroMajor < 6) {
+          logger.warn(
+            'Live content refresh in dev requires Astro 6 or newer. On Astro 5, newly created entries will not appear on the dev site until you restart the dev server. See https://github.com/tinacms/tinacms/issues/5611.'
+          );
+        }
         // Dev: serve the bridge straight from the package rather than writing
         // it into the user's source tree — config-time writes churn
         // public/admin on every `astro dev`/`astro build`, break on read-only
@@ -61,7 +75,7 @@ export default function tina(
           vite: {
             plugins: [
               bridgeDevPlugin(),
-              devContentInvalidationPlugin(astroMajorVersion()),
+              devContentInvalidationPlugin(astroMajor),
               cloudflareImportMetaUrlPlugin(() => isCloudflareAdapter),
             ],
           },
--- a/packages/@tinacms/astro/src/integration.ts
+++ b/packages/@tinacms/astro/src/integration.ts
@@ -61,6 +61,7 @@ export default function tina(
           vite: {
             plugins: [
               bridgeDevPlugin(),
+              devContentInvalidationPlugin(),
               cloudflareImportMetaUrlPlugin(() => isCloudflareAdapter),
             ],
           },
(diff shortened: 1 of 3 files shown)


packages/@tinacms/astro/src/integration.ts
```diff
@@ -120,6 +139,75 @@ function bridgeDevPlugin(): VitePlugin {
...
+  // formats only count inside a `content` directory, so an unrelated
+  // `package.json` / `tsconfig.json` write doesn't trigger a reload.
+  const MARKDOWN = /\.(?:md|mdx|markdown|mdoc)$/i;
+  const STRUCTURED = /\.(?:json|ya?ml|toml)$/i;
+  const IGNORED =
+    /[\\/](?:node_modules|\.git|\.astro|dist|\.vercel|\.netlify|\.cache)[\\/]/;
+  const isContentFile = (file: string): boolean => {
+    if (IGNORED.test(file)) return false;
+    if (MARKDOWN.test(file)) return true;
+    return STRUCTURED.test(file) && /[\\/]content[\\/]/.test(file);
```
JackDevAU, line 240: if people have content out side of a `content` path would this still work?

packages/@tinacms/astro/src/integration.ts
```diff
@@ -120,6 +139,75 @@ function bridgeDevPlugin(): VitePlugin {
...
+ *
+ * See https://github.com/tinacms/tinacms/issues/5611.
+ */
+function devContentInvalidationPlugin(
+  astroMajor: number | undefined
+): VitePlugin {
+  // Markdown-family files are content wherever they live. Structured-data
+  // formats only count inside a `content` directory, so an unrelated
+  // `package.json` / `tsconfig.json` write doesn't trigger a reload.
+  const MARKDOWN = /\.(?:md|mdx|markdown|mdoc)$/i;
```
JackDevAU, line 218: think these are currently global checks. (though only for creation of md(+more files) etc. I think this should be fine tbh but just worth pointing out

JackDevAU (commented):
small comments, but looks good to me

packages/@tinacms/astro/src/integration.ts
kulesy: Good catch, this was a real gap. The old check only treated `json`/`yaml`/`toml` as content when the path contained a `content` segment, so a structured-data collection configured at any other path (say `data/settings`) would have been missed and its new entries kept 404ing. Markdown was matched anywhere, so md/mdx collections were fine regardless.

Fixed in 5f3f3c5: the watcher now reads the collection roots from `tina/__generated__/_schema.json` (the same source `tinacms dev`'s own watcher globs in `watchContentFiles`) and matches content under those exact roots, wherever they live. The `content/` heuristic stays only as a fallback for when the schema isn't on disk yet (e.g. `astro dev` run without `tinacms dev`). Added a test covering a JSON collection at a non-`content` path.

packages/@tinacms/astro/src/integration.ts
kulesy: Right, markdown was matched globally and structured data was scoped to `content/` dirs. Rather than keep inferring content from the filename/location, 5f3f3c5 now derives the watch list from the collection `path`s in the generated schema, so detection matches whatever the schema actually declares (content-format files under each collection root). The path heuristic remains only for the no-schema case.

packages/@tinacms/astro/src/integration.ts
```diff
@@ -120,6 +170,103 @@ function bridgeDevPlugin(): VitePlugin {
...
+    return STRUCTURED.test(file) && /[\\/]content[\\/]/.test(file);
+  };
+
+  return {
+    name: '@tinacms/astro:dev-content-invalidation',
+    apply: 'serve',
+    configureServer(server) {
+      // The `astro:content-changed` signal below is only handled by Astro 6+.
+      // Bail on older (or unresolvable) majors so a no-op stays a no-op.
+      if (astroMajor === undefined || astroMajor < 6) return;
```
JackDevAU, line 249: If people are on Astro v5 -- should we let them know they are missing out on this?

kulesy: I was going to add this to the docs.

kulesy: Probably should add this as a CLI warning too.

kulesy: Done in 6cab57b: on Astro 5 the integration now logs a warning during `astro dev` (via the Astro logger, so it surfaces in the `tinacms dev` terminal) telling them live content refresh needs Astro 6+ and that new entries won't show until they restart. Gated to `command === 'dev'` so builds stay quiet, and it only fires when the installed Astro major actually resolves to < 6. Still planning the docs note for the Astro 5 caveat on top of this.

JackDevAU (approved):
:shipit:

Issue #7061, opened 2026-06-16
Url relative path error


Kacper-Slezak:
### The exact error message

```shell
I have a lot of markdown files with whitespace in it, and after upgrade form 3.7.2 to 3.9.3 i got url relative path error which is only appearing when the url contains %20 char may you fix it ?? 
[Error] GetDocument failed: Unable to fetch, errors: 
	Invalid path: relativePath contains invalid characters
```

### Steps to reproduce

Open tina and try to edit via local mode a file which url contains whitespaces 

### What you expected vs. what actually happened

I wrote it up 

### Your environment

    "@tinacms/cli": "^2.2.2",
    "tinacms": "^3.9.3"
 node js 

### Relevant sections of your schema file

```typescript

```

### Validations

- [x] Follow our [Code of Conduct](https://github.com/tinacms/tinacms/blob/main/CODE_OF_CONDUCT.md)
- [x] Read the [docs](https://tina.io/docs).
- [x] Check that there isn't [already an issue](https://github.com/tinacms/tinacms/issues) that reports the same bug.

PR #7073, opened 2026-06-21
fix(cli): fail build on GraphQL validation errors

### LINKED ISSUE #7052: tinacms build exits 0 when @graphql-codegen/core reports validation errors
### The exact error message

When the codegen pipeline hits validation errors during `tinacms build`, output like this prints to stderr:

```
Error: GraphQL Document Validation failed with 19 errors;
  Error 0: Fields "heading" conflict because they return conflicting
    types "String!" and "String". Use different aliases on the fields
    to fetch both if this was intentional.
    at tina/__generated__/frags.gql:79:7
    at tina/__generated__/frags.gql:497:7
  ... (more conflicts) ...
```

…and the process exits with code **0**.

### Steps to reproduce

1. In a project on the modern `tina/` layout, create a schema where the same field name appears across multiple templates in a section union with inconsistent nullability — e.g. one section template has `{ type: 'string', name: 'heading', required: true }` (codegen → `String!`) and another has `{ type: 'string', name: 'heading' }` (codegen → `String`).
2. Run `npx tinacms build`.
3. Inspect the exit code: `echo $?`.

**Expected:** non-zero exit code so CI fails and `&&`-chained commands (`tinacms build && next build`) short-circuit. Ideally also a one-line summary at the end of the output like `✗ Schema build failed: 19 codegen errors` so the failure is visible even when devs aren't reading the full log.

**Actual:** exit code 0. CI considers the build successful. Chained commands proceed. The generated artifacts under `tina/__generated__/` are written, but with the codegen-error documents potentially missing fields or types. Devs reasonably conclude the build worked and push; the downstream TinaCloud indexer then fails with a content-seeder error like `Block template "X" is not defined for field "sections"`, with no signal connecting the indexer failure back to the codegen validation that caused it.

### Your environment

- `@tinacms/cli` v2.4.x (the codegen path in `packages/@tinacms/cli/src/next/codegen/`)
- Any framework wrapping `tinacms build` — issue is independent of host framework

### Suggested fix

Two pieces, both small:

1. **Exit code:** capture validation-error count from `@graphql-codegen/core` inside `Codegen.execute()` (or in [`build-command/index.ts`](packages/@tinacms/cli/src/next/commands/build-command/index.ts)), set an internal flag, and exit non-zero at the end of `build` if any were seen. The catch handler at [`build-command/index.ts:97-108`](packages/@tinacms/cli/src/next/commands/build-command/index.ts#L97) is the natural place to surface this through the existing `ERR_BUILD_FAILED` errorCode.
2. **Summary line (optional, nice-to-have):** print one terminating line via the existing `summary()` helper used today for the success block — e.g. `✗ Schema build failed: 19 codegen errors`. Visible even when the developer is scrolled away from the per-error output.

### Why this matters

The exit-0 behaviour silently masks downstream failures and turns a 30-second-loud-fix into a multi-hour bisect on the indexer side. Reported via support after a customer burned ~6 hours bisecting field shapes/names on the indexer because their local `tinacms build` looked clean. Adjacent CLI defect (lockfile asymmetry between `dev` and `build`) is being fixed in #7051; this is the orthogonal half.

michaelsam94:
## What problem does this address?

Fixes #7052.

`tinacms build` could continue successfully even when generated GraphQL documents failed schema validation during the codegen step.

## What is changed?

- Validate loaded query and fragment documents before invoking `@graphql-codegen/core`.
- Throw a build-stopping error with the validation error count when invalid documents are found.
- Add a focused regression test for invalid loaded documents.
- Add a patch changeset for `@tinacms/cli`.

## Testing

- `git diff --check`
- Attempted `npx --yes pnpm@11 --filter @tinacms/cli test -- src/next/codegen/codegen/index.test.ts --runInBand`, but local dependency installation timed out while fetching unrelated workspace packages. Local shell also reports unsupported Node `v26.0.0`; repo expects Node `22.x || 24.x`.

### DIFF
--- a/.changeset/quiet-codegen-builds.md
+++ b/.changeset/quiet-codegen-builds.md
@@ -1,0 +1,5 @@
+---
+"@tinacms/cli": patch
+---
+
+Fail CLI builds when generated GraphQL documents do not pass schema validation.
--- a/packages/@tinacms/cli/src/next/codegen/codegen/index.test.ts
+++ b/packages/@tinacms/cli/src/next/codegen/codegen/index.test.ts
@@ -1,0 +1,60 @@
+import { codegen } from '@graphql-codegen/core';
+import { loadDocuments } from '@graphql-tools/load';
+import { buildSchema, parse } from 'graphql';
+import { generateTypes } from './index';
+
+jest.mock('@graphql-codegen/core', () => ({
+  codegen: jest.fn(),
+}));
+
+jest.mock('@graphql-tools/load', () => ({
+  loadDocuments: jest.fn(),
+}));
+
+jest.mock('@graphql-tools/graphql-file-loader', () => ({
+  GraphQLFileLoader: jest.fn(),
+}));
+
+const mockedCodegen = codegen as jest.MockedFunction<typeof codegen>;
+const mockedLoadDocuments = loadDocuments as jest.MockedFunction<
+  typeof loadDocuments
+>;
+
+describe('generateTypes', () => {
+  beforeEach(() => {
+    mockedCodegen.mockReset();
+    mockedLoadDocuments.mockReset();
+  });
+
+  it('throws when loaded documents fail schema validation', async () => {
+    const schema = buildSchema(`
+      type Query {
+        post: Post
+      }
+
+      type Post {
+        title: String!
+      }
+    `);
+
+    mockedLoadDocuments
+      .mockResolvedValueOnce([
+        {
+          document: parse(`
+            query InvalidQuery {
+              post {
+                missingField
+              }
+            }
+          `),
+          location: 'queries.gql',
+        },
+      ])
+      .mockResolvedValueOnce([]);
+
+    await expect(
+      generateTypes(schema, 'queries.gql', 'frags.gql', 'http://localhost:4001')
+    ).rejects.toThrow('GraphQL Document Validation failed with 1 errors');
+    expect(mockedCodegen).not.toHaveBeenCalled();
+  });
+});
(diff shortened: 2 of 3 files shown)


kulesy:
Thanks for picking this up, but I think this should be closed: the build already exits non-zero on this scenario, so the PR does not change the reported behavior.

I ran the published `@tinacms/[email protected]` (the version in #7052) and current `main` against the issue's exact schema. Both throw `Fields "heading" conflict ...` and exit `1` already, in [`BuildCommand.catch`](https://github.com/tinacms/tinacms/blob/e7a19239cb0b2d91075922d74f8d372d5f40d8f9/packages/@tinacms/cli/src/next/commands/build-command/index.ts#L97-L107), which calls `process.exit(1)`. `@graphql-codegen/core` validates and throws on its own.

The one thing this PR does change: the new [`validate(...)`](https://github.com/tinacms/tinacms/pull/7073/files#diff-415a78aa449217343e0fcbee618a95b5d33e05be8214703bbc43e3fab560a6efR29-R44) uses the full rule set, while codegen deliberately skips `NoUnusedFragments`, `NoUnusedVariables`, `KnownDirectives`. So it would start failing previously-green builds on those three rules, as a `patch`, while adding nothing for the hard-rule case the issue describes. Happy to be wrong if you have a repro where `main` exits 0.

michaelsam94:
Thanks for checking and for the detailed repro. If the issue is not reproducible on your side and this PR would broaden validation behavior unintentionally, I’m happy to close it. I’m ready to pick up another issue instead if there’s one you’d prefer me to work on.

kulesy:
Hi @michaelsam94, yes, close it please.

It'd be great if you could take a look at this issue!
https://github.com/tinacms/tinacms/issues/7061

PR #7074, opened 2026-06-22
add dependency review

joshbermanssw:
As per https://github.com/tinacms/tinacloud/issues/3575

- [ ] Adds dependency review GitHub workflow


.github/workflows/dependency-review.yml
```diff
@@ -0,0 +1,18 @@
...
+permissions:
+  contents: read
+  pull-requests: write
+
+jobs:
+  dependency-review:
+    runs-on: ubuntu-latest
+    steps:
+      - name: 'Checkout Repository'
+        uses: actions/checkout@v6
```
18-th: Could we SHA pin the versions, please?

joshbermanssw: Done — SHA-pinned both actions: `actions/checkout@9c091bb...` (v7.0.0, matching the rest of the workflows) and `actions/dependency-review-action@a1d282b...` (v5.0.0). Pushed in ab602dd06.

18-th (approved):
LGTM

PR #7080, opened 2026-06-22
📝 Document the dependency cooldown policy in SECURITY.md

joshbermanssw:
## TL;DR

Dependabot already enforces a 3-day cooldown on update PRs, but a manual `pnpm add <pkg>@latest` slips straight past it. This adds a Dependency Updates and Supply-Chain section to `SECURITY.md` stating the rule: no manual install of a version younger than 3 days without a security sign-off.

## Why

Compromised or malicious package releases are usually caught and yanked within the first couple of days. The Dependabot cooldown (`.github/dependabot.yml`) encodes that for automated updates; this writes down the human-facing half so the manual path has the same guardrail.

## Links

- Companion config: `.github/dependabot.yml` (`cooldown: default-days: 3`)

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -36,3 +36,16 @@ To help us triage and resolve the issue quickly, please include:
 - We will notify you when a fix has been released.
 
 Thank you for helping keep TinaCMS safe!
+
+## Dependency Updates and Supply-Chain Policy
+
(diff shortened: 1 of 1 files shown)


18-th (approved):
LGTM

PR #7088, opened 2026-06-22
Fix media mediaroot key scoping

Aibono1225:
## Summary

The first-party media adapters (s3, dos, azure, cloudinary) passed a caller-controlled object key straight to the storage SDK on their upload/delete paths without checking it falls under mediaRoot (CWE-639). mediaRoot was only enforced on listing, so any logged-in editor had write/delete reach over every key the IAM credential could touch — not just mediaRoot.

This PR adds a centralised resolveKey(mediaRoot, rawKey) helper and wires it into every write/delete entry point so the key reaching the SDK is always normalised and bounded.

## What this fixes

- S3 / DOS — upload (?key=) and delete keys are resolved against mediaRoot; out-of-root keys are conte, .. incl. percent-encoded, NUL, empty) return 400.
- S3 — expiresIn is now clamped (default/max 3600s); it was caller-controlled, bounded only by SigV4
- Azure / Cloudinary — no mediaRoot config today, so they pass an empty root, which still rejects tr(including the Cloudinary upload filename).

The helper is duplicated byte-for-byte across the four packages (avoids a cross-package publish dependency); a shared suite asserts the copies stay identical.

## Not in this PR (follow-ups)

- SEC-1 — constrain upload Content-Type (stored-XSS via uploaded HTML/SVG; still open on S3, payload is only confined to mediaRoot).
- SEC-2 — real mediaRoot boundary for Azure/Cloudinary (new config; Cloudinary switch needs a live env to verify).
- SEC-4 — README + scope IAM policy to mediaRoot/* (defence-in-depth).
- SEC-5 — Cloudinary listMedia interpolates directory into a search-expression DSL (separate injection issue).

## Behaviour & testing

- S3 with mediaRoot set: uploads now land under mediaRoot/ (also fixes the prior upload/list mismatch). No client/API/config changes; deployments without mediaRoot only gain traversal/absolute/empty rejection. Empty ?key= stilllists.
- tests/media-key.test.ts — 110 cases (adversarial vectors + anti-drift across all four copies), green on Node 22. Handler-level scoping verified separately.

### DIFF
--- a/.changeset/rude-ears-remember.md
+++ b/.changeset/rude-ears-remember.md
@@ -1,0 +1,22 @@
+---
+"tinacms-gitprovider-github": patch
+"@tinacms/vercel-previews": patch
+"next-tinacms-cloudinary": patch
+"@tinacms/schema-tools": patch
+"@tinacms/datalayer": patch
+"next-tinacms-azure": patch
+"@tinacms/graphql": patch
+"next-tinacms-dos": patch
+"@tinacms/search": patch
+"create-tina-app": patch
+"next-tinacms-s3": patch
+"@tinacms/astro": patch
+"tinacms-authjs": patch
+"tinacms-clerk": patch
+"@tinacms/app": patch
+"@tinacms/cli": patch
+"@tinacms/mdx": patch
+"tinacms": patch
+---
+
+Fix media upload/delete paths to prevent access to storage keys outside mediaRoot.
--- a/package.json
+++ b/package.json
@@ -19,7 +19,7 @@
 		"update-tina-lock": "pnpm -r --filter='{./**}' --workspace-concurrency=1 run --if-present update-tina-lock",
 		"test": "turbo run test --filter=\"./packages/**\"",
 		"lint:packages": "pnpm --filter './packages/**' --no-bail exec pnpm dlx publint --strict",
-		"test:packages": "vitest run tests/build-verification.test.ts --reporter=verbose && pnpm run lint:packages",
+		"test:packages": "vitest run --reporter=verbose && pnpm run lint:packages",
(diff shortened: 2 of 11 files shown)


kulesy (dismissed):
Approving. Reviewed the full diff and traced the access-control logic by hand (did not run the suite locally; author reports the 110-case media-key.test.ts green on Node 22).

Verified:
- resolveKey is sound: posix.normalize collapses `..` before the boundary check, the already-prefixed-vs-relative branch is idempotent (so the upload/list/delete round-trip stays consistent with mediaRoot set), and the trailing `startsWith(root + '/')` backstop holds. A `media-evil/` prefix-lookalike is correctly not treated as inside `media`.
- GET ?key= now passes { decode: false }, fixing the latent double-decode and matching the delete path. Confirmed against the client: raw `?key=` template on upload + encodeURIComponent on delete, each decoded once by Next.
- expiresIn clamp handles NaN / negative / over-limit.
- List traversal is closed on s3/dos/azure via resolveDirectory; cloudinary is validate-only and documents it as SEC-5.

Pushed one fixup (6a15d3c): the new cloudinary listMedia 400 returned { e: e.message } where every other handler returns { message: ... }, so a client reading .message got undefined.

Non-blocking follow-ups (changelog / separate issues, not for this PR):
- Release note: with mediaRoot set, uploads now correctly land under mediaRoot/ (fixes the prior upload/list mismatch); objects uploaded before this fix remain at bucket root and stay invisible. expiresIn > 3600 is now silently clamped.
- Pre-existing { e: message } at cloudinary handlers.ts:219 is the same anti-pattern, intentionally left out of scope.
- Confirm the widened `test:packages` glob doesn't sweep in tests that need creds/env.

kulesy (approved):

PR #7094, opened 2026-06-23
Fix encoded admin document paths

### LINKED ISSUE #7061: Url relative path error

michaelsam94:
## Summary

1. Decode URL-encoded admin route filenames before constructing document `relativePath` values for edit and duplicate flows.
2. Add a focused regression test for encoded route params and malformed escape fallback.
3. Add a patch changeset for `tinacms`.

Fixes #7061

## Testing

1. `pnpm exec biome check packages/tinacms/src/admin/pages/utils.tsx packages/tinacms/src/admin/pages/utils.test.ts packages/tinacms/src/admin/pages/CollectionUpdatePage.tsx packages/tinacms/src/admin/pages/CollectionDuplicatePage.tsx`
2. `pnpm --dir packages/tinacms exec vitest run src/admin/pages/utils.test.ts --reporter=verbose`

## Notes

`pnpm --filter tinacms test -- --reporter=verbose` currently fails before running several unrelated suites because workspace package entrypoints such as `@tinacms/schema-tools`, `@tinacms/mdx`, and `@tinacms/bridge/metadata` are not built/resolvable in this checkout.


joshbermanssw (commented):
Thanks for digging into this. The client-side decode is a reasonable and safe change, but I don't think this PR actually fixes #7061 on its own; the user would still hit the same error.

## Why it still fails

The error in #7061 comes from the server-side resolver, not the admin. When the edit form loads, `resolveRetrievedDocument` -> `getValidatedPath` -> `validateRelativePath` runs this allowlist regex on the path (`packages/@tinacms/graphql/src/resolver/index.ts:759`):

```js
// packages/@tinacms/schema-tools/src/util/relativePath.ts
RELATIVE_PATH_REGEX = /^[a-zA-Z0-9\-_./]+$/   // no space in the allowlist
```

A space isn't in that allowlist, so decoding `%20` into a literal space just swaps one rejected character for another:

```
"posts/hello%20world.md"  -> REJECT   (the '%')
"posts/hello world.md"    -> REJECT   (the space)   <- what this PR produces
"posts/helloworld.md"     -> PASS
```

So the resolver still throws `Invalid path: relativePath contains invalid characters`. This PR only touches `packages/tinacms/**` and never changes the regex, which is the thing rejecting the path.

## What's actually needed

This looks like a two-part fix, and the PR has the client half only:

1. **Server (missing):** widen `RELATIVE_PATH_REGEX` in `@tinacms/schema-tools` to allow spaces. Path-traversal is guarded separately (`..`, null bytes, `\`, and absolute paths are handled in `validatePath`/`sanitizePath`), so permitting a space doesn't reopen what #6664 closed. `validateRelativePath` and the inline UI message share that constant, so one change covers both.
2. **Client (this PR):** decode `%20` so the looked-up key matches the datalayer's indexed path `hello world.md`. Needed alongside the regex change, just not sufficient alone.

## Smaller notes

- The added test only asserts that `decodeRouteParam` decodes a string. It passes, but it doesn't exercise #7061. A test that a space-containing `relativePath` survives `validateRelativePath` would be the meaningful guard, and it would currently fail.
- react-router 6.30 doesn't decode the splat (`*`) param, which is why `%20` reaches the server, so the decode isn't a double-decode. That part is fine.

Happy to be wrong if I've missed a code path where the resolver validation is skipped in local mode, but from tracing `resolveRetrievedDocument` it runs on every document load. Could we add the regex change plus a resolver-level test proving `posts/hello world.md` now validates?

PR #7099, opened 2026-06-25
feat(@tinacms/tinacms): v4 field-plugin system + String field

### LINKED ISSUE #6915: Field Type Implementation - String
## Pain

The string field is one of the simplest field types, but it still needs to prove the new field plugin contract works for everyday scalar content. If it remains wired through legacy field paths, the most common text-editing case will not validate the new ingest, digest, validation and rendering shape.

## Approach

Implement `StringField` as a field plugin against the field contract. Reuse existing string field behaviour where it holds up, but wrap serialization, validation and rendering behind the new plugin interface.

## Expected outcomes

- Basic text fields work through the v4 field plugin system.
- The simplest scalar field becomes a reference implementation for other small field plugins.
- String field behaviour is isolated and testable without touching unrelated field code.

## Acceptance Criteria

- `StringField` exists as a field plugin that conforms to the field contract.
- Ingest and digest behaviour preserves the existing string field schema and stored value behaviour.
- Rendering supports the expected editor input for string values.
- Metadata wrapping works for custom string field configuration.
- Tests cover rendering, value updates, ingest, digest and validation.

JackDevAU:
Closes #6915

First slice of TinaCMS v4 — under `packages/v4/@tinacms/tinacms` (private alpha, src-only, nothing published; v3 untouched).

## What's added

- Field-plugin contract: `definePlugin` / `defineClientPlugin`, a field registry, and `t.string` schema authoring.
- The **String field** as the reference implementation — schema + zod validation + component + manifest, co-located in one folder.
- Generic `core/` — it knows no field type, so a new field plugin needs zero core changes.
- Form state on react-hook-form: ingest (load) → form → digest (save), plus the field hooks (`useFieldValue` / `useFieldErrors` / `useFieldAddress` / `useFieldActivation`).

## How to test

```bash
pnpm install
pnpm --filter @tinacms/tinacms test     # 10 pass
pnpm --filter @tinacms/tinacms types
```

## Notes for reviewers

- `packages/v4/**` + `pnpm-lock.yaml` — v3 packages are untouched. The lockfile change is only the new v4 deps (`react-hook-form`, `@hookform/resolvers`, React 19 dev deps); it's the monorepo dev/CI lockfile, never shipped to consumers.
- Form state uses react-hook-form, not the Zustand store from the v4 docs yet — `TODO(zustand)` markers show where it plugs in.
- `t` is the built-in default until `defineConfig` derives it from the plugin set. A few `TODO`s flag deliberately-deferred work (server segments, shadcn UI, parse/serialize transforms).


18-th (approved):
LGTM

PR #7100, opened 2026-06-25
feat: add deno support

brookjeynes-ssw:
```
This PR adds Deno support to the installer. 

It also resolves an esbuild issue that occurred when attempting to dev/build an
astro + deno project. Additional information is as follows:

    Esbuild 2 and 3 bundled `react` and `react-dom` into `config.build.mjs`.
    When the bundled react-dom called `require('react')` internally, the
    `createRequire` polyfill resolved a different copy of React from the
    filesystem. This lead to an undefined `ReactCurrentDispatcher` issue. react
    and react-dom are now resolved via the projects node modules on import for
    node-targeted builds.

EDIT: Removing the react pre-fill is probably not the right solution and needs more thought
```

### DIFF
--- a/packages/@tinacms/cli/src/cmds/init/apply.ts
+++ b/packages/@tinacms/cli/src/cmds/init/apply.ts
@@ -398,6 +398,7 @@ const addDependencies = async (
     npm: `npm install ${deps.join(' ')}`,
     yarn: `yarn add ${deps.join(' ')}`,
     bun: `bun add ${deps.join(' ')}`,
+    deno: `deno install ${deps.join(' ')}`,
   };
 
   if (packageManagers[packageManager] && deps.length > 0) {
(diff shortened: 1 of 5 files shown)


packages/@tinacms/cli/src/cmds/init/apply.ts
```diff
@@ -643,6 +645,7 @@ const other = ({ packageManager }: { packageManager: string }) => {
     npm: `npx`, // npx is the way to run executables that aren't in your "scripts"
     yarn: `yarn`,
     bun: `bun run`,
+    deno: `deno run`,
```
18-th, line 648:
```suggestion
    deno: `deno run -A npm:`,
```
`deno run tinacms` here will try to resolve tinacms as a file, if I'm not mistaken

Should this be something like above?

PR #7101, opened 2026-06-26
✨ Add source-agnostic `_sys.lastUpdated` with server-side collection sort

joshbermanssw:
## TL;DR

Adds a "Last updated" column to the collection list, sortable by recency or name via clickable headers or the dropdown. Sorting is server-side across the whole collection, sourced from a pluggable `Bridge.lastUpdated()`, so it works on any backend (self-hosted included).

Screenshot 2026-06-29 at 12 31 16 pm

**Figure: UI using tagged version on tina.io**

**Testing**

Tagged and built into a preview deployment of Tina.io : https://tina-io-git-jb-test-last-updated-tagged-version-tinacms.vercel.app/admin/index.html#/collections/page/~

## Notes

- Value is materialized into the index at index time (reserved `__lastUpdated`, stripped from content files); `_sys.lastUpdated` reads it.
- Defaults: git commit time (isomorphic) / file mtime (filesystem). Requires a reindex to backfill.

## Links

- Issue: https://github.com/tinacms/tinacloud/issues/3771
- TinaCloud timestamp source (merged): https://github.com/tinacms/tinacloud/pull/3779

### DIFF
--- a/.changeset/sys-last-updated.md
+++ b/.changeset/sys-last-updated.md
@@ -4,3 +4,5 @@
 ---
 
 Add a `lastUpdated` field to documents' `_sys` (`SystemInfo`) and a `lastUpdated` collection sort, both sourced from an optional `Bridge.lastUpdated(filepath)` capability that integrators can plug in (defaults: git commit time for the isomorphic bridge, file mtime for the filesystem bridge). The value is materialized into the index at index time, so `_sys.lastUpdated` and server-side `documents(sort: "lastUpdated")` work across any store/backend (including self-hosted), not just TinaCloud. The admin collection-listing page shows a "Last Updated" column and offers "Last Updated" in its sort control. Requires a reindex to populate the new sort index.
+
+Note: the value's meaning is bridge-defined and not identical across bridges — the isomorphic (git) bridge reports the file's last commit time (which is stale for uncommitted working-tree edits), while the filesystem bridge reports file mtime. Documents the bridge can't time are ordered last and shown empty.
--- a/.changeset/sys-last-updated.md
+++ b/.changeset/sys-last-updated.md
@@ -1,0 +1,6 @@
+---
+"@tinacms/graphql": minor
+"tinacms": minor
+---
+
+Add a `lastUpdated` field to documents' `_sys` (`SystemInfo`), sourced from the content store's last-write time when the store can report it. The admin collection-listing page now shows a "Last Updated" column with relative times and lets you sort the loaded page by filename or last updated via clickable column headers.
--- a/.changeset/sys-last-updated.md
+++ b/.changeset/sys-last-updated.md
@@ -3,4 +3,4 @@
 "tinacms": minor
 ---
 
-Add a `lastUpdated` field to documents' `_sys` (`SystemInfo`), sourced from the content store's last-write time when the store can report it. The admin collection-listing page now shows a "Last Updated" column with relative times and lets you sort the loaded page by filename or last updated via clickable column headers.
+Add a `lastUpdated` field to documents' `_sys` (`SystemInfo`) and a `lastUpdated` collection sort, both sourced from an optional `Bridge.lastUpdated(filepath)` capability that integrators can plug in (defaults: git commit time for the isomorphic bridge, file mtime for the filesystem bridge). The value is materialized into the index at index time, so `_sys.lastUpdated` and server-side `documents(sort: "lastUpdated")` work across any store/backend (including self-hosted), not just TinaCloud. The admin collection-listing page shows a "Last Updated" column and offers "Last Updated" in its sort control. Requires a reindex to populate the new sort index.
--- a/packages/@tinacms/graphql/src/database/bridge/isomorphic.ts
+++ b/packages/@tinacms/graphql/src/database/bridge/isomorphic.ts
@@ -549,6 +549,25 @@ export class IsomorphicBridge implements Bridge {
     return Buffer.from(blob).toString('utf8');
   }
 
+  public async lastUpdated(filepath: string) {
+    assertWithinBase(filepath, this.relativePath);
+    const ref = await this.getRef();
+    try {
+      const commits = await git.log({
+        ...this.isomorphicConfig,
+        ref,
+        filepath: this.qualifyPath(filepath),
+        depth: 1,
+        force: true,
+        cache: this.cache,
+      });
+      const timestamp = commits[0]?.commit?.committer?.timestamp;
+      return typeof timestamp === 'number' ? timestamp * 1000 : null;
+    } catch {
+      return null;
+    }
+  }
+
   public async put(filepath: string, data: string) {
     assertWithinBase(filepath, this.relativePath);
     const ref = await this.getRef();
--- a/packages/@tinacms/graphql/src/database/datalayer.ts
+++ b/packages/@tinacms/graphql/src/database/datalayer.ts
@@ -70,6 +70,7 @@ export const DEFAULT_COLLECTION_SORT_KEY = '__filepath__';
 export const REFS_COLLECTIONS_SORT_KEY = '__refs__';
 export const LAST_UPDATED_SORT_KEY = 'lastUpdated';
 export const LAST_UPDATED_FIELD = '__lastUpdated';
+export const LAST_UPDATED_SENTINEL = new Date(0).toISOString();
 export const REFS_REFERENCE_FIELD = '__tina_ref__';
 export const REFS_PATH_FIELD = '__tina_ref_path__';
 export const DEFAULT_NUMERIC_LPAD = 4;
--- a/packages/@tinacms/graphql/src/database/datalayer.ts
+++ b/packages/@tinacms/graphql/src/database/datalayer.ts
@@ -2,8 +2,10 @@
 
 */
 
-import { JSONPath } from 'jsonpath-plus';
+import path from 'path';
+import type { Collection } from '@tinacms/schema-tools';
 import sha from 'js-sha1';
+import { JSONPath } from 'jsonpath-plus';
 import {
   ARRAY_ITEM_VALUE_SEPARATOR,
   BatchOp,
@@ -12,8 +14,6 @@ import {
   Level,
   SUBLEVEL_OPTIONS,
 } from './level';
-import type { Collection } from '@tinacms/schema-tools';
-import path from 'path';
 import { normalizePath } from './util';
 
 export enum OP {
@@ -68,6 +68,8 @@ type StringEscaper = <T extends string | string[]>(input: T) => T;
 
 export const DEFAULT_COLLECTION_SORT_KEY = '__filepath__';
 export const REFS_COLLECTIONS_SORT_KEY = '__refs__';
+export const LAST_UPDATED_SORT_KEY = 'lastUpdated';
+export const LAST_UPDATED_FIELD = '__lastUpdated';
 export const REFS_REFERENCE_FIELD = '__tina_ref__';
 export const REFS_PATH_FIELD = '__tina_ref_path__';
 export const DEFAULT_NUMERIC_LPAD = 4;
--- a/packages/@tinacms/graphql/src/database/index.ts
+++ b/packages/@tinacms/graphql/src/database/index.ts
@@ -29,6 +29,8 @@ import {
   FOLDER_ROOT,
   FolderTreeBuilder,
   type IndexDefinition,
+  LAST_UPDATED_FIELD,
+  LAST_UPDATED_SORT_KEY,
   REFS_COLLECTIONS_SORT_KEY,
   REFS_PATH_FIELD,
   REFS_REFERENCE_FIELD,
@@ -944,6 +946,15 @@ export class Database {
                   },
                 ],
               },
+              [LAST_UPDATED_SORT_KEY]: {
+                fields: [
+                  {
+                    name: LAST_UPDATED_FIELD,
+                    type: 'datetime',
+                    list: false,
+                  },
+                ],
+              },
             };
 
             let fields: TinaField<true>[] = [];
@@ -1694,6 +1705,11 @@ const _indexContent = async ({
       // if we aren't able to load the data, we can't index it
       if (!aliasedData) {
         return;
+      }
+
+      const lastUpdatedMs = await database.bridge.lastUpdated?.(filepath);
+      if (typeof lastUpdatedMs === 'number') {
+        aliasedData[LAST_UPDATED_FIELD] = new Date(lastUpdatedMs).toISOString();
       }
 
       if (passwordFields?.length) {
--- a/packages/@tinacms/graphql/src/database/index.ts
+++ b/packages/@tinacms/graphql/src/database/index.ts
@@ -30,6 +30,7 @@ import {
   FolderTreeBuilder,
   type IndexDefinition,
   LAST_UPDATED_FIELD,
+  LAST_UPDATED_SENTINEL,
   LAST_UPDATED_SORT_KEY,
   REFS_COLLECTIONS_SORT_KEY,
   REFS_PATH_FIELD,
@@ -432,9 +433,10 @@ export class Database {
     }
 
     const lastUpdatedMs = await this.bridge?.lastUpdated?.(normalizedPath);
-    if (typeof lastUpdatedMs === 'number') {
-      dataFields[LAST_UPDATED_FIELD] = new Date(lastUpdatedMs).toISOString();
-    }
+    dataFields[LAST_UPDATED_FIELD] =
+      typeof lastUpdatedMs === 'number'
+        ? new Date(lastUpdatedMs).toISOString()
+        : LAST_UPDATED_SENTINEL;
 
     let level = this.contentLevel;
     if (collection?.isDetached) {
@@ -613,11 +615,10 @@ export class Database {
         }
 
         const lastUpdatedMs = await this.bridge?.lastUpdated?.(normalizedPath);
-        if (typeof lastUpdatedMs === 'number') {
-          dataFields[LAST_UPDATED_FIELD] = new Date(
-            lastUpdatedMs
-          ).toISOString();
-        }
+        dataFields[LAST_UPDATED_FIELD] =
+          typeof lastUpdatedMs === 'number'
+            ? new Date(lastUpdatedMs).toISOString()
+            : LAST_UPDATED_SENTINEL;
 
         const folderTreeBuilder = new FolderTreeBuilder();
         const folderKey = folderTreeBuilder.update(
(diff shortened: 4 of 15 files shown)


packages/tinacms/src/admin/pages/CollectionListPage.tsx
```diff
@@ -370,14 +417,20 @@ const CollectionListPage = () => {
...
-                ).filter((x) =>
-                  // only allow sortable fields
-                  ['string', 'number', 'datetime', 'boolean'].includes(x.type)
-                );
+                  ).filter((x) =>
+                    // only allow sortable fields
+                    ['string', 'number', 'datetime', 'boolean'].includes(x.type)
+                  ),
+                  {
+                    name: 'lastUpdated',
```
joshbermanssw: **Blocker: sort by Last Updated is a no-op in the admin.** This synthetic `lastUpdated` field only exists in this local `fields` array; it is never part of the real schema collection. `GetCollection.tsx:76` validates the chosen sort name with `isValidSortKey(name, collectionExtra)`, which checks `collectionExtra.fields`/`templates`. Since `lastUpdated` is not a real schema field, `isValidSortKey` returns `false`, `validSortKey` becomes `undefined`, and `fetchCollection` falls back to the default filepath sort. So clicking the Last Updated header or picking it in the dropdown flips the chevron and refetches, but returns the same default order. The DB-layer tests pass `sort: 'lastUpdated'` directly and bypass this gate, so they do not catch it. Fix: allow `lastUpdated` (and any other system sort keys) through `isValidSortKey`.

joshbermanssw (commented):
Automated review of the Last Updated feature. The backend index/sort mechanism is sound and well tested, but the headline sort does not currently reach the server from the admin UI, and the indexed value degrades on the first edit. Inline comments below, grouped by severity.

**Blockers**
1. Sort by Last Updated never reaches the server; the admin validation gate rejects the synthetic sort key.
2. Saving a document blanks its Last Updated and drops it from the sort until a full reindex.
3. Documents with a null timestamp vanish from the sort instead of ordering last.

**Secondary:** hardcoded title sort name, a duplicate sort option when a collection already has a field named `lastUpdated`, an always-on non-required banner, a colSpan regression in folder-only views, and a dead `TimestampedStore` code path.

Happy to draft fixes for the three blockers.

packages/tinacms/src/admin/pages/CollectionListPage.tsx
```diff
@@ -370,14 +417,20 @@ const CollectionListPage = () => {
...
+                    label: 'Last Updated',
```
joshbermanssw: **Unconditional append duplicates a user field named `lastUpdated`.** If a collection already defines a `datetime`/`string` field literally named `lastUpdated`, it passes the type filter above and is kept, then this synthetic entry is appended on top. The Sort `<Select>` then renders two options with identical `value` JSON, and `select.tsx` keys options by `option.value`, so React logs a duplicate-key warning and the dropdown shows the entry twice. Guard with `fields.some((f) => f.name === 'lastUpdated')` before appending.

packages/tinacms/src/admin/pages/CollectionListPage.tsx
```diff
@@ -370,14 +417,20 @@ const CollectionListPage = () => {
...
+                    type: 'datetime',
```
joshbermanssw: **Missing `required`, so the non-required-sort banner always fires for Last Updated.** The warning around line 871 renders when `sortField && !sortField.required`. Because this object omits `required`, selecting Last Updated always shows 'Sorting on a non-required field. Some documents may be excluded'. After a reindex the value is materialized for every doc, so the message is misleading in the normal case. Mark this `required: true` or special-case it. (It is accidentally accurate for the null-timestamp case; see the index-definition comment.)

packages/tinacms/src/admin/pages/CollectionListPage.tsx
```diff
@@ -865,9 +904,12 @@ const CollectionListPage = () => {
...
                                             colSpan={hasAnyTitles ? 1 : 2}
                                           >
-                                            {hasAnyTitles
-                                              ? 'Title'
-                                              : 'Filename'}
+                                            {sortableHeader(
+                                              hasAnyTitles
+                                                ? 'Title'
+                                                : 'Filename',
+                                              hasAnyTitles ? 'title' : ''
```
joshbermanssw: **The Title/Filename header hardcodes the sort name.** `'title'` only matches collections whose title field is literally named `title`; for any other name this header sorts by a field that does not exist (and, via the `isValidSortKey` gate noted on the synthetic field above, silently no-ops). The no-titles branch passes `''`, which only works by falling through to the default sort. Use the collection's actual title field name (e.g. via `getIsTitleFieldName`) instead of the literal `'title'`.

packages/tinacms/src/admin/pages/CollectionListPage.tsx
```diff
@@ -962,7 +1010,7 @@ const CollectionListPage = () => {
                                                 </td>
                                                 <td
                                                   className='px-3 py-3'
-                                                  colSpan={4}
+                                                  colSpan={5}
```
joshbermanssw, line 1033: **colSpan regression for folder-only listings.** Folder rows now span 6 columns (this `colSpan={5}` plus the name cell) and the breadcrumb row was bumped to `colSpan={6}`, but the folders-only header (the `!hasAnyDocuments && hasAnyFolders` thead, around line 947) still declares `Name` + `Path colSpan={4}` = 5 columns. In a folder that contains only subfolders, the Path header no longer lines up with the rows. Bump that header's Path `th` to `colSpan={5}`.

packages/@tinacms/graphql/src/database/index.ts
```diff
@@ -1672,6 +1707,11 @@ const _indexContent = async ({
         return;
       }
 
+      const lastUpdatedMs = await database.bridge.lastUpdated?.(filepath);
```
joshbermanssw, line 1723: **Blocker: `__lastUpdated` is materialized only here, not on the interactive write paths.** `database.put` (around lines 619-700) and `addPendingDocument` build their `putOps` from `dataFields`, which never contains `__lastUpdated`, while `delOps` are built from `existingItem` (which does carry it after a reindex). So every CMS save deletes the old `lastUpdated` index entry, adds none back (`makeKeyForField` returns null for the missing field), and overwrites the stored content record without the value. Result: editing any document blanks its Last Updated column (renders '—') and drops it from the Last Updated sort until the next full reindex, even though the file's mtime/commit time just changed. Recompute `bridge.lastUpdated(filepath)` and set `dataFields[LAST_UPDATED_FIELD]` in both write paths.

packages/@tinacms/graphql/src/database/index.ts
```diff
@@ -920,6 +946,15 @@ export class Database {
                   },
                 ],
               },
+              [LAST_UPDATED_SORT_KEY]: {
```
joshbermanssw, line 962: **Blocker: documents with a null timestamp are omitted from this sort entirely, not ordered last.** `makeIndexOpsForDocument` (datalayer.ts:818) only emits an index op for non-default sort keys `if (indexedValue)`. When `bridge.lastUpdated` returns null (isomorphic `git.log` on an uncommitted/new file, an `fs.stat` failure, or any doc not yet reindexed), `__lastUpdated` is absent, so the doc gets no entry in this sublevel and disappears from the listing when sorted by Last Updated. It also skews pagination/total counts. Consider materializing an epoch/0 sentinel so missing-timestamp docs still appear deterministically at one end.

packages/@tinacms/graphql/src/database/level.ts
```diff
@@ -14,6 +14,16 @@ export type Level = AbstractLevel<
   Record<string, any>
 >;
 
+export interface TimestampedStore {
```
joshbermanssw, line 17: **This `TimestampedStore` path looks dead.** No production store implements `getTimestamp` (only `TimestampedMemoryLevel` in the tests), so `isTimestampedStore(store)` is always false and `Database.getDocumentTimestamp` always returns null in real deployments. Meanwhile the resolver `await`s `getDocumentTimestamp` per document on every resolve (after `initLevel()` plus key computation), and its result is discarded whenever `__lastUpdated` is present, since the resolver prefers the materialized field. Either drop this mechanism in favor of the materialized `__lastUpdated`, or document it as an intentional public extension point. As written it reads as accidental and adds per-document overhead.

packages/@tinacms/graphql/src/database/bridge/isomorphic.ts
```diff
@@ -549,6 +549,25 @@ export class IsomorphicBridge implements Bridge {
     return Buffer.from(blob).toString('utf8');
   }
 
+  public async lastUpdated(filepath: string) {
```
joshbermanssw, line 552: **Worth noting in the changeset:** this returns the last commit time, while the filesystem bridge returns mtime. The same edit reads differently across bridges, and the git value is stale for uncommitted working-tree edits (a file edited but not committed shows its previous commit date). Not necessarily wrong, but the cross-bridge inconsistency will surprise editors comparing local dev to a git-backed deploy.

packages/tinacms/src/admin/pages/CollectionListPage.tsx
joshbermanssw: **Reply** TinaCMS has a safety check in the admin (`isValidSortKey`) that only lets you sort by a column that's a real field defined in your schema. "Last Updated" isn't a field you defined — it's a special/synthetic one we invented — so the check says "not a real field," throws the sort away, and the server just returns the default order. The button looks like it works (arrow flips, list reloads) but the rows don't actually reorder. I confirmed it by following the code from the button → GetCollection → useGetCollection → that check → the fetch, so it's the same path that draws your list, not a separate one.

joshbermanssw: **Reply** Fixed in 4682b0287. `isValidSortKey` now short-circuits to true for system sort keys (`SYSTEM_SORT_KEYS = ['lastUpdated']`) before the schema-field check, so `validSortKey` keeps `lastUpdated` instead of dropping to `undefined`. Verified against the network request: the GraphQL `variables` now include `sort: "lastUpdated"` (previously `sort` was absent entirely).

packages/tinacms/src/admin/pages/CollectionListPage.tsx
joshbermanssw: **Reply** Fixed in a2ea4c4ff. The synthetic `lastUpdated` entry is now only appended when `sortableFields.some((f) => f.name === 'lastUpdated')` is false, so a collection with a real `lastUpdated` field won't produce a duplicate Select option.

packages/tinacms/src/admin/pages/CollectionListPage.tsx
joshbermanssw: **Reply** Fixed in 77a56fa7b. Since `lastUpdated` is DB metadata, not a document field, the right fix isn't `required: true` — the banner is now suppressed for `SYSTEM_SORT_KEYS` (exported from GetCollection.tsx and checked in the banner condition). With B3's null sentinel, no doc is excluded when sorting by it anyway.

packages/tinacms/src/admin/pages/CollectionListPage.tsx
joshbermanssw: **Reply** Fixed in 8f76505e3. The Filename header now sorts by the system `__filepath__` key (`FILENAME_SORT_KEY`, allowlisted in `SYSTEM_SORT_KEYS`) — it's document metadata, not a field — and the Title header sorts by `getIsTitleFieldName(collectionName)` instead of the literal `'title'`. `__filepath__` is exactly what `query()` already uses as its default sort (`index.ts:1069`), so filename sort matches existing behaviour and is reversible.

packages/tinacms/src/admin/pages/CollectionListPage.tsx
joshbermanssw: **Reply** Fixed in f48e39f59 — bumped the folders-only header's Path `th` to `colSpan={5}` so `Name` (1) + `Path` (5) = 6, matching the document rows. (Confirmed this view has no Last Updated column — it's just Name + Path — so it's purely an alignment fix.)

packages/@tinacms/graphql/src/database/index.ts
joshbermanssw: **Reply** Fixed in 11ff7bb59. Both `put` and `addPendingDocument` now recompute `bridge.lastUpdated(normalizedPath)` after the bridge write and set `dataFields[LAST_UPDATED_FIELD]` before building the index ops, so a save no longer blanks the value or drops the doc from the sort until reindex. Primarily a self-hosted fix (TinaCloud's store auto-stamps its own Timestamp); it's a no-op on TinaCloud until GithubBridge implements lastUpdated.

packages/@tinacms/graphql/src/database/index.ts
joshbermanssw: **Reply** Fixed in 9d79e18c9. When the bridge returns null, the three write paths now materialize an epoch-0 sentinel (`LAST_UPDATED_SENTINEL`) so the doc still gets a `lastUpdated` index entry (sorts at the oldest end) instead of disappearing. The resolver maps the sentinel back to `null` so `_sys.lastUpdated` still renders empty (—). Added a test asserting a null-timestamp collection keeps all docs when sorted by lastUpdated.

packages/@tinacms/graphql/src/database/bridge/isomorphic.ts
joshbermanssw: **Reply** Added to the changeset in 9fcf9457d: noted that the value is bridge-defined (git commit time for isomorphic — stale for uncommitted edits — vs mtime for filesystem) and that untimed docs sort last and render empty.

kulesy (commented):
Review

packages/@tinacms/graphql/src/resolver/index.ts
```diff
@@ -400,12 +410,15 @@ export class Resolver {
         path: rawData['__folderPath'],
       };
     } else {
+      const lastUpdated = await this.database.getDocumentTimestamp(fullPath);
```
kulesy, line 413: Design question, non-blocking. This store-mtime fallback (`getDocumentTimestamp`) is a second timestamp source layered under the materialized `__lastUpdated`. It only buys a pre-reindex display window on TinaCloud, yet sorting needs the reindex regardless, so the column stays inconsistent until then. It's also what forces the extra `transformDocumentIntoPayload` param and the branchy ternary above. Worth weighing whether dropping it, and relying solely on the materialized field, is simpler for the same practical result.

packages/@tinacms/graphql/src/resolver/index.ts
```diff
@@ -267,6 +270,13 @@ export const transformDocumentIntoPayload = async (
         breadcrumbs,
         collection,
         template: lastItem(template.namespace),
+        lastUpdated:
```
kulesy, line 273: Minor, readability. This nested ternary is hard to parse. Normalizing each source to `string | null` first, then `materialized ?? fromStore`, reads cleaner, and collapses to a single coalesce if the store fallback is dropped.

packages/@tinacms/graphql/src/database/bridge/isomorphic.ts
```diff
@@ -549,6 +549,25 @@ export class IsomorphicBridge implements Bridge {
...
+    assertWithinBase(filepath, this.relativePath);
+    const ref = await this.getRef();
+    try {
+      const commits = await git.log({
```
kulesy, line 556: Minor, perf. This runs a per-file `git.log` walk for every document at index time, sequentially, and only on the self-hosted IsomorphicBridge (filesystem is a single `stat`; TinaCloud doesn't implement it). It's partly cache-amortized and bounded, but on a large repo with deep history the reindex cost could be noticeable. A single whole-tree log pass, or a quick benchmark on a big repo, would settle it.

packages/@tinacms/graphql/src/database/datalayer.ts
```diff
@@ -68,6 +68,9 @@ type StringEscaper = <T extends string | string[]>(input: T) => T;
 
 export const DEFAULT_COLLECTION_SORT_KEY = '__filepath__';
 export const REFS_COLLECTIONS_SORT_KEY = '__refs__';
+export const LAST_UPDATED_SORT_KEY = 'lastUpdated';
```
kulesy, line 71: Minor. The sort key is the bare `lastUpdated` while the stored field and resolver use `__lastUpdated`. In `getIndexDefinitions` the system seeds `indexDefinitions['lastUpdated']`, then the user-field loop does `indexDefinitions[field.name] = ...`, so a collection with a user field named `lastUpdated` silently overwrites the system sort with that field's index. Harmless for most schemas, but setting `LAST_UPDATED_SORT_KEY = '__lastUpdated'` to match the field name closes the gap and keeps the two names consistent.

packages/@tinacms/graphql/src/database/index.ts
```diff
@@ -405,6 +432,12 @@ export class Database {
       }
     }
 
+    const lastUpdatedMs = await this.bridge?.lastUpdated?.(normalizedPath);
```
kulesy, line 435: Minor, maintainability. This 'call bridge then convert-or-sentinel' block is duplicated inline across three write paths (here, the `put` path near line 614, and `_indexContent` near line 1720). The history already shows one path was missed and patched later. A small `stampLastUpdated(record, path)` helper would give one place to keep in sync.

.changeset/sys-last-updated.md
```diff
@@ -0,0 +1,8 @@
+---
+"@tinacms/graphql": minor
+"tinacms": minor
+---
+
+Add a `lastUpdated` field to documents' `_sys` (`SystemInfo`) and a `lastUpdated` collection sort, both sourced from an optional `Bridge.lastUpdated(filepath)` capability that integrators can plug in (defaults: git commit time for the isomorphic bridge, file mtime for the filesystem bridge). The value is materialized into the index at index time, so `_sys.lastUpdated` and server-side `documents(sort: "lastUpdated")` work across any store/backend (including self-hosted), not just TinaCloud. The admin collection-listing page shows a "Last Updated" column and offers "Last Updated" in its sort control. Requires a reindex to populate the new sort index.
+
+Note: the value's meaning is bridge-defined and not identical across bridges — the isomorphic (git) bridge reports the file's last commit time (which is stale for uncommitted working-tree edits), while the filesystem bridge reports file mtime. Documents the bridge can't time are ordered last and shown empty.
```
kulesy, line 8: Minor. 'Ordered last and shown empty' only holds for a descending sort. The sentinel is epoch (`new Date(0)`) and the column's first click defaults to ascending, so untimed rows (the empty-value rows) sort to the front. Two follow-ons: default the date column's first click to descending, and note that pre-reindex (empty or partial index) a lastUpdated sort can silently drop un-indexed docs while the 'some documents may be excluded' banner is suppressed for system sort keys. Gating the sort until the index is populated would avoid that.

joshbermanssw:
Closing PR: As per my conversation with @18-th , @kulesy @wicksipedia and @adamcogan we will not be continuing with this feature - see ADR https://github.com/tinacms/tinacloud/pull/3805

PR #7103, opened 2026-06-26
chore(tinacms): move moment-timezone out of the shipped admin bundle (~39 KB gzip)

### LINKED ISSUE #7102: 🤖 chore(tinacms): move moment-timezone out of the shipped admin bundle (~39 KB gzip)
### Summary

The `tinacms` package declares `moment-timezone` as a **runtime dependency** and loads it via a side-effect import (`import 'moment-timezone'`) in `toolkit/components/ui/date-time-picker.tsx`. Production code never calls any `moment.tz.*` API — only a unit test uses `moment.tz.setDefault('UTC')`. Because a side-effect import can't be tree-shaken, the full IANA timezone database ships in the admin client bundle.

### Impact

Dropping the side-effect import and moving the dependency to `devDependencies` removes it from the shipped bundle:

- **~732 KB uncompressed / ~39 KB gzipped** less JavaScript on first admin load (measured against `moment-timezone@0.6.0`'s `moment-timezone-with-data.min.js`).

That payload is the entire timezone database (every IANA zone + historical transitions) that production code does not use.

### Proposed change

- Remove `import 'moment-timezone'` from `date-time-picker.tsx`.
- Remove a dead `import moment` from `toolkit/fields/plugins/date-format.ts` (imported, never called).
- Move `moment-timezone` from `dependencies` → `devDependencies`; the unit test keeps it via its own explicit import.
- Keep `moment` itself — it is still used for moment-token date formatting (`moment(date).format(...)`). Fully removing `moment` would change the public date-format-token contract and is a separate, breaking change.

### Notes

- No behavior change: the existing `date-time-picker.test.tsx` suite (415 tests) stays green.

PR incoming.

wicksipedia:
## What

Moves `moment-timezone` from `dependencies` to `devDependencies` in the `tinacms` package and removes the production side-effect import that pulled the full IANA timezone database into the admin bundle.

## Why

`moment-timezone` was loaded via `import 'moment-timezone'` in `toolkit/components/ui/date-time-picker.tsx`, but production code never calls any `moment.tz.*` API — only the unit test does (`moment.tz.setDefault('UTC')`). Because a side-effect import can't be tree-shaken, the entire timezone database shipped to every admin user.

## Impact

- **~732 KB uncompressed / ~39 KB gzipped** less JavaScript on first admin load (measured against `moment-timezone@0.6.0`'s `moment-timezone-with-data.min.js`).
- No behavior change — the existing `date-time-picker.test.tsx` suite (415 tests) stays green; the test keeps `moment-timezone` via its own explicit import.

## Scope / non-goals

- `moment` itself is **kept** — it still renders dates with moment-token formats (`moment(date).format(...)`), and the defaults plus user-configured format strings use moment tokens. Fully removing `moment` would be a separate, breaking change.

## Changes

- Remove `import 'moment-timezone'` from `date-time-picker.tsx`
- Remove dead `import moment` from `toolkit/fields/plugins/date-format.ts`
- Add explicit `import 'moment-timezone'` to `date-time-picker.test.tsx`
- Move `moment-timezone` `dependencies` → `devDependencies`
- Add a patch changeset

Closes #7102


JackDevAU (approved):
:shipit:

PR #7111, opened 2026-06-29
Reject cross-origin state-changing requests on the dev server

Aibono1225:
Closes https://github.com/tinacms/tinacloud/issues/3601

## Summary

The `@tinacms/cli` Vite dev server currently uses `cors({ origin: corsOriginCheck })`, but the `cors` middleware only controls CORS response headers. It does not reject disallowed origins server-side.

This matters for state-changing routes such as `POST /media/upload`: `multipart/form-data` can be sent as a CORS simple request without a preflight, so a malicious page can trigger the request from a developer's browser. The browser cannot read the response, but the server may still process the request and write a file into the configured media root.

## Fix

* Extracted the origin allow-list logic into `isOriginAllowed(origin, allowedOrigins)` in `cors.ts`.
* Added a server-side origin gate in `plugins.ts` before route handling.
* Disallowed cross-origin state-changing requests now return `403` before reaching handlers.

The gate applies to:

* `POST /media/upload`
* `DELETE /media`
* `POST` / `DELETE` search index routes

Existing CORS header behavior is unchanged.

Requests without an `Origin` header remain allowed to preserve existing behavior for non-browser clients such as curl, CLI tooling, and same-origin/server-side calls. Browser-driven cross-origin state-changing requests include an `Origin` header and are rejected when disallowed.

## Testing

Added coverage for:

* `isOriginAllowed` and the existing CORS origin callback behavior.
* The plugin middleware rejecting disallowed cross-origin state-changing requests before handlers run.
* Allowed cases, including no-Origin requests, localhost, and configured `allowedOrigins`.

The behavior was also verified locally with real HTTP multipart uploads:

* Before the fix, a disallowed cross-origin upload completed and wrote a file.
* After the fix, the same request returns `403` and no file is written.

### DIFF
--- a/.changeset/media-upload-cross-origin.md
+++ b/.changeset/media-upload-cross-origin.md
@@ -1,0 +1,5 @@
+---
+"@tinacms/cli": patch
+---
+
+Reject cross-origin state-changing requests on the dev server. Previously the `cors` middleware only suppressed response headers, so an attacker-controlled page could still drive a cross-origin multipart upload to completion (writing files into the media root). State-changing routes (`/media/upload`, `/media` DELETE, and `searchIndex` POST/DELETE) now reject disallowed origins server-side with a 403.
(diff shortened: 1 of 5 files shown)


18-th (approved):
LGTM

PR #7113, opened 2026-06-29
Remove the unused legacy Express dev server from @tinacms/cli

Aibono1225:
Deletes the old Express-based development server in `@tinacms/cli` (`packages/@tinacms/cli/src/server`), along with its own media router, media model, and tests.

## Impact

- No change to the running `tinacms dev` server or any user-facing behavior.
- No public API is removed (`@tinacms/cli/dist/server` was never a documented entry point).

## Notes

This is a follow-up cleanup related to the cross-origin media-upload hardening in #7111. That PR fixed the live (Vite) dev server; this one removes the parallel legacy server so the vulnerable handler is no longer shipped at all.


18-th (approved):
LGTM

PR #7122, opened 2026-06-30
feat(create-tina-app): stream live install activity on the spinner

### LINKED ISSUE #7120: `create-tina-app`: install step shows no progress for minutes, looks like a hang
### TL;DR
`create-tina-app`'s "Installing packages." step renders as a single `ora` spinner with no progress for the whole install. On a heavy starter (the Astro starter pulls 957 packages / 1.2 GB) that is minutes of an unchanging line, indistinguishable from a hang.

### Pain
The success path gives no liveness beyond the spinner glyph: no package count, no elapsed time, no package-manager output. A cold-cache Astro starter install takes around 5 minutes, during which the CLI looks frozen. This is the success-path counterpart to #7031, which fixed only the failure path by surfacing stderr.

### Reproduction
```
npm create tina-app@latest demo -- --template tina-astro-starter --pkg-manager npm
```
Default (no `--verbose`) output for the entire install is just:
```
- Installing packages.
✔ Installing packages.
```
Nothing in between. Warm cache: ~22s. Cold first run: ~5 min. The starter installs 957 packages / 1.2 GB.

### Cause
`packages/create-tina-app/src/util/install.ts` ignores the package manager's stdout unless `--verbose`:
```js
stdio: verboseOutput ? 'inherit' : ['ignore', 'ignore', 'pipe'],
```
So by default there is nothing to show but the spinner; `--verbose` is the only passthrough, and it dumps full npm output rather than a concise signal.

### Acceptance Criteria
1. The default install step shows a liveness signal that changes over time (e.g. elapsed time ticking next to the spinner) so a multi-minute install is clearly working, not hung.
2. No regression to the #7031 stderr error-surfacing or the `--verbose` passthrough.

kulesy:
Closes #7120

**TL;DR** The install step now reflects the package manager's live output on the spinner, instead of a static "Installing packages." that gave no sign of progress.

**Pain:** `create-tina-app` ran `<pm> install` with stdout ignored, so "Installing packages." showed no progress for the whole install (the Astro starter pulls ~900 packages / 1.2 GB, minutes on a cold cache). An animated spinner alone doesn't tell you the child is actually making progress.

**Solution:** Pipe the package manager's output and show its latest line on the spinner, e.g. `Installing packages, npm http fetch GET 200 .../vite`. npm is silent when piped so it gets `--loglevel=http`; pnpm and yarn stream on their own. Both stdout and stderr are piped (npm logs to stderr, pnpm/yarn to stdout), and `NO_COLOR` keeps the captured lines plain. `--verbose` still streams the full output, and the #7031 stderr-tail error capture is unchanged.


kulesy:
### Verification

What each package manager emits to a pipe (non-TTY), and the line it puts on the spinner:

| PM | activity stream | example spinner line |
|---|---|---|
| npm `--loglevel=http` | stderr | `npm http fetch GET 200 .../vite` |
| pnpm | stdout | `Progress: resolved 1491, downloaded 1315, added 457` |
| yarn 1.x | stdout | `[3/5] Fetching packages...` |
| bun | (silent during download) | `Installing packages.` |

Both stdout and stderr are piped because npm logs to stderr while pnpm/yarn log to stdout; `NO_COLOR` keeps the lines plain (verified the piped output has zero escape bytes, so no stripping needed). Driven end to end through the built CLI for npm and yarn; pnpm and bun via stream probes.

`tsc` and `tinacms-scripts build` pass; Biome clean on the changed lines. `--verbose` still streams the full output, and the #7031 stderr-tail error capture is preserved (a forced E404 still lands in the tail, with the plain `"npm install"` command).

kulesy:
The spinner reflects each package manager's own live output (real captures scaffolding `tina-astro-starter`):

install spinner activity across npm, pnpm, yarn, bun

Figure: the install step showing live activity across all four supported package managers. npm needs `--loglevel=http` to stream; pnpm/yarn/bun stream on their own. The line is shown raw (no per-manager parsing), so npm reads as a fetch URL, pnpm as resolved/downloaded counts, yarn as its current phase or warning, and bun as its resolve step.

packages/create-tina-app/src/index.ts
```diff
@@ -442,8 +442,12 @@ export async function run() {
 
   spinner.start('Installing packages.');
   try {
-    await install(pkgManager as PackageManager, opts.verbose);
-    spinner.succeed();
+    // Reflect the package manager's latest line on the spinner as live activity.
+    await install(pkgManager as PackageManager, opts.verbose, (line) => {
+      const text = line.length > 56 ? `${line.slice(0, 55)}…` : line;
```
18-th: Could we move the magic number into a named constant?

kulesy: Done, pulled it out into a `MAX_ACTIVITY_LINE` constant.

18-th (approved):
LGTM

joshbermanssw (approved):

PR #7123, opened 2026-06-30
🧹 Replace error-message string checks with shared constants

### LINKED ISSUE #6777: 🤖 🛠️ Refactor error message string checks into shared constants
CC:

### Description

Refactor error handling logic across `tinacms`, `@tinacms/graphql`, and other related packages to replace hardcoded string comparisons such as `error.message.includes('already exists')` with shared error constants or standardized error identifiers.

The current implementation relies on matching raw error message text, which is fragile and can break when error wording changes. Introduce a centralized set of constants, enums, or error codes that can be referenced consistently throughout the codebase to make error handling more maintainable and reliable.

we already have common packages shared between all, so we can look at adding these constants there.

### Acceptance Criteria

* [ ] Identify all usages of hardcoded error message checks across the repository
* [ ] Define a shared set of error constants, enums, or error codes for common error states
* [ ] Refactor existing `error.message.includes(...)` checks to use the shared constants instead of raw strings
* [ ] Ensure the shared error identifiers are accessible from both `tinacms` and `@tinacms/graphql`
* [ ] Add or update tests to verify error handling behavior continues to work correctly
* [ ] Document the preferred pattern for future error handling implementations to avoid direct string matching

🤖

joshbermanssw:
## TL;DR

Code decided *which* error occurred by substring-matching raw message prose (`error.message.includes('already exists')`), which breaks when a message is reworded. This adds shared error-identifier constants in `@tinacms/schema-tools` and points every in-repo check at them. Closes #6777.

## Notes

- Matched message bytes are unchanged, so existing resolver, filesystem-bridge, and CLI security assertions pass untouched.
- Constants are grouped by enforceability: genuine in-repo producer/consumer pairs (`ERR_ALREADY_EXISTS`, `ASYNC_POLLER_ERROR`); external TinaCloud contracts this repo can't enforce (`ERR_BRANCH_EXISTS`, `ERR_BRANCH_CONFLICT`, `ERR_HAS_REFERENCES`, `ERR_NOT_INDEXED`); and a documentation-only reference (`ERR_PATH_TRAVERSAL`, whose producers inline the literal).
- Fixes a latent delete-document bug: the check used `indexOf` (truthy even when the phrase was absent). It now uses `includes`, and the handler resolves instead of re-throwing to an un-catching modal.
- `@tinacms/cli` is left as an independent producer (canonical owner of its path message; its jest setup can't load the schema-tools ESM build).

Start at `@tinacms/schema-tools/src/errors.ts`.


18-th (dismissed):
LGTM

18-th (approved):
LGTM

PR #7128, opened 2026-06-30
Validate message sender in useEditState

Aibono1225:
## Summary

Aligns the `useEditState` hook's `message` handling with the pattern already used by `useTina`: it now validates the sender of incoming `message` events via `isFromAdmin(event, trustedAdminOrigins)` before acting on them, and cleans up its `message` listener on unmount.

## Changes

- `useEditState` gates its `message` handler on `isFromAdmin(...)`, consistent with `useTina`.
- The `message` listener is removed on unmount.
- Adds regression tests covering trusted vs. untrusted senders, the `TinaAdminOriginProvider` override, unrelated message types, and listener cleanup.


18-th (approved):
LGTM

Issue #7112, opened 2026-06-29
🐛 Bug - Edit link panel is clipped against the left edge in narrow sidebar


uly1:
Requested by: @uly1 via [YakShaver.ai](https://YakShaver.ai) 🦬
cc: @uly1, @kulesy, @JackDevAU

Hi Team!

&nbsp;&nbsp;[🟥&nbsp;&nbsp;Watch the video (55 sec)](https://www.youtube.com/watch?v=QhguCw7-jZA)

&nbsp;&nbsp;Url from screen share: [https://ssw.com.au/rules/admin/#/rules/best-ai-tools](https://ssw.com.au/rules/admin/#/rules/best-ai-tools)

### Pain
When editing a link in TinaCMS, the edit link panel appears flush against the left edge of the sidebar when the sidebar is narrow. This creates a clipped and visually uncomfortable experience, making it look like content may be cut off. The panel should include appropriate left padding/margin and adapt gracefully to narrower sidebar widths without requiring manual resizing.

### Acceptance Criteria
1. The Edit Link panel always includes visible left padding/margin and is not flush against the sidebar edge
2. The Edit Link panel renders correctly and remains readable at narrow sidebar widths
3. No content within the Edit Link panel appears clipped or cut off
4. The UI remains visually consistent when the sidebar is resized wider or narrower
5. No regression in the Edit Link experience at standard or wide sidebar widths

### Reproduce Steps
1. Open a page in TinaCMS with editable content containing a link
2. Ensure the TinaCMS sidebar is set to a narrow width
3. Click on the link in the content
4. Select 'Edit link'
5. Observe that the Edit Link panel appears flush against the left edge with no padding

### Screenshot
![Screenshot](https://sayakshaverproduction.blob.core.windows.net/images/images/a11522b3-d7a3-4866-9d1a-b6e3fd0d4f26_frame_40.png?sv=2025-05-05&se=2046-06-29T08%3A57%3A04Z&sr=b&sp=rd&sig=TsEFY2T7bzLsSn1Oo1GS3rXuLpyVNYLXSTY7g4e2dZU%3D)
**Figure: Edit link panel clipped against left edge in narrow sidebar**

joshbermanssw:
✅ Done 


https://github.com/user-attachments/assets/d3e6bcdd-5fa3-4afd-957b-4e5005be2618

**Figure: ❌ Before - bugs are out and about**


[...]

PR #7131, opened 2026-07-01
🐛 move floating edit bar into React Portal

joshbermanssw:
As per #7112 

- Moves the entire `LinkFloatingToolbar` into a `<PortalBody/>` so that the we move the Toolbar outside of its parent (the LHS editor) up into the `<body>` 

https://github.com/user-attachments/assets/d3e6bcdd-5fa3-4afd-957b-4e5005be2618

**Figure: ❌ Before - bugs are out and about**

https://github.com/user-attachments/assets/e9b5fc19-04a5-46bd-931e-947fce40b4ee

**Figure: ✅ After - moved to a `<Portal/>`**

### DIFF
--- a/.changeset/gentle-planets-fail.md
+++ b/.changeset/gentle-planets-fail.md
@@ -1,0 +1,5 @@
+---
+"tinacms": patch
+---
+
+move floatingtoolbar for links to a react portal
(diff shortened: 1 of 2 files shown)


18-th (approved):
LGTM

Issue #4150, opened 2023-08-17
Deleting or renaming a page causes a GraphQL error in Local Mode


jerefrer:
### Describe the issue that you're seeing. Any Loom videos or screenshots usually help a lot!

Hi there,

When I delete or rename a page **in Local Mode** the whole admin section goes down :
![Capture d’écran (45)](https://github.com/tinacms/tinacms/assets/140954/debdd040-2878-4e38-bb59-8f885ab0c689)

With this error in the console:
````
GraphQLError [Object]: Unable to find record content/Tagazok.md
   at Database.get (E:\Programming\songtsen\node_modules\.pnpm\@tinacms+graphql@1.4.22_react@18.2.0\node_modules\@tinacms\graphql\dist\index.js:5110:17)
   at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
   at async Resolver.getDocumentOrDirectory (E:\Programming\songtsen\node_modules\.pnpm\@tinacms+graphql@1.4.22_react@18.2.0\node_modules\@tinacms\graphql\dist\index.js:4323:23)
   at async E:\Programming\songtsen\node_modules\.pnpm\@tinacms+graphql@1.4.22_react@18.2.0\node_modules\@tinacms\graphql\dist\index.js:5587:26
   at async sequential (E:\Programming\songtsen\node_modules\.pnpm\@tinacms+graphql@1.4.22_react@18.2.0\node_modules\@tinacms\graphql\dist\index.js:78:18)
   at async Database.query (E:\Programming\songtsen\node_modules\.pnpm\@tinacms+graphql@1.4.22_react@18.2.0\node_modules\@tinacms\graphql\dist\index.js:5585:16)
   at async Resolver.resolveCollectionConnection (E:\Programming\songtsen\node_modules\.pnpm\@tinacms+graphql@1.4.22_react@18.2.0\node_modules\@tinacms\graphql\dist\index.js:4688:22)
   at async Promise.all (index 4)
   at async Promise.all (index 0)
   at async resolve (E:\Programming\songtsen\node_modules\.pnpm\@tinacms+graphql@1.4.22_react@18.2.0\node_modules\@tinacms\graphql\dist\index.js:4853:17)
   at async E:\Programming\songtsen\node_modules\.pnpm\@tinacms+cli@1.5.26_@babel+core@7.22.9_react@18.2.0\node_modules\@tinacms\cli\dist\index.js:1211:26
[...]

Issue #4018, opened 2023-06-19
Delete not working in windows


kldavis4:
### Describe the issue that you're seeing. Any Loom videos or screenshots usually help a lot!

```
GraphQLError [Object]: Unable to find record content/post/dasdfsdf.md
    at Database.get (C:\Users\kldav\Documents\projects\tinacms\packages\@tinacms\graphql\dist\index.js:5133:17)
    at processTicksAndRejections (node:internal/process/task_queues:96:5)
    at async Resolver.getDocumentOrDirectory (C:\Users\kldav\Documents\projects\tinacms\packages\@tinacms\graphql\dist\index.js:4347:23)
    at async C:\Users\kldav\Documents\projects\tinacms\packages\@tinacms\graphql\dist\index.js:5610:26
    at async sequential (C:\Users\kldav\Documents\projects\tinacms\packages\@tinacms\graphql\dist\index.js:76:18)
    at async Database.query (C:\Users\kldav\Documents\projects\tinacms\packages\@tinacms\graphql\dist\index.js:5608:16)
    at async Resolver.resolveCollectionConnection (C:\Users\kldav\Documents\projects\tinacms\packages\@tinacms\graphql\dist\index.js:4712:22)
    at async Promise.all (index 4)
    at async Promise.all (index 0)
    at async resolve (C:\Users\kldav\Documents\projects\tinacms\packages\@tinacms\graphql\dist\index.js:4878:17)
    at async C:\Users\kldav\Documents\projects\tinacms\packages\@tinacms\cli\dist\index.js:1217:26
Error querying file content/post/dasdfsdf.md from collection post. Please run "tinacms audit" or add the --verbose option  for more info

GraphQL request:8:11
8 |           documents(sort: $sort, after: $after, first: $limit, filter: $filter,
  |           ^
  | folder: $folder) @include(if: $includeDocuments) {
event - compiled client and server successfully in 30.4s (2133 modules)
```

To reproduce: 

### Reproduction

https://github.com/tinacms/tinacms

### Steps to reproduce

On Windows, run the basic-iframe starter in the monorepo. Create a new post. Delete the post.

### System Info

```shell
windows 10 home
```

### Validations

- [X] Follow our [Code of Conduct](https://github.com/tinacms/tinacms/blob/main/CODE_OF_CONDUCT.md)
- [X] Read the [docs](https://tina.io/docs).
[...]

PR #7133, opened 2026-07-01
fix(graphql): tolerate a dangling sort-index entry in collection queries

kulesy:
**TL;DR** A stale (dangling) entry in a collection's field-sort index made `Database.query` throw for the entire collection list; this skips the dangling entry instead of failing, so the collection loads (`tinacms audit` / re-index still purges the orphan).

**Pain:** Index entries live in per-sort sublevels (`contentLevel -> <collection> -> <sortKey>`). The default `__filepath__` index is keyed by the bare path and can't go stale, but a field-sort index is keyed by `` `<value>\x1D<path>` `` and a key can only be removed by a `del` derived from that value. A value-change or reconcile that doesn't `del` the old value's key (the incremental indexer's `if (item)` cleanup gap, or an index batch interrupted by a crashed/timed-out process) strands the old entry; a later delete clears the path-keyed entries (the `__filepath__` index + the `CONTENT_ROOT` record) but not the stranded value-keyed one. `Database.query` then hydrates the dangling entry, `Database.get` throws `NotFoundError`, and it surfaces as `TinaQueryError` ("Error querying file ... from collection ...") for the whole collection. It presents as intermittent and per-user because the selected sort is cached per-browser in `localStorage`, so only editors on the affected field sort hit it while default-sort editors are fine. Previously closed as Windows / local re-index timing (#4018, #4150), but it is neither Windows-specific nor fixed by re-indexing a different index.

**Solution:** In `Database.query`, when hydrating a node throws `NotFoundError` (its record is gone), drop that edge instead of rethrowing `TinaQueryError`, so the collection loads and the orphan is skipped; `tinacms audit` / re-index still removes it permanently. This is resilience, not prevention (never stranding a value-keyed index entry is a deeper change to index maintenance, left as a follow-up). Adds `database.sort-index-ghost.test.ts` (real `Database` + `MemoryLevel` + in-memory bridge) reproducing the exact error and asserting the tolerant behavior, plus a control test showing the in-band delete path is unaffected.

### DIFF
--- a/packages/@tinacms/graphql/src/database/database.sort-index-ghost.test.ts
+++ b/packages/@tinacms/graphql/src/database/database.sort-index-ghost.test.ts
@@ -1,0 +1,150 @@
+/**
+ * Repro: "Error querying file <path> from collection <collection>" (surfaced as
+ * "GetCollection failed: Unable to fetch"), broken ONLY for users sorting by a
+ * field while default-sort users are fine.
+ *
+ * Index entries live in per-sort sublevels (contentLevel -> <collection> -> <sortKey>).
+ * The default `__filepath__` index is keyed by the bare path, so it can't go stale.
+ * A field-sort index (e.g. `score`) is keyed by `<value>\x1D<path>` and is removable
+ * only by a `del` derived from that value -- so a value change/reconcile that fails to
+ * `del` the old value (the incremental indexer's `if (item)` gap, or an interrupted
+ * batch) strands the old key. A later delete clears the path-keyed entries but not the
+ * stranded one: default list stays healthy while the field-sorted list breaks. The fix
+ * makes Database.query skip an index entry whose record is gone, so these tests assert
+ * the tolerant behavior (field-sorted list returns the healthy items, no throw).
+ *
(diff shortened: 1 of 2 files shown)


kulesy:
### Verification

- The regression test passes with the fix and **fails without it** with the exact reported error: `Error querying file content/posts/ghost.json from collection post. Please run "tinacms audit" or add the --verbose option  for more info`.
- **368** existing `@tinacms/graphql` database + resolver tests pass, no pagination / sort / filter / cursor regressions.
- Type-clean in the edited files (the only `tsc` errors are pre-existing in the sibling `@tinacms/mdx` package, untouched here).

Before the fix, the reproduction throws on the real code path:

```
NotFoundError: Unable to find record content/posts/ghost.json
    at Database.get (packages/@tinacms/graphql/src/database/index.ts)
    at Database.query (packages/@tinacms/graphql/src/database/index.ts)
```

which `Database.query` surfaces as `TinaQueryError` ("Error querying file ... from collection ..."). After the fix, the dangling entry is skipped and the collection loads. A control test confirms the normal in-band delete path already cleans the field-sort index (no strand).

PR #7137, opened 2026-07-02
Adding a boolean field scaffold to the v4

### LINKED ISSUE #6916: Field Type Implementation - Boolean
## Pain

Boolean values are simple, but they often expose inconsistencies in form defaults, empty values and validation. If `BooleanField` is not implemented as a field plugin, checkbox/toggle behaviour remains tied to legacy form wiring rather than the shared ingest, digest, validation and rendering contract.

## Approach

Implement `BooleanField` as a field plugin against the field contract. Reuse existing boolean field behaviour where possible, while moving default handling, serialization, validation and rendering behind the plugin interface.

## Expected outcomes

- Boolean fields work through the v4 field plugin system.
- Toggle or checkbox rendering is owned by the boolean field plugin.
- Boolean defaults and empty-state behaviour are consistent across forms.
- Boolean field behaviour is isolated and testable.

## Acceptance Criteria

- `BooleanField` exists as a field plugin that conforms to the field contract.
- Ingest and digest behaviour preserves existing boolean schema and stored value behaviour.
- Rendering supports the expected editor control for boolean values.
- Default, false and empty values are handled without ambiguity.
- Tests cover rendering, true/false updates, default handling, ingest, digest and validation.

18-th:
Closes #6916 

* Wires in the client, schema, plugin and ui for the boolean field
* Boolean field is a 2 state by design
* Tests
* Updated docos

### DIFF
--- a/packages/v4/@tinacms/tinacms/_docs/boolean-field.md
+++ b/packages/v4/@tinacms/tinacms/_docs/boolean-field.md
@@ -1,0 +1,123 @@
+# The `boolean` field
+
+One of the field plugins v4 ships: a single checkbox backed by a Zod type guard.
+
+## Files
+
+All four live in `plugins/fields/boolean/`:
+
+| File | Role |
+|---|---|
+| `boolean-field.schema.ts` | the `t.boolean()` helper + the `booleanSchema` guard |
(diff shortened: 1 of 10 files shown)


JackDevAU (approved):
:shipit:

PR #7139, opened 2026-07-02
Adding a number field

### LINKED ISSUE #6917: Field Type Implementation - Number
## Pain

Number fields need consistent handling for parsing, empty values, validation and stored numeric output. Today that behaviour can be coupled to form implementation details, which makes it easy for number inputs to drift between string-like editor values and numeric stored values.

## Approach

Implement `NumberField` as a field plugin against the field contract. The plugin owns numeric rendering, parsing, ingest, digest and validation while preserving existing schema and storage behaviour.

## Expected outcomes

- Number fields work through the v4 field plugin system.
- Numeric parsing and empty-state handling are owned in one place.
- Stored values remain numeric rather than leaking input-string behaviour.
- Number field behaviour is isolated and testable.

## Acceptance Criteria

- `NumberField` exists as a field plugin that conforms to the field contract.
- Ingest and digest behaviour preserves the existing number schema and stored value behaviour.
- Rendering supports the expected editor input for numeric values.
- Empty, zero, negative and decimal values are handled correctly.
- Tests cover rendering, value updates, parsing, empty values, ingest, digest and validation.

18-th:
Closes #6917 

* Number field plugin (schema, client, ui, plugin) with parse/serialize, min/max, step
* Adds useFieldSchema() hook so a field can read its own node (wires step)
* Tests + docs

This notably wires a schema into the field to be exposed through the context

### DIFF
--- a/packages/v4/@tinacms/tinacms/_docs/field-plugins.md
+++ b/packages/v4/@tinacms/tinacms/_docs/field-plugins.md
@@ -26,7 +26,10 @@ below — substitute your own `type` for `string` throughout. For that field's
 exact config options and validation semantics, see
 [`string-field.md`](./string-field.md). The shipped `boolean` field
 ([`boolean-field.md`](./boolean-field.md)) is a second example — a two-state
-checkbox where `required` is a no-op.
+checkbox where `required` is a no-op. The shipped `number` field
+([`number-field.md`](./number-field.md)) is a third — a string editor value ↔
+numeric stored value via `parse`/`serialize`, reading its own config
+(`step`) through `useFieldSchema`.
 
 ### 1. Manifest (`.plugin.ts`)
 
@@ -89,10 +92,11 @@ export const string = (config: Omit<StringFieldSchema, 'type'>): StringFieldSche
 
 ### 4. The component (`.ui.tsx`)
 
-The component receives **only its address** via context and pulls value/errors
-through address-keyed hooks (`editor/hooks.ts`). No `value`/`onChange` props:
-that keeps each field an O(1) react-hook-form subscription instead of
-re-rendering the whole form on a keystroke.
+The component receives **its address and its resolved schema node** via context
+(both provided by `<Field>`) and pulls value/errors through address-keyed hooks
+(`editor/hooks.ts`). No `value`/`onChange` props: that keeps each field an O(1)
+react-hook-form subscription instead of re-rendering the whole form on a
+keystroke.
 
 ```tsx
 import { useRef } from 'react';
@@ -123,6 +127,7 @@ Hooks:
 | Hook | Does |
 |---|---|
 | `useFieldAddress()` | this field's address |
+| `useFieldSchema<T>()` | this field's resolved schema node (render hints, e.g. `step`) |
 | `useFieldValue<T>(address)` | `[value, setValue]` via the RHF controller |
 | `useFieldErrors(address)` | validation messages at the address |
 | `useFieldActivation(handler)` | run `handler` when this becomes the active field (visual editing) |
--- a/packages/v4/@tinacms/tinacms/_docs/number-field.md
+++ b/packages/v4/@tinacms/tinacms/_docs/number-field.md
@@ -30,16 +30,29 @@ Config (`NumberFieldSchema`, extends `BaseFieldSchema`):
 | `max` | `number` | maximum **value** |
 | `step` | `number` | the input's `step` (render hint; no validation role) |
 
+## Descriptor
+
+The client segment (`number-field.client.tsx`) claims the `number` key:
+
+```tsx
+defineClientPlugin({
+  field: {
+    type: 'number',            // NUMBER_FIELD_TYPE
+    Component: NumberField,
+    // no defaultValue — an absent field stays absent
+    metadata: { layout: 'inline' },
+    schema: numberSchema,
+    parse: (stored) => (stored == null ? undefined : String(stored)),  // load: number → string
+    serialize: (value) => Number(value),                               // save: string → number
+  },
+});
+```
+
 ## Editor value vs stored value
 
 The DOM gives `<input type="number">` a **string**, but the document stores a
-**number**. The descriptor bridges them, so partial entries (`-`, `1.`) survive
-typing and the document always gets a clean number back:
-
-```ts
-parse: (stored) => (stored == null ? undefined : String(stored)),  // load: number → string
-serialize: (value) => Number(value),                                // save: string → number
-```
+**number**. `parse`/`serialize` (above) bridge them, so partial entries
+(`-`, `1.`) survive typing and the document always gets a clean number back.
 
 Empty is `undefined`, converted in **one place** — the component's `onChange`
 (`'' → undefined`). There is no `defaultValue`, and `digestDocument` drops
--- a/packages/v4/@tinacms/tinacms/_docs/number-field.md
+++ b/packages/v4/@tinacms/tinacms/_docs/number-field.md
@@ -1,0 +1,109 @@
+# The `number` field
+
+A shipped field plugin: a numeric `<input type="number">` whose **stored value is
+a number** but whose **editor value is the raw input string**. Source:
+`plugins/fields/number/`.
+
+## Authoring
+
+`t.number({...})` stamps `type: 'number'` (`NUMBER_FIELD_TYPE`) onto the config:
+
+```ts
+import { t } from '@tinacms/tinacms';
+
+const collection = {
+  name: 'post',
+  fields: [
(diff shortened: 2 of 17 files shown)


packages/v4/@tinacms/tinacms/src/plugins/fields/number/number-field.schema.ts
```diff
@@ -0,0 +1,49 @@
...
+// Coerce the editor string before validating. Guard empty explicitly: `0` is valid
+// and a falsy test would smuggle empty in as zero (`Number('') === 0`).
+const toNumber = (value: unknown): unknown =>
+  value === '' || value == null ? undefined : Number(value);
+
+// `min`/`max` are value bounds (not string length); a non-numeric string becomes
+// `NaN`, which `z.number()` rejects.
+export const numberSchema = (node: FieldSchema): ZodType => {
+  const field = node as NumberFieldSchema;
+  let schema = z.number({
```
JackDevAU: Typing `1e999` (valid in a number input) coerces to `Infinity`, which zod's `z.number()` accepts — and `JSON.stringify` then writes it as `null` in the saved doc. Worth adding `.finite()` here.

packages/v4/@tinacms/tinacms/src/plugins/fields/number/number-field.schema.ts
```diff
@@ -0,0 +1,49 @@
...
+export const number = (
+  config: Omit<NumberFieldSchema, 'type'>
+): NumberFieldSchema => ({ ...config, type: NUMBER_FIELD_TYPE });
+
+const labelOf = (node: NumberFieldSchema): string => node.label ?? node.name;
+
+// Coerce the editor string before validating. Guard empty explicitly: `0` is valid
+// and a falsy test would smuggle empty in as zero (`Number('') === 0`).
+const toNumber = (value: unknown): unknown =>
+  value === '' || value == null ? undefined : Number(value);
```
JackDevAU: `Number(' ') === 0`, so a whitespace-only value slips past the `''` guard, passes `required`, and saves as `0` — the exact empty-as-zero case the comment above guards against. Trimming before the check would close it (plus a test to pin it).

packages/v4/@tinacms/tinacms/src/plugins/fields/number/number-field.test.tsx
```diff
@@ -0,0 +1,187 @@
...
+    renderField('rating');
+    const input = await screen.findByLabelText('rating');
+    await userEvent.type(input, '0');
+    expect(await screen.findByRole('alert')).toHaveTextContent(
+      'Rating must be at least 1'
+    );
+  });
+});
+
+describe('NumberField validation', () => {
```
JackDevAU, line 102: The docs promise an optional empty value passes via `.optional()`, but none of the validation tests hit the optional `weight` node with `''`/`undefined`. One small test would pin that branch.

packages/v4/@tinacms/tinacms/src/plugins/fields/number/number-field.ui.tsx
```diff
@@ -0,0 +1,44 @@
...
+
+  // TODO(shadcn): swap this raw input/markup for shared, themed primitives from
+  // src/ui/ (shadcn — Input/Label/form-field wrapper, added via the shadcn CLI) so
+  // every field looks consistent and is re-themeable without per-field redesign.
+  return (
+    <div>
+      <input
+        ref={inputRef}
+        type='number'
+        step={field.step}
```
JackDevAU: Without a configured `step` the browser default is `1`, so stored decimals like `-1.5` hit `stepMismatch` (`:invalid`, spinner snaps to integers). `step={field.step ?? 'any'}` fixes it.

packages/v4/@tinacms/tinacms/src/plugins/fields/number/number-field.ui.tsx
```diff
@@ -0,0 +1,44 @@
...
+  useFieldActivation(() => inputRef.current?.focus());
+
+  // TODO(shadcn): swap this raw input/markup for shared, themed primitives from
+  // src/ui/ (shadcn — Input/Label/form-field wrapper, added via the shadcn CLI) so
+  // every field looks consistent and is re-themeable without per-field redesign.
+  return (
+    <div>
+      <input
+        ref={inputRef}
+        type='number'
```
JackDevAU, line 29: Small UX footgun: Chrome mutates a focused number input on wheel scroll, so scrolling the form can silently edit a value. Blurring on wheel is the usual guard.

packages/v4/@tinacms/tinacms/src/plugins/fields/number/number-field.client.tsx
```diff
@@ -0,0 +1,16 @@
...
+
+export default defineClientPlugin({
+  field: {
+    type: NUMBER_FIELD_TYPE,
+    Component: NumberField,
+    // No defaultValue: empty stays `undefined`, so `serialize` only ever sees a number.
+    metadata: { layout: 'inline' },
+    schema: numberSchema,
+    // Normalise a stored `null` to empty rather than "null"/`NaN`.
+    parse: (stored) => (stored == null ? undefined : String(stored)),
```
JackDevAU, line 13: Now that parse/serialize are wired to a built-in field, the comments in `contract.ts` ("not wired to any built-in field yet") and `ingest.ts` ("Every field is identity today") are both stale — worth updating them in this PR.

packages/v4/@tinacms/tinacms/src/plugins/fields/number/number-field.client.tsx
```diff
@@ -0,0 +1,16 @@
...
+    serialize: (value) => Number(value),
```
JackDevAU, line 16: If a document already holds a non-number here (`"abc"`, `true`), `parse` stringifies it and an unedited save runs it through `Number()` → `NaN`, which JSON writes as `null`. `digestDocument` never validates, so a NaN guard in `serialize` would be safer.

packages/v4/@tinacms/tinacms/src/plugins/fields/number/number-field.ui.tsx
```diff
@@ -0,0 +1,44 @@
...
+        step={field.step}
+        aria-label={address}
+        value={value ?? ''}
+        onChange={(event) =>
+          setValue(event.target.value === '' ? undefined : event.target.value)
```
JackDevAU, line 37: Number inputs report `''` on badInput (e.g. `5e`), so this stores `undefined` while the stale text stays visible — an optional field then saves as absent with no error shown. Checking `validity.badInput` would catch it.

packages/v4/@tinacms/tinacms/src/editor/context.ts
```diff
@@ -10,9 +10,12 @@ export interface ActiveField {
 
 // TODO(zustand): RegistryContext and ActiveFieldContext become reads off the global
 // store (ADR-003) once it exists. CollectionContext is form-scoped (the open
-// document's schema), and FieldAddressContext is intentionally plain React context —
-// per ADR-009 the only thing a field receives is its address, passed down by <Field>.
+// document's schema). FieldAddressContext and FieldSchemaContext are intentionally
```
JackDevAU, line 13: This rewrites what ADR-009 is cited as saying (address-only → address + schema node), but the ADR itself wasn't amended. Could you either update ADR-009 or word this as an extension of it?

packages/v4/@tinacms/tinacms/_docs/field-plugins.md
```diff
@@ -123,6 +127,7 @@ Hooks:
 | Hook | Does |
 |---|---|
 | `useFieldAddress()` | this field's address |
+| `useFieldSchema<T>()` | this field's resolved schema node (render hints, e.g. `step`) |
```
JackDevAU, line 130: architecture.md still describes the address-only field contract and its hooks table omits `useFieldSchema` — the two docs now disagree. Worth updating architecture.md alongside this.

packages/v4/@tinacms/tinacms/_docs/number-field.md
```diff
@@ -0,0 +1,109 @@
...
+| Key | Type | Effect |
+|---|---|---|
+| `name` | `string` (required) | field key in the document; also the fallback label |
+| `label` | `string` | display label; used in validation messages |
+| `required` | `boolean` | empty value fails validation |
+| `min` | `number` | minimum **value** (not length) |
+| `max` | `number` | maximum **value** |
+| `step` | `number` | the input's `step` (render hint; no validation role) |
+
+## Editor value vs stored value
```
JackDevAU: string-field.md and boolean-field.md both have a `## Descriptor` section showing the full `defineClientPlugin` block — adding one here would keep the trio consistent.

JackDevAU (approved):

PR #7138, opened 2026-07-02
✨ Replace editorial-workflow draft/ready toggle with a save-options dropdown

joshbermanssw:
## TL;DR

Replaces the draft / ready-for-review **toggle** in the "Save changes to new branch" modal with a **save-options dropdown**: Save draft, Save (ready for review), and Save and publish. The split button's main action is the editor's last choice (default **Save draft**, remembered via `localStorage`); the caret menu holds the other two and opens downward.

Screenshot 2026-07-02 at 3 11 01 pm

**Figure: Mockup of UI**

Screenshot 2026-07-02 at 3 11 08 pm

**Figure: Hover over disabeld option**

Screenshot 2026-07-02 at 3 34 09 pm

**Figure: Actual UI on stagged + tagged version**

https://tina-io-git-jb-test-tagged-v-tinacms.vercel.app/admin#/~

Screenshot 2026-07-02 at 3 35 39 pm

**Figure: LocalStorage saving**

## Reviewer notes

- **Save and publish** (direct commit) is disabled with a tooltip when the branch is TinaCloud-protected (`usingProtectedBranch()`). If the remembered choice is a now-disabled publish, the main button falls back to Save draft.
- **`DropdownButton`** gained an optional `tooltip` on items, shown even when the item is disabled (drives the disabled-publish tooltip).
- **Media flow unchanged**: `CreateBranchPromptModal` is shared, so the new options are gated behind `showSaveOptions`; the media modal keeps its legacy button.
- Pure choice resolution (`resolveSaveOptions`: default + fallback + menu ordering) is unit-tested; the old toggle component and its localStorage helpers are removed.

## Links

- Follows the merged draft/ready feature (#7136)

PR #7140, opened 2026-07-02
✨ Track editorial-workflow save option choice in PostHog

joshbermanssw:
As per https://github.com/tinacms/tinacloud/issues/3815

## TL;DR

Adds a single PostHog event, `editorial-workflow-save`, that records which save option an editor used in the "Save changes to new branch" modal (`draft` / `review` / `publish`), whether it succeeded, and the failure reason.

## Reviewer notes

- **All three choices in one event.** draft/review fire from the modal with the workflow result; publish fires from the form-builder save path (`safeSubmit`) so its outcome is tagged too.
- **`executeWorkflow` now returns `{ success, error }`** (was `boolean`) so draft/review failures carry a message; both callers (`create-branch-modal`, `branch-deleted-modal`) were updated.
- Fires on the **outcome**, so a failed save is recorded as `success: false` (never counted as a save).

Screenshot 2026-07-03 at 10 09 42 am

Screenshot 2026-07-03 at 10 14 23 am

**Figure: PostHog events w/ Properties**

## Links

- Metric for the save-options modal (#7138)

### DIFF
--- a/packages/tinacms/src/toolkit/form-builder/create-branch-modal.tsx
+++ b/packages/tinacms/src/toolkit/form-builder/create-branch-modal.tsx
@@ -148,6 +148,10 @@ export const CreateBranchModal = ({
       branchGuardAbortRef.current = null;
     }
 
+    // Cancelled mid-run (modal closed, branch renamed, another save started, or
+    // unmounted) — treat as a no-op and record nothing.
+    if (abortController.signal.aborted) return;
+
     captureEvent(EditorialWorkflowSaveEvent, {
       choice: isDraft ? 'draft' : 'review',
       success,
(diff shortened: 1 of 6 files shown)


packages/tinacms/src/toolkit/form-builder/create-branch-modal.tsx
```diff
@@ -146,6 +148,12 @@ export const CreateBranchModal = ({
       branchGuardAbortRef.current = null;
     }
 
+    captureEvent(EditorialWorkflowSaveEvent, {
```
18-th, line 155: How does this handle the cancellation before and after the submit?

18-th (approved):
LGTM

PR #7141, opened 2026-07-03
⬆️ Bump @radix-ui/* deps to latest and drop unused react-checkbox

joshbermanssw:
## TL;DR

Refreshes the `@radix-ui/*` packages used by `tinacms` to their newest same-major releases (all patch/minor, no breaking changes) and removes the one Radix dependency that isn't imported anywhere.

## What changed

- **Bumped 8 Radix packages** via the `catalog:` block in `pnpm-workspace.yaml` (`tinacms` is the only consumer): dialog `1.1.18`, dropdown-menu `2.1.19`, popover `1.1.18`, separator `1.1.11`, select `2.3.2`, slot `1.3.0`, toolbar `1.1.14`, tooltip `1.2.11`.
- **Switched `react-slot`** in `packages/tinacms` from a hard-pinned `^1.1.2` to `catalog:` so it tracks the catalog like its siblings.
- **Removed `@radix-ui/react-checkbox`** (see below).
- `@radix-ui/react-icons` was already current — untouched.

## Why remove `react-checkbox`

It was declared in `packages/tinacms/package.json` but **imported in zero source files**:

- The checkbox *field* (`toolkit/fields/components/checkbox-group.tsx`) is a custom implementation — it never used the Radix primitive.
- The `Checkbox` references in the dropdown code are `DropdownMenuCheckboxItem`, which ships inside `@radix-ui/react-dropdown-menu` — a different package.

`tinacms` was the sole reference, so the dep and its now-orphaned catalog entry both come out. No import sites change.

## Verification

`pnpm install` re-resolved cleanly; `turbo build --filter=tinacms` and `tsc` both green (13/13 type tasks).

## Links

- Tracking PBI: tinacms/tinacloud#3820

### DIFF
--- a/.changeset/bump-radix-ui.md
+++ b/.changeset/bump-radix-ui.md
@@ -2,4 +2,4 @@
 "tinacms": patch
 ---
 
-Update `@radix-ui/*` dependencies to their latest patch/minor releases
+Update `@radix-ui/*` dependencies to their latest patch/minor releases and remove the unused `@radix-ui/react-checkbox` dependency
--- a/.changeset/bump-radix-ui.md
+++ b/.changeset/bump-radix-ui.md
@@ -1,0 +1,5 @@
+---
+"tinacms": patch
+---
+
+Update `@radix-ui/*` dependencies to their latest patch/minor releases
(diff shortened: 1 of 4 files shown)


18-th (approved):
LGTM

PR #7143, opened 2026-07-03
fix(tinacms): reject invalid folder names inline via shared relativePath allowlist

### LINKED ISSUE #7142: Unify folder-name validation with the document-filename/backend relativePath allowlist
### TL;DR

Creating a folder with an invalid name (e.g. spaces) passes the admin UI's non-empty-only check, is sent to the backend, and fails with a generic "please contact the developer" error. Document filenames already validate the character allowlist inline. Unify both on a single shared allowlist so folders get the same friendly inline validation.

### Pain

- The Create Folder modal only checks that the name is non-empty unless a project sets `ui.regexValidation.folderNameRegex`. A name like `a b c d` is accepted, the `createFolder` mutation is sent, and the backend rejects it with `Invalid path: relativePath contains invalid characters`. That surfaces as the generic "There was an error creating the folder. Please try again. If the problem continues, please contact the developer" alert, which tells the editor nothing.
- Document filenames don't have this problem: the create-document form always validates against the allowlist and shows "Must contain only a-z, A-Z, 0-9, -, _, ., or /." inline before submit.
- The same allowlist is encoded independently in three places (the doc-filename UI, the backend `validateRelativePath`, and the optional folder regex), and they have drifted: the folder modal has no default.

### Approach

- Add a single shared allowlist (`RELATIVE_PATH_REGEX`) plus its human-readable message to `@tinacms/schema-tools`.
- Backend `validateRelativePath` imports the shared regex instead of an inline literal.
- The create-document filename validation imports the shared regex + message.
- The Create Folder modal applies the shared allowlist as a baseline, so invalid characters are rejected inline with a specific message and the Create button stays disabled. A project's `folderNameRegex` becomes an additional constraint on top of the baseline, never a replacement, so the client can only be stricter than the backend, never looser.
- Backend enforcement stays as the authoritative path-traversal gate; the client change is UX only.

### Acceptance Criteria

- Creating a folder whose name contains a space (or any character outside the allowlist) is blocked in the modal with the allowed-characters message; the Create button stays disabled and no `createFolder` request is sent.
- Valid folder names (`a-b-c-d`, `a_b_c_d`, nested `parent/child`) still create successfully.
- Document filename validation behaviour is unchanged.
- A project-level `folderNameRegex` still applies, layered on top of the baseline allowlist.
- One shared allowlist constant is the source of truth for the admin UI and the backend.

kulesy:
Closes #7142

**TL;DR** Unify folder-name validation with the document-filename and backend `relativePath` allowlist so invalid folder names (e.g. spaces) are rejected inline instead of failing on the backend with a generic error.

**Pain:** The Create Folder modal only checked that a name was non-empty unless a project set `ui.regexValidation.folderNameRegex`. A name like `a b c d` passed that check, the `createFolder` mutation was sent, and the backend rejected it with `Invalid path: relativePath contains invalid characters`, surfaced as the unhelpful "There was an error creating the folder. Please try again. If the problem continues, please contact the developer" alert. Document filenames already validated the same allowlist inline, so the two flows had drifted.

**Solution:** Add a single shared allowlist (`RELATIVE_PATH_REGEX` + `RELATIVE_PATH_ALLOWED_CHARS_MESSAGE` + `isValidRelativePath`) to `@tinacms/schema-tools`, and consume it from the resolver's `validateRelativePath`, the create-document filename field, and the Create Folder modal. The folder modal applies the allowlist as a baseline and layers any project-level `folderNameRegex` on top, so the UI can only ever be stricter than the backend, never looser. Backend enforcement is unchanged and remains the authoritative path-traversal gate; the client change is UX only.


kulesy:
**Verification**

- `@tinacms/schema-tools` unit test for `isValidRelativePath` (accepts allowed chars, rejects spaces and other disallowed chars): passing.
- `@tinacms/graphql` resolver suite (`src/resolver/index.test.ts`, 116 tests incl. existing interior-space rejection): passing after the constant swap.
- `tsc --noEmit` on the touched files: clean (the pre-existing `@tinacms/mdx` errors are unrelated and also present on `main`).
- Biome check on all changed files: clean.
- Added an e2e (`playwright/tina-playwright/tests/action/create-folder.spec.ts`) mirroring the existing filename-rejection test. Not run locally (the tina-playwright suite needs a full monorepo build + `pnpm dev` server); it will run in CI.

kulesy:
**Before / after in the real admin**

Captured from the `playwright/tina-playwright` app running against a local Tina content server (Next on :3000, Tina GraphQL on :4001), same input `a b c d` on the `author` collection.

Pre-fix: modal accepts a name with spaces and enables Create

Figure: BEFORE - the modal accepts `a b c d`, shows no inline validation, and enables the Create button.

Pre-fix: clicking Create produces a generic error

Figure: BEFORE - clicking Create sends the `createFolder` mutation, which the backend rejects with `Invalid path: relativePath contains invalid characters`, surfaced as the generic "please contact the developer" dialog.

Post-fix: the same name is rejected inline and Create is disabled

Figure: AFTER - the same `a b c d` is rejected inline with the allowed-characters message and the Create button is disabled, so no request is sent.

The Network panel confirmed a real `POST http://localhost:4001/graphql` round-trip on the pre-fix build and zero requests on the fixed build.

joshbermanssw (dismissed):
LGTM — clean, well-scoped fix. Nice call sharing the allowlist from `@tinacms/schema-tools` so the UI can only ever be stricter than the backend, and good coverage with the unit + Playwright tests.

One minor non-blocking nit: when the user clears the folder-name field after typing, `validationMessage` is set to `null` so the message disappears, but the input keeps its red border (`!isFolderNameValid && isInteracted`). Red border with no explanation on an empty field is slightly inconsistent — feel free to leave it, just flagging.

packages/tinacms/src/admin/pages/CollectionListPage.tsx
```diff
@@ -1510,10 +1537,8 @@ const FolderModal = ({
                 validateFolderName(event.target.value);
               }}
             />
-            {!isFolderNameValid && isInteracted && (
-              <p className='text-red-500 text-sm pl-1'>
-                Folder name is not valid – please enter a valid folder name.
-              </p>
+            {isInteracted && validationMessage && (
```
joshbermanssw, line 1540: Non-blocking nit: the message here now gates on `validationMessage`, but the input's red border a few lines up still gates on `!isFolderNameValid && isInteracted`. So when the user clears the field after typing, `validationMessage` is `null` (message hidden) yet `isFolderNameValid` is `false` (border stays red) — a red border with no explanation on an empty field.

Keying both off the same signal fixes it. On the border `className`:

```tsx
className={`mb-4 ${
  isInteracted && validationMessage ? 'border-red-500' : ''
}`}
```

Totally fine to leave as-is if you'd rather keep the empty-field border.

kulesy: Good catch, fixed in 132ef8ad7. Gated the border on the same `isInteracted && validationMessage` signal as the message, so an emptied field no longer shows an orphan red border. Verified in the admin UI that an actual invalid name still gets the red border plus the message.

joshbermanssw (approved):

PR #7145, opened 2026-07-03
fix(graphql): clear stale sort-index entries on delete (deleteIndexEntriesForPaths)

kulesy:
**TL;DR** Adds `Database.deleteIndexEntriesForPaths(paths)`, which removes every index entry for the given paths by matching on the path (not the stored value), so a caller can ensure a deleted document doesn't leave a stale field-sort entry that fails the whole collection list. Prevention complement to #7133 (read-time tolerance). **Mechanism only: this is not wired into `delete()` / `deleteContentByPaths()` and has no caller, so on its own it is inert.** Activation is a follow-up: the tinacloud content-api calling it on delete, and/or wiring it into the OSS delete path so self-hosted is covered too.

**Pain:** Field-sort index entries are keyed by `<value>\x1D<path>`, and `delete()` / `deleteContentByPaths()` only remove the *current* value's key (reconstructed from the stored record). A value change that never removed the old key strands `<oldValue>\x1D<path>`; a later delete can't target it because it only knows the current value. `Database.query` then hydrates that dangling entry, `Database.get` throws `NotFoundError`, and the whole field-sorted collection list fails (surfaced as `TinaQueryError`, "Error querying file ... from collection ..."). #7133 makes the read tolerate it; this provides the mechanism to stop it being left behind in the first place.

**Solution:** `deleteIndexEntriesForPaths(paths)` matches on the *path* instead of the value: for each collection it scans the value-keyed sort sublevels (the collection index plus each affected folder index) and deletes any key whose trailing path segment is one of the removed paths, then point-deletes the default index and the record store. So when a caller runs it for a removed document, that document leaves no `<*>\x1D<path>` orphan, stranded or current. Reference indexes are keyed under the referenced collection and left to `delete()`. It is deliberately a standalone method rather than being called from the existing delete paths, because the field-sort sweep is O(sublevel size) per sort field (value-keyed, so the path can't be seeked), so the caller decides where to run it rather than taxing every delete. Includes a test (real `Database` + `MemoryLevel` + in-memory bridge) that strands a value key and asserts every entry for the path is cleared and the field-sorted query resolves cleanly.

### DIFF
--- a/packages/@tinacms/graphql/src/database/database.delete-index-entries.test.ts
+++ b/packages/@tinacms/graphql/src/database/database.delete-index-entries.test.ts
@@ -1,0 +1,148 @@
+/**
+ * Prevention for the field-sort "ghost" that fails a collection list: a field-sort
+ * index is keyed by `<value>\x1D<path>`, and delete()/deleteContentByPaths() only
+ * remove the CURRENT value's key. A value change that never removed the old key
+ * strands `<oldValue>\x1D<path>`; a later delete can't target it, so a field-sorted
+ * query later loads its missing record and throws.
+ *
+ * deleteIndexEntriesForPaths() matches on the path instead of the value, so it clears
+ * every entry for a removed path, including stranded ones, from the collection index
+ * and its folder index. Real Database + MemoryLevel + in-memory Bridge.
+ */
+
+import { describe, expect, it } from 'vitest';
+import { MemoryLevel } from 'memory-level';
+import { buildSchema, createDatabaseInternal } from '..';
+import type { Bridge } from './bridge';
+import type { Schema } from '@tinacms/schema-tools';
+import {
+  CONTENT_ROOT_PREFIX,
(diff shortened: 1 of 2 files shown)


kulesy:
### Verification

- 2 new tests pass (real `Database` + `MemoryLevel` + in-memory bridge): a stranded `<oldValue>\x1D<path>` key plus the current entry and the record are all cleared, and the field-sorted query resolves cleanly. This branch has no read-time tolerance, so a surviving orphan would throw — the clean resolve proves the sweep is complete.
- **368** existing database + resolver tests pass, no regressions.
- Type-clean in the edited files.

Scope: sweeps the collection index and each affected folder index for the removed paths, then point-deletes the default index and the record store. Reference indexes are keyed under the referenced collection and left to `delete()`. Intended caller: the tinacloud content-api after `deleteContentByPaths(removed)` (a follow-up once this releases and the dep is bumped).

PR #7146, opened 2026-07-03
Harden media upload filename handling

Aibono1225:
https://github.com/tinacms/tinacloud/issues/3816

## What

Both media upload handlers pass the uploaded filename straight into multer's disk storage, and use a misspelled `directory` key (not a real multer option) instead of `destination`. This tightens both:

- Strip any directory component from the filename before the disk write, via a small shared `safeUploadName()` helper, so `path.join(os.tmpdir(), name)` can't point outside the temp dir.
- Replace the inert `directory` key with the correct `destination`, pointing at `os.tmpdir()` (its current effective value).

## Why

Defense in depth. This is **not exploitable as shipped** — multer's parser (busboy) already reduces the filename to its basename before our callback runs, and neither handler enables busboy's `preservePath`. The change just removes the reliance on a transitive dependency for that guarantee, so it holds even if the parser or its version changes.

## Changes

- `next-tinacms-dos`: add `safeUploadName`, wire it into disk storage, fix `destination`
- `next-tinacms-cloudinary`: same
- Shared regression tests in `tests/upload-filename.test.ts`

## Testing

`pnpm exec vitest run` — 371 passing. No behavior change for normal uploads (same temp location, same filenames); no breaking changes.


18-th (approved):
LGTM

Issue #6909, opened 2026-05-19
Forms System Rebuild - Zustand Store


18-th:
## Pain

Dirty and pristine form state is tracked inconsistently across the editor. Editors cannot always trust the unsaved-changes indicator, and different views can disagree about whether content has changed. Because there is no single source of truth, navigation warnings, save buttons and editor-state UI all have to infer form state in their own way.

## Approach

Introduce a Zustand-backed form state store that owns clean, dirty and pristine state for editor forms. The store becomes the shared source of truth used by form rendering, navigation warnings, save actions and editor-state indicators.

## Expected outcomes

- The editor has one trustworthy source of truth for whether a form is dirty or clean.
- Unsaved-change indicators are consistent across every view.
- Save and navigation behaviours can rely on the same form state instead of duplicating logic.
- Later persistence and draft-state work can build on the same store.

## Acceptance Criteria

- A Zustand store exists for form clean, dirty and pristine states.
- Save success resets the relevant form state back to clean.
- Navigation warnings and save controls read from the store rather than the local duplicate state.
- Multiple open or nested forms are tracked without overwriting each other's state.
- Tests cover dirty state, pristine state, save reset and multiple-form behaviour.

18-th:
Completed through the stack of #7150, #7151, #7152, #7153

PR #7150, opened 2026-07-06
feat(v4): form-state store + plugin store-slice registration

JackDevAU:
> **Stack 1/5** (base) · next: #7151
> `feat/v4-form-store` → `feat/v4-playground` → `feat/v4-preview-adapter` → `feat/v4-form-continuity` → `feat/v4-shadcn-ui`

## TL;DR

The v4 client-store foundation: a per-document form-state Zustand store (single source of truth for pristine/dirty/clean, ADR-010, #6909) plus boot-time plugin store-slice registration.

## Feats

- Per-document form-store keyed by `formId` — pristine/dirty/clean authority; save re-baselines to clean, untouched scopes re-seed on reload.
- Boot-time slice registration: a client segment may declare a `slice`; `composePluginSlices` mounts each at its namespace and rejects duplicate singleton providers unless one declares an override.
- `createTinaStore` composes core + plugin slices once at boot under the fixed devtools → persist stack; form-store rides devtools only (form state is volatile).
- `core/invariant` util with stable grep-able error codes, adopted at all guard sites.
- Type tightening: branded `FormValues` (keyed by `FieldAddress`), `Partial<Record<FormId, …>>` forms map, registry base-vs-base conflict no longer masked by overrides.

## Notes

Field values stay owned by react-hook-form — this store is the status/subscription surface, not a second value authority. Editor wiring is deliberately left to the next PR in the stack.

## Testing

Unit suites for form-store (incl. hooks, persist round-trip, composite values), create-store, compose-slices, and overridable-registry all pass.

## Status

Complete and nm-reviewed; ready for review once the stack order is agreed.


18-th (approved):
LGTM

PR #7152, opened 2026-07-06
feat(v4): visual-editing wire protocol — tinacms/preview entry + editor connection

JackDevAU:
> **Stack 3/5** · prev: #7151 · next: #7153
> `feat/v4-form-store` → `feat/v4-playground` → `feat/v4-preview-adapter` → `feat/v4-form-continuity` → `feat/v4-shadcn-ui`

## TL;DR

The real visual-editing wire protocol — the v4 slice of #6944. A `tinacms/preview` subpath entry gives sites `useTina` + click-to-activate; the editor side gets `usePreviewConnection`; the playground drops its prototype bridge for the public entry.

## Feats

- Protocol module: `tina:ready` / `tina:values` / `tina:activate` with payload guards before anything reaches `toFieldAddress`, plus a spreadable `tinaField` marker helper.
- Site side (`tinacms/preview`, react + protocol only — independent of the editor bundle): `useTina` is inert standalone; inside the editor iframe streamed values win and `isEditing` flips.
- Editor side: `usePreviewConnection` answers every ready with idempotent full state (covers iframe reload and ready-vs-registerForm races), reposts when the store scope's values reference changes, and activates fields on the existing rail.
- Security: incoming messages must match both origin and window source on both ends; `'*'` origins are rejected at construction (cross-origin is explicit opt-in).

## Notes

One document per connection for now — pinned in the `useTina` header. `toDocument` joins `toFormValues` in the form-store as its inverse for the wire.

## Testing

Protocol, connection, use-tina, and editor preview-connection suites pass — including form-switch streaming, early-ready recovery, and custom-allowedOrigin.

## Status

Complete and nm-reviewed.

PR #7153, opened 2026-07-06
feat(v4): form continuity — kept-edit re-adoption, error mirror, unkeyed-switch fixes

JackDevAU:
> **Stack 4/5** · prev: #7152 · next: #7154
> `feat/v4-form-store` → `feat/v4-playground` → `feat/v4-preview-adapter` → `feat/v4-form-continuity` → `feat/v4-shadcn-ui`

## TL;DR

Form continuity across document switches: kept (unsaved) edits re-adopt on mount, RHF errors mirror one-way into the store so they outlive the form's mount, and the unkeyed path-switch holes are closed.

## Feats

- Re-adopt kept edits on mount (ADR-012): navigating away and back seeds RHF from the store's kept values, with one `trigger()` to re-derive errors — but only for edited scopes, so a pristine revisit never shadows fresh content.
- Error mirror: RHF stays the error authority; the store carries a one-way mirror on the edited scope (same chokepoint pattern as values), with a clobber guard so kept errors can't be wiped even transiently.
- `useFormValues` / `useFormErrors` join the public read-hook family — collection views can badge dirty/error for any open form while another is hosted.
- Unkeyed-switch fixes: formId-scoped reset (identical-content documents still reset) and a mirror ownership guard (one form's errors can't bleed into another's scope).
- Playground document switcher with per-document dirty/error dots driven by the store's mirror.

## Notes

Also carries two follow-up refactors: `useErrorMirror` extracted from `FormProvider` (the flag/guard invariant lives in one hook), and the store's internal `FormScope` union renamed `OpenForm` to stop colliding with the context type of the same name.

## Testing

Continuity suite pins the behaviors: A→B→A round trips, save isolation, mirror persistence across teardown, error-transition recorders proving no clobber window; preview-connection suite back to zero act warnings.

## Status

Complete and nm-reviewed.


18-th (approved):
LGTM

PR #7156, opened 2026-07-06
⬆️ Bump postcss to 8.5.16

joshbermanssw:
## TL;DR

All five example apps hardcoded `postcss` at `^8.5.2` (resolving 8.5.6), ten bug-fix patches behind. This adds `postcss: ^8.5.16` to the workspace catalog and points every consumer at `catalog:`, so they stay in lockstep and future bumps are a one-line edit.

Dev/build-only dependency of private example apps — no changeset needed; the example build jobs in CI exercise it.

## Links

- Dependency audit: tinacms/tinacloud#3820


18-th (approved):
LGTM

PR #7157, opened 2026-07-06
⬆️ Bump @tailwindcss/typography to 0.5.20

joshbermanssw:
## TL;DR

One-line catalog bump, `^0.5.16` → `^0.5.20` (was resolving 0.5.19; one bug-fix release between). All seven consumers already use `catalog:`, including `@tinacms/cli`, where the plugin generates the `tina-prose` styles for the admin rich-text editor — so this ships to users and carries a patch changeset for the CLI.

## Links

- Dependency audit: tinacms/tinacloud#3820

### DIFF
--- a/.changeset/light-buses-repair.md
+++ b/.changeset/light-buses-repair.md
@@ -1,0 +1,5 @@
+---
+'@tinacms/cli': patch
+---
+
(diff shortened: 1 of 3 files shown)


18-th (approved):
LGTM

PR #7159, opened 2026-07-06
ci: sign the changesets release commit via the GitHub API

### LINKED ISSUE #7158: Release PR ("Version Packages") can't merge: changesets commit is unsigned but main requires signed commits
### The problem

`main` now enforces **required signed commits**. The automated *Version Packages* release PR created by `changesets/action` is committed with a plain local `git commit` + `git push` as `Tina Release Bot <bot@tina.io>`, which GitHub cannot sign. GitHub only auto-signs commits it creates through its own API/web UI, so the release commit is **Unverified** and the PR is blocked at the merge button with:

> Merging is blocked - Commits must have verified signatures.

This affects every release going forward (current example: #7130).

### Steps to reproduce

1. Merge any PR carrying a changeset to `main`.
2. The release workflow opens/updates the *Version Packages* PR.
3. Try to squash-merge it - GitHub blocks it because the single bot commit is unsigned.

### What I expected vs. what happened

Expected: the release PR is mergeable like every prior release. Actual: the merge button is disabled because `main`'s required-signatures rule rejects the unsigned bot commit.

### Fix

Recreate the release commit through the GitHub API (GraphQL `createCommitOnBranch`) after `changesets/action` runs, reusing the existing App token. GitHub signs API-created commits, so the head becomes Verified and the tree stays byte-identical. No new secret or bot account needed.

kulesy:
Closes #7158

**TL;DR** The automated *Version Packages* release PR can no longer merge now that `main` requires signed commits, because `changesets/action` pushes that commit with plain git and GitHub never signs it; this recreates the commit through GitHub's API so it is signed, with no new secret or bot account.

**Pain:** `main` enforces required signed commits, but `changesets/action` creates the *Version Packages* commit with a local `git commit` + `git push` as `Tina Release Bot <bot@tina.io>`, so it carries no signature and shows Unverified. GitHub only auto-signs commits made through its own API/web UI, so every release PR is now blocked at the merge button ("Commits must have verified signatures"), e.g. #7130.

**Solution:** After `changesets/action` opens or updates the release PR, a new "Sign the release pull request commit" step recreates that single commit via the GraphQL `createCommitOnBranch` mutation, reusing the existing `release-bot-allow-prs-and-push` App token; GitHub signs API-created commits, so the head becomes Verified and mergeable. It resets the branch to its base and re-lays the exact same tree copied blob-for-blob, so release contents are byte-identical and only the signature and committer change; the step is idempotent (skips an already-signed head), gated to the release-PR path (`hasChangesets == 'true' && published != 'true'`), and uses the git blob API so it keeps working as CHANGELOGs grow past the 1MB inline-content limit.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -289,6 +289,79 @@ jobs:
           GITHUB_TOKEN: ${{ steps.generate-token.outputs.token  }}
           NPM_CONFIG_PROVENANCE: true
 
+      # changesets/action pushes "Version Packages" with plain git, which GitHub can't
+      # sign, so main's required-signatures rule blocks the release PR. Recreate that
+      # commit through the API, which GitHub signs, leaving the tree byte-identical.
+      - name: Sign the release pull request commit
+        if: steps.changesets.outputs.hasChangesets == 'true' && steps.changesets.outputs.published != 'true'
+        uses: actions/github-script@3a2844b7e9c422d3c10d287c895573f7108da1b3 # v9.0.0
+        with:
+          github-token: ${{ steps.generate-token.outputs.token }}
+          script: |
+            const branchName = `changeset-release/${context.ref.replace('refs/heads/', '')}`;
+            const { owner, repo } = context.repo;
+
+            const { data: ref } = await github.rest.git.getRef({ owner, repo, ref: `heads/${branchName}` });
+            const headSha = ref.object.sha;
+
+            const { data: head } = await github.rest.repos.getCommit({ owner, repo, ref: headSha });
+            if (head.commit.verification?.verified) {
+              core.info('Release commit is already signed; nothing to do.');
+              return;
+            }
+            if (head.parents.length !== 1) {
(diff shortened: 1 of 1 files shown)


kulesy:
Notes for review:

- **Rollout / #7130** - merging this is itself a push to `main`, which re-runs the release workflow and regenerates the open *Version Packages* PR (#7130) branch through the new step, so #7130 heals automatically on the same run - no manual step needed.
- **Testing status** - the step can't run in isolation, so I validated the workflow YAML and the embedded `github-script` JS syntactically and walked the API sequence by hand against #7130's commit (re-laying its tree reproduces a byte-identical tree). It runs live for the first time on the next release.
- **Alternative considered** - signing in CI with a dedicated bot account's key is fewer moving parts in the step, but needs a new machine account + a stored private-key secret; the API approach was chosen to avoid standing that up.

kulesy:
Superseded by #7160, which does this natively in one line via changesets/action's `commitMode: github-api` (server-side GitHub signing) instead of a custom post-step that rewrites the branch. Closing in favour of that, cleaner and less brittle.

PR #7161, opened 2026-07-06
Fix folder collection auto-open

### LINKED ISSUE #7148: Folder-based collection with create:false + delete:false is unnavigable — single-document auto-open fires inside folder views
### The exact error message

```shell
No error is thrown — the admin performs a silent history.replaceState back into the document editor. Console and network are clean; no GraphQL request fires. Observable symptom: entering a folder in the collection immediately redirects back into the single document's editor.
```

### Steps to reproduce

1. Define a create/delete-locked, folder-based collection:

    {
      name: 'gallery',
      path: 'src/content/gallery',
      format: 'md',
      match: { include: '*/index' },
      ui: { allowedActions: { create: false, delete: false } },
      fields: [ /* ... */ ],
    }

2. Create at least two album folders, each containing exactly one index.md (so the collection ROOT lists 2+ folders).
3. Open /admin, open one album document (…/collections/gallery/<folder>/index), then click the "← Gallery" collection breadcrumb (the back-to-collection breadcrumb added in #6942 / PR #6939).
4. A brief "Please wait, Tina is loading data…" appears, then the app history.replaceState's straight back into the same document editor (~1s later). The folder list never stays, and the folder browser is unreachable.

The left sidebar collection link works, because the collection ROOT resolves to >1 document; only folder views (exactly one */index doc each) bounce.

Expected: clicking the collection breadcrumb returns to the folder/album list, and folder views stay navigable.

Actual: every folder view immediately redirects back into its single document editor, so folder navigation is impossible for a create/delete-locked folder collection — the newly added breadcrumb is effectively dead for this collection type.

Root cause (traced in source): GetCollection's singleton auto-open effect (dist/index.js ~L74221 in 3.9.4) calls navigate(..., { replace: true }) into the sole document whenever a collection has allowedActions.create === false AND delete === false AND the list resolves to exactly one non-folder document. The condition does not exclude folder views, so in a folder-based collection every */index folder (which contains exactly one document) satisfies it and bounces. It is purely client-side — no GraphQL request is involved.

### Your environment

- tinacms 3.9.4 (latest), @tinacms/cli ^2.5.2
- Framework: Astro 6, static output; admin served at /admin
- Non-default: hybrid MediaStore (Cloudflare R2 via a small Worker) and Tina Cloud auth — neither is involved here (the repro is pure client-side navigation, no GraphQL request)
- Reproduces on both the Tina Cloud production admin and local `tinacms dev`

### A way for us to reproduce

_No response_

```typescript
// tina/config.ts — the affected collection (create/delete-locked + folder-based)
const gallery = {
  name: 'gallery',
  label: 'Gallery',
  path: 'src/content/gallery',
  format: 'md',
  match: { include: '*/index' },   // each album is a folder containing one index.md
  ui: {
    allowedActions: { create: false, delete: false },  // albums managed outside Tina
  },
  fields: [ /* title, date, cover, photos[], ... */ ],
};
```

### Client ID

dcbeb84c-65dc-4d6e-8dad-03ff2cfd7542

### Validations

ahfoysal:
Fixes #7148.

This keeps the single-document auto-open behavior at the collection root, but leaves folder views open so nested items can still be selected.

Checked with:
- pnpm exec vitest run src/admin/components/GetCollection.test.ts
- pnpm exec biome check packages/tinacms/src/admin/components/GetCollection.tsx packages/tinacms/src/admin/components/GetCollection.utils.ts packages/tinacms/src/admin/components/GetCollection.test.ts

### DIFF
--- a/.changeset/fuzzy-cameras-wait.md
+++ b/.changeset/fuzzy-cameras-wait.md
@@ -1,0 +1,5 @@
+---
+"tinacms": patch
+---
+
+Keep folder collection views open when they only contain one document.
--- a/packages/tinacms/src/admin/components/GetCollection.test.ts
+++ b/packages/tinacms/src/admin/components/GetCollection.test.ts
@@ -1,0 +1,46 @@
+import { describe, expect, it } from 'vitest';
+
+import type { CollectionResponse } from '../types';
+import { shouldAutoOpenCollectionDocument } from './GetCollection.utils';
+
+const collectionWithNode = (node: { __typename: string }) =>
+  ({
+    documents: {
+      edges: [{ node }],
+    },
+  }) as CollectionResponse;
+
+describe('shouldAutoOpenCollectionDocument', () => {
+  it('auto-opens a single document at the collection root when create and delete are disabled', () => {
+    expect(
+      shouldAutoOpenCollectionDocument({
+        allowCreate: false,
+        allowDelete: false,
+        collection: collectionWithNode({ __typename: 'Document' }),
+        folder: { fullyQualifiedName: '~' },
+      })
+    ).toBe(true);
+  });
+
+  it('does not auto-open a single document inside a folder view', () => {
+    expect(
+      shouldAutoOpenCollectionDocument({
+        allowCreate: false,
+        allowDelete: false,
+        collection: collectionWithNode({ __typename: 'Document' }),
+        folder: { fullyQualifiedName: '~/album-one' },
+      })
+    ).toBe(false);
+  });
+
+  it('does not auto-open folders', () => {
+    expect(
+      shouldAutoOpenCollectionDocument({
+        allowCreate: false,
+        allowDelete: false,
+        collection: collectionWithNode({ __typename: 'Folder' }),
+        folder: { fullyQualifiedName: '~' },
+      })
+    ).toBe(false);
+  });
+});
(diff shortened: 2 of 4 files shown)


joshbermanssw:
Thanks for the fix! One thing needed before this can ship: a **changeset**. Without one the PR will merge but the fix never gets released to npm.

Can you add a `.changeset/*.md` marking `tinacms` as a `patch`? The changeset-bot comment above has a one-click link that pre-fills it for you.

ahfoysal:
Added the patch changeset.

joshbermanssw (approved):
Manually QA'd this against the branch base (`ebafd8f`), building at the root and driving the kitchen-sink admin.

Reproduced the bug on the baseline with a `create:false`/`delete:false` folder collection: opening a folder view bounced straight into the doc editor (`gallery/~/album-one` → `edit/gallery/album-one/index`). On this branch the folder view stays put. I also added a root-level locked singleton collection to check the auto-open feature this guard protects, and that still fires on fresh load and on normal cross-collection navigation. Unit tests pass.

Nice fix, and thanks for pulling the condition out into a tested helper.

Non-blocking nit for a follow-up: the effect now reads `folder.fullyQualifiedName`, but its dependency array is still `[collection?.name || '', loading]`, so `folder` can be stale when the effect fires. I hit this navigating from a folder view to a locked singleton collection via the sidebar; the auto-open silently doesn't fire and you land on a one-row list instead of the document. It's cosmetic and only reachable on a path the old code couldn't get to, so it's fine to leave, but adding `folder.fullyQualifiedName` to the deps would close it.

PR #7166, opened 2026-07-07
feat(@tinacms/astro): forward containerOptions to island route container

dfan1028:
## Problem

`experimental_createIslandRoute` renders each island through a fresh `AstroContainer` created with `AstroContainer.create()` — no renderers. Unlike Astro's page pipeline, the container does **not** inherit renderers from the project's Astro config, so any island that renders a **UI-framework component** (React, Vue, Svelte, …) fails to render during visual editing. The island-refresh request 500s and the editor silently falls back to form-only editing (no live preview).

This doesn't affect islands that render plain Astro/markdown, which is why it hasn't surfaced in the `examples/astro/visual-editing` app — but it reliably bites projects that put framework components inside visually-edited regions (in our case, a React component library rendered in TinaMarkdown blocks and an interactive search island).

## Fix

Add an optional `containerOptions` argument to `experimental_createIslandRoute`, forwarded to `AstroContainer.create`. Callers register renderers explicitly:

```ts
import reactRenderer from '@astrojs/react/server.js';

export const ALL = experimental_createIslandRoute(islands, {
  containerOptions: { renderers: [{ name: '@astrojs/react', ssr: reactRenderer }] },
});
```

Passing renderers explicitly (rather than auto-loading via `astro:container`'s `loadRenderers`) keeps this working in a plain Node/SSR context — the `astro:container` virtual module only resolves inside Vite.

The argument is optional, so existing callers are unchanged.

## Changes

- `island-route.ts`: `IslandRouteOptions` type; forward `options?.containerOptions` to `AstroContainer.create`.
- `experimental.ts`: export `IslandRouteOptions`.
- Test: assert `containerOptions` is forwarded to `AstroContainer.create`.
- Changeset (minor).

## Validation

Verified in a downstream Astro + React project: with this change the island route renders React islands in visual editing; without it the same islands 500. I did not run the full monorepo suite locally — deferring to CI — but the added unit test follows the existing `island-route.test.ts` patterns.

## Notes

Happy to adjust the API shape (e.g. accept `renderers` directly instead of the nested `containerOptions`) or add auto-loading behind a flag if preferred.

Made with [Cursor](https://cursor.com)

### DIFF
--- a/packages/@tinacms/astro/src/__tests__/island-route.test.ts
+++ b/packages/@tinacms/astro/src/__tests__/island-route.test.ts
@@ -1,6 +1,7 @@
 import { PREVIEW_CONTENT_TYPE, PRIME_HEADER } from '@tinacms/bridge/preview';
 import type { APIContext } from 'astro';
-import { describe, expect, it } from 'vitest';
+import { experimental_AstroContainer as AstroContainer } from 'astro/container';
+import { describe, expect, it, vi } from 'vitest';
 import { requestWithMetadata } from '../data';
 import { experimental_createIslandRoute } from '../island-route';
 import IslandStub from './IslandStub.astro';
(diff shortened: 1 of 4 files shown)


packages/@tinacms/astro/src/__tests__/island-route.test.ts
```diff
@@ -93,6 +94,46 @@ describe('experimental_createIslandRoute', () => {
     expect(globalHtml).not.toContain('data-tina-primary');
   });
 
+  it('forwards containerOptions to AstroContainer.create', async () => {
+    const spy = vi.spyOn(AstroContainer, 'create');
```
joshbermanssw, line 98: Nit: consider restoring this spy via `try`/`finally` or an `afterEach` cleanup so a failure before `mockRestore()` does not leave `AstroContainer.create` mocked for later tests. Not a blocker.

PR #7167, opened 2026-07-07
fix: omit empty optional image fields

### LINKED ISSUE #5229: Bug - optional image field still adds frontmatter key

Hi, 

When you have an optional image field and you don't give assign media to it (so you leave it blank) it still adds the name of the field as key to the frontmatter, with an empty string as value. This is inconsistent (and odd) behaviour, as all other field types remove the key from the frontmatter when it doesn't have a value.

### Reproduction

-

### Steps to reproduce

1. Simply add an image field (e.g. coverImage) to your collection with `required: false`
2. During creation/editing don't assign an image to the field.
3. Submit

Now if you look at the frontmatter of the file, it still contains the key (e.g. coverImage) with a value of `' '`.

### System Info

Latest version of tina cms

### Validations

ahfoysal:
Fixes #5229.

Omit empty optional image fields from document payloads so blank image fields are not written to frontmatter.

Tests:
- pnpm --filter @tinacms/graphql exec vitest run src/resolver/index.test.ts --coverage.enabled
- pnpm exec biome check packages/@tinacms/graphql/src/resolver/index.ts packages/@tinacms/graphql/src/resolver/index.test.ts
- pnpm turbo run build --filter=@tinacms/graphql

### DIFF
--- a/packages/@tinacms/graphql/src/resolver/index.ts
+++ b/packages/@tinacms/graphql/src/resolver/index.ts
@@ -38,6 +38,11 @@ import {
   resolveMediaRelativeToCloud,
 } from './media-utils';
 
+const isEmptyOptionalImageValue = (
+  value: unknown,
+  field: Pick<TinaField<true>, 'type' | 'required' | 'list'>
+) => field.type === 'image' && !field.required && !field.list && !value;
+
 interface ResolverConfig {
   config?: GraphQLConfig;
   database: Database;
@@ -90,6 +95,9 @@ export const resolveFieldData = async (
       };
       break;
     case 'image':
+      if (isEmptyOptionalImageValue(value, field)) {
+        break;
+      }
       accumulator[field.name] = resolveMediaRelativeToCloud(
         value as string,
         config,
@@ -1553,6 +1561,9 @@ export class Resolver {
       const field = template.fields.find((field) => field.name === fieldName);
       if (!field) {
         throw new Error(`Expected to find field by name ${fieldName}`);
+      }
+      if (isEmptyOptionalImageValue(fieldValue, field)) {
+        continue;
       }
       switch (field.type) {
         case 'displayOnly':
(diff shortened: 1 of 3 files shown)


packages/@tinacms/graphql/src/resolver/index.ts
```diff
@@ -1554,6 +1562,9 @@ export class Resolver {
       if (!field) {
         throw new Error(`Expected to find field by name ${fieldName}`);
       }
+      if (isEmptyOptionalImageValue(fieldValue, field)) {
+        continue;
```
joshbermanssw, line 1572: Requesting changes here: this handles a blank optional image by omitting the field from the mutation values, but update persistence later merges those values back onto the existing document (`{ ...legacyValues, ...params }` in `resolveUpdateDocument()` and `{ ...oldDoc, ...values }` in the deprecated update path). That means an existing `coverImage: "foo.png"` followed by an editor submit of `coverImage: ""` can produce empty mutation values for this field and preserve the old image instead of clearing it from frontmatter. Please add coverage for clearing an existing optional image on update and make the explicit clear remove the key from the merged payload, while keeping truly omitted fields as unchanged.

joshbermanssw (changes requested):
Requesting changes for the update clear case.

ahfoysal:
Fixed the update clear case. Empty optional image values are still omitted on create, but now clear an existing image on update. Added a resolver regression test for that path.

packages/@tinacms/graphql/src/resolver/index.ts
```diff
@@ -90,6 +95,9 @@ export const resolveFieldData = async (
       };
       break;
     case 'image':
+      if (isEmptyOptionalImageValue(value, field)) {
```
joshbermanssw, line 98: Question on this one: why change the read path too?

The frontmatter bug only needs the write path. This makes an existing document with `avatar: ''` return `null` instead of `""` over GraphQL, which is an API-visible change for existing content shipping as a patch.

If it's load-bearing for something I've missed, could you note it in the changeset? If not, dropping it would make the PR strictly safer.

PR #7171, opened 2026-07-07
chore: tidy bug report template and strip empty issue sections

kulesy:
**TL;DR** Tidy the Bug Report issue template, give reporters a real description field, trim what's required, and stop omitted optional fields from cluttering filed issues.

**Pain:** the intro heading leaned on internal framing (our Discord `#ask-for-help` post guidelines and "AI triage tooling") that does not belong on a public reporting form; there was no place to describe the problem, so reporters used "The exact error message" as a catch-all description; the form asked for too much up front; the "Relevant sections of your schema file" field mostly added noise; and every omitted optional field still renders as a `### <label>` / `_No response_` block on the filed issue with no template-level way to suppress it.

**Solution:** adds a required "Describe the bug" field at the top and makes it and "Your environment" the only required fields (the exact error and steps to reproduce become optional, and "What you expected vs. what actually happened" is removed). Also trims the intro, drops the schema-file section, and adds a workflow that on issue open strips any `_No response_` section from bug-report submissions. The workflow is scoped to the form's auto-applied `bug` label so it can only touch form submissions, fenced code blocks are respected so a `### ` line inside a pasted error is not mistaken for a heading, and required fields plus issues with nothing to strip are left untouched.


18-th (approved):
LGTM

PR #7172, opened 2026-07-07
ci: strip omitted optional sections from filed issues

kulesy:
**TL;DR** Auto-remove `_No response_` sections from an issue body when it's opened.

**Pain:** GitHub issue forms render every omitted optional field as a `### <label>` heading followed by `_No response_`. On the Bug Report form that leaves empty "A way for us to reproduce" and "Client ID" sections cluttering most filed issues, and there is no template-level switch to suppress them.

**Solution:** adds a workflow that runs on issue open, strips any section whose only content is `_No response_` (fenced code blocks are respected so a `### ` line inside a pasted error is never mistaken for a heading), and rewrites the body only when something was removed. Required fields are untouched, and issues with nothing to strip are left as-is.

### DIFF
--- a/.github/workflows/strip-empty-issue-sections.yml
+++ b/.github/workflows/strip-empty-issue-sections.yml
@@ -1,0 +1,58 @@
+name: Strip empty issue sections
+
+on:
+  issues:
+    types: [opened]
+
+permissions:
+  issues: write
+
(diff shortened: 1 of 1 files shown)


kulesy:
Folded into #7171.

PR #7176, opened 2026-07-09
🐛 Improve CLI error when tinacms/@tinacms packages can't be resolved

### LINKED ISSUE #4730: 🐛 CLI - Lack of useful error message upon 'Could not resolve "tinacms"'
CC @wicksipedia @JackDevAU @ozairashfaqueSSW 

There have been several occurrences where time has been wasted trying to track down why the "tinacms" could not be resolved when attempting to build the Tina config. The typical cause is that artifacts such as `package.json`, `node_modules/` or a malformed `.pnp.cjs` exist in one of the parent directories of the source tree, causing the resolution of the 'tinacms' package to fail.

The error message produced by `tinacms build` does not help the user find the cause.

## Tasks

- [ ] Improve the error messages produced by [`ConfigureManager`](https://github.com/tinacms/tinacms/blob/main/packages/%40tinacms/cli/src/next/config-manager.ts) so that the user is aware that a failure to find the "tinacms" package may be due to artifacts further up the directory tree.

Image
**Figure: An example of the output seen by a user when TinaCMS packages could not be resolved due to conflicting package manager data further up the directory path**

joshbermanssw:
## TL;DR

When a stray `package.json`, `node_modules`, or `.pnp.cjs` in a *parent* directory breaks module resolution, `tinacms build`/`dev` failed with esbuild's bare `Could not resolve "tinacms"` and no hint at the cause (see the screenshot in #4730). The CLI now catches that failure and prints the package that failed, the directory Tina searched from, and the parent-directory artifacts to check.

## How

- New `formatConfigBuildError()` detects `Could not resolve "<pkg>"` for `tinacms` and any `@tinacms/*` package, and rewrites it with actionable guidance.
- Routes both the **config** and **database** esbuild builds through it (the database build resolves `@tinacms/datalayer`, which hits the same parent-dir problem).
- Anchors esbuild package resolution at the project root via `absWorkingDir`. Note: this is a partial mitigation, not a cure - a `.pnp.cjs` above the project root can still hijack resolution, which is exactly why the improved message matters.

## Links

- Closes #4730

### DIFF
--- a/packages/@tinacms/cli/src/next/config-build-error.ts
+++ b/packages/@tinacms/cli/src/next/config-build-error.ts
@@ -1,0 +1,75 @@
+type EsbuildErrorLike = {
+  message?: string;
+  errors?: Array<{
+    text?: string;
+  }>;
+};
+
+const isObject = (value: unknown): value is Record<string, unknown> =>
+  typeof value === 'object' && value !== null;
+
+// A package specifier belongs to TinaCMS if it is the `tinacms` package itself
+// or one of the scoped `@tinacms/*` packages (e.g. @tinacms/schema-tools,
+// @tinacms/datalayer). Lookalikes such as `my-tinacms-plugin` are excluded.
+const isTinaPackage = (specifier: string) =>
+  specifier === 'tinacms' || specifier.startsWith('@tinacms/');
+
+// Pull every `Could not resolve "<specifier>"` target out of an esbuild error
+// string. esbuild surfaces this both as structured `errors[].text` entries and
+// as the flattened `Error.message`, so callers feed us both.
+const collectResolveTargets = (text: string | undefined): string[] => {
+  if (!text) {
+    return [];
+  }
+  const targets: string[] = [];
+  const pattern = /Could not resolve "([^"]+)"/g;
+  for (const match of text.matchAll(pattern)) {
+    targets.push(match[1]);
+  }
+  return targets;
+};
+
+// Returns the first unresolved TinaCMS package specifier found in the error, or
+// null if the error isn't a TinaCMS resolution failure.
+const getUnresolvedTinaPackage = (error: unknown): string | null => {
+  if (!isObject(error)) {
+    return null;
+  }
+
+  const esbuildError = error as EsbuildErrorLike;
+  const targets = [
+    ...(esbuildError.errors ?? []).flatMap((item) =>
+      collectResolveTargets(item.text)
+    ),
+    ...collectResolveTargets(esbuildError.message),
+  ];
+
+  return targets.find(isTinaPackage) ?? null;
+};
+
+export const isTinacmsResolveError = (error: unknown): boolean =>
+  getUnresolvedTinaPackage(error) !== null;
+
+export const formatConfigBuildError = ({
+  error,
+  rootPath,
+}: {
+  error: unknown;
+  rootPath: string;
+}) => {
+  const unresolvedPackage = getUnresolvedTinaPackage(error);
+  if (!unresolvedPackage) {
+    return error;
+  }
+
+  return new Error(
+    [
+      `Unable to resolve the "${unresolvedPackage}" package while building your Tina config.`,
+      '',
+      `Tina looked from: ${rootPath}`,
+      '',
+      'Make sure the TinaCMS packages are installed in this project and that you are running the CLI from the project root.',
+      'If they are installed, check parent directories for package-manager files that can interfere with module resolution, such as package.json, node_modules, yarn.lock, or .pnp.cjs.',
+    ].join('\n')
+  );
+};
--- a/packages/@tinacms/cli/src/next/config-build-error.ts
+++ b/packages/@tinacms/cli/src/next/config-build-error.ts
@@ -1,8 +1,16 @@
+type EsbuildMessageLike = {
+  text?: string;
+  location?: {
+    file?: string;
+    line?: number;
+    column?: number;
+    lineText?: string;
+  } | null;
+};
+
 type EsbuildErrorLike = {
   message?: string;
-  errors?: Array<{
-    text?: string;
-  }>;
+  errors?: EsbuildMessageLike[];
 };
 
 const isObject = (value: unknown): value is Record<string, unknown> =>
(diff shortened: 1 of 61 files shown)


packages/@tinacms/cli/src/next/config-manager.ts
```diff
@@ -429,21 +438,28 @@ export class ConfigManager {
...
+        target: ['esnext'],
+        platform: 'browser',
+        format: 'esm',
+        logLevel: 'silent',
+        packages: 'external',
+        ignoreAnnotations: true,
+        outfile: preBuildConfigPath,
+        loader: loaders,
+        metafile: true,
+        absWorkingDir: esbuildRootPath,
```
kulesy: **Major:** This `absWorkingDir` addition is out of scope for #4730 (which only asks for a better error message) and it introduces a dev-server regression. The metafile inputs feed `watchList`, which goes straight to `chokidar.watch()`; esbuild reports those paths relative to `absWorkingDir` while chokidar resolves them relative to `process.cwd()`, so `tinacms dev --rootPath <dir other than the invocation cwd>` (for example `--rootPath packages/site` from a monorepo root) watches the wrong paths and config live-reload silently stops. The common case (cwd equals rootPath) is unaffected, and I could not find a case where `absWorkingDir` actually turns a failing resolve into a passing one, so I'd recommend dropping the `absWorkingDir`/`esbuildRootPath` changes and shipping just the error-message improvement. If you keep them, make the watch list cwd-independent with `path.resolve(esbuildRootPath, key)`.

packages/@tinacms/cli/src/next/config-build-error.ts
```diff
@@ -0,0 +1,75 @@
...
+    ...(esbuildError.errors ?? []).flatMap((item) =>
+      collectResolveTargets(item.text)
+    ),
+    ...collectResolveTargets(esbuildError.message),
+  ];
+
+  return targets.find(isTinaPackage) ?? null;
+};
+
+export const isTinacmsResolveError = (error: unknown): boolean =>
```
kulesy: **Minor:** `isTinacmsResolveError` is exported and unit-tested but never called in production (only `formatConfigBuildError` is wired into the catch sites), so it is dead public surface. Either drop it and its test block, or wire it in where a boolean check reads better than the rewrite.

packages/@tinacms/cli/src/next/config-build-error.ts
```diff
@@ -0,0 +1,75 @@
...
+}: {
+  error: unknown;
+  rootPath: string;
+}) => {
+  const unresolvedPackage = getUnresolvedTinaPackage(error);
+  if (!unresolvedPackage) {
+    return error;
+  }
+
+  return new Error(
```
kulesy: **Minor:** Returning a fresh `new Error(...)` discards esbuild's `location`, which carries the failing file, line, column, and the offending import line (I captured `{file:'tina/config.ts', line:1, column:29, lineText:'import { defineConfig } from "tinacms"'}`). Consider appending that one-line location to the guidance so users keep the precise "which import, which file" detail alongside the parent-directory hint.

packages/@tinacms/cli/src/next/config-build-error.ts
```diff
@@ -0,0 +1,75 @@
...
+
+// Pull every `Could not resolve "<specifier>"` target out of an esbuild error
+// string. esbuild surfaces this both as structured `errors[].text` entries and
+// as the flattened `Error.message`, so callers feed us both.
+const collectResolveTargets = (text: string | undefined): string[] => {
+  if (!text) {
+    return [];
+  }
+  const targets: string[] = [];
+  const pattern = /Could not resolve "([^"]+)"/g;
```
kulesy, line 33: **Minor:** Detection keys on esbuild's exact English string `Could not resolve "..."`. That is deterministic against the pinned esbuild but brittle across versions, and it fails silently: a future esbuild reword would just revert users to the raw error. The unit tests feed hardcoded strings, so they would stay green through such a reword; an integration test that runs a real failing `esbuild.build`, catches it, and asserts `formatConfigBuildError` rewrites it would turn that drift into a loud CI failure at the next esbuild bump. A fully decoupled alternative for a later robustness pass is an `onResolve` filter on the `tinacms`/`@tinacms/*` specifier plus `build.resolve().errors.length`, which uses esbuild's own resolver instead of its prose.

packages/@tinacms/cli/src/next/config-manager.ts
joshbermanssw: Dropped the `absWorkingDir`/`esbuildRootPath` changes entirely in 2a8e525. As you noted, they never turned a failing resolve into a passing one (esbuild resolves the bare specifier from the importing file, walking up, so `absWorkingDir` had no effect there), and they made esbuild emit metafile inputs relative to `rootPath` while `chokidar.watch()` resolves the watch list relative to `cwd`, breaking `--rootPath <dir != cwd>` live-reload. Only the error-message wiring remains.

packages/@tinacms/cli/src/next/config-build-error.ts
joshbermanssw: Removed `isTinacmsResolveError` and its `describe` block in 2a8e525. Only `formatConfigBuildError` is wired into the catch sites, so the boolean was dead public surface.

packages/@tinacms/cli/src/next/config-build-error.ts
joshbermanssw: `formatConfigBuildError` now pulls the `location` off the failing esbuild entry and appends `Failing import: <file>:<line>:<column>` plus the offending source line when present (2a8e525), so the precise "which import, which file" detail survives the rewrite alongside the parent-directory hint.

packages/@tinacms/cli/src/next/config-build-error.ts
joshbermanssw: Added an integration test in 2a8e525 that runs a real `esbuild.build` against a bare `@tinacms/*` specifier that cannot resolve, then asserts `formatConfigBuildError` still rewrites the thrown error. An esbuild reword of the `Could not resolve "..."` string now fails CI instead of silently reverting users to the raw error. Left the `onResolve` resolver-based detection for the later robustness pass you flagged, since it is a larger change.

kulesy (approved):
Reviewed. All review feedback is addressed in 2a8e525: the out-of-scope `absWorkingDir` change (and its `dev --rootPath` watch regression) is dropped, esbuild's `location` is now preserved in the guidance message, the dead `isTinacmsResolveError` export is pruned, and a new integration test runs a real `esbuild.build` so the `Could not resolve` string detection fails loudly if esbuild ever rewords it. Verified locally: full `src/next` suite green (163/163), including the new tests, and the changed files typecheck clean. Cleanly satisfies #4730. LGTM.

PR #7178, opened 2026-07-10
Pin npm to 11 in the Publish workflow

18-th:
* Latest right now is npm 12, using that requires bumping engines for tina, pinning to 11 instead

Triggered by https://github.com/tinacms/tinacms/actions/runs/29057020703/job/86250518374


joshbermanssw (dismissed):

wicksipedia:
instead of pinning in the actions, can we instead reference a field in the package.json?
that way we can avoid drift

joshbermanssw (approved):

18-th:
@wicksipedia Good idea, done that

Issue #7185, opened 2026-07-10
Rename tina-barebones-starter to tina-react-starter and update all references


joshbermanssw:
## Goal

Rename the starter repo `tinacms/tina-barebones-starter` to `tinacms/tina-react-starter` and update every reference across the TinaCMS org so nothing points at the old name.

Context: the starter is a minimal React/Next.js starter, so "React Starter" also disambiguates it from the dedicated `tina-nextjs-starter`. GitHub keeps an automatic redirect from the old repo slug, so existing links will not 404 the moment we rename; the tasks below still update references so the code, docs, and analytics stay accurate.

## Acceptance criteria

- [ ] The GitHub repo is renamed and its own identity files are updated.
- [ ] `create-tina-app`, the TinaCloud quickstart, and the tina-dashboard analytics all resolve to the new name.
- [ ] Live docs and community lists point at the new repo.
- [ ] Each external-identifier decision (telemetry keys, feature flag, Checkly project) is made deliberately rather than by accident.

### 1. The starter repo (`tinacms/tina-barebones-starter`)

- [ ] Rename the GitHub repository to `tina-react-starter` (repo admin action; GitHub adds a redirect from the old slug automatically).
- [ ] Verify the `use-dynamodb-store` branch survives the rename (TinaCloud links to it; see Task 4).
- [ ] `package.json`: the `name` field is `"starter-basic"`. Update it to `"tina-react-starter"` (or `"starter-react"` to keep the `starter-*` pattern).
- [ ] `content/page/home.mdx` (line 1): "Welcome to the barebones starter for TinaCMS!" copy update.
- [ ] Optional: add a project title to `README.md` naming it the React starter.

### 2. `create-tina-app` (in `tinacms/tinacms`)

- [ ] `packages/create-tina-app/src/templates.ts` (line ~191): update `gitURL` to `https://github.com/tinacms/tina-react-starter`.
- [ ] `packages/create-tina-app/src/templates.ts` (line ~172): rename the display `title: 'Bare bones starter'` (e.g. to `'React Starter'`).
- [ ] Decide on the template `value` (line ~175, currently `'basic'`). It is the `--template basic` CLI flag value and the PostHog telemetry key. Changing it breaks anyone scripting `create-tina-app --template basic` and splits telemetry history. Recommendation: keep `'basic'`, or add the new value as an alias. If it is changed, also update the JSDoc examples in `packages/create-tina-app/src/util/posthog.ts` (lines ~256, ~317).

[...]

PR #7186, opened 2026-07-10
🧹 Reference renamed tina-react-starter in create-tina-app and DoD

joshbermanssw:
## TL;DR

Part of the `tina-barebones-starter` → `tina-react-starter` rename (#7185). Updates `create-tina-app` and the Definition of Done to reference the renamed repo, and renames the CLI template's display title to "React Starter".

## Reviewer notes

- **Template `value: 'basic'` kept.** It is the `--template basic` CLI flag value and the PostHog telemetry key; renaming it would break existing scripts and split analytics history. Only the display title and git URL change.

## Note

Merge after the GitHub repo rename lands (GitHub keeps a redirect from the old slug afterward).

## Links

- Tracking PBI: #7185

### DIFF
--- a/packages/create-tina-app/src/templates.ts
+++ b/packages/create-tina-app/src/templates.ts
@@ -171,7 +171,7 @@ export const TEMPLATES: Template[] = [
   {
     title: 'React Starter',
     description:
-      'Stripped down to essentials, this starter is the canvas for pure, unadulterated code creativity. Built with Next.js.',
+      'Stripped down to essentials, this starter is the canvas for pure, unadulterated code creativity. Built with Vite and React.',
     value: 'basic',
     isInternal: false,
     features: [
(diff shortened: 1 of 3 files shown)


18-th (dismissed):
LGTM

packages/create-tina-app/src/templates.ts
```diff
@@ -169,7 +169,7 @@ export const TEMPLATES: Template[] = [
     devUrl: 'http://localhost:3000',
   },
   {
-    title: 'Bare bones starter',
+    title: 'React Starter',
     description:
       'Stripped down to essentials, this starter is the canvas for pure, unadulterated code creativity. Built with Next.js.',
```
wicksipedia: I think this description should be updated to mention that it is using Vite and whatever else is in the mods - https://github.com/tinacms/tina-barebones-starter/pull/341

JackDevAU (approved):
:shipit:

PR #7192, opened 2026-07-10
fix(tinacms): spin the create-branch save button during pre-flight

### LINKED ISSUE #7191: Editorial workflow: create-branch save button is unresponsive for ~1s before the progress modal
### The exact error message
No error is thrown. This is a UI responsiveness issue.

### Steps to reproduce
1. Using the TinaCloud editorial workflow on a protected branch, edit a document and click **Save**.
2. In the "Save changes to new branch" modal, click the save button (e.g. **Save draft** / **Save to a new branch**).

Expected: the click gives immediate feedback.
Actual: the button sits unresponsive for about a second before the progress modal appears. The click runs two sequential branch-existence checks (`checkBaseBranchExists`, then `checkTargetBranchExists` inside `executeWorkflow`) before `isExecuting` flips and the progress modal renders, and during that window the button is only `disabled` with no spinner, so it reads as frozen.

### Your environment
- `tinacms` 3.10.0 (`main`)
- Editorial workflow / protected branch (TinaCloud), Next.js starter

kulesy:
Closes #7191

**TL;DR** Clicking the create-branch save button in the editorial workflow left it unresponsive for about a second before the progress modal appeared; this keeps a spinner on the button through the whole pre-flight so the click feels immediate.

**Pain:** On a protected branch, saving opens the create-branch modal, and clicking the save button runs two sequential branch-existence checks (`checkBaseBranchExists`, then `checkTargetBranchExists` inside `executeWorkflow`) before `isExecuting` flips and the progress modal renders. During that window the button is only `disabled` with no feedback, and the checking flag was cleared before the target check, so it reads as a frozen one second gap.

**Solution:** Keep the submit button in a `busy` spinner state for the entire async span (guard checks plus workflow start) via a renamed `isSubmitting` flag and a new `busy` prop on the shared `CreateBranchPromptModal` that swaps the button icon for a spinner. Both the content flow (`isSubmitting`) and the media-store flow (`isChecking`) feed it, and a validation failure during pre-flight drops the spinner back to the prompt with the error instead of showing a premature "Creating branch" state. Adds a regression test covering the spinner behavior.

### DIFF
--- a/.changeset/spin-create-branch-save-button.md
+++ b/.changeset/spin-create-branch-save-button.md
@@ -1,0 +1,5 @@
+---
+"tinacms": patch
+---
+
+Show a spinner on the editorial-workflow "Save to a new branch" button while the pre-flight branch checks run, so the click gives immediate feedback instead of a brief unresponsive gap before the progress modal appears.
--- a/packages/tinacms/src/toolkit/components/media/media-workflow-overlay.tsx
+++ b/packages/tinacms/src/toolkit/components/media/media-workflow-overlay.tsx
@@ -1,5 +1,12 @@
-import * as React from 'react';
-import { BiError } from 'react-icons/bi';
+import { CreateBranchPromptModal } from '@toolkit/form-builder/create-branch-modal';
+import { EditorialWorkflowProgressModal } from '@toolkit/form-builder/editorial-workflow-progress-modal';
+import {
+  type MediaWorkflowConfirmBranchEvent,
+  TARGET_BRANCH_EXISTS_ERROR,
+  checkBaseBranchExists,
+  checkTargetBranchExists,
+} from '@toolkit/form-builder/editorial-workflow-utils';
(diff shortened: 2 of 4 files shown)


kulesy:
### Testing
- **Unit** — new `create-branch-modal.test.tsx` asserts the submit button shows a spinner while the pre-flight check is in flight and does **not** jump to the progress modal. Verified it fails without the `busy` spinner and passes with it.
- **Static** — `tsc --noEmit`, `biome check`, and `vitest` are green for the changed files.
- **Flow** — the create-branch editorial-workflow modal was exercised against a local TinaCloud instance on a protected `main` (the modal opens on save; the spinner render itself is covered by the unit test above).

kulesy:
Updated: switched the busy indicator from a custom `AiOutlineLoading` spinner to the existing `<LoadingDots />` component, matching the house pattern already used on the form's own Save button (`form-builder.tsx`) and `AuthModal` — so the create-branch button reads consistently with the Save button that opens it. Test now asserts the loading-dots render instead of a spinner class.

PR #7194, opened 2026-07-10
refactor(tinacms): render loading dots automatically on busy buttons

### LINKED ISSUE #7193: Consolidate the busy-button loading spinner into the Button component
### Problem
`Button`'s `busy` prop only dims the button (`opacity` + `cursor-wait`); it does not render a spinner. Every busy button hand-rolls `{busy ? <LoadingDots/> : children}` (form-builder Save button, `AuthModal`, media `UploadButton`), and the account password form (`update-password.tsx`) passes `busy` but renders no spinner at all, so it only dims. The convention has drifted.

### Proposed change
Render the shared `<LoadingDots />` indicator automatically inside `Button` when `busy`, and drop the per-call-site duplication. Dots should inherit the button text color (`currentColor`) so they stay visible on every variant (the current default-white dots are invisible on `secondary`/`ghost`). The split `DropdownButton` caret should keep its chevron (disable it rather than mark it busy).

### Notes
This also fixes the missing spinner on the account password form.

### Your environment
- `tinacms` 3.10.0 (`main`)

kulesy:
Closes #7193

**TL;DR** `Button`'s `busy` prop only dimmed the button, so every busy button hand-rolled its own loading spinner and some were missing it entirely; this renders the shared loading dots automatically whenever a button is `busy`.

**Pain:** The `busy` prop applied only `opacity`/`cursor-wait`, so each caller had to add `{busy ? <LoadingDots/> : children}` itself (form-builder Save button, `AuthModal`, media `UploadButton`), and the account password form passed `busy` with no spinner at all. The default-white dots were also invisible on `secondary`/`ghost` variants.

**Solution:** `Button` now renders `<LoadingDots color="currentColor" />` in place of its children when `busy`, so the dots inherit the button text color and show on every variant, and the per-call-site duplication is removed. `form-builder` narrows its `busy` to `submitting` (its `isGuardChecking` already sits in `disabled`, preserving today's dim-without-dots during the guard check). The split `DropdownButton` caret keeps its chevron by disabling rather than busying it. This also fixes the account password form's missing spinner. Adds a `Button` unit test.


kulesy:
### Testing
- Full `tinacms` vitest suite green (**444 passed, 5 skipped, 0 failed**) — no regressions in the touched buttons (form Save, auth, media upload).
- New `button.test.tsx` asserts a `busy` `Button` swaps its children for the shared loading dots.
- `tsc --noEmit` and `biome check` clean.

### Coordinates with #7192
#7192 (the create-branch save-button fix) currently hand-rolls the same `{busy ? <LoadingDots/> : ...}` snippet in `create-branch-modal.tsx`. Once this PR lands, `Button` renders the dots automatically, so that snippet becomes redundant — whichever merges second should drop the manual dots from the create-branch buttons so they don't double up.

joshbermanssw (approved):

kulesy:
The behavior this PR adds, in the running admin: a primary `Button` in its `busy` state now renders the shared loading dots automatically, with no per-call-site `{busy ? <LoadingDots/> : children}`.

Busy Save button rendering loading dots

Figure: The content form's primary Save button in its `busy` state (TinaCMS admin, local dev). `Button` swaps the label for `LoadingDots`, colored via `currentColor` so the dots match the button's text.

PR #7160, opened 2026-07-06
🔏 Use changesets github-api commit mode so release commits are signed

joshbermanssw:
## TL;DR

"Version Packages" PRs (e.g. #7130) can never merge because `main` requires signed commits and `changesets/action` commits via git CLI on the runner, unsigned. Setting `commitMode: github-api` (native in the pinned v1.9.0) has GitHub create and sign the commit server-side; no key to provision, no extra workflow step.

One-line alternative to #7159, which recreates the commit with a custom script after the fact. Commits will be attributed to the `release-bot-allow-prs-and-push[bot]` app identity instead of "Tina Release Bot".

## Links

- Closes #7158
- Alternative to #7159
- Blocked example: #7130

PR #7203, opened 2026-07-13
🐛 Hard-reset before the @dev snapshot so beta stops publishing release versions

joshbermanssw:
## TL;DR

Since #7160 switched changesets to `commitMode: github-api`, the release-commit version bump is committed server-side and left **uncommitted in the runner's working tree**. The "Release to @dev channel" step then does `git checkout ${{ github.sha }}` (which *keeps* those dirty files), so `version:snapshot` finds no changesets and publishes the **clean release version** to the `beta` tag instead of a throwaway `0.0.0-<sha>`. That pre-claims the version the Version Packages publish needs, so the `latest` publish finds it already on npm and no-ops.

## Fix

Replace `git checkout` with `git reset --hard ${{ github.sha }}` so the snapshot runs against the pristine release commit (changesets restored) and emits a disposable `0.0.0-<sha>` version.

## Reviewer notes

- **Impact (separate from this PR):** the last release is already stranded, `latest` for ~16 packages (tinacms `3.10.1`, `@tinacms/cli` `2.5.5`, `create-tina-app` `2.1.11`, …) sits on `beta`. That needs a one-time `npm dist-tag add …@… latest` retag; this PR only stops it recurring.
- **Validate:** on a branch that still has a changeset, run `git reset --hard <sha> && pnpm version:snapshot deadbee` and confirm the bumped `package.json` reads `0.0.0-deadbee-…`, not a clean semver.

## Links

- Regression source: #7160

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -295,7 +295,12 @@ jobs:
       - name: Release to @dev channel
         if: steps.changesets.outputs.hasChangesets == 'true'
         run: |
-          git checkout ${{ github.sha }}
+          # Hard reset (not checkout) so the snapshot runs against the pristine
+          # release commit. changesets/action's github-api commit mode leaves the
+          # version bumps uncommitted in the working tree; a plain checkout keeps
+          # them, so `version:snapshot` finds no changesets and publishes the real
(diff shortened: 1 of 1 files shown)


18-th (approved):
LGTM

PR #7214, opened 2026-07-13
🐛 Stop shipping monaco-editor — 73 MB of never-executed code

### LINKED ISSUE #7209: 🤖 Stop shipping monaco-editor — it's a types-only import loaded from a CDN (~73 MB)
A stock Astro + TinaCMS blog installs **1412 MB** of `node_modules`. **73.3 MB of it is `monaco-editor` — and none of that code is ever used.**

## Finding

`monaco-editor` is a production dependency of `tinacms` and `@tinacms/app`. But **every import of it in shipped source is types-only:**

```
packages/tinacms/src/toolkit/fields/plugins/mdx-field-plugin/monaco/index.tsx:2
packages/@tinacms/app/src/fields/rich-text/monaco/index.tsx:10
    import type * as monaco from 'monaco-editor';
```

The actual editor is fetched **from jsdelivr at runtime** via `@monaco-editor/react`'s loader:

```ts
// packages/tinacms/src/toolkit/fields/plugins/mdx-field-plugin/monaco/index.tsx:22
loader.config({
  paths: { vs: 'https://cdn.jsdelivr.net/npm/monaco-editor@0.31.1/min/vs' },
});
```

There is no `loader.config({ monaco })`, no `monaco-editor/esm` import, and no worker bundling (`vite-plugin-monaco-editor`, `MonacoEnvironment`, `getWorker`) anywhere in the CLI or app build config. **The 73 MB on disk does nothing.**

## Two things make this bigger than a one-line dependency removal

**1. The copy in `packages/tinacms` is dead code.**

`packages/tinacms/src/toolkit/fields/plugins/mdx-field-plugin/monaco/index.tsx` is a stale duplicate:
- its `parseMDX`/`stringifyMDX` are **stubs returning empty values** (lines 16-17), so it would destroy content if it ever ran
- nothing imports it, and it is not exported from the package
- only its sibling `error-message.tsx` is actually used (by `create-invalid-markdown-plugin`)

The live raw editor is `packages/@tinacms/app/src/fields/rich-text/monaco/index.tsx` (lazy-loaded at `App.tsx:14`, real `parseMDX` from `@tinacms/mdx`), which is injected into `tinacms` as the `rawEditor` **prop** (`mdx-field-plugin/index.tsx:128`).

So `packages/tinacms` can drop both monaco packages simply by deleting dead code.

**2. Removing `monaco-editor` from our `dependencies` is not sufficient on its own.**

`monaco-editor` is a **required `peerDependency` of `@monaco-editor/react`**:

```
@monaco-editor/react@4.7.0-rc.0 -> peerDependencies: { "monaco-editor": ">= 0.25.0 < 1" }
```

It is not marked optional, and **npm 7+ auto-installs required peers**. Deleting it from our `dependencies` alone changes nothing on disk.

## Proposed fix

1. **`packages/tinacms`** — delete the dead `monaco/index.tsx` (keep `error-message.tsx`), then remove both `monaco-editor` and `@monaco-editor/react` from `dependencies`.
2. **`packages/@tinacms/app`** — replace `@monaco-editor/react` with **`@monaco-editor/loader`** (its only dependency is `state-local`, ~4 KB). We already use its `loader` export directly; `MonacoEditor`/`useMonaco` are a thin wrapper worth ~40 lines to inline. Then remove `monaco-editor` + `@monaco-editor/react`.
3. Drop the `monaco-editor` entry from the `pnpm-workspace.yaml` catalog.

The feature (raw-mode MDX editing when MDX fails to parse) keeps working exactly as it does today — it already loads from the CDN.

## Worth deciding separately, not in this issue

Loading the editor from a public CDN means raw mode breaks offline/air-gapped and adds third-party supply-chain surface. Self-hosting it properly would mean paying the 73 MB back. This issue only removes weight that is **already unused** and changes no behaviour; the hosting question is a separate call.

## Acceptance

- `monaco-editor` absent from a fresh install (`ls node_modules/monaco-editor` → not found).
- Raw mode still opens (paste malformed MDX into a rich-text field) and the editor loads.
- ~73 MB off a fresh install.

wicksipedia:
Fixes #7209.

## The bug

Every `monaco-editor` import in shipped source is **types-only**:

```ts
import type * as monaco from 'monaco-editor';   // erased at compile time
```

The editor itself has always been fetched from a CDN at runtime by `@monaco-editor/loader`. There is no `loader.config({ monaco })`, no `monaco-editor/esm` import, and no worker bundling anywhere in the CLI or app build config.

So the **73 MB** we put on every user's disk has never been executed. It isn't even the same version as the editor that actually runs: we ship `monaco-editor@0.31.0`, while the loader fetches `monaco-editor@0.55.1` from jsdelivr.

## Why removing the dependency wasn't enough

`monaco-editor` is a **required `peerDependency`** of `@monaco-editor/react`:

```
@monaco-editor/react@4.7.0-rc.0 -> peerDependencies: { "monaco-editor": ">= 0.25.0 < 1" }
```

It is not marked optional, and **npm 7+ auto-installs required peers** — so dropping it from our own `dependencies` would have changed nothing on disk.

`@tinacms/app` now uses **`@monaco-editor/loader`** directly instead. That is the very package `@monaco-editor/react` wraps (it's its only dependency), so **the CDN and the editor version are unchanged** — this is not a downgrade or a behaviour change. `MonacoEditor`/`useMonaco` are replaced with a small local equivalent. `monaco-editor` stays on as a **devDependency** for its types.

## Dead code in `packages/tinacms`

`packages/tinacms/.../mdx-field-plugin/monaco/index.tsx` turned out to be a stale duplicate:

- nothing imported it, and it was not exported from the package
- its `parseMDX`/`stringifyMDX` were **stubs returning empty values** — it would have destroyed content had it ever run

The live raw editor is in `@tinacms/app` (lazy-loaded at `App.tsx:14`, real `parseMDX` from `@tinacms/mdx`) and is injected into `tinacms` as the `rawEditor` **prop**. Deleting the dead copy lets `tinacms` drop both monaco packages outright. `error-message.tsx` is kept — it's the one file in that directory still used.

## Verification

- `pnpm build` — 23/23 tasks pass.
- **Typecheck unchanged: 586 errors on `main`, 586 on this branch** (the package has a large pre-existing baseline; the point is this adds none).
- `tinacms build` on `examples/next/kitchen-sink` produces a working admin.
- The built admin bundle contains the loader's CDN require and **no monaco source**:
  ```
  t(["vs/editor/editor.main"], ...)        <- loader's AMD require
  monaco-editor@0.55.1/min/vs             <- fetched at runtime, unchanged
  monaco-editor/esm/vs      -> not found  <- monaco is NOT bundled
  StandaloneCodeEditor      -> not found
  ```

Raw MDX mode still loads the same editor, from the same CDN, at the same version.

## Follow-ups, deliberately not in this PR

- The types we compile against (0.31.0) don't match the editor that runs (0.55.1). Pre-existing; aligning them is a separate change with its own type-error surface.
- Loading the editor from a public CDN means raw mode breaks offline/air-gapped and carries third-party supply-chain surface. Self-hosting would mean paying the 73 MB back — a maintainer call, tracked in #7209. This PR only removes weight that is already unused.


packages/@tinacms/app/src/fields/rich-text/monaco/index.tsx
```diff
@@ -29,6 +29,70 @@ export const uuid = () => {
...
+      return;
+    }
+    const uri = monaco.Uri.parse(`inmemory://tina/${path}`);
+    const model =
+      monaco.editor.getModel(uri) ||
+      monaco.editor.createModel(value, language, uri);
+    const editor = monaco.editor.create(containerRef.current, {
+      model,
+      ...options,
+    });
```
joshbermanssw, line 87: `@monaco-editor/react`'s `Editor` created the editor with `automaticLayout: true` by default (`editor.create(el, { model, automaticLayout: !0, ...opts })`). This local reimplementation drops it, and the `options` passed from `RawEditor` don't set it either.

Consequence: monaco no longer installs its container ResizeObserver, so the editor won't reflow when its container **width** changes without a content edit (e.g. resizing the browser window or the form pane). Height growth and post-keystroke reflow still work via the `onDidContentSizeChange` → `editor.layout()` handler, so it's a minor, self-healing glitch rather than a break. But it is a behaviour difference from the package being removed, which slightly undercuts the "No behaviour change" claim.

Re-adding it restores exact parity at negligible cost (ResizeObserver-based in monaco 0.55):

```suggestion
    const editor = monaco.editor.create(containerRef.current, {
      model,
      automaticLayout: true,
      ...options,
    });
```

18-th (approved):
LGTM

PR #7215, opened 2026-07-13
🐛 Gate quick-edit hover wash behind `@media (hover: hover)`

joshbermanssw:
As per #7217

## TL;DR

The quick-edit hover highlight painted a full-bleed blue wash (`inset 100vi 100vh` box-shadow, plus the overlay's `::after` reveal) with no hover-capability guard, so on a tablet it stayed stuck over the last-tapped field (touch `:hover` latches and never clears). This gates only the wash behind `@media (hover: hover)`.

## Notes

- **Only the fill is gated.** The solid outline still appears on tap as selection feedback, and the resting dashed outline is unchanged. On touch you get the outline, just not the screen-filling blue.
- **One constant, two consumers.** Both the React `useTina` hook and the vanilla `initClickToFocus` bridge import `QUICK_EDIT_CSS`, so both pick up the fix. The click-to-select handler is untouched.

## Before 

https://github.com/user-attachments/assets/efcf9b8d-98e8-4b38-b669-fc1b41b1be54

**Figure: ❌ Tablet view - hover:* effect on background wash gets stuck when I click**

## After 

https://github.com/user-attachments/assets/8cc506be-b812-4b5a-8530-0913f39368dd

**Figure: ✅ Tablet view - hover:* effect is gated behind @media so hover doesnt stick around**

## Tests

New CSSOM structural test (`quick-edit-css.test.ts`) asserts the wash rules (inset box-shadow, overlay `::after` reveal) are gated behind `@media (hover: hover)` while the solid outline stays available without hover. Full bridge suite passes (39 tests), `tsc` clean.

Verified structurally and via typecheck, not on physical hardware. Worth a manual tap-test on a real tablet before release.


18-th (approved):
LGTM

PR #7216, opened 2026-07-13
🐛 Consolidate duplicate UI libraries — 3 icon libs → lucide, headless-UI → Radix (~135 MB)

### LINKED ISSUE #7208: 🤖 Collapse three icon libraries into lucide-react (~85 MB)
A stock Astro + TinaCMS blog — no libraries beyond Astro and TinaCMS — installs **1412 MB** of `node_modules`. After deduplication (see #7207) it is still **1091 MB**, and the largest single item is an icon library:

| Package | On disk |
|---|---:|
| **react-icons** | **84.6 MB** |
| lucide-react | 31.9 MB |
| @heroicons/react | small |

`tinacms` declares **three** icon libraries as production dependencies. Every user downloads all three.

## Finding

Icon usage in shipped packages is **100% internal admin chrome** — sidebar, media manager, branch switcher, rich-text toolbar, status badges:

| Library | Files | Distinct icons | Subpaths |
|---|---:|---:|---|
| `react-icons` | 47 | 76 | 13 (`bi`, `md`, `fa`, `bs`, `io`, `ai`, `fi`, `gr`, `im`, `pi`, `ri`, `tb`, `vsc`) |
| `@heroicons/react` | 5 | 3 | `/solid`, `/outline` |
| `lucide-react` | 29 | 99 | — |

Verified **not** present anywhere:
- no icon-picker UI
- no `icon` field type in the schema (`@tinacms/schema-tools`)
- no `react-icons/all-files` or dynamic whole-set import

`IconType` from `react-icons/lib` is used as an internal prop type in `sidebar.tsx`, `AdminNav.tsx`, and `cloud-config-plugin.tsx`, but `ScreenPlugin.Icon` is typed `any` — so it is **not** part of the public plugin API.

**Therefore removing `react-icons` and `@heroicons/react` is not a breaking change for user content, schemas, or plugins.** `lucide-react` is already the de-facto house set (99 icons, and the entire rich-text toolbar).

## Proposed fix

Collapse to `lucide-react` only.

**`@heroicons/react` — trivial.** All 3 icons already have lucide equivalents imported elsewhere in this codebase:
`ChevronDownIcon` → `ChevronDown`, `XCircleIcon` → `CircleX`, `FolderIcon` → `Folder`.

**`react-icons` — mechanical.** The large majority map 1:1 (`BiSearch`→`Search`, `BiX`→`X`, `MdCheckCircle`→`CircleCheck`, `BiGitBranch`→`GitBranch`, …). **5 of 76 need a cosmetic judgment call:**

| Current | Suggested | Note |
|---|---|---|
| `BsCheckCircleFill` | `CircleCheck` | lucide is outline-only; filled look is lost |
| `BsExclamationOctagonFill` | `OctagonAlert` | same |
| `MdSyncProblem` | `RefreshCwOff` | no exact match |
| `TbLogs` | `ScrollText` | no exact match |
| `BiRename` | `SquarePen` | no "rename" glyph |
| `MdOutlineDataSaverOff` | `Info` | used once for an "info" badge; semantics already loose |

Swap `IconType` for lucide's `LucideIcon`. Then drop `react-icons` and `@heroicons/react` from `packages/tinacms/package.json` and `packages/@tinacms/app/package.json`.

Both libraries accept `size`/`className`, so props are largely compatible.

Scope: ~50 files.

## Acceptance

- `rg "react-icons|@heroicons" packages/ --glob '!**/node_modules/**'` returns zero hits outside `examples/`.
- `pnpm build` clean.
- Admin renders with no missing glyphs — check sidebar, media manager, branch switcher, rich-text toolbar, and the 6 substituted icons above.
- ~85 MB off a fresh install.

### LINKED ISSUE #7219: 🤖 Drop @headlessui/react — 50 MB of React Aria for components Radix already covers
After the install-size fixes in #7213, #7214, #7216 and #7218, a stock Astro + TinaCMS blog installs **958 MB** of `node_modules` (down from 1412 MB). **50 MB of what's left is `@headlessui/react` — and we don't need it.**

## The problem

`tinacms` ships **three** headless component libraries as production dependencies:

| Library | Files using it |
|---|---|
| `@radix-ui/*` (8 packages: dialog, dropdown-menu, popover, select, separator, slot, toolbar, tooltip) | 22 |
| `@headlessui/react` | 11 |
| `@ariakit/react` | 1 |

`@headlessui/react` is only 1.9 MB itself, but v2 pulls in the entire React Aria stack:

| Package | On disk |
|---|---:|
| `react-aria` | 34.8 MB |
| `react-stately` | 9.4 MB |
| `@internationalized/*` | 1.9 MB |
| `@headlessui/react` | 1.9 MB |
| **total** | **~50 MB** |

## We already have the replacements installed

Every Headless UI component in use has an existing-dependency equivalent — **this requires no new packages**:

| Headless UI | Files | Replace with | Already a dep? |
|---|---:|---|---|
| `Popover` / `PopoverButton` / `PopoverPanel` | 4 | `@radix-ui/react-popover` | ✅ |
| `Menu` / `MenuButton` / `MenuItem` / `MenuItems` | 2 | `@radix-ui/react-dropdown-menu` | ✅ |
| `Combobox` (+ Button/Input/Option/Options) | 1 | `cmdk` | ✅ |
| `Disclosure` (+ Button/Panel) | 1 | local `useState` + conditional render | n/a |
| `Transition` / `TransitionChild` | 11 | CSS transitions / Radix `data-state` animations | n/a |

### The files

```
packages/tinacms/src/admin/pages/CollectionListPage.tsx                                    Menu, Transition
packages/tinacms/src/toolkit/react-sidebar/components/sidebar.tsx                          Transition, TransitionChild
packages/tinacms/src/toolkit/react-sidebar/components/sidebar-body.tsx                     Transition
packages/tinacms/src/toolkit/fields/plugins/group-field-plugin.tsx                         Transition
packages/tinacms/src/toolkit/fields/plugins/blocks-field-plugin/block-selector.tsx         Popover, Transition
packages/tinacms/src/toolkit/fields/plugins/blocks-field-plugin/block-selector-big.tsx     Disclosure, Transition, TransitionChild
packages/tinacms/src/toolkit/fields/plugins/mdx-field-plugin/monaco/error-message.tsx      Popover, Transition
packages/tinacms/src/toolkit/fields/plugins/mdx-field-plugin/plate/plugins/create-mdx-plugins/component.tsx  Popover, Transition
packages/tinacms/src/toolkit/fields/plugins/mdx-field-plugin/plate/plugins/ui/autocomplete.tsx              Combobox, Transition
packages/tinacms/src/toolkit/fields/plugins/mdx-field-plugin/plate/plugins/ui/dropdown.tsx                  Menu, Transition
packages/@tinacms/app/src/fields/rich-text/monaco/error-message.tsx                        Popover, Transition
```

## Proposed fix

Migrate the 11 files to Radix (+ `cmdk` for the one Combobox), then drop `@headlessui/react` from `packages/tinacms/package.json` and `packages/@tinacms/app/package.json`.

This is a real component migration, not a mechanical import swap — Radix has a different composition model (`Root`/`Trigger`/`Content`) and its own focus/a11y behaviour. It's worth doing carefully, but it consolidates us onto **one** headless library (Radix, already the majority at 22 files) instead of three.

`Transition` is the easy part: Radix primitives expose `data-state="open|closed"`, so the enter/leave animations become CSS (`data-[state=open]:animate-in` etc.) rather than a JS component.

## Acceptance

- `rg "@headlessui" packages/ --glob '!**/node_modules/**'` → zero hits
- `@headlessui/react`, `react-aria`, `react-stately` absent from a fresh install
- `pnpm build` clean, typecheck no worse
- Admin verified by hand: the collection-list overflow menu, the block-selector popover and its big variant, the MDX autocomplete combobox and dropdown, the rich-text error popover, and the sidebar open/close transition all still work — including keyboard nav and focus trapping.
- ~50 MB off a fresh install.

## Context

Part of an install-size audit. Related: #7207 / #7213 (exact-pin duplication, ~320 MB), #7208 / #7216 (icon libraries, ~85 MB), #7209 / #7214 (monaco-editor, ~73 MB), #7210 / #7218 (TypeScript compiler), #7211 (typedoc/uvu), #7212 (vite 4).

Remaining after this: `mermaid` (~124 MB, needs a CDN-vs-optional-peer decision), duplicate `lodash`/`zod`/`prettier` (~86 MB), duplicate `esbuild` (~60 MB, #7212), `posthog-js` + `core-js` (~57 MB).

wicksipedia:
Tagged preview on tina.io https://tina-io-git-jb-icons-tag-tinacms.vercel.app/admin

## Follow-up (after review, rebased on `main`)

Fixes below were verified live in the running admin with Playwright; `main` was then merged to clear conflicts.

- **Block picker rendered at 0 height.** The local `Transition` kept a `transform` after the enter animation settled, which turned the wrapper into a containing block and collapsed the absolutely-positioned picker panel. Added a `settled` phase that drops the transition classes once the animation lands (`0 → 715px`), fixing the `+` doing nothing on `visualSelector` blocks fields.
- **`fill-current` on lucide `Pencil`** removed (the group-list / block-list edit icons were flooding to solid gray blobs).
- **Icon sizing.** lucide renders at a fixed 24px and ignores `font-size`; gave the Add Folder/Add File, media-usage, and nav-info icons explicit dimensions.
- **`modal={false}`** on the migrated dropdown menus, matching Headless UI's non-modal `Menu`.
- **Removed dead code:** `Autocomplete` and the toolbar `Dropdown` had no importers anywhere in the repo.

---

Fixes #7208. Fixes #7219. (Supersedes #7220 — folded in here; the two changes edit the same 11 files.)

TinaCMS shipped **three icon libraries and three headless-component libraries** at once. This consolidates onto one of each: **lucide** for icons, **Radix** for components. **No new dependencies are added** — both winners were already installed.

**~135 MB off a fresh install.**

---

## Part 1 — Three icon libraries → `lucide-react` (~85 MB)

| Library | Files | Icons | Action |
|---|---:|---:|---|
| `react-icons` | 47 | 76 (13 subpacks) | **remove** — 84 MB |
| `@heroicons/react` | 5 | 3 | **remove** |
| `lucide-react` | 29 | 99 | **keep** — already the house set |

Icon usage is **100% internal admin chrome**. Verified absent: no icon-picker UI, no `icon` schema field type, no `react-icons/all-files` import. `ScreenPlugin.Icon` is typed `any` and `CloudConfigOptions` isn't exported — **nothing here is public API**, so no break for user content, schemas, or plugins.

Six icons have no exact lucide counterpart (all cosmetic): `BsCheckCircleFill`/`BsExclamationOctagonFill` → `CircleCheck`/`OctagonAlert` (lucide is outline-only), `MdSyncProblem` → `RefreshCwOff`, `TbLogs` → `ScrollText`, `BiRename` → `SquarePen`, `MdOutlineDataSaverOff` → `Info`.

### The `fill` trap (two separate bugs, both caught by screenshotting)

react-icons glyphs are **solid shapes with negative space** cut out for the internal mark. lucide icons are **stroke-based**. So a `fill` that was correct before now floods the shape and erases its detail.

**1. Explicit `fill-current` on icons.** `CircleAlert` would have rendered as a featureless red disc, `CircleCheck` as a green disc with no tick. Removed from the detail-bearing icons; `fill-*` colours on the file/folder icons became `text-*` (lucide colours the *stroke*).

Kept only on solid silhouettes — and **`FormStatus` needed `fill-current` *added***, since lucide's `Circle` is a hollow ring by default and the pristine/dirty dot would otherwise have become an empty circle.

**2. Inherited `fill` from an ancestor** (09ba989). `index.css` had:

```css
.icon-parent svg { fill: currentColor; }
```

That exists for the hand-written Tina SVGs (`AddIcon`, `CloseIcon`), which *are* solid shapes — and **`icon-parent` sits on the base `Button` component**, so it hit every icon in every button. The media manager's Refresh arrow rendered as a blue blob, New Folder as a solid orange rectangle, Upload as a featureless cloud. The icons themselves had **no fill class at all**, so an audit of the icon elements never surfaced it. Fixed by scoping: `.icon-parent svg:not(.lucide)`.

**Before/after screenshots and the full visual review: [see this comment](https://github.com/tinacms/tinacms/pull/7216#issuecomment-4955674639).**

---

## Part 2 — Drop `@headlessui/react` (~50 MB)

Headless UI is 1.9 MB itself, but v2 pulls in the whole React Aria stack: `react-aria` (34.8 MB) + `react-stately` (9.4 MB) + `@internationalized/*` — for components Radix already covers (Radix was already the majority, at 22 files vs 11).

| Headless UI | Files | Now |
|---|---:|---|
| `Popover` / `PopoverButton` / `PopoverPanel` | 4 | `@radix-ui/react-popover` *(already a dep)* |
| `Menu` / `MenuButton` / `MenuItem` / `MenuItems` | 2 | `@radix-ui/react-dropdown-menu` *(already a dep)* |
| `Combobox` | 1 | Radix popover + input + `PopoverAnchor` |
| `Disclosure` | 1 | `useState` + conditional render |
| `Transition` / `TransitionChild` | 11 | local component, same prop API |

Headless UI's `focus` render prop becomes Radix's `data-[highlighted]:` CSS. Menu items use `asChild` so the real `<button>`/`<Link>` survives and `onClick` still gets a genuine `MouseEvent`.

### The local `Transition`

Same prop API (`show`, `appear`, `enter*`/`leave*`), so the call sites only change their import.

The subtle part: `sidebar.tsx` has a `<Transition show={menuIsOpen}>` with **no transition classes of its own** — it's a pure mount controller, and a `<TransitionChild>` does the animating. So the parent must not unmount until the *child* finishes leaving. `element.getAnimations({ subtree: true })` gives that natively.

My first version passed build, typecheck and lint — and was **wrong**: it unmounted instantly on close, skipping the leave animation, because it asked `getAnimations()` before React had painted the leave classes. Only driving it in a browser caught it. After the fix:

| moment | state |
|---|---|
| open +40 ms | mounted, opacity **0.106** (animating in) |
| close +60 ms | mounted, opacity **0.971**, `leaveTo` applied |
| close +140 ms | mounted, opacity **0.675** — parent still waiting on the child |
| close settled | **unmounted** |

---

## Verification

- `pnpm build` — 23/23
- Typecheck **unchanged: 586 on `main`, 586 here**
- `biome check` clean
- `rg "react-icons|@heroicons|@headlessui" packages/` → **zero**
- `react-aria` / `react-stately` confirmed gone from the store
- **Measured in a real install** (packed tarballs into a live project): this PR is worth **~127 MB** of the 496 MB that the full stack (#7213, #7214, #7216, #7218) removes — 1412 MB → 916 MB.
- Asserted in the running admin: lucide inside `.icon-parent` → `fill: none`; custom Tina SVGs → still filled; `FormStatus` dot → still a solid green disc; **zero** lucide icons filled without asking.

## Not exercised locally — worth a human pass

- **`Disclosure` → block category groups**: only renders when block templates define a `category`; kitchen-sink's don't, so it never mounted.
- **`TemplateMenu` "Create New" dropdown**: needs a collection with multiple `templates`.
- **MDX/rich-text paths**: the `Autocomplete` language picker, the toolbar `Dropdown`, the embed `DotMenu`, the raw-editor error popover.
- **Branch switcher / sync status** — including the two filled Bootstrap glyphs, where filled-vs-outline is a judgment call.
- **Keyboard nav / focus trapping** in the new Radix menus — Radix's focus model differs from Headless UI's.


wicksipedia:
## Visual review — before/after

Ran the admin on `main` and on this branch and captured the same regions at the same viewport, so these are directly comparable.

**Reviewing this found a real bug, which is now fixed in 09ba989.** Details at the bottom — the build, typecheck and lint were all green while it was broken, so it was only visible in a screenshot.

<sub>Images are hosted on the `review-assets/icons-7216` branch (screenshots only, not for merge).</sub>

---

### 1. Sidebar navigation

| Before (`main`) | After |
|---|---|
|  |  |

Two differences worth calling out:

- **Media Usage** was `MdImage` — a *solid filled* icon sitting among outline icons. It's now an outline `Image`, which makes the whole nav consistent. I'd argue this is an improvement, but it is a visible change.
- The **ⓘ** next to COLLECTIONS is slightly bolder (`FiInfo` → `Info`).

Everything else (Files, Globe, Images, LogOut) maps cleanly.

---

### 2. Collection list — toolbar and rows

| | Before (`main`) | After |
|---|---|---|
| Toolbar |  |  |
| Rows |  |  |

---

### 3. Media manager

| Before (`main`) | After |
|---|---|
|  |  |

The folder and clapperboard tiles go from solid to outline. That's the `fill-*` → `text-*` change (lucide colours the *stroke*, not the fill) and it reads correctly.

---

### 4. The bug this review caught 🐛

The media toolbar came out like this:

**`main`:**


**Broken (before 09ba989):**


Refresh became a blue blob, New Folder a solid orange rectangle, and Upload a featureless cloud with the arrow gone.

**Cause.** `packages/@tinacms/app/src/index.css` has:

```css
.icon-parent svg { fill: currentColor; }
```

That rule exists for the hand-written Tina SVGs (`AddIcon`, `CloseIcon`, `TinaIcon`), which are solid shapes coloured via `fill`. And **`icon-parent` is on the base `Button` component** — so it applies to every icon inside every button.

react-icons glyphs are solid and expect exactly that. lucide icons are stroke-based and ship `fill="none"` **as a presentation attribute** — and CSS beats a presentation attribute. So every lucide icon inside a button got flooded with `currentColor`, erasing its internal detail.

Note the icons themselves had **no fill class at all**:

```jsx
<RefreshCw className='w-6 h-full ml-2 opacity-70 text-blue-500' />
```

which is why an audit of the icon elements didn't surface it — the fill was inherited from an ancestor.

**Fix** (09ba989):

```css
.icon-parent svg:not(.lucide) { fill: currentColor; }
```

Custom Tina SVGs keep their fill; lucide icons stay outlines. Icons that genuinely *want* a fill still opt in explicitly via the `fill-current` utility.

**Fixed:**


---

### 5. The form status dot

The one icon where a fill is load-bearing. `FaCircle` is a solid disc; lucide's `Circle` is a hollow ring, so this needed `fill-current` **added** or the pristine/dirty indicator would have rendered as an empty circle.

| Before (`main`) | After |
|---|---|
|  |  |

Computed style is identical on both:

```
main   : class="h-3 text-green-500"                                fill: rgb(34, 197, 94)
branch : class="lucide lucide-circle h-3 fill-current text-green-500"  fill: rgb(34, 197, 94)
```

---

### Post-fix assertions (in the running admin)

- lucide icons inside `.icon-parent` → `fill: none` ✅
- custom Tina SVGs inside `.icon-parent` → still a real fill ✅
- `FormStatus` dot → still a solid green disc ✅
- **zero** lucide icons filled without an explicit `fill-*` class ✅

`pnpm build` 23/23 · typecheck unchanged (586 = 586) · biome clean.

### Still not exercised

The rich-text/MDX icon paths (toolbar, code-block language picker, embed menus) and the branch switcher / sync-status icons — including the two filled Bootstrap glyphs (`BsCheckCircleFill` → `CircleCheck`, `BsExclamationOctagonFill` → `OctagonAlert`), which are the remaining substitutions where "filled vs outline" is a judgment call. Worth a look when someone has a branch-enabled setup.

packages/tinacms/src/toolkit/components/transition.tsx
```diff
@@ -0,0 +1,220 @@
...
+    direction === 'enter'
+      ? stage === 'start'
+        ? c.enterFrom
+        : c.enterTo
+      : stage === 'start'
+        ? c.leaveFrom
+        : c.leaveTo;
+  const settled =
+    direction === 'enter' && stage === 'end' ? c.entered : undefined;
+  return [base, phase, settled].filter(Boolean).join(' ');
```
joshbermanssw: **Bug: this collapses the visual block picker so its `+` looks dead.** On a blocks field with `visualSelector: true` (e.g. kitchen-sink `page.tsx` `blocks`), clicking `+` renders `BlockSelectorBig`'s slide-in panel but it comes up at **zero height**, so nothing appears.

**Root cause is here.** For a *settled* enter (`direction: 'enter'`, `stage: 'end'`), `classesFor` returns `[c.enter, c.enterTo, c.entered]` and keeps `enter`/`enterTo` applied forever. In `BlockSelectorBig` the `enter` classes are `transform transition-all …` and `enterTo` is `translate-x-0`. `translate-x-0` computes to `transform: translateX(0)`, which is **not** `none`, so the `<TransitionChild>` wrapper stays a containing block for its `absolute left-0 top-0 h-full w-full` panel. The wrapper is height:auto (its only child is out of flow) → `h-full` resolves to 0 → the panel is invisible.

Headless UI does not hit this: once the enter transition finishes it **removes** `enter`/`enterFrom`/`enterTo` and applies only `entered` (that is the whole reason the `entered` prop exists), so no leftover transform.

**Verified** with an isolated repro of this exact `classesFor` + the `BlockSelectorBig` DOM structure: the panel mounts at `400×0`, the wrapper carries class `enter enterTo` with computed `transform: matrix(1,0,0,1,0,0)`. Dropping the enter classes once settled restores `400×360` and `transform: none`.

**Fix direction:** add a settled phase. After the enter transition's animations finish (reuse the `getAnimations`/`waitForExit` you already have, on the parent ref), advance to a state where `classesFor` returns only `entered` and drops `enter`/`enterFrom`/`enterTo`. Dropping them the instant `stage === 'end'` also fixes the collapse but skips the slide-in animation, so gate it on the animation actually completing.

This is a regression from the Headless UI → local `Transition` swap: `block-selector-big.tsx` is otherwise unchanged (only the import moved). It affects any `<TransitionChild>` whose `enter` classes include a `transform` around absolutely-positioned content, so the mobile sidebar slide-in (`sidebar.tsx`) is the other place worth a look.

packages/tinacms/src/toolkit/fields/plugins/group-list-field-plugin.tsx
```diff
@@ -230,7 +230,7 @@ const Item = ({
               }}
             >
               <GroupLabel>{title}</GroupLabel>
-              <BiPencil className='h-5 w-auto fill-current text-gray-200 group-hover:text-inherit transition-colors duration-150 ease-out' />
+              <Pencil className='h-5 w-auto fill-current text-gray-200 group-hover:text-inherit transition-colors duration-150 ease-out' />
```
joshbermanssw: **Fill trap not caught here.** `Pencil` is a lucide stroke icon that ships `fill="none"`, but this kept `fill-current` from the old solid `BiPencil`. CSS `fill-current` overrides lucide's `fill="none"` attribute, so this edit-pencil renders as a solid gray shape instead of an outline. It is the same trap your `index.css` `svg:not(.lucide)` rule fixes, but that rule can't help when `fill-current` sits directly on the icon. Same issue at `blocks-field-plugin/index.tsx:265`. Fix: drop `fill-current` (the stroke is already colored via `text-*`).

packages/tinacms/src/toolkit/fields/plugins/blocks-field-plugin/index.tsx
```diff
@@ -262,7 +262,7 @@ const BlockListItem = ({
                   NEW
                 </span>
               )}
-              <BiPencil className='h-5 w-auto fill-current text-gray-200 group-hover:text-inherit transition-colors duration-150 ease-out' />
+              <Pencil className='h-5 w-auto fill-current text-gray-200 group-hover:text-inherit transition-colors duration-150 ease-out' />
```
joshbermanssw: **Same fill trap as `group-list-field-plugin.tsx:233`.** lucide `Pencil` ships `fill="none"`; the leftover `fill-current` overrides it and floods the outline into a solid gray blob. Drop `fill-current` here too.

packages/tinacms/src/admin/pages/CollectionListPage.tsx
```diff
@@ -751,7 +730,7 @@ const CollectionListPage = () => {
                                               : 0
                                           }
                                         >
-                                          <FaFolder className='mr-2' />
+                                          <Folder className='mr-2' />
```
joshbermanssw: **Size regression: lucide is a fixed 24px and ignores `font-size`.** react-icons rendered at `1em`, so the old `FaFolder`/`FaFile` scaled with the button's font-size (~14-16px). lucide `Folder`/`File` render at 24px no matter what, so the Add Folder / Add File icons jump noticeably larger. Same on the Add File icon at line 789. Add an explicit size (e.g. `className='mr-2 w-4 h-4'` or `size={16}`).

This is a general pattern in the PR: any icon that was sized only by `text-*` or inherited font-size (no `w`/`h`/`size`) is now 24px. Other spots: `media-usage-table.tsx:176/179/629`, `nav.tsx:166`.

packages/tinacms/src/toolkit/components/dashboard/media-usage-dashboard/media-usage-table.tsx
```diff
@@ -167,10 +173,10 @@ const getMediaColumns = (
                   onPreview?.(row.original);
                 }}
               >
-                <BiMovie className='text-3xl' />
+                <Clapperboard className='text-3xl' />
```
joshbermanssw: **Size regression.** `text-3xl` set font-size, which sized the old react-icons (`1em`) to ~30px. lucide ignores font-size and renders at its fixed 24px, so these media-type icons shrink. Same at line 179, and the untyped `<File>` at line 629 (was ~1em, now 24px). Use `w-8 h-8` / `size={30}` instead of `text-3xl`.

packages/tinacms/src/toolkit/react-sidebar/components/nav.tsx
```diff
@@ -165,7 +163,7 @@ export const Nav = ({
                       href='https://tina.io/docs/r/content-modelling-collections'
                       target='_blank'
                     >
-                      <FiInfo />
+                      <Info />
```
joshbermanssw: **Size regression.** Bare `<Info/>` with no size class renders at lucide's fixed 24px; the old `<FiInfo/>` was `1em` and scaled with the surrounding font-size. Add an explicit `w`/`h` (or `size`) if the prior size mattered.

packages/tinacms/src/admin/pages/CollectionListPage.tsx
```diff
@@ -91,64 +84,50 @@ const TemplateMenu = ({
...
-                    )}
-                  </MenuItem>
-                ))}
-              </div>
-            </MenuItems>
-          </Transition>
-        </div>
-      )}
-    </Menu>
+    <DropdownMenuPrimitive.Root>
```
joshbermanssw: **Behavior change: Radix `DropdownMenu` is modal by default; Headless UI `Menu` was not.** With no `modal` prop, opening this menu scroll-locks the page, traps focus, sets `body { pointer-events: none }`, and `aria-hidden`s the rest of the admin until it closes (the scrollbar-compensation can also nudge layout). Selection and keyboard both work correctly; this is purely the modal wrapper. For parity with the old non-modal menu, pass `modal={false}` here and on the toolbar `Dropdown` (`.../plate/plugins/ui/dropdown.tsx:22`).

packages/tinacms/src/toolkit/fields/plugins/mdx-field-plugin/plate/plugins/ui/autocomplete.tsx
```diff
@@ -41,56 +52,60 @@ export const Autocomplete: React.FC<AutocompleteProps> = ({
...
-                >
-                  {item.render(item)}
-                </button>
-              )}
-            </ComboboxOption>
-          ))}
-        </ComboboxOptions>
-      </Transition>
-    </Combobox>
+        {filteredItems.map((item) => (
```
joshbermanssw: **Two regressions vs the old Headless UI `Combobox`.**

1. **Keyboard nav is gone.** The options are plain `<button>`s inside a Radix Popover (no roving focus), and with `onOpenAutoFocus` prevented, arrow keys and Enter no longer move through or select an option. The format/language picker is mouse-only now; the old `Combobox` supported full keyboard use.
2. **Focusing blanks the selection.** `value={open ? query : displayValue}` with `query` reset to `''` means focusing the field clears the shown label (it survives only as greyed placeholder) instead of showing the current value.

If keyboard use matters here, this wants a roving-focus pattern (a listbox/command menu) rather than a popover of buttons.

packages/tinacms/src/toolkit/components/transition.tsx
```diff
@@ -0,0 +1,251 @@
...
+const waitForAnimations = (el: HTMLElement | null): Promise<void> => {
+  if (!el?.getAnimations) {
+    return Promise.resolve();
+  }
+  const animations = el.getAnimations({ subtree: true });
+  if (animations.length === 0) {
+    return Promise.resolve();
+  }
+  // allSettled, not all: a cancelled transition still counts as "done".
+  return Promise.allSettled(animations.map((a) => a.finished)).then(() => {});
```
kulesy, line 71: **Major:** `waitForAnimations` gates both the leave-unmount ([transition.tsx:177](https://github.com/tinacms/tinacms/pull/7216/files#diff-94188b20f05f6c0047037703845c0fdf82c5062df66cb1931e64b712f4c3f9c0R177)) and the enter settle on `Promise.allSettled` over every subtree animation's `.finished`, and an infinite animation never settles that promise, so any spinner in the subtree stalls the gate. Concrete path: close the sidebar menu while the version check is still in flight, [VersionInfo.tsx:42](https://github.com/tinacms/tinacms/blob/35414c3850dfbd238144049713f91cd493aeb64b/packages/tinacms/src/toolkit/react-sidebar/components/VersionInfo.tsx#L42) renders `animate-spin` inside the leave-gating [sidebar.tsx:263](https://github.com/tinacms/tinacms/pull/7216/files#diff-777440cc4bad9de315188b7c1574cb290e1a37a8f6b3f785b294ff0f6d49f56bR263) Transition, and the invisible `fixed inset-0` overlay ([sidebar.tsx:320](https://github.com/tinacms/tinacms/pull/7216/files#diff-777440cc4bad9de315188b7c1574cb290e1a37a8f6b3f785b294ff0f6d49f56bR320)) stays mounted and swallows every click until the fetch settles, indefinitely on a hung network. Reproduced in headless Chromium with this file verbatim in a harness mirroring the sidebar structure: with the spinner running, real clicks at a page button's coordinates were still intercepted by the leftover overlay at T+0.5s, T+1.5s and T+3s after close, then landed 0.4s after the spinner unmounted (cancellation rejects `.finished`, which `allSettled` treats as done, so it only self-heals once the spinner leaves; a control run without the spinner unmounted within 0.5s). Headless UI keyed off the transitioning element's own duration, so descendant spinners never delayed unmount; skipping animations with `iterations === Infinity` in `waitForAnimations` (or racing a max-duration timeout) restores that behavior.

brookjeynes-ssw (approved):

Issue #7211, opened 2026-07-13
🤖 @tinacms/mdx ships typedoc and uvu as production dependencies


wicksipedia:
`@tinacms/mdx` declares development tooling in its **`dependencies`**, so every TinaCMS user installs it.

## Finding

From `packages/@tinacms/mdx/package.json` `dependencies`:

### `typedoc` — not used at all

A documentation generator. **It is not imported anywhere in `src/`.** Its only reference in the whole package is the `docs` npm script:

```json
"scripts": {
  "docs": "typedoc --plugin typedoc-plugin-markdown src/parse/plate.ts --theme markdown && concat-md ..."
}
```

It is a pure `devDependency` sitting in `dependencies`. Note `typedoc-plugin-markdown` and `concat-md` — used by the *same script* — are already correctly in `devDependencies`, which makes this look like a straightforward slip.

Free to move. Measured at 7.7 MB across copies.

### `uvu` — a test framework, imported by production code

`uvu` is in `dependencies` and *is* imported from `src/` — but only for a single assertion helper, in two files:

```
src/extensions/tina-shortcodes/shortcode-container.ts:9:  import { ok as assert } from 'uvu/assert';
src/next/shortcodes/lib/factory-tag.ts:1:            import { ok as assert } from 'uvu/assert';
```

We are shipping a test runner to production to get `assert(cond, msg)`. Replace with three lines:

```ts
function assert(value: unknown, message?: string): asserts value {
  if (!value) throw new Error(message ?? 'Assertion failed');
}
```

### `prettier`

`prettier` is also in `dependencies`, but it is legitimately used at runtime to stringify MDX (`src/stringify/acorn.ts`, `src/next/stringify/acorn.ts`). **Not part of this issue.**

There is a separate PR to remove prettier - #7409 

## Proposed fix

1. Move `typedoc` from `dependencies` to `devDependencies` in `packages/@tinacms/mdx/package.json`.
2. Replace `uvu/assert` with a local `assert` helper in the two files above; remove `uvu` from `dependencies`.
3. While in there, audit the rest of `@tinacms/mdx`'s `dependencies` for other dev-only entries.

Small in isolation, but it is a correctness issue as much as a size one — a docs generator and a test runner have no business in a consumer's production dependency tree.

## Context

[...]

Issue #7210, opened 2026-07-13
🤖 @tinacms/cli ships the TypeScript compiler as a prod dep for a one-time init command (~88 MB)


wicksipedia:
A stock Astro + TinaCMS blog installs **1412 MB** of `node_modules`. **The TypeScript compiler accounts for 87.9 MB of it** (×4 copies), and it is only ever used by a command most users run once.

## Finding

`@tinacms/cli` declares `typescript` in **`dependencies`**, not `devDependencies`:

```json
"dependencies": { "typescript": "^5.7.3" }
```

It is genuinely imported at runtime — but **only by the `tinacms init` scaffolding command**:

```
packages/@tinacms/cli/src/cmds/init/codegen/index.ts:1:  import ts from 'typescript';
packages/@tinacms/cli/src/cmds/init/codegen/util.ts:1:   import ts from 'typescript';
```

Nothing in `tinacms dev` or `tinacms build` touches it. So every user ships the full TypeScript compiler to support a one-time scaffold — and, because it is a hard dependency of the CLI, it is **duplicated against the project's own `typescript`** rather than shared (measured: 4 copies, 87.9 MB).

## Proposed fix

Move `typescript` to an **optional `peerDependency`** and `import()` it lazily inside the init codegen path.

```jsonc
// packages/@tinacms/cli/package.json
"peerDependencies": { "typescript": ">=5.0.0" },
"peerDependenciesMeta": { "typescript": { "optional": true } },
"devDependencies": { "typescript": "^5.7.3" }
```

```ts
// cmds/init/codegen/index.ts
const ts = await import('typescript').catch(() => {
  throw new Error(
    'tinacms init requires TypeScript. Install it with: npm i -D typescript'
  );
});
```

Almost every Astro/Next/Remix project already has `typescript` installed, so in practice the peer resolves to the copy the user already has — which is also the *correct* one to generate their config against. Projects that genuinely have no TypeScript get a clear, actionable error from the one command that needs it, instead of a silent 23 MB tax on every install.

**Note this is not a simple `dependencies` → `devDependencies` move.** The init command needs it at runtime; the point is to stop *bundling* a compiler and start *borrowing* the user's.

## Related

- #7207 — internal deps publish as exact pins, which is what turns this one copy into four.

## Acceptance

- `typescript` is not installed by `npm i @tinacms/cli` in a project that has no TypeScript.
- `tinacms init` works in a project that has TypeScript; errors clearly in one that does not.
[...]

Issue #7207, opened 2026-07-13
🤖 Internal deps publish as exact pins (workspace:*), forcing ~320 MB of duplicate trees


wicksipedia:
A stock Astro + TinaCMS blog — no libraries beyond Astro and TinaCMS — installs **1412 MB** of `node_modules` (npm 10.9.3, Node 22.20, macOS).

The single largest cause is not any one fat package. It is **duplicate copies of `tinacms` and its ~85 dependencies**, caused by how we publish our internal package references.

## Root cause

`@tinacms/cli` and `@tinacms/app` both declare:

```json
"dependencies": { "tinacms": "workspace:*" }
```

pnpm expands `workspace:*` at publish time to an **exact version**, not a range:

```
@tinacms/cli@2.5.3  ->  "tinacms": "3.10.0"
@tinacms/app@2.5.8  ->  "tinacms": "3.10.0"
```

An exact pin cannot dedupe against a consumer's own semver range. As soon as the user's `tinacms` resolves to *anything* other than the pinned version, npm cannot hoist a shared copy and instead nests a **second and third full `tinacms` tree**, each with its own complete dependency set.

**This will bite every user continuously, not just those on an old version.** A user installs `tinacms@^3.10.0` today and dedupes fine. The moment `tinacms@3.11.0` ships, their range resolves to 3.11.0, the CLI still demands exactly 3.10.0, and they silently gain a duplicate tree on their next install.

## Measured impact

Observed in a real project (`tinacms@2.10.1` + `@tinacms/cli@2.5.1`, which pins `tinacms@3.9.3`) — three `tinacms` trees on disk:

| Package | Copies | Total on disk |
|---|---:|---:|
| mermaid | **×3** | 186.5 MB |
| date-fns | **×5** | 150.7 MB |
| typescript | **×4** | 87.9 MB |
| esbuild + native binary | ×3 | 57.4 MB |
| prettier | ×4 | 42.0 MB |
| better-sqlite3 | ×3 | 35.3 MB |
| tinacms itself | ×3 | 18.7 MB |

Aligning the version so the pin is satisfiable collapses this to **one copy each** and takes the install from **1412 MB → 1091 MB (−321 MB)** with no other change.

## Proposed fix

Change `workspace:*` to `workspace:^` for internal dependencies, so they publish as a caret range (`^3.10.0`) that can dedupe with the consumer's own resolution:

```diff
- "tinacms": "workspace:*"
+ "tinacms": "workspace:^"
```

[...]

PR #7218, opened 2026-07-13
🐛 Remove typescript from @tinacms/app's production dependencies (~23 MB)

wicksipedia:
Tagged [tina-io-git-jb-test-7218-tinacms.vercel.app](https://tina-io-git-jb-test-7218-tinacms.vercel.app/)

Partial fix for #7210.

## This PR

`@tinacms/app` declared `typescript` in **both** `dependencies` and `devDependencies`. Nothing in `src/` imports it — the package ships raw source and is compiled by `@tinacms/cli`. So every TinaCMS user installed a second copy of the TypeScript compiler (~23 MB) for nothing.

Removed from `dependencies`; it stays a devDependency for type-checking.

- `pnpm build` — 23/23 tasks pass
- Typecheck unchanged: **586 errors on `main`, 586 on this branch**

## Why the rest of #7210 is now blocked

#7210 proposed moving `@tinacms/cli`'s `typescript` to an **optional `peerDependency`** and lazily importing it, so we borrow the user's copy instead of shipping our own. **That plan has a problem, and it's worth writing down before anyone picks this up.**

`@tinacms/cli` genuinely uses the TypeScript compiler API — but only in the `tinacms init` codegen path:

```
packages/@tinacms/cli/src/cmds/init/codegen/index.ts:1:  import ts from 'typescript';
packages/@tinacms/cli/src/cmds/init/codegen/util.ts:1:   import ts from 'typescript';
```

It uses ~18 APIs: `ts.factory`, `ts.transform`, `ts.createSourceFile`, `ts.visitEachChild`, `ts.visitNode`, `ts.ScriptTarget`, and a dozen `ts.isX` type guards.

**TypeScript 7 — the native Go port — is now `latest` on npm** (`7.0.2`; `6.0.3` is the last JS release). And in TS 7 the main entry point no longer exports the compiler API at all:

```jsonc
// typescript@7.0.2 package.json
"exports": {
  ".": "./lib/version.cjs",          // <- exports ONLY { version, versionMajorMinor }
  "./unstable/ast": "./dist/ast/index.js",
  "./unstable/ast/factory": "./dist/ast/factory.generated.js",
  "./unstable/ast/visitor": "./dist/ast/visitor.js",
  "./unstable/ast/is": "./dist/ast/is.js",
  // ...
}
```

Verified against a real install:

```
import ts from 'typescript';
ts.createSourceFile  ->  undefined
ts.factory           ->  undefined
Object.keys(ts)      ->  [ 'version', 'versionMajorMinor' ]
```

So if we make `typescript` a peer dependency, a user who runs `npm i -D typescript` gets **7.x**, we import it, and `tinacms init` breaks. It works *today* precisely because we bundle our own TS 5.

**Also note: bumping to TS 7 would not save disk anyway.** It installs at **30 MB** (3.5 MB of JS + a 26.5 MB native Go binary, shipped as 20 per-platform optional deps, the esbuild model) versus ~23 MB for TS 5.x. It's *bigger*.

### The path forward

The AST API does still exist in TS 7, under the explicitly-`unstable` subpaths (`typescript/unstable/ast/factory` — 370 exports, `/is` — 347, `/visitor` — 8). So porting `cmds/init/codegen/` is possible, but it means depending on an API Microsoft has namespaced "unstable", and it should be its own issue rather than a rider on a size fix.

Until then, the CLI's `typescript` stays a regular dependency. Suggested sequencing:

1. **This PR** — drop the unused copy in `@tinacms/app`. Free, no risk. ✅
2. **#7211** — removing `typedoc` from `@tinacms/mdx` kills *another* typescript copy (`@tinacms/mdx` → `typedoc@0.26.11` → `typescript@5.6.3`). Also free.
3. **#7207** — collapsing the duplicate `tinacms` trees removes the rest of the duplication.
4. **Later** — port the init codegen to `typescript/unstable/ast/*`, *then* make it an optional peer.

Steps 1–3 recover most of the measured 88 MB (4 copies) with no behaviour change and no TS 7 exposure.


joshbermanssw (dismissed):

18-th (approved):
LGTM

Issue #7221, opened 2026-07-13
🤖 Duplicate copies of identical packages — lodash ×6, zod ×5, prettier ×3 (~83 MB)


wicksipedia:
After the fixes in #7213, #7214, #7216 and #7218, a stock Astro + TinaCMS blog still installs **916 MB** of `node_modules` (down from 1412 MB). For scale: **the same blog with Astro and no TinaCMS at all is 173 MB** — so TinaCMS still accounts for ~743 MB of it.

A chunk of that isn't a big dependency at all. It's the **same package installed several times over**.

## The finding

In a real install:

| Package | Copies | Versions |
|---|---:|---|
| `lodash` | **6** | 5 × `4.17.23` (identical) + 1 × `4.18.1` |
| `@graphql-codegen/plugin-helpers` | **6** | 5 × `5.1.1` (identical) + 1 × `7.1.0` |
| `zod` | **5** | 4 × `3.25.76` (identical) + 1 × `4.4.3` |
| `prettier` | **3** | 2 × `2.8.8` (identical) + 1 × `3.9.5` |

npm hoists one copy of a package to the top of `node_modules` and shares it. Anything that can't use the hoisted version gets its own nested copy. In every row above, **the top-level slot is held by a different major**, so each consumer of the other major nests its own — even though those nested copies are byte-identical to one another.

## Cause 1 — we evict our own dependency (fixed in #7221)

`@tinacms/cli` declares `@graphql-codegen/plugin-helpers` **directly** at `^7.0.1`. But the five codegen plugins it actually uses — `core`, `typescript`, `typescript-operations`, `visitor-plugin-common`, and the transitive `schema-ast` — all want `^5`.

Our v7 pin wins the top-level slot, so all five plugins nest their own `plugin-helpers@5.1.1`. Each of those needs `lodash` at `~4.17.0`, which cannot use the hoisted `lodash@4.18.1` either — hence five nested lodashes.

**We were causing this ourselves.** Aligning our direct dependency to `^5.1.0` lets one `plugin-helpers` hoist and be shared: **lodash goes from 6 copies to 1.**

Also in scope: `@graphql-codegen/core` was on `^2.6.8` (2023) while the other plugins were on v4, and `@graphql-inspector/core` was `^4.2.2` in the CLI but `^6.2.1` in `tinacms` — so both majors were installed.

**Measured: 1091 MB → 1065 MB (~26 MB), with byte-identical codegen output.**

[...]

PR #7222, opened 2026-07-13
🐛 Deduplicate the CLI's graphql-codegen tree — lodash ×6 → ×1 (~26 MB)

wicksipedia:
Tag [tina-io-git-jb-test-7222-tinacms.vercel.app](https://tina-io-git-jb-test-7222-tinacms.vercel.app/)

Addresses the first half of #7221.

## The bug

`@tinacms/cli` declares `@graphql-codegen/plugin-helpers` **directly** at `^7.0.1`. But the five codegen plugins it actually uses — `core`, `typescript`, `typescript-operations`, `visitor-plugin-common`, and the transitive `schema-ast` — all want `^5`.

npm hoists one copy of a package to the top of `node_modules`. **Our v7 pin wins that slot**, so all five plugins nest their own `plugin-helpers@5.1.1`. And each of those needs `lodash` pinned at `~4.17.0`, which can't use the hoisted `lodash@4.18.1` either — so each nests a lodash too.

The result, in a real install:

```
  node_modules/@graphql-codegen/plugin-helpers                      7.1.0   <- ours, hoisted
  node_modules/@graphql-codegen/core/.../plugin-helpers             5.1.1
  node_modules/@graphql-codegen/schema-ast/.../plugin-helpers       5.1.1
  node_modules/@graphql-codegen/typescript/.../plugin-helpers       5.1.1
  node_modules/@graphql-codegen/typescript-operations/.../…         5.1.1
  node_modules/@graphql-codegen/visitor-plugin-common/.../…         5.1.1
```

Six copies of `plugin-helpers` and six of `lodash` — five of each byte-identical to the others. **We were causing it ourselves.**

## The fix

Align our direct dependency **down** to `^5.1.0` so one `plugin-helpers` hoists and is shared by everything.

Also aligned, same theme:
- `@graphql-codegen/core` `^2.6.8` (2023) → `^4.0.2`, so it wants the same `plugin-helpers` major as the other plugins
- `@graphql-inspector/core` `^4.2.2` → `^6.2.1` in the CLI, matching `tinacms` (both majors were being installed)

Three version numbers. No code changes.

## Verification

**Behaviour is unchanged — proven, not assumed.** Ran `tinacms build` on `examples/next/kitchen-sink` before and after and diffed the codegen output:

| file | result |
|---|---|
| `types.ts` | ✅ **byte-identical** |
| `frags.gql` | ✅ **byte-identical** |
| `queries.gql` | ✅ **byte-identical** |

- `pnpm build` — 23/23
- Typecheck unchanged (586 = 586)
- Measured on a real project install: **1091 MB → 1065 MB**
- `lodash`: **6 copies → 1**. `plugin-helpers`: **6 → 1**.

## What I tried and rejected

Worth recording so nobody repeats it.

**Bumping the codegen plugins to their latest majors** looks like a bigger win — v7 of `plugin-helpers` drops lodash entirely. But it **changes the generated types**:

- the `JSON` and `Reference` scalars go from `{ input: any; output: any }` to `{ input: unknown; output: unknown }` — that breaks user code at compile time
- `Exact`, `MakeOptional`, `MakeMaybe`, `MakeEmpty` stop being exported from `types.ts`
- `types.ts` grows 2155 → 2408 lines with reordered content

That's a breaking change for anyone consuming the generated types, and belongs in a major release rather than a size fix.

**Bumping `@graphql-codegen/core` alone** (2 → 4) changed nothing at all — the top-level slot was still held by our v7 pin. That's what pointed at the real cause.

## Not in this PR

`zod` (~21 MB, ×5) and `prettier` (~20 MB, ×3) have the same *shape* of problem but a different cause: we're on the older major while the ecosystem is on the newer one (Astro pulls `zod@4` to the top; every modern project has `prettier@3`), so our copies can never win the slot. Both need real API migrations — tracked in #7221.


wicksipedia:
Added a test for `generateTypes` (11f84de), and it turned up something worth knowing.

## The codegen path had no tests

`src/next/codegen/index.test.ts` mocks `./codegen` out entirely:

```ts
jest.mock('./codegen', () => ({
  generateTypes: jest.fn().mockResolvedValue(''),
}));
```

So `generateTypes` — the graphql-codegen invocation that produces `types.ts` — had **zero coverage**. A dependency bump that silently changed the generated output would go green.

That isn't hypothetical. While exploring whether we could bump graphql-codegen to its **current** majors (v6/v7, which drop lodash entirely), the generated output changed in a way that **breaks users**:

- unmapped scalars flip from `any` to `unknown`. Rich-text bodies (`_body`, `excerpt`, `text`) ride on the `JSON` scalar and get handed straight to `<TinaMarkdown content={...} />`, which takes `TinaMarkdownContent` — `unknown` doesn't assign to that. Six call sites in `examples/next/kitchen-sink` alone, i.e. the canonical pattern from every doc.
- `Exact`, `MakeOptional`, `MakeMaybe` and `MakeEmpty` stop being exported.

**`pnpm test` would not have caught any of it.**

The new test asserts the contract the generated module has always had — `JSON` and `Reference` are `any`, and `Exact`/`Maybe`/`InputMaybe` are exported. It passes on `main` and on this branch. It fails on the v6/v7 bump, which is the point.

Full CLI suite: **24 suites, 256 tests, green.**

## A side-finding about the v6/v7 bump

I have a working spike of the full v6/v7 upgrade — `defaultScalarType: 'any'` plus a one-line re-export of `Exact` makes it **provably non-breaking** (kitchen-sink: 895 type errors before, 895 after, zero new), and it saves ~33 MB instead of the ~26 MB here, because `plugin-helpers@7` drops lodash outright.

But it has a cost that only showed up once the test existed: **codegen v6 can't be loaded under jest at all.** Its CJS build `require()`s `auto-bind@5`, which is ESM-only:

```
SyntaxError: Unexpected token 'export'
  at @graphql-codegen/typescript/cjs/introspection-visitor.js
```

The CLI is ESM at runtime so it works in production — but the test harness can't load it, which means the upgrade also needs jest ESM work. v4 uses `auto-bind@4` (CJS) and is fine.

So this PR stays as-is: the safe, byte-identical 26 MB. The v6/v7 upgrade is a bigger piece of work than a version bump, and now we have a test that will hold it honest when someone takes it on.

wicksipedia:
Superseded by #7228 — same bug, better fix.

This PR aligned `@graphql-codegen/plugin-helpers` **down** to `^5` so one copy could hoist and be shared: safe, byte-identical output, **~26 MB**.

#7228 goes **up** to v7 instead. `plugin-helpers@7` drops lodash outright, so there's nothing left to align around: **~33 MB**, and the surviving lodash is `4.18.1` (the Plate editor's) rather than being dragged back to the `4.17.x` line by our pin — which matters for advisories. It also gets us off a 2023-era codegen major rather than pinning us to it.

v7 needed three things, all handled there:

- jest couldn't load codegen v6 at all (`auto-bind@5` is ESM-only; Node 22 can `require()` it, jest's module registry can't). Babel now down-levels that one package.
- `@graphql-inspector/core` v6 turns `ChangeType` from an enum into a const — the same fix I pushed to this branch.
- codegen v5+ defaults unmapped scalars to `unknown` and stops exporting `Exact`. Both pinned back, so the generated output is unchanged.

The two are mutually exclusive, so I'd close this in favour of #7228. Keeping it open for now as a fallback in case the jest/babel addition isn't wanted — this branch is green on the full suite (58/58, uncached) after the `ChangeType` fix.

---

Worth recording: **this PR was broken when I first pushed it.** The `@graphql-inspector/core` v4 → v6 bump changed `ChangeType` from an enum (usable as value *and* type) to a const, and `pnpm test` failed on it. I missed it because I ran `pnpm build` and `pnpm --filter @tinacms/cli test` — neither runs the `types` task, but turbo's `test` task `dependsOn` it. Fixed in 18dc547; the full suite is green now.

joshbermanssw (dismissed):

joshbermanssw (approved):

PR #7233, opened 2026-07-13
⚡ Give @tinacms/mdx a light sanitize-url subpath so rich-text skips the parser

### LINKED ISSUE #7232: 🤖 TinaMarkdown's rich-text renderer pulls in @tinacms/mdx's full ~2MB markdown parser just for sanitizeUrl
### The problem

`packages/tinacms/src/rich-text/index.tsx` and `packages/tinacms/src/rich-text/static.tsx` (the source of the `TinaMarkdown` / `StaticTinaMarkdown` components that every TinaCMS site renders, including at a `'use client'` boundary) both do:

```ts
import { sanitizeUrl } from '@tinacms/mdx';
```

`sanitizeUrl` is a small (~40 line, including comments/whitespace) URL-scheme allowlist check defined in `packages/@tinacms/mdx/src/parse/remarkToPlate.ts`. But `@tinacms/mdx` has a single entry point (`src/index.ts`), built with `buildConfig.entryPoints[0].bundleDeps: true` in its `package.json` — so importing anything from the package root, including `sanitizeUrl`, pulls in the entire remark / remark-gfm / remark-mdx / mdast-util-* / micromark-* / prettier markdown-parsing toolchain that the rest of `@tinacms/mdx` bundles for MDX parsing.

### Measured

Built `packages/@tinacms/mdx` from the current `main` (commit `711ba30f6`) with `pnpm exec turbo run build --filter=@tinacms/mdx`:

- `packages/@tinacms/mdx/dist/index.js` (Node/SSR entry): **2,013,349 bytes** (~1.92 MB)
- `packages/@tinacms/mdx/dist/index.browser.js` (browser-condition entry): **1,976,351 bytes** (~1.88 MB)

Built `packages/tinacms` on the same commit and inspected the emitted `dist/rich-text/index.js`:

```js
import { sanitizeUrl } from "@tinacms/mdx";
```

Since `@tinacms/mdx` is listed as an external dependency (not bundled) of `tinacms`, this import is resolved by the consuming app's own bundler at build time — meaning every site's rich-text/client bundle transitively pulls in the full ~1.9–2.0 MB `@tinacms/mdx` bundle just to get a ~40-line URL sanitizer.

### Proposed fix

Give `@tinacms/mdx` a dedicated, dependency-free subpath export (e.g. `@tinacms/mdx/sanitize-url`) with its own tiny entry point containing only `sanitizeUrl`, following the existing multi-entry-point pattern already used by `@tinacms/bridge` (`exports` map + `buildConfig.entryPoints`, each built as a separate bundle by `@tinacms/scripts`). Point `packages/tinacms/src/rich-text/index.tsx` and `static.tsx` at the new subpath instead of the package root. Keep `sanitizeUrl` exported from the root `@tinacms/mdx` entry too, so this is purely additive/non-breaking.

### Your environment

- This is a source-level architecture/bundle-size issue in the `@tinacms/mdx` and `tinacms` packages themselves, not specific to any one site's setup — reproducible in the reference `examples/next/kitchen-sink` app or any site that imports `TinaMarkdown` from `tinacms/dist/rich-text`.
- Affects any consumer on the current `@tinacms/mdx` / `tinacms` versions published from `main`.

wicksipedia:
Tagged: [tina-io-git-jb-test-7233-tinacms.vercel.app](https://tina-io-git-jb-test-7233-tinacms.vercel.app/)

## Summary

- `packages/tinacms/src/rich-text/index.tsx` and `static.tsx` (source of `TinaMarkdown` / `StaticTinaMarkdown`) imported `sanitizeUrl` from the root `@tinacms/mdx` entry, which is built with `bundleDeps: true` and inlines the entire remark/mdast/micromark/prettier markdown-parsing toolchain (~2MB) — even though `sanitizeUrl` is a ~40-line URL-scheme check.
- Adds a dedicated, dependency-free `@tinacms/mdx/sanitize-url` subpath (own entry point in `buildConfig.entryPoints`, own `exports` map entry — same multi-entry pattern `@tinacms/bridge` already uses) and points the rich-text renderer at it instead.
- The root `@tinacms/mdx` export of `sanitizeUrl` is unchanged (still re-exported from `parse/remarkToPlate.ts`), so this is purely additive/non-breaking.

Fixes #7232

## Measured before/after

Built both packages from `main` (`711ba30f6`) vs. this branch with `pnpm exec turbo run build --filter=@tinacms/mdx --filter=tinacms --force`, then measured the actual emitted files:

| | Before | After |
|---|---|---|
| `tinacms/dist/rich-text/index.js` import | `from "@tinacms/mdx"` | `from "@tinacms/mdx/sanitize-url"` |
| what that import resolves to | `@tinacms/mdx/dist/index.js` — **2,013,349 bytes** (Node/SSR) or `dist/index.browser.js` — **1,976,351 bytes** (browser condition) | `@tinacms/mdx/dist/sanitize-url.js` — **658 bytes** |
| `rich-text/index.js` itself | 12,493 bytes | 12,506 bytes (+13 bytes, longer import specifier) |
| **total pulled in by a site's rich-text import** | **~1.99–2.03 MB** | **13,164 bytes (~12.9 KB)** |

That's a ~99.3% reduction — the new `sanitize-url.js` was also grepped to confirm it contains no `remark`/`mdast`/`micromark`/`prettier` references.

## Test plan

- [x] `pnpm test` from repo root (full suite, not filtered) — **58/58 tasks passed**, including `@tinacms/mdx`'s existing `sanitizeUrl` unit tests (`remarkToPlate.test.ts`, 21 tests, unchanged) and `tinacms`'s full vitest suite (444 passed, 5 skipped)
- [x] Verified `packages/@tinacms/mdx/dist/sanitize-url.js` is 658 bytes and free of remark/mdast/micromark/prettier imports
- [x] Verified `packages/tinacms/dist/rich-text/index.js` now imports `@tinacms/mdx/sanitize-url`, not the `@tinacms/mdx` root
- [x] Verified `tsc` (the `types` task) emits a real `.d.ts` for the new entry point
- [x] Added a changeset (`@tinacms/mdx`: minor, `tinacms`: patch)

### DIFF
--- a/.changeset/mdx-sanitize-url-subpath.md
+++ b/.changeset/mdx-sanitize-url-subpath.md
@@ -1,0 +1,6 @@
+---
+"@tinacms/mdx": minor
+"tinacms": patch
+---
+
+Add a dedicated `@tinacms/mdx/sanitize-url` subpath export containing just the URL-scheme sanitizer, and point `tinacms`'s rich-text renderer (`TinaMarkdown` / `StaticTinaMarkdown`) at it instead of the root `@tinacms/mdx` entry. Previously, importing `sanitizeUrl` pulled in `@tinacms/mdx`'s full remark/mdast/micromark markdown-parsing bundle (~2MB) into every site's client bundle, even though rich-text rendering only needs the ~15-line sanitizer. The root `@tinacms/mdx` export of `sanitizeUrl` is unchanged and still works.
--- a/packages/@tinacms/mdx/package.json
+++ b/packages/@tinacms/mdx/package.json
@@ -14,6 +14,10 @@
 			"types": "./dist/index.d.ts",
 			"browser": "./dist/index.browser.js",
 			"default": "./dist/index.js"
+		},
+		"./sanitize-url": {
+			"types": "./dist/sanitize-url.d.ts",
+			"default": "./dist/sanitize-url.js"
 		}
 	},
 	"license": "Apache-2.0",
@@ -23,7 +27,8 @@
 				"name": "src/index.ts",
 				"target": "node",
 				"bundleDeps": true
-			}
+			},
+			"src/sanitize-url.ts"
 		]
 	},
 	"scripts": {
--- a/packages/@tinacms/mdx/src/index.ts
+++ b/packages/@tinacms/mdx/src/index.ts
@@ -6,4 +6,4 @@ export * from './parse/plate';
 
 export { parseMDX };
 export { serializeMDX };
-export { sanitizeUrl } from './parse/remarkToPlate';
+export { sanitizeUrl } from './sanitize-url';
--- a/packages/@tinacms/mdx/src/parse/remarkToPlate.ts
+++ b/packages/@tinacms/mdx/src/parse/remarkToPlate.ts
@@ -8,8 +8,9 @@ import type { RichTextType } from '@tinacms/schema-tools';
 import type * as Md from 'mdast';
 import type { ContainerDirective } from 'mdast-util-directive';
 import type { MdxJsxFlowElement, MdxJsxTextElement } from 'mdast-util-mdx-jsx';
-import { directiveElement, mdxJsxElement as mdxJsxElementDefault } from './mdx';
+import { sanitizeUrl } from '../sanitize-url';
 import { parseMarkMdxText } from './mark';
+import { directiveElement, mdxJsxElement as mdxJsxElementDefault } from './mdx';
 import type * as Plate from './plate';
 
 export type { Position, PositionItem } from './plate';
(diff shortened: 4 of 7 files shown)


joshbermanssw (approved):
Approving. Built the branch and verified the numbers rather than reading them: `dist/sanitize-url.js` is 658 bytes with no remark/mdast/micromark/prettier in it, `dist/index.js` is untouched at 2,013,373 bytes, both rich-text entries emit the subpath import, the subpath and root export resolve identically at runtime, and the `types` task emits a real `.d.ts` (not the build-time stub). mdx tests and the tinacms typecheck pass. The sanitizer body is a byte-identical move, which is the right call for a security-relevant function.

Two follow-ups, neither blocking:

1. **Latent footgun worth a comment.** `@tinacms/scripts` (`src/index.ts:411-435`) hardcodes `dist/index.js` / `dist/index.browser.js` as outfiles in the `@tinacms/mdx` branch, ignoring the entry's computed output path. The new entry is safe *only because* it's a plain string, which normalizes to `target: 'browser'` and takes the Vite path instead. If someone later rewrites it in object form with `target: 'node'` to match its sibling (which reads as harmless tidying), it silently overwrites the 2MB root bundle with the 658-byte sanitizer. Worth a note on the entry in `package.json`, or fixing the script to use `outInfo.outfile`.

2. **The description's mechanism isn't quite right.** `bundleDeps` isn't read anywhere in `@tinacms/scripts`; the `Entry` type has `bundle: string[]` (graphql/datalayer only) and `target`. mdx's dep bundling comes from that hardcoded package-name branch. The outcome is correct, but the causal story will mislead whoever touches this next.

Nits: the `sanitizeUrl` tests still live in `parse/remarkToPlate.test.ts` and import from `./remarkToPlate`, so they exercise the re-export shim rather than the new module (`src/sanitize-url.test.ts` would be the natural home). And the changeset says "~15-line sanitizer" while the PR body says "~40-line"; the changeset is the one that lands in the public CHANGELOG.

brookjeynes-ssw (approved):

PR #7251, opened 2026-07-15
fix(tinacms): show the rich-text link popover above form fields

### LINKED ISSUE #7250: 🐛 Rich-text link popover does not appear (hidden behind form fields)
### The exact error message

No error is thrown. Clicking the link button in the rich-text editor does nothing visible, the URL input popover never appears.

### Steps to reproduce

1. Open any document with a `rich-text` (body) field in the CMS.
2. In the editor, select some text.
3. Click the "insert link" (chain) button in the toolbar (or the link button in the floating toolbar).

The popover mounts and is correctly positioned, but it is painted behind the form and is invisible, so nothing appears to happen. The text selection also clears (the popover's URL input steals focus).

**Expected:** a small popover with a "Paste link" URL input, a "text to display" field, and Cancel/OK buttons appears next to the selection.

**Actual:** nothing visible appears; no link can be added or edited.

### Your environment

- `tinacms` 3.10.0 / 3.10.1
- Any framework; reproduced in local dev and standalone admin.

### Root cause

Regression from #7131. That PR moved the link popover into a `PortalBody` on `document.body` with a `z-[999999]` class. `@udecode/plate-floating` writes an inline `z-index: 50` that overrides the class, and at the body level `50` loses to Tina's form field wrappers (`wrap-field-with-meta` sets `z-index: 1000 - index`, up to 1000), so the opaque field paints over the popover. Before #7131 the popover lived inside the editor's own stacking context, where `z-50` was fine (the pre-#7131 bug was purely off-screen positioning on long bodies).

kulesy:
Closes #7250

**TL;DR** Adding or editing a link in the rich-text editor did nothing visible; the popover opened but rendered behind the form. This puts it back on top. Regression from #7131, reported on Discord (tinacms 3.10.0).

**Pain:** Clicking the link button in the rich-text toolbar (or floating toolbar) appeared to do nothing, no URL input showed up. Since #7131 moved the popover into a `PortalBody` on `document.body`, `plate-floating`'s inline `z-index: 50` overrides its `z-[999999]` class. At the body level that loses to Tina's form field wrappers (`wrap-field-with-meta` sets `z-index: 1000 - index`, up to 1000), so the opaque field paints over the popover. It was mounting, correctly positioned, and interactive the whole time, just hidden behind the form.

**Solution:** Override the inline z-index on both the insert and edit popovers so the intended `z-[999999]` wins and the popover sits above the form field wrappers.

### DIFF
--- a/.changeset/link-popover-behind-form-fields.md
+++ b/.changeset/link-popover-behind-form-fields.md
@@ -1,0 +1,7 @@
+---
+"tinacms": patch
+---
+
+Fix the rich-text link popover not appearing when adding or editing a link.
+
+Since the popover was moved into a portal on `document.body`, `plate-floating`'s inline `z-index: 50` overrode its `z-[999999]` class, so it rendered behind the form field wrappers (which use z-index up to 1000) and was invisible. It now sits above them, so clicking the link button shows the URL input as expected.
--- a/packages/tinacms/src/toolkit/fields/plugins/mdx-field-plugin/plate/components/plate-ui/link-floating-toolbar.tsx
+++ b/packages/tinacms/src/toolkit/fields/plugins/mdx-field-plugin/plate/components/plate-ui/link-floating-toolbar.tsx
@@ -237,7 +237,10 @@ export function LinkFloatingToolbar({
         {...insertProps}
         // Override plate-floating's inline z-index (which beats the z-[999999]
         // class) so the portaled popover sits above the form field wrappers.
-        style={{ ...(insertProps.style as React.CSSProperties), zIndex: 999999 }}
+        style={{
+          ...(insertProps.style as React.CSSProperties),
+          zIndex: 999999,
+        }}
       >
         {input}
       </div>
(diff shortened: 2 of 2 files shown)


kulesy:
### Evidence

Reproduced on a long article (tall body with a bulleted list) at tinacms 3.10.1. Selecting a list item and clicking the toolbar link button.

**Before** (real UI): nothing appears. The popover is actually in the DOM (`display:block`, `opacity:1`, positioned right below the selection, URL input focused, stable across a 50ms-interval time sample), but `document.elementFromPoint` at its center returns the editor's `<ul>`, not the popover: it is occluded. The occluder is the field wrapper at `z-index: 966`; the popover's computed z-index is `50`.

before

Figure: 3.10.1 as shipped. The link popover is mounted and correctly positioned, but painted behind the form field, so nothing is visible.

**After**: raising the popover's z-index above the field wrappers (the exact effect of this PR's one-line change) makes it appear in place. Screenshot below is a runtime z-index override in the same live repro (not a from-source build); `elementFromPoint` at the center now returns the popover's `INPUT`.

after

Figure: With the popover z-index raised above the form field wrappers, the "Paste link" input, "text to display" field, and Cancel/OK buttons render in place.

isaaclombardssw (approved):
✅ 999999 is extreme but this is fine

PR #7282, opened 2026-07-16
Update the env variable fron next to vite

18-th:
Finishing the starter migration from Next to React + Vite by renaming the env variable passed to the starter to execute correctly


KahaMason (approved):
LGTM 🚀

Issue #7114, opened 2026-06-29
global collection regression or `ui.global` documentation out of sync


joeyfigaro:
### The exact error message

```shell
There's no error message this time—it's just impossible to get a single-document global collection to be treated differently in the sidebar even with the exact example cited in the docs for global collections.
```

<img width="2090" height="1230" alt="Image" src="https://github.com/user-attachments/assets/cda6dbf3-9759-464d-b964-b6cfdfb68b9c" />

<img width="695" height="945" alt="Image" src="https://github.com/user-attachments/assets/2b85ab71-1ff5-4c1b-af96-3d0e9341c3c3" />

### Steps to reproduce

1. Setup fresh TinaCMS install in a vanilla React v19 Vite app
2. Copy the snippet for a global collection [from the docs](https://tina.io/docs/reference/collections#global-collections)
3. Update the collections in your `tina/config.ts` with the snippet and save
4. Confirm that a new `Global` collection is added to the sidebar under `Collections` instead of `Site`

### What you expected vs. what actually happened

Expected: `Global` shows up under `Site` or is treated/marked differently than collections with multiple documents
Actual: `Global` shows up under `Collections` 

### Your environment

- TinaCMS v3.9.3
- @tinacms/cli v2.5.1
- Vite v8.0.12 + TypeScript v6.0.3 + React v19
- Local only—no deployed environment(s)

### Relevant sections of your schema file

```typescript
schema: {
    collections: [
      ...,
      {
        label: 'Global',
        name: 'global',
        path: 'content/global',
        format: 'json',
        ui: {
          global: true,
          allowedActions: {
            create: false,
            delete: false,
          },
        },
        fields: [
          {
            name: 'bundle_download_url',
            label: 'Bundle Download URL',
            type: 'string'
          }
        ],
      },
      {
        label: 'Navigation',
        name: 'navigation',
        path: 'content/navigation',
        format: 'json',
        ui: {
          global: true,
          allowedActions: {
            create: false,
            delete: false,
          },
        },
        fields: [
          {
            name: 'enabled',
            label: 'Enable Navigation',
            type: 'boolean'
          }
        ],
      },
    ],
  },
```

### Validations

- [x] Follow our [Code of Conduct](https://github.com/tinacms/tinacms/blob/main/CODE_OF_CONDUCT.md)
- [x] Read the [docs](https://tina.io/docs).
- [x] Check that there isn't [already an issue](https://github.com/tinacms/tinacms/issues) that reports the same bug.

[...]

PR #7283, opened 2026-07-16
feat(@tinacms/cli): warn on version skew between tinacms packages at startup

### LINKED ISSUE #7284: Warn when tinacms and @tinacms/* package versions are skewed
### Pain

When a package manager holds one core package back (partial upgrade, pnpm `minimumReleaseAge`, a stale lockfile entry, or a stray old copy of `tinacms` in the dependency tree), nothing fails: the schema compiles and the admin builds, but the served admin UI is an older `tinacms` where documented features quietly don't exist. Users have no signal to debug from. Two support threads hit this: #7114 (docs described the `ui.global` Site-sidebar behaviour shipped in tinacms 3.9.4, user was on 3.9.3) and a Discord report on 13/07/2026 where a stray old `tinacms` dependency kept the old sidebar after upgrading. A docs callout landed separately (tina.io PR #4714); this is the product-side fix.

### Approach

During `tinacms dev` and `tinacms build` startup, compare the versions of `tinacms`, `@tinacms/graphql`, and `@tinacms/schema-tools` resolved from the user's project against the ranges the installed `@tinacms/cli` was published with (pnpm rewrites the internal `workspace:^` refs to concrete ranges at publish, so the CLI's own package.json already carries the expected versions). Also detect two different resolved copies of `tinacms` (project root vs the copy the admin bundle dedupes to via `@tinacms/app`). Print a clear warning on skew. Warning only, never a hard failure.

### Acceptance Criteria

- [ ] `tinacms dev` and `tinacms build` print a warning at startup when a resolved core package falls outside the CLI's declared range
- [ ] A warning is printed when two different copies of `tinacms` are resolved in the tree
- [ ] The check never fails or blocks the command, and stays silent inside the monorepo (unrewritten `workspace:` ranges)
- [ ] Unit tests cover the range check and both warning rules

kulesy:
Closes #7284

**TL;DR** `tinacms dev` and `tinacms build` now warn at startup when the installed `tinacms`/`@tinacms/*` packages don't come from a coherent release.

**Pain:** when a package manager holds one core package back (partial upgrade, pnpm `minimumReleaseAge`, a stale lockfile entry, or a stray old copy of `tinacms` in the tree), nothing fails: the schema compiles and the admin builds, but the served admin UI is an older `tinacms` where documented features quietly don't exist. Two support threads hit this (#7114, plus a Discord report where an old `tinacms` dependency kept the old sidebar after an upgrade), and users have no signal to debug from.

**Solution:** the published `@tinacms/cli` already declares concrete ranges of its release siblings (pnpm rewrites the `workspace:^` refs at publish, exact pins on older releases), so the installed CLI knows which versions it expects. On dev/build startup a new `version-coherence` module resolves `tinacms`, `@tinacms/graphql` and `@tinacms/schema-tools` from the project and prints a warning when one falls outside the CLI's declared range, or when two different copies of `tinacms` are resolved (the project copy vs the one the admin bundle dedupes to via `@tinacms/app`). Warning only - it never fails the command, and it stays silent inside the monorepo where the refs are unrewritten workspace ranges.


kulesy:
Verification: 12 new unit tests cover the range check (exact and caret, 0.x carets, unparseable specs return "can't tell"), the warning rules, and an end-to-end fixture with a real directory layout; the full `@tinacms/cli` jest suite passes (264 tests) and `tsc` adds no new errors over main. Also smoke-tested the built CLI against a scratch project containing a stray `tinacms@3.7.0`: `tinacms dev` printed the warning below and then continued exactly as before, while a clean project printed nothing.

```
🦙 TinaCMS Dev Server is initializing...
WARN: TinaCMS package version mismatch detected:
  - multiple copies of tinacms are installed: 3.7.0 (.../fake-skewed-project/node_modules/tinacms) and 3.10.1 (.../packages/tinacms) - the admin UI bundles only one of them
A held-back package (stale lockfile entry, partial upgrade, pnpm minimumReleaseAge) can leave the admin UI on an older tinacms where newer documented features are silently missing. Upgrade the packages above to matching releases and reinstall.
```

packages/@tinacms/cli/src/next/commands/baseCommands.ts
```diff
@@ -82,6 +86,34 @@ export abstract class BaseCommand extends Command {
...
 
+  warnOnVersionSkew(rootPath: string) {
+    try {
+      const moduleDir = path.dirname(fileURLToPath(import.meta.url));
+      const require = createRequire(import.meta.url);
+      const warnings = collectVersionCoherenceWarnings({
+        rootPath,
+        cliModuleDir: moduleDir,
+        resolveEntry: (packageName, fromDir) =>
+          require.resolve(packageName, { paths: [fromDir] }),
```
18-th, line 97: The `@tinacms/graphql` check never actually fires. Its published exports map only has `types` and `import` conditions, no `require` or `default` (see `npm view @tinacms/graphql@latest exports`), so `require.resolve('@tinacms/graphql', ...)` throws `ERR_PACKAGE_PATH_NOT_EXPORTED` even when the package is sitting right there in `node_modules`. `resolvePackage` then catches that as "not installed" and skips it.

Easy to reproduce with real packages:

```bash
npm install --ignore-scripts @tinacms/cli@2.5.6 tinacms@3.11.0 @tinacms/graphql@2.4.8
```

That leaves graphql at 2.4.8 against the CLI's declared `^2.4.9`, but `collectVersionCoherenceWarnings` comes back empty. If you also hold `tinacms` back to 3.9.3, both of its warnings fire, so the machinery works. It's specifically graphql that can't be resolved.

The tests don't catch this because the fixture resolver is a hand-rolled walker that resolves anything with an `index.js`, which makes it more capable than the real `require.resolve` it stands in for. I'd fall back to walking `node_modules/<name>/package.json` up from `fromDir` when `require.resolve` throws, and add a regression test that uses the real resolver against graphql's actual exports map.

kulesy: Good catch, confirmed - `require.resolve` throws `ERR_PACKAGE_PATH_NOT_EXPORTED` on @tinacms/graphql's exports map (only `types`/`import` conditions) and the check swallowed it as not-installed, so the graphql range warning never fired. Fixed in 37e567966 with your suggested fallback: when the resolver throws, `resolvePackage` now walks `node_modules/<name>/package.json` up from `fromDir`; regression tests cover both the real-resolver failure (a node subprocess asserts the exports-map fixture really does throw `ERR_PACKAGE_PATH_NOT_EXPORTED`, since jest patches `require.resolve` in-process) and the fallback recovery, and I re-ran your exact npm repro (`@tinacms/cli@2.5.6` + `@tinacms/graphql@2.4.8`) which now emits `@tinacms/graphql@2.4.8 is installed, but @tinacms/cli@2.5.6 expects @tinacms/graphql@^2.4.9`.

18-th (approved):
LGTM

PR #7285, opened 2026-07-16
fix(@tinacms/cli): capture telemetry as anonymous events

kulesy:
No linked issue (stems from an internal telemetry cost investigation).

**TL;DR** Capture CLI build telemetry as anonymous events instead of minting a person profile per run.

**Pain:** every `tinacms build` sends its telemetry with a fresh random UUID as the distinct id and person profiles enabled, so each run creates a single-use PostHog person and the events bill at the identified-event rate (roughly 4x the anonymous rate). Nothing ever references these person profiles - they're pure junk data with a billing multiplier attached.

**Solution:** set `$process_person_profile: false` on every CLI capture, so events bill as anonymous and create no person profiles.


kulesy:
Test plan: 19 jest tests pass in `@tinacms/cli` (`src/utils/buildTelemetryThrottle.test.ts` covers first-failures-send, suppression past the threshold, hourly sampling with `suppressedCount`, errorCode-change reset, success reset, and quiet-loop expiry; `src/utils/posthog.test.ts` asserts `$process_person_profile: false` on every capture), and `tsc`/biome are clean on the changed files.

kulesy:
Slimmed this down since the earlier test-plan comment: the failure-loop throttle was dropped as not worth the code (runaway hosts can be handled with a PostHog drop transformation, and the anonymous flag already cuts any flood's billing impact ~4x). What remains is the `$process_person_profile: false` flag, its test, and the changeset - 18 added lines, 12 jest tests pass, biome clean.

18-th (approved):
LGTM

PR #7286, opened 2026-07-16
fix(tinacms): persist repo-relative paths for git-backed media on cloud uploads

kulesy:
Closes tinacms/tinacloud#3858

**TL;DR** Cloud uploads with git-backed media now persist the repo-relative path, matching local mode.

**Pain:** uploading media on an editorial-workflow branch baked the staging-scoped hosted URL (`.../__staging/{branch}/__file/...`) into committed content. It renders fine in the staging preview, then 404s on the statically-built production site after merge, because the static build serves the committed file from `public/{mediaRoot}/` and never touches the assets CDN.

**Solution:** for git-media configs (`media.tina` with `mediaRoot` + `publicFolder`, the same discriminator `persist_local` uses), `TinaMediaStore.parse` reconstructs `/{mediaRoot}/{directory}/{filename}` from the `Media` object instead of returning the hosted `src`, so the persisted value is branch-independent by construction. This covers both the image field and rich-text inserts, which funnel through `parse`. The image field keeps the hosted `src` for the in-editor preview only, since a freshly staged file isn't servable at its repo path until merge. Hosted-media configs (no `mediaRoot`/`publicFolder`) still persist the absolute URL.


kulesy:
Verified with `vitest run` on the two touched suites: 59/59 pass, including 7 new `parse()` tests (staging-branch upload persists `/{mediaRoot}/{name}`, nested directories, mediaRoot slash/empty normalization, hosted-media configs unchanged, filename-less fallback, and a parity test asserting the cloud-parse value equals `persist_local`'s for the same file) plus 3 display-src tests for the field preview; `biome lint`/`format` and `tsc` are clean on the changed files.

packages/tinacms/src/toolkit/fields/plugins/image-field-plugin.tsx
```diff
@@ -18,7 +18,12 @@ export const ImageField = wrapFieldsWithMeta<InputProps, ImageProps>(
     const ref = React.useRef(null);
     const cms = useCMS();
     const { value } = props.input;
-    const src = value;
+    const [selected, setSelected] = useState<
+      { value: string; src: string } | undefined
+    >();
+    // A just-picked image displays its hosted src: the persisted value may be
+    // a repo-relative path that isn't servable until the branch merges.
+    const src = selected && selected.value === value ? selected.src : value;
```
Aibono1225, line 26: Reloading/reopening the document may cause the image field to lose the hosted preview URL and fall back to the saved relative path.

I like that this PR keeps the fix simple and scoped. We could keep that scope and follow up separately, or consider resolving the saved relative path back to the hosted `Media.src` at display time.

kulesy:
Closing as I could not reproduce the issue

PR #7297, opened 2026-07-21
fix(@tinacms/cli): bump vite off EOL 4.x to 6.4.3 (#7212)

### LINKED ISSUE #7212: 🤖 Bump @tinacms/cli off vite 4
`@tinacms/cli` pins **`vite@^4.5.9`** (released 2023) and **`@vitejs/plugin-react@3.1.0`**. Vite is a real production dependency — the CLI uses it to build/serve the `@tinacms/app` admin SPA on the user's machine.

## The bump is cheaper than it looks

**The CLI is already pure ESM.** `packages/@tinacms/cli/package.json` has `"type": "module"`; `@tinacms/scripts` builds every Node entry with esbuild `format: 'esm'`; `bin/tinacms` uses a dynamic `import()`. There is no `require('vite')` anywhere.

This matters because the usual blocker for a Vite 4 → 6+ bump is that **Vite 6 removed the CJS Node API**. That does not apply here — **there is nothing to port.**

Actual breaking changes to handle:

| Change | Where | Effort |
|---|---|---|
| `splitVendorChunkPlugin` removed in Vite 5 | `src/next/vite/index.ts` | Delete the call — Rollup vendor-splits automatically now |
| Node floor rises (v5 ≥18, v6 ≥18.19/20.10, v7 ≥20.19/22.12) | `bin/tinacms` still declares Node ≥14.18 | Bump the engine check |
| `@vitejs/plugin-react` 3.x is far behind | `package.json` | Bump to a version matching the chosen Vite major |

Everything else the CLI uses from Vite — `createServer`, `build`, `transformWithEsbuild`, `InlineConfig`/`Plugin` types — is unchanged across these majors.

## Be clear about what this does *not* buy

**It will not deduplicate Vite with the user's framework.** Measured against a current Astro project:

| | Vite | esbuild |
|---|---|---|
| Astro 7 | `^8.0.13` | `^0.28` |
| `@tinacms/cli` | `^4.5.9` | `^0.24` (and its `vite@4` pulls `^0.18`) |

`^7.x ∩ ^8.x = ∅`, so bumping to Vite 7 still cannot hoist a shared copy with Astro 7. Chasing Vite 8 would simply break dedup for Astro 6 users (Vite 7) instead — and Next.js users have no Vite at all, so they carry ours regardless. **You cannot win a dedup race against a moving framework target, and this issue should not be sold as a size fix.**

What it *does* buy:
- closes ~2 years of outdated-dependency and security-advisory exposure on a production dependency
- lets the CLI's own `esbuild` pin align with its Vite's, removing one of the **three** esbuild native binaries currently installed (~10 MB, measured at 57.4 MB across 3 copies)

The real answer to "why does the CLI ship Vite at all" is architectural — `@tinacms/app` publishes raw `.tsx` and is compiled on the user's machine — and is out of scope here.

## Context

Part of an install-size audit: a stock Astro + TinaCMS blog installs **1412 MB** of `node_modules`. See #7207, #7208, #7209.

## Acceptance

- `tinacms dev` works on the kitchen-sink examples: admin loads, HMR works, forms save.
- `tinacms build` produces a working `public/admin`.
- One esbuild native binary instead of two in the CLI's own tree.

wicksipedia:
Closes #7212.

`@tinacms/cli` and `@tinacms/scripts` pinned `vite@^4.5.9`, which caps at the last-ever **4.5.14** (EOL 4.x) and leaves upstream path-traversal / file-disclosure advisories unpatched — none were backported to 4.x. This bumps them to `vite@^6.4.3` and `@vitejs/plugin-react` 3 → 4.

## esbuild dedupe (the size win)

The `esbuild` catalog pin moves `^0.24.2` → `^0.25.0` to match the version vite 6 bundles, so the CLI's **direct** esbuild dedupes onto vite's transitive copy. `@tinacms/cli` now installs **one** esbuild native binary instead of two.

**`@tinacms/cli` production tree — resolved versions** (`git show <ref>:pnpm-lock.yaml`):

| | direct `esbuild` | direct `vite` | `esbuild` bundled by that vite | esbuild copies in CLI prod tree |
|---|---|---|---|---|
| **main** | `0.24.2` | `4.5.14` | `0.18.20` | **2** (`0.24.2` + `0.18.20`) |
| **this PR** | `0.25.12` | `6.4.3` | `0.25.12` | **1** (`0.25.12`) |

Live proof on this branch:

```
$ pnpm --filter @tinacms/cli why esbuild --prod   # → single version
esbuild 0.25.12
```

Whole-lockfile effect — the `0.24.2` copy is eliminated entirely:

```
distinct esbuild versions in pnpm-lock.yaml:  main 6  →  this PR 5   (0.24.2 gone)
pnpm-lock.yaml size:                          34435  →  34232 lines  (−203)
```

(`@tinacms/scripts` still resolves a second esbuild via tsup's own bundled copy — that's a build-tooling package users never install, so it's outside the CLI criterion.)

## Dev-server fix (regression the bump would otherwise ship)

Vite 6 serves its dev endpoints (`@vite/client`, `@react-refresh`, the SPA entry) **under the configured `base`** (`/admin/`) — Vite 4 served them at the server root. The injected dev HTML hard-coded them at root, so on Vite 6 all three 404 and `tinacms dev` fails with *"Failed loading TinaCMS assets"*. `tinacms build` was unaffected (static SPA, no dev endpoints), which is why type/unit/build checks all passed it.

Fix: a shared `getBasePath()` (used by both `createConfig` and `devHTML` so they can't drift) prefixes the three URLs. Guarded by a new `html.test.ts`.

## Other fallout from the majors

- drop `splitVendorChunkPlugin` (removed in Vite 5; Rollup vendor-splits automatically now)
- drop the `fastRefresh: false` option (removed in plugin-react 4) — **Fast Refresh is now always on in `tinacms dev`**
- `process.env` define switches from `new Object(...)` to a plain JSON literal (esbuild ≥0.25 rejects the old form)
- raise the `tinacms` bin Node floor 14.18 → 18 to match vite 6's engines

## Verification

- `pnpm build` 23/23 · `@tinacms/cli` jest **272/272** (+3 base-path guards) · CLI `tsc` clean (the 146 `@tinacms/mdx` type errors are pre-existing on `main`, none in changed files) · Biome clean
- kitchen-sink `tinacms dev` browser-checked end-to-end: admin mounts, the three dev endpoints return 200 under `/admin/`, HMR updates the live preview, and a form save writes to disk


JackDevAU (approved):
:shipit:

PR #7304, opened 2026-07-22
refactor: facilitate token refresh through tinacloud

brookjeynes-ssw:
Token refresh logic is hard-coded to Cognito and wont work when auth is migrated. This updates token refresh logic to go through TinaCloud


packages/tinacms/src/internalClient/authProvider.ts
```diff
@@ -166,44 +170,54 @@ export class TinaCloudAuthProvider extends AbstractAuthProvider {
...
+        const val = await res.json();
+        if (res.status !== 200) {
+          throw new Error(
+            `Unable to refresh auth tokens. Status: ${res.status} - Body: ${JSON.stringify(val)}`
+          );
+        }
+        const newToken = {
+          access_token: val.access_token,
+          id_token: val.id_token,
+          refresh_token: refresh_token,
```
KahaMason:
```suggestion
          refresh_token: val.refresh_token,
```

packages/tinacms/src/internalClient/authProvider.ts
```diff
@@ -166,44 +170,54 @@ export class TinaCloudAuthProvider extends AbstractAuthProvider {
...
+    refresh_token?: string;
+  }): Promise<TokenObject> {
+    const { access_token, id_token, refresh_token } = tokens;
+    if (!access_token) {
+      throw new Error('Unable to refresh auth tokens: missing access_token');
+    }
+    const { client_id, exp } = this.parseJwt(access_token);
 
     // if the token is going to expire within the next two minutes, refresh it now
     if (Date.now() / 1000 >= exp - 120) {
```
KahaMason, line 185: We might need an in-flight refresh guard here.

`getToken()` runs per request and each one checks the access token's expiry. When it's close to expiring, every parallel request fires its own refresh with the same refresh_token.

Cognito is fine with that, but WorkOS invalidates the old refresh token on every refresh — so the first request succeeds and gets a new one, and the rest are left holding a token that's already been spent.

packages/tinacms/src/internalClient/authProvider.ts
brookjeynes-ssw: A refresh token is not returned from the endpoint, you must use the one given previously
edit: this is true for cognito - workos may or may not provide a new one
https://workos.com/docs/authkit/session-resilience#refresh-token-rotation

packages/tinacms/src/internalClient/authProvider.ts
brookjeynes-ssw: while a nice to have, i dont think it's necessary as workos has a grace period as per their docs

https://workos.com/docs/authkit/session-resilience#the-replay-grace-period

KahaMason (dismissed):
LGTM 🚀

18-th:
I believe we've missed one token usage in packages/@tinacms/app/src/Playground.tsx

image

18-th (approved):
LGTM

Issue #7246, opened 2026-07-14
🤖 ⚡ posthog: telemetry chunk loads only when enabled — true opt-out, not a skipped init()


wicksipedia:
**Maintainer decision this implements:** *"posthog → optional, opt-OUT."* On by default; a user who disables it should have it **gone**, not merely dormant. Same reframe as #8: post-split (#12), posthog-js's 42 MB (maintainer-measured) is off the user's disk automatically; "removable" then means the browser never downloads it and no capture code runs.

**Verified mechanics (2026-07-14):** `posthog-js` `^1.347.1`, direct dep at `packages/tinacms/package.json:109`. `telemetry: 'disabled'` today only short-circuits init: the guard at `packages/tinacms/src/lib/posthog/posthogProvider.ts:86-91` returns before `posthog.init()` (`:100-119`) — but the **static import at `posthogProvider.ts:1`** (`import posthog, { PostHog } from 'posthog-js'`) means the library is bundled and shipped regardless. (`isDisabled` at `:88` is set and never read — dead.) Config flows from `client?.schema?.config?.config?.telemetry`, default `'anonymous'`, at `packages/tinacms/src/admin/index.tsx:99-102`. Call sites are already centralized: ~20 files use the `captureEvent` wrapper (`posthogProvider.ts:164-180`), which no-ops when the client is null; the only direct `posthog.capture` outside the provider is `admin/index.tsx:114`.

**What it does:** Make `posthogProvider.ts` the *only* module that touches `posthog-js`, and make its import dynamic: `initializePostHog` resolves the mode first; when `'disabled'`, `import('posthog-js')` is never evaluated — zero telemetry code fetched or executed. `captureEvent` already guards on a null client, so the ~20 call sites need no changes; fold `admin/index.tsx:114`'s direct `posthog.capture` into the wrapper. Default behaviour (`'anonymous'`) unchanged — opt-out, exactly as the maintainer specified. In the prebuilt shell this becomes a lazy chunk fetched only when telemetry is on.

**Files touched:** `packages/tinacms/src/lib/posthog/posthogProvider.ts`, `packages/tinacms/src/admin/index.tsx` (fold the direct capture; the type-only `PostHog` import becomes `import type`).

[...]

Issue #7245, opened 2026-07-14
⚡ Lazy-load mermaid: rich-text editor stops shipping 83 MB-closure diagrams to every code block


wicksipedia:
**Maintainer decision this implements:** *"mermaid → optional."* **Honest framing:** after the runtime split (#12), mermaid is off the user's disk regardless — it's an editor dep and the editor lives in our shell. What "optional" means *then* is browser payload: the admin shouldn't download mermaid unless a document actually uses it. Landing this on `main` now improves today's admin bundle **and** carries into the shell automatically.

**Verified mechanics (2026-07-14):** `mermaid` (`^11.12.2`, direct dep at `packages/tinacms/package.json:107`) has exactly two importers, both **static**:
- `packages/tinacms/src/toolkit/fields/plugins/mdx-field-plugin/plate/components/plate-ui/mermaid-element.tsx:1`
- `packages/tinacms/src/toolkit/fields/plugins/mdx-field-plugin/plate/components/plate-ui/code-block/code-block-element.tsx:13`

In `code-block-element.tsx`, a `useEffect` guard (`:45-47`, `if ((element.lang as string) !== 'mermaid') return`) means `mermaid.parse()` (`:51-55`) only *executes* for mermaid-language blocks — a nuance to the epic brief's "validates every code block" — but the **static top-level import makes the generic code-block component pull mermaid into the bundle unconditionally**: any document with any code block ships it. Net effect on payload is as assumed.

**What it does:** Convert both to lazy: `const mermaid = await import('mermaid')` inside the `lang === 'mermaid'` branch of the validation effect, and a `React.lazy`/dynamic boundary around `mermaid-element.tsx`'s renderer. Validation becomes async-on-demand (it already is async — `mermaid.parse` is awaited); render shows a lightweight placeholder while the chunk loads. In the prebuilt shell (#5/#6) this materializes as a separate lazy chunk the browser fetches only when a mermaid block exists.

**Files touched:** the two files above; possibly a small shared `lazy-mermaid.ts` helper.

**Verification:** unit test asserting no static `mermaid` import in the built output's main chunk (metafile/grep assertion in the app or tinacms build); Playwright: a doc with a plain code block never requests the mermaid chunk (`page.waitForRequest` negative assertion), a doc with a mermaid block renders the diagram and shows validation errors for invalid syntax.

[...]

Issue #7237, opened 2026-07-14
⚡ Epic: Prebuilt admin + runtime split — stop shipping the editor to every site


wicksipedia:
**Suggested title:** `📦 Epic: prebuild the admin shell, then split the runtime (−388 MB remaining of a measured −630 MB / 61%)`
**Labels:** `perf` + `v4`

### Tracklist

**Phase 0 — baselines (land first)**
- [ ] #7238 — 🧪 Size-baseline CI: per-PR install/tarball/bundle budgets + duplicate-copy watchlist
- [ ] #7239 — 🧪 First production-build admin browser suite + the hostile fixture project

**Phase 1 — prebuild the shell → scenario C (−68.3 MB)**
- [ ] #7240 — 🧹 The runtime seam: @tinacms/app reads project + env at runtime (dual-mode, vite path preserved)
- [ ] #7241 — 🧹 Tailwind two-pass split: export tinaTheme/tinaTwPlugins, make prebuildFilePath optional
- [ ] #7242 — 📦 @tinacms/app gets a real build: dist/ + manifest.json + shell contract tests (dual-publish src+dist)
- [ ] #7243 — ⚡ Flag-gated prebuilt production build: 33 ms project chunk, frozen externals ABI, skew checks
- [ ] #7244 — ⚡ tinacms dev without vite: connect server + esbuild watch + reload channel (flag-gated)

**Browser-payload wins (independent, `main`-safe)**
- [ ] #7245 — ⚡ Lazy-load mermaid: rich-text editor stops shipping 83 MB-closure diagrams to every code block
- [ ] #7246 — 🤖 ⚡ posthog: telemetry chunk loads only when enabled — true opt-out, not a skipped init()
- [ ] #7247 — ✨ Version the admin↔site postMessage handshake

**Phase 2 — the split → scenario D (−320.0 MB, v4)**
- [ ] #7248 — 🧹 Flip the default: prebuilt admin everywhere, delete the vite path (v4)
- [ ] #7249 — ⚡ The runtime split: sites import a slim runtime, the 81-dep editor leaves user installs (v4)

Hard edges: #7240→#7242→#7243→#7244; #7241→#7243; #7243+#7244→#7248→#7249. Parallel lanes: {#7238, #7239} · {#7240, #7241} · {#7245, #7246, #7247}.
---

### Problem

[...]

PR #7316, opened 2026-07-23
🧪 Add CI size budgets, and stop next-tinacms-azure bundling a second Next.js

### LINKED ISSUE #7238: 🧪 Size-baseline CI: per-PR install/tarball/bundle budgets + duplicate-copy watchlist
**Why now:** No CI job measures size today (verified: no size-limit/bundlewatch/`du` anywhere in `.github/workflows/`, root `package.json`, or `scripts/`; the only adjacent check is `publint --strict`). The 1412→1037 MB baseline drift went unnoticed until manually re-measured; the incident behind it — `@tinacms/cli` exact-pinning `tinacms@3.10.0` after `3.10.1` shipped, npm nesting 3 copies of tinacms/mermaid/date-fns — had no detection mechanism at all. This issue is the maintainer's stated top priority for the epic: discover regressions *ahead of time*. It lands before any prebuilt-admin code.

**What it does:** Adds a CI job + checked-in baseline as specified in the epic's baseline-test plan §(a): (1) verdaccio-backed real-`npm` install of a fixture Astro site measured with `du`; (2) `npm ls --all --json` duplicate-copy watchlist (`tinacms`, `mermaid`, `date-fns`, `typescript`, `react`, `graphql`, `lodash`) — copy-count >1 fails unconditionally; (3) per-package `npm pack --dry-run --json` unpacked sizes; (4) kitchen-sink `tinacms build` admin output size. Baselines in `tests/size-baselines.json`, seeded from the measured ladder (A=1037.8 MB today's publish state; fixture-local numbers will differ — seed from the job's own first green run, *not* by copying the ladder blindly). Tolerance: >5% or >5 MB over baseline fails; `pnpm size:update` re-baselines with the JSON diff reviewed in-PR. A cron variant runs the same harness against the real registry to catch published-state drift (the 1412 class).

**Files touched:** `.github/workflows/main.yml` (new job), `tests/size-fixture/` (new minimal Astro fixture), `tests/size-baselines.json` (new), `scripts/size-baseline.mjs` (new), root `package.json` (`size:check` / `size:update` scripts).

**Verification:** job goes red when a PR adds a chunky dep to any `dependencies` (test by draft PR adding e.g. `lodash` to `@tinacms/app`); job goes red when a watchlist package is forced to duplicate (pin an internal dep to a stale exact version in the fixture); green run time ≤ ~4 min.

**Size delta:** 0 — this is regression *prevention*. Same reframe as #7213: worth $0 on the day it lands, worth the whole ladder the day something regresses.

**What breaks if it's wrong:** flaky thresholds train people to rubber-stamp re-baselines (mitigate: absolute+relative tolerance, watchlist has zero tolerance); verdaccio limitation means registry-history bugs only surface in the cron mode, not per-PR — documented, not hidden.

---

Part of #7237

wicksipedia:
Closes #7238. Part of #7237, phase 0, which adds baselines before any risky work.

This PR adds the repo's first size check, and fixes a bug that the check found on its first run.

- `scripts/size-baseline.mjs` measures install size, package tarball sizes and the admin bundle, and counts copies of watchlist packages.
- `tests/size-baselines.json` holds the baseline.
- A new CI job fails a PR that grows past the baseline or adds a second copy of a watchlist package.

## Bug fix: `next-tinacms-azure` bundled Next.js

The first run measured `next-tinacms-azure` at 806 KB against a 358 KB baseline.

`@tinacms/scripts` builds esbuild's `external` list from a package's `dependencies` and `peerDependencies`. `next-tinacms-azure` listed `next` in `devDependencies`. Its `handlers.ts`, `delivery-handlers.ts` and `auth.ts` import values, not only types, from `next/server` and `next/headers`. So esbuild inlined Next.js and part of React into all three, and anyone who installed the package got a second Next.js runtime next to their own.

`next-tinacms-s3`, `-dos` and `-cloudinary` import only types from Next.js. TypeScript erases those imports, so these packages do not have the bug.

This PR declares `next` as a peer dependency, which puts it back on the external list. `dist/` drops from 868 KB to 44 KB, and the entry points contain `import … from "next/server"` again. The fix has its own changeset.

## What the check measures

1. **Install size** of a minimal Astro fixture (`tests/size-fixture/`). The harness installs it with npm from a temporary verdaccio registry, which it seeds with packed workspace tarballs. It uses npm because npm users get the nested duplicates this check looks for. pnpm's content-addressed store hides them.
2. **Copies of watchlist packages** on disk: `tinacms`, `mermaid`, `date-fns`, `typescript`, `react`, `graphql`, `lodash`. The harness counts physical copies, because `npm ls` counts one hoisted copy more than once. Any increase over the baseline fails.
3. **Unpacked size** of each of the 23 publishable packages.
4. **Admin output**: the kitchen-sink `public/admin` after `tinacms build`. The format allows a later split into shell and project budgets.

Baseline on ubuntu-latest: install 900.88 MB, admin output 9.69 MB, one copy of each watchlist package. The baseline was refreshed on 2026-09-14. On main, #7433 moved graphiql from 3.0.0-alpha.1 to 4.1.2, which added about 61 MB. This PR also moved the fixture from astro 7.1.1 to 7.3.2 to clear two advisories.

## Tolerance per metric

The metrics range from a 16 KB tarball to an install near 1 GB, so one tolerance cannot fit all of them. A 5 MB floor sized for the install would let `@tinacms/metrics` (16.5 KB) grow 317 times and still pass. A percentage tight enough for the install would fail a small tarball on a README edit.

| metric | tolerance | fails at |
|---|---|---|
| install (900.9 MB) | 1% or 5 MB | +9.0 MB, which catches typescript (23 MB), monaco (73 MB) or a nested `tinacms` |
| admin output (9.69 MB) | 5% or 250 KB | +496 KB, which catches a re-bundled mermaid or posthog chunk (#7245, #7246) |
| each package (14 KB to 3.2 MB) | 5% or 25 KB | +25 KB on small packages, +5% on large ones |

The 25 KB floor catches a bundled-in dependency and ignores normal file edits. For every metric, a decrease past 10% gives a warning.

## Baselines must come from CI

Sizes differ by platform. `du` counts blocks differently on each platform, and some packages install platform-specific optional binaries. A baseline taken on macOS reads 83 MB lower than ubuntu-latest, so the job would fail on every run. A dev machine can also hold stale `dist/` output that a clean build does not make. Locally, three packages measured larger than on CI, `@tinacms/schema-tools` by 483 KB.

To update the baseline, run the workflow with `update_size_baseline: true`. The job runs the same measurement and uploads a `size-baselines` artifact, and a person commits it. The job uploads an artifact instead of pushing a commit, so CI keeps `contents: read`.

## Fixture pins: `package-lock.fixture.json`

The fixture needs pinned versions. Without pins, the install size and the copy counts change when an upstream package ships a point release, even if the PR changes nothing.

The fixture installs `tinacms`, so its lockfile holds the product's whole dependency tree, about 1,400 entries. The `dependency-review` action compares lockfiles between base and head. If we committed the file as `package-lock.json`, the action would count all of those entries as new dependencies, and an advisory anywhere in the tree would fail the PR. Three advisories failed it while I built this. The last was in `react-router`, which has no patched release in the major version the product uses, so a version bump could not fix it.

So the harness copies the pins to `package-lock.json` for the install and deletes that file afterwards. The pins and the numbers stay the same, and the working tree is clean after a run.

The harness does not use `npm ci`. It rebuilds the workspace tarballs from HEAD on each run, so their integrity hashes change on every commit this job checks. The harness removes the workspace entries from the lockfile and resolves them again. The third-party pins do not change.

## Reporting

The job writes its report to the GitHub Actions job summary on every run, including passing runs. Reviewers can see a size change before it reaches a threshold. A drift from 1,412 MB to 1,037 MB went unnoticed because the numbers appeared only in the raw log of passing runs.

## Verification

- CI passes: `0 failing, 1 warning — 92.3s`. Between two CI runs, the size moved by −4.0 KB. The same check on a dev machine moved by +432 KB.
- Failure paths: a conflicting `lodash` pin raises the copy count and exits 1. A fixture `typescript` pin takes its count from 2 to 3 and exits 1.
- 55 unit tests (`tests/size-baseline.test.ts`) cover the tolerance math, the tar parser, and the lockfile strip and normalise helpers.
- A weekly cron runs `--registry-mode=real` against npmjs. It catches changes in published packages that verdaccio cannot show. The harness header documents this limit.

## Please review: the copy-count rule

Issue #7238 asks for "watchlist copy-count >1 fails". This PR fails on any increase above the baseline count instead. If the baseline already has a count above 1, the job shows a warning, not a failure.

When I wrote the harness, the fixture had two copies of `typescript` (#7211), so the literal rule would have failed CI from the first run. Now `typescript` and `lodash` each have one copy, so every watchlist package starts at 1 and the two rules give the same result. I kept the softer rule for the future: a duplicate that is already in the baseline shows as a warning and does not block the merge queue. The case this check exists for, `tinacms` going from 1 copy to more, fails under both rules.

## Trade-offs

- The weekly `schedule` trigger is at workflow level, so the other jobs in the workflow also run on the cron. A YAML comment notes this.
- CI runs with a cold cache can take longer than the ~4 minute target. Install and build take most of the time. The harness takes about 90 s.
- Update mode installs the fixture twice: once to resolve new pins, and once from those pins. A check run installs from the committed pins, so the second install makes the baseline measure the same tree. This adds one npm install to an update run.
- The harness uses `pnpm pack` and a small tar parser instead of `npm pack --dry-run --json`. npm cannot resolve the `workspace:` protocol, and pnpm's `--json` output has no size field.


kulesy:
The red Size baseline check is a real catch, not flakiness. The review commits aren't the cause either - they only touch CI wiring and comments.

What happened: #7433 merged to main shortly after this branch's last green run. It bumps graphiql to v4 in @tinacms/app, and graphiql 4 ships Monaco. PR builds include main, so the fixture install now pulls in the whole monaco-editor tree. That's 902 MB against the 840 MB baseline, about 62 MB over.

So the gate did its job. #7433 just landed before the gate existed to catch it.

Two ways forward:
1. Raise the +62 MB against #7237 and look at lazy-loading graphiql. This check goes green once that lands.
2. Accept the growth: run the "Size Baseline" workflow on this branch with `update_size_baseline: true` and commit the artifact.

.github/workflows/size-baseline.yml
```diff
@@ -0,0 +1,136 @@
...
+  # On the weekly schedule it runs the identical harness against the real npm
+  # registry (--registry-mode=real) instead of the throwaway verdaccio one.
+  size-baseline:
+    name: Size baseline (${{ github.event_name == 'schedule' && 'real registry' || 'verdaccio' }})
+    needs: size-changes
+    if: needs.size-changes.outputs.should_run == 'true'
+    runs-on: ubuntu-latest
+    timeout-minutes: 20
+    env:
+      TURBO_TOKEN: ${{ secrets.TURBO_TOKEN }}
```
kulesy: The fixture npm install runs third-party install scripts with the turbo cache token in reach - could the TURBO env move onto the Build step, which is the only step that needs it?

scripts/size-baseline.mjs
```diff
@@ -0,0 +1,925 @@
...
+  adminOutput: { pct: 0.05, abs: 250 * 1024 },
+  package: { pct: 0.05, abs: 25 * 1024 },
+};
+
+// A decrease warns only to prompt `pnpm size:update` so wins get locked in;
+// deliberately looser than the fail bands so it does not fire on every
+// dead-code removal.
+const WARN_DECREASE_PCT = 0.1;
+
+const VERDACCIO_VERSION = '6';
```
kulesy: Everything else in this CI path is pinned, but this fetches whatever verdaccio 6.x resolves to at run time, in the same step that holds the turbo token - worth pinning the exact version?

scripts/size-baseline.mjs
```diff
@@ -0,0 +1,925 @@
...
+const ROOT = path.resolve(__dirname, '..');
+const FIXTURE_DIR = path.join(ROOT, 'tests', 'size-fixture');
+// Two paths on purpose. npm only reads/writes `package-lock.json`, so that name
+// is transient: staged before the install, deleted after. The committed pins live
+// in `package-lock.fixture.json` - a name dependency-review does NOT recognise.
+// Committed as package-lock.json, the fixture's ~1400-entry closure reads as
+// newly-introduced dependencies and every advisory anywhere in it fails the PR
+// (see tests/size-fixture/.gitignore).
+const FIXTURE_LOCK_PATH = path.join(FIXTURE_DIR, 'package-lock.json');
+const COMMITTED_LOCK_PATH = path.join(FIXTURE_DIR, 'package-lock.fixture.json');
```
kulesy, line 59: Renaming the lockfile also hides these pins from npm audit and Dependabot, so a pin that later turns out vulnerable keeps installing in CI with no alert - the weekly cron could run npm audit against package-lock.fixture.json to cover that.

.github/workflows/size-baseline.yml
wicksipedia: Moved `TURBO_TOKEN` and `TURBO_TEAM` from the job onto the Build step. The fixture npm install, `pnpm install` and verdaccio no longer see the token. The size check does not call turbo, so no step loses the remote cache.

scripts/size-baseline.mjs
wicksipedia: Pinned to `verdaccio@6.10.3`. Its dependencies still resolve at run time, but the turbo token is no longer in that step.

Kept it outside of the normal package.json since this is only used for comparing the build sizes and not for regular development.

scripts/size-baseline.mjs
wicksipedia: Added an `Audit fixture lockfile` step to the weekly run. It audits a temp copy of the pins with `npm audit --package-lock-only` and writes the report to the job summary.

The advisories it reports are not a problem here, so the step does not fail on them. The fixture only exists to measure install size. None of it ships, and apart from install scripts none of its code runs, so bugs like XSS in dompurify have nothing to exploit.

The audit is still worth having. If a pinned version later turns out to be malicious, its install script would run in CI, and the weekly report would flag it.

kulesy (approved):

Issue #7243, opened 2026-07-14
⚡ Flag-gated prebuilt production build: 33 ms project chunk, frozen externals ABI, skew checks


wicksipedia:
**What it does:** `tinacms build` behind `build.experimental.prebuiltAdmin` (or `TINA_PREBUILT_ADMIN=1`): stages B/C/D from the spike, productionized with every gap in `.spike-scope.md` closed —

- **Stage B (project chunk):** esbuild of one generated entry (config + `_graphql.json` + `static-media.json` + optional `queries`) against the frozen externals, with: the **interop-correct CJS bridge** (namespace with `__esModule`-marked default-merged object per module — the spike's `m.default ?? m` heuristic silently breaks `require('tinacms').AbstractAuthProvider` because `tinacms`'s default export is `TinaCMSProvider2`, `packages/tinacms/src/index.ts:22`); the full loader map + emitted-asset copying (`config-manager.ts:517-541` — spike passed only `.json`); `process.platform` define (`vite/index.ts:192` — picomatch dies without it); node-builtin shims matching `vite/index.ts:140-143`; `shouldSkipSDK()` handling (`vite/index.ts:146-151`); `filterPublicEnv()` for `process.env` (CVE-2023-25164 — the filter is non-negotiable, `vite/index.ts:117-119`); Windows-safe entry generation (relative specifiers or `pathToFileURL`, not raw absolute paths in `import` strings — `.spike-scope.md` §1.8).
- **Stage C:** the user-only Tailwind pass (child #4's exports) over the **unminified** chunk — scan order pinned: esbuild → tailwind → optional minify (`.spike-scope.md` §1.9).
- **Stage D:** `index.html` with import map (before any module script — asserted, not assumed) + `__TINA_ENV__`, with asset/import-map/`projectUrl` URLs **absolutized** under `build.basePath`/`build.outputFolder` exactly like today's vite base (`vite/index.ts:167-169`) — the shell stays relative, the per-project HTML doesn't; this is the no-trailing-slash trap (`.spike-scope.md` §1.7). Keep writing the output-folder `.gitignore` (`build-command/index.ts:343`).
[...]

PR #7317, opened 2026-07-23
🧪 First production-build admin browser suite + hostile fixture

### LINKED ISSUE #7239: 🧪 First production-build admin browser suite + the hostile fixture project
**Why now:** Every Playwright suite in the repo boots `tinacms dev` (verified across all six `playwright.config.ts` — even `playwright/tina-playwright`'s `build` script wraps `tinacms dev`). The spike's four traps compiled clean and failed only in a browser, only in production output — today's CI is structurally blind to that entire class. This suite baselines *today's* vite-built admin first; child #6 then runs the identical suite with the prebuilt flag on. Two modes, one spec set.

**What it does:** New `playwright/prebuilt-admin/` fixture app (layout cloned from `playwright/tina-playwright`, Next.js so `next/image` is naturally present) implementing the full nasty-fixture ingredient list from the epic's baseline plan §(b): colocated custom field component, a component importing CJS `next/image`, arbitrary Tailwind class + `bg-blue-500`, real `media.loadCustomStore` via dynamic `import()` (pattern from `packages/@tinacms/schema-tools/src/types/index.ts` JSDoc), `UsernamePasswordAuthJSProvider` from `tinacms-authjs/dist/tinacms` (drags CJS `next-auth/react`), `cmsCallback` + `createScreen`, `build.basePath: 'my-site'`. `webServer` = `tinacms build` + static serve + GraphQL API — **not** `dev`. Specs: `boot.spec` (collection list renders; `__REACT_DEVTOOLS_GLOBAL_HOOK__.renderers.size === 1`; zero `console.error`), `custom-field.spec` (save round-trips to disk; reuse `examples/next/kitchen-sink/e2e/utils/admin-helpers.ts`), `tailwind.spec` (`getComputedStyle`: fixture class exists, `bg-blue-500` = Tina's `rgb(0,132,255)`, not stock), `basepath.spec` (with **and without** trailing slash), `media.spec`/`auth.spec`. New workflow cloned from `playwright-next-kitchen-sink.yml`, path-filtered on `packages/@tinacms/{cli,app}/**` + `packages/tinacms/**`.

**Files touched:** `playwright/prebuilt-admin/` (new), `.github/workflows/playwright-prebuilt-admin.yml` (new).

**Verification:** suite green on `main` today (that *is* the baseline); deliberately re-introduce one spike trap locally (e.g. break the base URL) and confirm the matching spec goes red.

**Size delta:** 0 — behavioural coverage.

**What breaks if it's wrong:** a fixture that's accidentally tame (e.g. the CJS dep tree-shaken out) passes forever while real configs die — mitigate with a meta-assertion that the built config chunk actually *contains* the CJS module and the custom store (grep the output in a setup step).

---

Part of #7237

wicksipedia:
Closes #7239 · Part of #7237 (Phase 0 — baselines land before any risky work)

The repo's first Playwright suite that runs against a **production** `tinacms build` output — every existing suite boots `tinacms dev`, which is structurally blind to bugs that only exist in production bundles. This baselines today's vite-built admin; #7243 will rerun the identical suite with the prebuilt flag on.

## The hostile fixture (`playwright/prebuilt-admin/`)

All seven ingredients from the issue, each guarding a named failure class:

- custom field component colocated with the schema
- a component importing `next/image` (CJS dep → CJS `require('react')` against the bundle's React)
- `aspect-w-9` (class the admin never uses) + `bg-blue-500` (must resolve to Tina's themed `rgb(0,132,255)`)
- real `media.loadCustomStore` behind a dynamic `import()`
- `UsernamePasswordAuthJSProvider` from `tinacms-authjs/dist/tinacms` (drags CJS `next-auth/react`); runtime-selected against `LocalAuthProvider` so both stay in the bundle while specs choose which is active
- `cmsCallback` registering a `createScreen` screen (react-router path)
- `build.basePath: 'my-site'`

## Specs (8 files, 12 tests)

`boot` (collection list; exactly one React renderer via an injected devtools-hook stub — production React has no hook; zero console errors), `custom-field` (save round-trips to disk), `screen` (navigates to the registered screen **through the admin's own sidebar nav**, so `useNavigate`/`NavLink` actually execute rather than a URL deep-link), `tailwind` (computed styles for both probe classes), `basepath` (**with and without** trailing slash), `media`, `auth`, plus a setup **meta-assertion** that greps the built chunks for `next/image`/custom-store/`next-auth`/screen markers — so the fixture can never go tame silently.

`webServer` recipe: `tinacms build --local` emits the production SPA into `public/admin` and keeps the local GraphQL/media server on :4001; a zero-dep static server (`serve-admin.mjs`) serves the built SPA under `/my-site/admin` (bare + trailing slash) on :3000.

## Notes for a reviewer

**The renderer-count + console-error check runs on the edit form, not just the list.** That check is the suite's sharpest instrument, but `FixtureField` — the component importing `next/image` — only mounts on the edit form. Scoped to the collection list it would miss a render-time throw from the very ingredient it exists to catch. Factored into `trackConsoleErrors` / `assertHealthyRender` (`e2e/utils/admin-helpers.ts`) and called from every spec that reaches the form.

**The mutating spec is isolated and resets per attempt.** `custom-field` writes to disk, so it owns `content/post/roundtrip.mdx`; the read-only specs use `hello.mdx`. Snapshot/restore is `beforeEach`/`afterEach`, not `beforeAll`/`afterAll` — with `retries: 2`, an attempt that saves successfully then times out on the disk read would otherwise leave the file mutated and make the retry fail deterministically, turning a transient flake into a guaranteed red.

**`next` is pinned to 15.5.21.** 14.2.35 carries seven high-severity advisories and `dependency-review` reads the fixture's manifest like any other. `next/image` is still CJS and still does `require('react')` in 15, so the trap is intact; React stays on 18.3.1.

## Verified

- **CI green**, `prebuilt-admin-e2e` included, on the current head
- 12/12 green on repeated clean runs
- Trap re-introduction: breaking bare-path serving turns `basepath.spec` red; removing the `next/image` import turns the meta-assertion red
- Workflow path-filtered on `packages/@tinacms/{cli,app}/**`, `packages/tinacms/**`, `playwright/prebuilt-admin/**`; traces and the HTML report upload on failure

## Finding (feeds #7243)

`tinacms build --local` does **not** bake the local apiUrl into the admin — `buildProductionSpa` always uses `codegen.productionUrl`, so with an empty clientId the SPA crashes on boot. The fixture works around it with `contentApiUrlOverride: 'http://localhost:4001/graphql'` (documented in its README). Worth an independent check on whether kitchen-sink local builds are similarly pointed at TinaCloud.


.github/workflows/playwright-prebuilt-admin.yml
```diff
@@ -0,0 +1,54 @@
+name: Playwright — prebuilt admin (production build)
+
+on:
+  pull_request:
+    branches:
+      - main
+    paths:
```
joshbermanssw, line 7: Would we also include @tinacms/schema-tools, @tinacms/graphql and @tinacms/mdx  are also bundled into the admin?

playwright/prebuilt-admin/playwright.config.ts
```diff
@@ -0,0 +1,43 @@
...
+  projects: [
+    {
+      name: 'chromium',
+      use: { ...devices['Desktop Chrome'] },
+    },
+  ],
+  webServer: {
+    command: 'pnpm serve:prod',
+    url: 'http://localhost:3000/my-site/admin/',
+    reuseExistingServer: !process.env.CI,
```
joshbermanssw: Should we ever be reusing an existing server? Port 3000 is super common and can cause some funky results

.github/workflows/playwright-prebuilt-admin.yml
wicksipedia: done

playwright/prebuilt-admin/playwright.config.ts
wicksipedia: done

joshbermanssw (approved):

PR #7322, opened 2026-07-26
🔧 Point media manager listing at the v2 assets-api endpoint

joshbermanssw:
## TL;DR

Media **listing** now calls the `/v2` assets-api endpoint. **Uploads and deletes stay on `/v1`** (v2 is list-only). v2 runs the same handler as v1, so listing behaviour is unchanged — this is a transport-only move that unblocks media search (the `?search=` param already lives on v2).

## What changed

- `TinaMediaStore` derives a `listUrl` (`…/v2/<clientId>`) and uses it only in `list()`. `url` (v1) still drives `upload_url` and `delete`.
- Local/self-hosted (`/media` on content-api) is untouched.
- Tests lock in the contract: list + subdir → v2, delete + upload_url → v1, post-upload `list()` → v2.

## The one thing to verify before release

**CORS on `/v2` has never been exercised by a real browser.** The server derives the allowed origin from the request path, so `/v2` needed its own matchers (added backend-side). Personal stages bypass that check, so this PR is the first real browser request carrying an `Origin` to `/v2`. Confirm the media manager loads against a TinaCloud project (prod has v2 as of `2026.07.1`) before this ships in a release.

## Links

- Parent: tinacms/tinacloud#2051
- Backend v2 endpoint: tinacms/tinacloud#3862 · search param: tinacms/tinacloud#3867

### DIFF
--- a/packages/tinacms/src/toolkit/core/media-store.default.ts
+++ b/packages/tinacms/src/toolkit/core/media-store.default.ts
@@ -803,7 +803,7 @@ export class TinaMediaStore implements MediaStore {
     if (!this.isLocal) {
       const encodedBranch = this.encodedBranchParam();
       res = await this.api.authProvider.fetchWithToken(
-        `${this.listUrl || this.url}/list/${options.directory || ''}?limit=${
+        `${this.listUrl}/list/${options.directory || ''}?limit=${
           options.limit || 20
         }${options.offset ? `&cursor=${options.offset}` : ''}${
           encodedBranch ? `&branch=${encodedBranch}` : ''
--- a/packages/tinacms/src/toolkit/core/media-store.default.ts
+++ b/packages/tinacms/src/toolkit/core/media-store.default.ts
@@ -106,6 +106,7 @@ export class TinaMediaStore implements MediaStore {
   private cms: TinaCMS;
   private isLocal: boolean;
   private url: string;
+  private listUrl: string;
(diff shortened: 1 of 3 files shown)


packages/tinacms/src/toolkit/core/media-store.default.ts
```diff
@@ -801,7 +803,7 @@ export class TinaMediaStore implements MediaStore {
     if (!this.isLocal) {
       const encodedBranch = this.encodedBranchParam();
       res = await this.api.authProvider.fetchWithToken(
-        `${this.url}/list/${options.directory || ''}?limit=${
+        `${this.listUrl || this.url}/list/${options.directory || ''}?limit=${
```
18-th: I don't believe it's possible for the `listUrl` to be unset when the `url` is set. It looks like dead code to me. Could you double-check please?

joshbermanssw: Confirmed — it's dead code. The constructor sets `isStatic` and `staticMedia` together (only when a non-empty `staticMedia` is passed), so `list()` returns early via `if (this.staticMedia)` on every static case. The non-local fetch branch is only reachable when `!isStatic && !isLocal`, which is exactly when `setup()` assigns `listUrl`. Dropped the `|| this.url` fallback in dd105dc.

18-th (approved):
LGTM

PR #7323, opened 2026-07-27
✨ Add media manager search and folder/file filter

joshbermanssw:
Tagged: https://tina-io-git-jb-test-tagged-media-tinacms.vercel.app/admin/index.html#/~/

## TL;DR

Adds a **search box** and a **folder/file filter** to the media manager. Search filters the library by filename (debounced, recursive across folders); the `All | Folders | Files` toggle narrows the view. Works against **both TinaCloud media** (via the `?search=` param from tinacms/tinacloud#3867) **and local dev media** (a matching `search` param added to the CLI dev-server). Builds on the v2 listing switch from #7322.

## How

- **`?search=` plumbing** — `search?: string` on `MediaListOptions`, appended (URL-encoded) to both the cloud and local list requests.
- **Local parity (`@tinacms/cli`)** — the dev-server media endpoint now honours `search`, matching filenames recursively within the requested folder (case-insensitive, dir-relative names) to mirror cloud. The recursive walk skips entries whose real path escapes the media root (symlink guard).
- **Search box** — debounced 300ms. On commit it resets paging and refetches; results *replace* the list (page-1 loads now replace instead of append) so the input keeps focus and doesn't flash a full-screen loader.
- **Opt-in per store** — the box only shows when the store sets a new `searchable` flag on the `MediaStore` interface. `TinaMediaStore` (TinaCloud + local dev) opts in; custom/self-hosted stores stay hidden until they set it and read `options.search`, so no store gets a box that returns unfiltered results.
- **Folder/file filter** — client-side over loaded items (folders arrive on page 1, files paginate). "Folders" view suppresses infinite-scroll since extra pages are only files.
- **Folders + active search → empty message**, since search never returns folders.

## Tests

- **Client:** store tests — `search` is URL-encoded, omitted when unset/empty, and `TinaMediaStore` advertises `searchable` (45 pass). Mutation-checked.
- **CLI dev-server:** 6 tests — recursive match + directory suppression, case-insensitivity, dir-scoped relative filenames, src prefix, no-match; plus a security test that search does **not** follow a symlink escaping the media root (22 pass). The symlink guard is mutation-checked as load-bearing.
- `tsc` clean on touched files, biome lint/format clean, `pnpm --filter tinacms build` and `--filter @tinacms/cli build` pass.

## Screenshots

Screenshot 2026-07-27 at 4 11 36 pm

Screenshot 2026-07-27 at 4 11 45 pm

https://github.com/user-attachments/assets/abd4f83a-5f06-4919-a5cd-9e6e7bfccc8f

## Links

- Parent: tinacms/tinacloud#2051
- Depends on: #7322 (v2 listing) · backend `?search=`: tinacms/tinacloud#3867

### DIFF
--- a/packages/@tinacms/cli/src/next/commands/dev-command/server/media.test.ts
+++ b/packages/@tinacms/cli/src/next/commands/dev-command/server/media.test.ts
@@ -118,6 +118,41 @@ describe('MediaModel (Vite dev server)', () => {
       expect(names).toEqual(['nested/deep/LLAMA.jpg']);
     });
 
+    it('returns matching directories as results, not only files', async () => {
+      await seed();
+      const model = new MediaModel(config);
+      const result = await model.listMedia({ searchPath: '', search: 'deep' });
+      expect(result.directories).toEqual(['/nested/deep']);
+    });
+
+    it('returns matching directories only on the first page', async () => {
+      const base = path.join(tmpDir, 'public', 'uploads');
+      await fs.mkdirp(path.join(base, 'match-dir'));
+      for (let i = 0; i < 15; i++) {
+        await fs.writeFile(
+          path.join(base, `match-${String(i).padStart(2, '0')}.png`),
+          'x'
+        );
+      }
+      const model = new MediaModel(config);
+
+      const page1 = await model.listMedia({
+        searchPath: '',
+        search: 'match',
+        limit: '10',
+      });
+      expect(page1.directories).toEqual(['/match-dir']);
+      expect(page1.files).toHaveLength(10);
+
+      const page2 = await model.listMedia({
+        searchPath: '',
+        search: 'match',
+        limit: '10',
+        cursor: '10',
+      });
+      expect(page2.directories).toEqual([]);
+    });
+
     it('paginates results with limit and cursor', async () => {
       const base = path.join(tmpDir, 'public', 'uploads');
       for (let i = 0; i < 25; i++) {
--- a/packages/@tinacms/cli/src/next/commands/dev-command/server/media.test.ts
+++ b/packages/@tinacms/cli/src/next/commands/dev-command/server/media.test.ts
@@ -60,6 +60,63 @@ describe('MediaModel (Vite dev server)', () => {
     });
   });
 
+  describe('listMedia search', () => {
+    const seed = async () => {
+      const base = path.join(tmpDir, 'public', 'uploads');
+      await fs.mkdirp(path.join(base, 'nested', 'deep'));
+      await fs.writeFile(path.join(base, 'llama.png'), 'x');
+      await fs.writeFile(path.join(base, 'cat.png'), 'x');
+      await fs.writeFile(path.join(base, 'nested', 'llama-baby.png'), 'x');
+      await fs.writeFile(path.join(base, 'nested', 'deep', 'LLAMA.jpg'), 'x');
+    };
+
+    it('matches filenames recursively and suppresses directories', async () => {
+      await seed();
+      const model = new MediaModel(config);
+      const result = await model.listMedia({ searchPath: '', search: 'llama' });
+      const names = result.files.map((f) => f.filename).sort();
+      expect(names).toEqual([
+        'llama.png',
+        'nested/deep/LLAMA.jpg',
+        'nested/llama-baby.png',
+      ]);
+      expect(result.directories).toEqual([]);
+    });
+
+    it('is case-insensitive', async () => {
+      await seed();
+      const model = new MediaModel(config);
+      const result = await model.listMedia({ searchPath: '', search: 'LLAMA' });
+      expect(result.files).toHaveLength(3);
+    });
+
+    it('scopes to the requested directory with dir-relative filenames', async () => {
+      await seed();
+      const model = new MediaModel(config);
+      const result = await model.listMedia({
+        searchPath: 'nested',
+        search: 'llama',
+      });
+      const names = result.files.map((f) => f.filename).sort();
+      expect(names).toEqual(['deep/LLAMA.jpg', 'llama-baby.png']);
+    });
+
+    it('prefixes src with the media root', async () => {
+      await seed();
+      const model = new MediaModel(config);
+      const result = await model.listMedia({ searchPath: '', search: 'cat' });
+      expect(result.files[0].src).toBe('/uploads/cat.png');
+    });
+
+    it('returns empty when nothing matches', async () => {
+      await seed();
+      const model = new MediaModel(config);
+      const result = await model.listMedia({ searchPath: '', search: 'zzz' });
+      expect(result.files).toEqual([]);
+      expect(result.directories).toEqual([]);
+    });
+  });
+
   describe('symlink traversal', () => {
     let outsideDir: string;
 
@@ -90,6 +147,15 @@ describe('MediaModel (Vite dev server)', () => {
       await expect(
         model.deleteMedia({ searchPath: 'escape/secret.txt' })
       ).rejects.toThrow(PathTraversalError);
+    });
+
+    it('search does not follow a symlink escaping the media root', async () => {
+      const model = new MediaModel(config);
+      const result = await model.listMedia({
+        searchPath: '',
+        search: 'secret',
+      });
+      expect(result.files).toEqual([]);
     });
   });
 
--- a/packages/@tinacms/cli/src/next/commands/dev-command/server/media.test.ts
+++ b/packages/@tinacms/cli/src/next/commands/dev-command/server/media.test.ts
@@ -70,7 +70,7 @@ describe('MediaModel (Vite dev server)', () => {
       await fs.writeFile(path.join(base, 'nested', 'deep', 'LLAMA.jpg'), 'x');
     };
 
-    it('matches filenames recursively and suppresses directories', async () => {
+    it('matches file paths recursively and suppresses directories', async () => {
       await seed();
       const model = new MediaModel(config);
       const result = await model.listMedia({ searchPath: '', search: 'llama' });
@@ -81,6 +81,69 @@ describe('MediaModel (Vite dev server)', () => {
         'nested/llama-baby.png',
       ]);
       expect(result.directories).toEqual([]);
+    });
+
+    it('matches a directory name against the files beneath it', async () => {
+      await seed();
+      const model = new MediaModel(config);
+      const result = await model.listMedia({ searchPath: '', search: 'deep' });
+      const names = result.files.map((f) => f.filename);
+      expect(names).toEqual(['nested/deep/LLAMA.jpg']);
+    });
+
+    it('paginates results with limit and cursor', async () => {
+      const base = path.join(tmpDir, 'public', 'uploads');
+      for (let i = 0; i < 25; i++) {
+        await fs.writeFile(
+          path.join(base, `match-${String(i).padStart(2, '0')}.png`),
+          'x'
+        );
+      }
+      const model = new MediaModel(config);
+
+      const page1 = await model.listMedia({
+        searchPath: '',
+        search: 'match',
+        limit: '10',
+      });
+      expect(page1.files).toHaveLength(10);
+      expect(page1.files[0].filename).toBe('match-00.png');
+      expect(page1.cursor).toBe('10');
+
+      const page3 = await model.listMedia({
+        searchPath: '',
+        search: 'match',
+        limit: '10',
+        cursor: page1.cursor as string,
+      });
+      expect(page3.files[0].filename).toBe('match-10.png');
+
+      const last = await model.listMedia({
+        searchPath: '',
+        search: 'match',
+        limit: '10',
+        cursor: '20',
+      });
+      expect(last.files).toHaveLength(5);
+      expect(last.cursor).toBeNull();
+    });
+
+    it('returns a structured result instead of throwing on a broken symlink', async () => {
+      await seed();
+      const base = path.join(tmpDir, 'public', 'uploads');
+      await fs.symlink(path.join(tmpDir, 'gone'), path.join(base, 'dangling'));
+      const model = new MediaModel(config);
+      const result = await model.listMedia({ searchPath: '', search: 'llama' });
+      expect(result.files).toHaveLength(3);
+    });
+
+    it('terminates on a symlink cycle inside the media root', async () => {
+      await seed();
+      const base = path.join(tmpDir, 'public', 'uploads');
+      await fs.symlink(base, path.join(base, 'loop'), 'dir');
+      const model = new MediaModel(config);
+      const result = await model.listMedia({ searchPath: '', search: 'llama' });
+      expect(result.files).toHaveLength(3);
     });
 
     it('is case-insensitive', async () => {
--- a/packages/@tinacms/cli/src/next/commands/dev-command/server/media.test.ts
+++ b/packages/@tinacms/cli/src/next/commands/dev-command/server/media.test.ts
@@ -58,6 +58,33 @@ describe('MediaModel (Vite dev server)', () => {
         model.listMedia({ searchPath: '%2e%2e/%2e%2e/%2e%2e/etc' })
       ).rejects.toThrow(PathTraversalError);
     });
+
+    it('returns all directories on the first page and paginates files', async () => {
+      const base = path.join(tmpDir, 'public', 'uploads');
+      for (let i = 0; i < 5; i++) {
+        await fs.mkdirp(path.join(base, `dir-${i}`));
+      }
+      for (let i = 0; i < 10; i++) {
+        await fs.writeFile(
+          path.join(base, `file-${String(i).padStart(2, '0')}.png`),
+          'x'
+        );
+      }
+      const model = new MediaModel(config);
+
+      const page1 = await model.listMedia({ searchPath: '', limit: '4' });
+      expect(page1.directories).toHaveLength(5);
+      expect(page1.files).toHaveLength(4);
+      expect(page1.cursor).toBe('4');
+
+      const page2 = await model.listMedia({
+        searchPath: '',
+        limit: '4',
+        cursor: '4',
+      });
+      expect(page2.directories).toEqual([]);
+      expect(page2.files).toHaveLength(4);
+    });
   });
 
   describe('listMedia search', () => {
@@ -144,6 +171,28 @@ describe('MediaModel (Vite dev server)', () => {
       const model = new MediaModel(config);
       const result = await model.listMedia({ searchPath: '', search: 'llama' });
       expect(result.files).toHaveLength(3);
+    });
+
+    it('keeps matches from readable folders when a nested folder cannot be read', async () => {
+      await seed();
+      const realReaddir = fs.readdir;
+      const spy = jest
+        .spyOn(fs, 'readdir')
+        .mockImplementation((dirPath: any, ...rest: any[]) =>
+          typeof dirPath === 'string' && dirPath.endsWith('nested')
+            ? Promise.reject(new Error('EACCES: permission denied'))
+            : (realReaddir as any).call(fs, dirPath, ...rest)
+        );
+      try {
+        const result = await new MediaModel(config).listMedia({
+          searchPath: '',
+          search: 'llama',
+        });
+        expect(result.files.map((f) => f.filename)).toEqual(['llama.png']);
+        expect(result.error).toBeUndefined();
+      } finally {
+        spy.mockRestore();
+      }
     });
 
     it('is case-insensitive', async () => {
--- a/packages/@tinacms/cli/src/next/commands/dev-command/server/media.test.ts
+++ b/packages/@tinacms/cli/src/next/commands/dev-command/server/media.test.ts
@@ -125,6 +125,16 @@ describe('MediaModel (Vite dev server)', () => {
       expect(result.directories).toEqual(['/nested/deep']);
     });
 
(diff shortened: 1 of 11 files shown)


packages/@tinacms/cli/src/next/commands/dev-command/server/media.ts
```diff
@@ -321,6 +324,17 @@ export class MediaModel {
           directories: [],
         };
       }
+
+      const search = args.search?.trim().toLowerCase();
+      if (search) {
+        return this.searchMedia({
```
joshbermanssw: **The missing `await` makes this whole path skip the `catch` block below.**

`return promise` inside a `try` does not route rejections through that `try`'s `catch`. So any error from the recursive walk escapes `listMedia` entirely and skips the handler at L397, whose comment says it exists specifically so internal details aren't returned to the client.

Confirmed against the real router: a dangling symlink under `public/uploads` throws `ENOENT` straight out of `handleList`, and `res.end` is never called, so the request goes unanswered and falls through to Vite's error middleware. `EACCES` on an unreadable subdirectory and `ELOOP` on an internal symlink loop behave the same way.

The contrast with the non-search path is the tell: given the identical dangling symlink, plain listing degrades gracefully and returns `{ files: [], directories: [], error }`. A stale symlink under `public/uploads` isn't exotic (one pointing into a sibling checkout that has since moved), and this takes search from "degrades" to "unhandled".

```suggestion
        return await this.searchMedia({
```

Worth a test asserting `listMedia` resolves rather than throws when the tree contains a dangling symlink, so this stays pinned.

packages/@tinacms/cli/src/next/commands/dev-command/server/media.ts
```diff
@@ -393,6 +407,51 @@ export class MediaModel {
...
+        let src = `/${relPath}`;
+        if (searchPath) src = `/${searchPath}${src}`;
+        if (this.mediaRoot) src = `/${this.mediaRoot}${src}`;
+        files.push({ src, filename: relPath, size: stat.size });
+      }
+    };
+
+    await walk(validatedPath, '');
+
+    return { files, directories: [] };
```
joshbermanssw: `searchMedia` never applies `args.limit` or `args.cursor`, and returns no cursor. Verified: 50 matching files requested with `limit: '20'` returns all 50, `cursor: undefined`.

The client sends `limit=20` on every search request (`media-store.default.ts:823`), so it's being silently dropped, and the response is unbounded on a library of any real size. The non-search path directly above slices at L385 and returns a cursor; reusing that same slice here would keep the two consistent, since right now the same endpoint paginates or doesn't depending on one query param.

packages/@tinacms/cli/src/next/commands/dev-command/server/media.ts
```diff
@@ -393,6 +407,51 @@ export class MediaModel {
...
+        } catch {
+          continue;
+        }
+        const relPath = relPrefix ? `${relPrefix}/${entry}` : entry;
+        const stat = await fs.stat(absPath);
+        if (stat.isDirectory()) {
+          await walk(absPath, relPath);
+          continue;
+        }
+        if (!relPath.toLowerCase().includes(search)) continue;
```
joshbermanssw, line 476: This matches the full relative path, not the filename. Verified: `search=invoice` returns `invoices/photo.png`, whose filename contains no match.

That contradicts the changeset ("matching filenames recursively"), the PR body, and the name of the test `matches filenames recursively and suppresses directories`. Both existing search fixtures happen to have the term in the basename, so nothing currently covers it.

It may well be the behaviour you want, but as it stands it's undocumented and untested, which is exactly how it drifts from whatever the cloud `?search=` does. Either match on `path.basename(relPath)`, or document path matching and add a case for it.

packages/@tinacms/cli/src/next/commands/dev-command/server/media.ts
```diff
@@ -393,6 +407,51 @@ export class MediaModel {
...
+        const absPath = join(dir, entry);
+        // @security Skip entries whose real path escapes the media root
+        // (symlink/junction), matching resolveWithinBase for the recursive walk.
+        try {
+          assertSymlinkWithinBase(absPath, resolvedBase, absPath);
+        } catch {
+          continue;
+        }
+        const relPath = relPrefix ? `${relPrefix}/${entry}` : entry;
+        const stat = await fs.stat(absPath);
```
joshbermanssw: Two performance notes on the walk, both amplified by the missing limit:

- `await fs.stat` runs serially per entry here, where the non-search path uses `Promise.all`.
- `assertSymlinkWithinBase` calls `realpathSync` **synchronously** once per entry, so the dev server's event loop is blocked for the duration of the traversal, not just slowed.

Separately, the walk has no cycle guard. An internal symlink loop (a link inside the media root pointing at an ancestor still inside it) passes the containment check, and recursion currently terminates only because the OS raises `ELOOP` at ~32 levels. Tracking visited real paths, or using `lstat` and skipping symlinked directories, would handle the cycle case and the dangling-symlink case in one change.

packages/tinacms/src/toolkit/components/media/media-manager.tsx
```diff
@@ -175,11 +186,12 @@ export function MediaPicker({
           { w: 1000, h: 1000 },
         ],
         filesOnly: !loadFolders,
+        search: debouncedSearch || undefined,
       });
-      setList({
-        items: [...list.items, ..._list.items],
+      setList((prev) => ({
```
joshbermanssw: Page-1 loads now replace instead of append, which is correct for search, but it removes the accidental safety appending gave us. There's no sequence guard or `AbortController` here, so if the request for `"a"` resolves after the request for `"ab"`, it overwrites the newer results and nothing re-fetches to correct it. The user ends up looking at `"ab"` in the box with results for `"a"`, or at an empty state reading `No media matches "ab"` when matches do exist.

The 300ms debounce bounds request *spacing*, not response *ordering*, so it narrows the window without closing it. A monotonically increasing ref captured before the call and compared before `setList` is enough.

packages/tinacms/src/toolkit/components/media/media-manager.tsx
```diff
@@ -383,6 +406,16 @@ export function MediaPicker({
...
+  const canSearch = !cms.media.store.isStatic && !!cms.media.store.searchable;
+  const visibleItems = list.items.filter((item) => {
+    if (mediaFilter === 'folders') return item.type === 'dir';
+    if (mediaFilter === 'files') return item.type === 'file';
+    return true;
+  });
+  const folders = visibleItems.filter((item) => item.type === 'dir');
+  const files = visibleItems.filter((item) => item.type === 'file');
+
   if ((listState === 'loading' && !list?.items?.length) || uploading) {
```
joshbermanssw: This guard undercuts the "input keeps focus and doesn't flash a full-screen loader" goal, though only on the zero-result path.

Once a search returns no matches, `list.items` is `[]`. The next keystroke sets `listState = 'loading'` while the list is still empty, this fires, and the entire picker (search box included) is replaced by `LoadingMediaList`. The input unmounts mid-typing and loses focus; `search` state survives since it lives on `MediaPicker`, so the value comes back but the caret doesn't.

Repro: search for something with no matches, then keep typing. Gating this on `!isSearching`, or on a separate initial-load flag, keeps the toolbar mounted.

packages/tinacms/src/toolkit/components/media/media-manager.tsx
```diff
@@ -482,48 +515,97 @@ export function MediaPicker({
...
-                } ${isDragActive ? `border-2 border-blue-500 rounded-lg` : ``}`}
+                className={`h-full grow overflow-y-auto p-4 transition duration-150 ease-out bg-gradient-to-b from-gray-50/50 to-gray-50 ${
+                  isDragActive ? `border-2 border-tina-orange rounded-lg` : ``
+                }`}
               >
                 <input {...getInputProps()} />
 
-                {listState === 'loaded' && list.items.length === 0 && (
-                  <EmptyMediaList />
+                {listState === 'loaded' && visibleItems.length === 0 && (
```
joshbermanssw: Two cases fall through to the default copy:

- `Folders` selected, not searching, and the directory has no subfolders, so it reads "Drag and drop assets here".
- `Files` selected on a folder-heavy page 1, where `visibleItems` is empty and this flashes while the next page loads.

packages/tinacms/src/toolkit/components/media/media-manager.tsx
```diff
@@ -760,37 +843,102 @@ const DocsLink = ({ title, message, docsLink, ...props }) => {
...
+  clear: () => void;
+}) => {
+  return (
+    <div className='relative flex flex-1 items-center min-w-[200px] max-w-md'>
+      <BiSearch className='absolute left-3 w-5 h-5 text-gray-400 pointer-events-none' />
+      <input
+        type='text'
+        value={value}
+        onChange={(e) => setValue(e.target.value)}
+        placeholder='Search this library...'
```
joshbermanssw, line 891: `placeholder` isn't an accessible name, so this input has none. It needs an `aria-label` or a visually hidden `<label>`.

Related in this file: the `MediaFilterToggle` and `ViewModeToggle` buttons have no `aria-pressed`, so their selected state is conveyed visually only.

packages/tinacms/src/toolkit/components/media/media-item.tsx
```diff
@@ -75,58 +75,94 @@ export function ListMediaItem({ item, onClick, active }: MediaItemProps) {
   );
 }
 
+const typeBadge = (filename: string): string | null => {
```
joshbermanssw, line 79: Dotfiles get a badge built from their own name: `.gitkeep` renders `GITKEEP`. `split('.').pop()` returns `gitkeep`, and the `ext === filename` guard doesn't catch it because the leading dot makes the two differ.

Skipping names that start with `.`, or capping badge length, handles it.

packages/tinacms/src/toolkit/components/media/media-item.tsx
```diff
@@ -75,58 +75,94 @@ export function ListMediaItem({ item, onClick, active }: MediaItemProps) {
...
         {item.new && (
-          <span className='absolute top-1 right-1 rounded shadow bg-green-100 border border-green-200 text-[10px] tracking-wide font-bold text-green-600 px-1.5 py-0.5 z-10'>
+          <span className='absolute top-2 right-2 z-10 rounded bg-green-100 border border-green-200 text-[10px] tracking-wide font-bold text-green-600 px-1.5 py-0.5'>
             NEW
           </span>
         )}
-        <div className='w-full overflow-hidden aspect-[1/1]'>
+        <div
+          className='relative w-full overflow-hidden bg-gray-50'
+          style={{ aspectRatio: '1 / 1' }}
```
joshbermanssw, line 137: This is the only inline style in the file, replacing the previous `aspect-[1/1]` class. If the arbitrary-value class was what wasn't working, Tailwind's built-in `aspect-square` should behave identically and keep this consistent with the rest of the styling here.

joshbermanssw (commented):
Review of the search + filter work. The architecture is sound and the `searchable` opt-in is the right call; a custom store that ignores `options.search` can't end up with a box that returns unfiltered results. The symlink test on the new recursive walk is a good instinct too.

Two things I'd fix before merge, both in the CLI dev-server search path (one is a one-word change), plus two user-visible bugs in the search flow that are self-contained in `media-manager.tsx`. Details inline. Server-side findings were verified by running the real code rather than read off the diff.

packages/@tinacms/cli/src/next/commands/dev-command/server/media.ts
joshbermanssw: Fixed in 2627cbfd3 — `return await`. Test added that the walk returns a structured result instead of throwing.

packages/@tinacms/cli/src/next/commands/dev-command/server/media.ts
joshbermanssw: Fixed in 2627cbfd3 — `searchMedia` takes `cursor`/`limit` and returns a cursor, with a stable sort so paging is deterministic. Test covers three pages.

packages/@tinacms/cli/src/next/commands/dev-command/server/media.ts
joshbermanssw: Kept path matching as intended. Changeset and test name reworded in 2627cbfd3, plus a test pinning folder-name matching.

packages/@tinacms/cli/src/next/commands/dev-command/server/media.ts
joshbermanssw: Fixed in 2627cbfd3 — `stat` calls parallelised, unstattable entries skipped, and a `visitedDirs` real-path set breaks cycles. Test added.

packages/tinacms/src/toolkit/components/media/media-manager.tsx
joshbermanssw: Fixed in 2627cbfd3 — `listRequestRef` counter; stale responses bail before both `setList` and `setListError`.

packages/tinacms/src/toolkit/components/media/media-manager.tsx
joshbermanssw: Fixed in 2627cbfd3 — guard now gated on `!isSearching`.

packages/tinacms/src/toolkit/components/media/media-manager.tsx
joshbermanssw: Fixed in 2627cbfd3 — copy extracted to `emptyMessage()` covering both filter cases, and suppressed when the loader will render.

packages/tinacms/src/toolkit/components/media/media-manager.tsx
joshbermanssw: Fixed in 2627cbfd3 — `aria-label` on the input, `aria-pressed` on both toggle groups, labels on the view buttons. Left `type=text`, since `type=search` adds a native clear button that collides with the existing one.

packages/tinacms/src/toolkit/components/media/media-item.tsx
joshbermanssw: Fixed in 2627cbfd3 — basename first, dotfiles skipped, extension length capped. Also fixes `my.folder/file`.

packages/tinacms/src/toolkit/components/media/media-item.tsx
joshbermanssw: Not changed. `aspect-square` only appears in vendored shadcn, and 070182ff0 deliberately fixed squareness, so leaving the inline style until someone can eyeball it.

packages/@tinacms/cli/src/next/commands/dev-command/server/media.test.ts
```diff
@@ -60,6 +60,126 @@ describe('MediaModel (Vite dev server)', () => {
...
+      const base = path.join(tmpDir, 'public', 'uploads');
+      for (let i = 0; i < 25; i++) {
+        await fs.writeFile(
+          path.join(base, `match-${String(i).padStart(2, '0')}.png`),
+          'x'
+        );
+      }
+      const model = new MediaModel(config);
+
+      const page1 = await model.listMedia({
```
18-th, line 176: Passes with the fix reverted. Swapping `return await` back to `return` at media.ts:330 leaves all 26 tests green, because the dangling symlink gets absorbed by the per-entry catch at :443-447. A readdir that throws on a nested dir would actually exercise it.

packages/@tinacms/cli/src/next/commands/dev-command/server/media.ts
```diff
@@ -393,6 +409,81 @@ export class MediaModel {
...
+    search: string;
+    cursor?: string;
+    limit?: string;
+  }): Promise<ListMediaRes> {
+    const resolvedBase = path.resolve(mediaBase);
+    const files: File[] = [];
+    const visitedDirs = new Set<string>([resolveRealPath(validatedPath)]);
+
+    const walk = async (dir: string, relPrefix: string) => {
+      const entries = await fs.readdir(dir);
```
18-th: `fs.readdir` isn't guarded like the stat and containment checks below it, so one unreadable subdirectory empties the whole result set. chmod 000 on a nested folder gives back `{files: [], directories: [], error: "EACCES"}`, which throws away matches already found elsewhere, and the client never reads `error`. A try/catch returning from this walk invocation would match how entries are handled.

packages/tinacms/src/toolkit/components/media/media-manager.tsx
```diff
@@ -383,7 +411,32 @@ export function MediaPicker({
...
+  const isSearching = !!debouncedSearch;
+  const canSearch = !cms.media.store.isStatic && !!cms.media.store.searchable;
+  const visibleItems = list.items.filter((item) => {
+    if (mediaFilter === 'folders') return item.type === 'dir';
+    if (mediaFilter === 'files') return item.type === 'file';
+    return true;
+  });
+  const folders = visibleItems.filter((item) => item.type === 'dir');
+  const files = visibleItems.filter((item) => item.type === 'file');
+  const showLoader = !!list.nextOffset && mediaFilter !== 'folders';
```
18-th, line 453: When a directory has more subfolders than pageSize, clicking Folders before anything has paged shows only the first page. `showLoader` suppresses the loader here, so page 2 never gets requested. `filesOnly` doesn't make later pages files-only either, since TinaMediaStore never sends it. Either drop the `mediaFilter !== 'folders'` condition, or partition in listMedia so directories all land on page 1.

packages/@tinacms/cli/src/next/commands/dev-command/server/media.ts
```diff
@@ -393,6 +409,81 @@ export class MediaModel {
...
+      const stats = await Promise.all(
+        entries.map(async (entry) => {
+          const absPath = join(dir, entry);
+          // @security Skip entries whose real path escapes the media root
+          // (symlink/junction), matching resolveWithinBase for the recursive walk.
+          try {
+            assertSymlinkWithinBase(absPath, resolvedBase, absPath);
```
18-th, line 448: Following up on the perf thread above. The realpath half is still synchronous. assertSymlinkWithinBase calls fs.realpathSync twice per entry (:218-219 via :190) plus once per directory at :458, so Promise.all doesn't help here. Flagging only because the thread reads as fully resolved.

packages/tinacms/src/toolkit/components/media/media-item.tsx
```diff
@@ -75,58 +75,96 @@ export function ListMediaItem({ item, onClick, active }: MediaItemProps) {
...
+            </span>
+          )}
+          {badge && (
+            <span className='absolute bottom-2 left-2 rounded bg-black/65 text-white text-[10px] font-semibold tracking-wide px-1.5 py-0.5'>
+              {badge}
+            </span>
+          )}
         </div>
-        <div className='mt-auto w-full px-2 py-1 text-sm truncate'>
+        <div className='w-full px-2.5 py-2 text-sm text-gray-700 truncate'>
```
18-th: `truncate` now clips the identifying part of path-shaped search results, so several hits from one folder all render as `nested/deep/...`. `title={item.filename}` covers it without touching layout. Same at :70 and :99.

packages/@tinacms/cli/src/next/commands/dev-command/server/media.ts
joshbermanssw: Fixed in 5bef64d — the walk's `readdir` is now wrapped, so an unreadable subfolder is skipped instead of emptying the whole result; matches from other folders are preserved. Mutation-checked: removing the guard fails the new test.

packages/@tinacms/cli/src/next/commands/dev-command/server/media.test.ts
joshbermanssw: Added a test in 5bef64d that makes `readdir` reject on the nested folder (via a spy, to stay deterministic under root in CI) and asserts the root match still comes back. It fails with the guard removed.

packages/tinacms/src/toolkit/components/media/media-manager.tsx
joshbermanssw: Took the partition option (5bef64d): `listMedia` now returns all directories on the first page and paginates only files, so Folders view shows every subfolder regardless of page size — and it avoids auto-paging through file pages just to surface later folders. Server test added covering >limit dirs on page 1.

packages/tinacms/src/toolkit/components/media/media-item.tsx
joshbermanssw: Fixed in 5bef64d — added `title={item.filename}` on the grid, list, and folder name elements, so the full path shows on hover without touching layout.

packages/@tinacms/cli/src/next/commands/dev-command/server/media.ts
joshbermanssw: Correct — the containment check (`assertSymlinkWithinBase` → `realpathSync`) is synchronous, so `Promise.all` parallelises the stats but not the realpath work. Leaving it as-is on purpose: that helper is the shared, CodeQL-inlined security primitive the non-search list/delete paths also call synchronously, and this runs over a small local dev-server tree, so forking an async copy isn't worth the security-surface churn. Flagging here so the thread isn't read as fully resolved.

18-th (dismissed):
LGTM

18-th (approved):
LGTM

PR #7327, opened 2026-07-27
feat: workos redirect auth

brookjeynes-ssw:
This PR adds support for workos redirect authentication when enabled.

Depends on https://github.com/tinacms/tinacms/pull/7304
Depends on changes in TinaCloud <TODO: link PR>


packages/tinacms/src/internalClient/authProvider.ts
```diff
@@ -23,8 +23,9 @@ export abstract class AbstractAuthProvider implements AuthProvider {
   async fetchWithToken(input: Input, init: Init): FetchReturn {
     const headers = init?.headers || {};
     const token = await this.getToken();
-    if (token?.id_token) {
-      headers['Authorization'] = 'Bearer ' + token?.id_token;
+    const accessToken = token?.id_token ?? token?.access_token;
```
18-th: You have an access token first in e.g. Playground.tsx, here it is reversed, which one is correct?

packages/tinacms/src/auth/authenticate.test.ts
```diff
@@ -80,72 +101,142 @@ const settleViaPopupClose = async (result: Promise<unknown>) => {
   await expectation;
 };
 
-describe('authenticate origin/source validation', () => {
```
18-th, line 83: Why did we delete old tests?

packages/tinacms/src/auth/authenticate.ts
```diff
@@ -1,39 +1,57 @@
...
 
-export const authenticate = (
+let workosEnabledPromise: Promise<boolean> | null = null;
+
+export async function getWorkosEnabled(
+  identityApiUrl: string
+): Promise<boolean> {
+  if (!workosEnabledPromise) {
+    workosEnabledPromise = (async () => {
+      const res = await fetch(`${identityApiUrl}/v2/auth/config`);
```
18-th, line 24: If this fails - you would get the workosEnabledPromise stuck in a rejected state, would you not?

packages/tinacms/src/auth/useTinaAuthRedirect.tsx
```diff
@@ -1,24 +1,82 @@
...
-      code: urlParams.get('code') || '',
-      scope: urlParams.get('scope') || 'email',
-      state: urlParams.get('state'),
-    };
+    if (!code || !state) {
+      return;
+    }
 
-    if (!config.code) {
+    const pkceData = JSON.parse(
```
18-th, line 31: Can we get some test coverage for this file? It's new work, and it's around security

18-th:
Can we get a changeset?

18-th:
I believe we have removed the last usage of `crypto.js` in this. Do you reckon we could remove the package now?

packages/tinacms/src/auth/TinaCloudProvider.tsx
```diff
@@ -345,7 +356,19 @@ export const TinaCloudProvider = (
...
+  const authRedirectParams: AuthRedirectParams = {
+    code: urlParams.get('code'),
+    state: urlParams.get('state'),
+    error: urlParams.get('error'),
+  };
+
+  const isAuthRedirect = !!(
+    authRedirectParams.code && authRedirectParams.state
+  );
+  useTinaAuthRedirect(authRedirectParams);
```
18-th: I believe this and the above will mess with the auth for the self-hosted since this is not guarded

packages/tinacms/src/auth/authenticate.ts
brookjeynes-ssw: From my testing, an error thrown rejects the promis

packages/tinacms/src/internalClient/authProvider.ts
brookjeynes-ssw: resolved by https://github.com/tinacms/tinacms/pull/7327/changes/4085774248c1391f504d1dbcc3fc998de4a02b97

packages/tinacms/src/auth/TinaCloudProvider.tsx
brookjeynes-ssw: resolved by https://github.com/tinacms/tinacms/pull/7327/changes/be98a8c02333dc75aeb52af2359c02894847a0ce

packages/tinacms/src/auth/useTinaAuthRedirect.tsx
brookjeynes-ssw: resolved by https://github.com/tinacms/tinacms/pull/7327/changes/d6a5153c6558ef261da3a1f3f0dcccafe1240e2d

KahaMason:
✅ Looks good, tested this locally.

Testing Checklist:
- [x] TinaCMS login redirects to AuthKit (WorkOS enabled)
- [x] Expired Token triggers a refresh successfully through TinaCloud
- [x] Cognito login behavior unchanged.

KahaMason (dismissed):
LGTM 🚀

packages/tinacms/src/auth/authenticate.ts
18-th: Yes, that's correct. The problem is that the rejected promise is cached until the page reloads. So every call through this would just hit a cached rejected promise. Is that by design?

packages/tinacms/src/internalClient/authProvider.ts
```diff
@@ -170,11 +175,7 @@ export class TinaCloudAuthProvider extends AbstractAuthProvider {
     this.setToken(null);
   }
 
-  async getRefreshedToken(tokens: {
-    access_token?: string;
-    id_token?: string;
-    refresh_token?: string;
-  }): Promise<TokenObject> {
+  async getRefreshedToken(tokens: TokenObject): Promise<TokenObject> {
```
18-th, line 178: With WorkOS here, clientId on the TokenObject no longer matches the project id, so the refresh fails for WorkOS.

We could use `this.clientId` instead, and that should do it for WorkOS. In my testing, by setting the expiry to now, the refresh fails with a CORS issue.

We just need to make sure that it still works for Cognito though.

packages/tinacms/src/auth/useTinaAuthRedirect.tsx
```diff
@@ -1,24 +1,89 @@
...
+
+    const tokenParams = new URLSearchParams({
+      grant_type: 'authorization_code',
+      code,
+      redirect_uri: redirectUri,
+      client_id,
+      code_verifier,
+    });
+
+    fetch(`${identity_api_url}/oauth/token`, {
```
18-th, line 58: We need an `res.ok` check for this call, otherwise errors go through as 200 since TinaCloud returns `{"message": blah...}` as an error. The mock in the test for this is wrong I reckon, could you double-check?

brookjeynes-ssw: Resolved by a9a6008cb76087c92ce6dab559ff5779428021ee

packages/tinacms/src/internalClient/authProvider.ts
brookjeynes-ssw: Resolved by a9a6008cb76087c92ce6dab559ff5779428021ee

18-th (approved):
LGTM

PR #7362, opened 2026-07-30
feat(v4): plugin runtime — capability graph, RPC transport, plugin lifecycle

JackDevAU:
**TLDR:** The v4 plugin runtime: plugins declare capabilities, the graph resolves and boots them, and a typed RPC transport serves the server half.

**Feats:**
- Capability graph resolution, with cycles rejected at validate
- Typed RPC handler and proxy client, with sanitized dispatch failures and a JSON content-type CSRF guard
- onInit/onDestroy plugin lifecycle in dependency order, reverse-order teardown
- RpcHandler gets destroy() and a pinned mount path
- Plugin manifests normalized at definePlugin, so the runtime drops its ?? [] guards

**How to test:** The vitest suite covers the runtime end to end.
- Run `pnpm install`.
- Run `pnpm test`.
- Run `pnpm types`.


packages/v4/@tinacms/tinacms/src/rpc/handler.ts
```diff
@@ -0,0 +1,369 @@
...
+  const composeAfterCurrentTurn = (): Promise<ServerRuntime> => {
+    const composing = lifecycleTurn.then(() => composeServerRuntime(plugins));
+    lifecycleTurn = composing.then(
+      () => undefined,
+      () => undefined
+    );
+    return composing.catch((cause) => {
+      runtimePromise = null;
+      throw cause;
+    });
```
18-th: This catch nulls `runtimePromise` without checking it still owns the slot, so a compose that fails after a `destroy()` cleared it and a later request refilled it wipes the newer promise. I reproduced it with a deferred `server()` rejection — request A pins P1, `destroy()` clears the slot, request B assigns P2, then P1 rejects and nulls it, so request C composes a third runtime: `onInit` ran twice against a single `onDestroy`, orphaning P2's plugins, which is the leak `destroy()` exists to prevent. Naming the returned promise and guarding with `if (runtimePromise === composed)` fixed it with the existing 33 handler tests still green.

packages/v4/@tinacms/tinacms/src/core/resolve.ts
```diff
@@ -0,0 +1,180 @@
...
+    if (failures.length > 1) {
+      throw new AggregateError(failures, 'Plugin teardown failed.');
+    }
+  };
+  for (const plugin of orderPluginsByDependencies(plugins)) {
+    try {
+      await plugin.onInit?.();
+      initialized.push(plugin);
+    } catch (cause) {
+      await destroyInitialized().catch(() => {});
```
18-th, line 159: This discards every rollback teardown failure, which sits oddly with the "Surface every teardown failure, not just the first" comment a few lines above — and handler.ts:104 says initializePlugins logs the teardown that rolls back a failed init, which it doesn't. Logging the cause here the way the other `[tinacms]` sites do would make that comment true; if silence is deliberate, then the comment is the thing to change.

packages/v4/@tinacms/tinacms/src/rpc/proxy.ts
```diff
@@ -0,0 +1,119 @@
...
+    ?.toLowerCase()
+    .includes('application/json');
+  if (response.ok && !isJson) {
+    throw new RpcError(
+      response.status,
+      'rpc-not-json',
+      `RPC ${namespace}/${opName} returned ${response.status} but not JSON.`
+    );
+  }
+  const payload = await response.json().catch(() => null);
```
18-th, line 110: A malformed 200 lands here as `null` and resolves the call as an empty success — the exact case the `rpc-not-json` guard above is written to prevent, since that guard only inspects the header. Reading the text and throwing when the response was ok, keeping the null fallback for the error path, would close it. Related: the `includes('application/json')` on line 102 is the substring check the handler moved off in 2294199, so the two ends of the transport now disagree on what counts as JSON.

packages/v4/@tinacms/tinacms/src/rpc/proxy.ts
```diff
@@ -0,0 +1,119 @@
...
+// which the compiler erases, so no server code and no secret reaches the browser. This
+// file runs in the browser, and it must not import from ../server or from ./handler.
+
+export type RpcProxy<TSegments> = {
+  [Namespace in keyof TSegments]: {
+    [Op in keyof TSegments[Namespace]]: TSegments[Namespace][Op] extends (
+      input: infer TInput
+    ) => Promise<infer TResult>
+      ? (input: TInput) => Promise<TResult>
+      : never;
```
18-th, line 12: `extends (input: infer TInput) => …` forces a parameter, so a zero-arg op can't be called with zero args: `client.search.reindex()` is TS2554, which is why proxy.test.ts has to write `reindex(undefined)`. `(...args: infer TArgs) => Promise<infer TResult> ? (...args: TArgs) => …` preserves arity — I checked it compiles both call forms. Worth taking before anything consumes `./client`, since it turns `op(undefined)` into a compile error.

packages/v4/@tinacms/tinacms/src/rpc/handler.ts
JackDevAU: ✅ Done

packages/v4/@tinacms/tinacms/src/core/resolve.ts
JackDevAU: Not Done as this has been fixed in #7376

packages/v4/@tinacms/tinacms/src/rpc/proxy.ts
JackDevAU: Not Done as this has been fixed in #7381

packages/v4/@tinacms/tinacms/src/rpc/proxy.ts
JackDevAU: ✅ Done in #7383

18-th (approved):
LGTM

PR #7366, opened 2026-07-30
feat(v4): defineConfig — one composition root, compiled to tina-lock.json

JackDevAU:
**TLDR:** defineConfig is the one composition root: a collection is declared once, the plugin graph fails at import, and the schema compiles to tina-lock.json behind a contract-version gate.

**Feats:**
- defineConfig folds in the built-in field plugins and validates the capability graph at import
- compileSchema writes tina-lock.json and pins only the field types the schema uses (ADR-016)
- checkLock regenerates a stale lock and hard-stops on a contract-version mismatch
- Templates enter the schema, and the compile walks their fields for the gate and the pins
- A field plugin declares its type key on the manifest, so the compile never imports client segments

**How to test:** The suite covers the composition root and the compile.
- Run `pnpm test`.
- Run `pnpm types`.

Note: this PR is based on `wip-lockfile`, a parking branch that holds a pnpm-lock refresh from a pull. It has no PR of its own.


JackDevAU:

PR #7367, opened 2026-07-30
feat(v4): tinacms codegen — the CLI that writes tina-lock.json

JackDevAU:
**TLDR:** The `tinacms` bin arrives with one command: codegen compiles the schema to tina-lock.json, conservatively, in the runtime package (ADR-001).

**Feats:**
- An already-current lock is left untouched; a stale lock regenerates and says why
- A contract-version mismatch stops without touching the disk
- Bad flags exit 1 with a message instead of a stack trace; Windows-style --root paths resolve
- An unreadable config surfaces as an error instead of reading as "no config"
- The bin boots one Vite server and lends it to the config loader

**How to test:** Run codegen against the playground.
- Run `pnpm test`.
- Run `node bin/tinacms.mjs codegen --root playground`.
- Confirm it reports the lock as up to date.


JackDevAU:

PR #7369, opened 2026-07-30
fix(v4): nothing in v4 claims a fixed port

JackDevAU:
**TLDR:** Reading a config no longer binds Vite's HMR socket on a fixed port, and the e2e harness probes a free port instead of pinning 5174.

**Feats:**
- loadTinaConfig sets ws: false, so a config read opens no socket
- The playwright harness asks the OS for a port and publishes it to the workers; TINA_E2E_PORT still pins it
- A test asserts the property itself: a config load opens no server

**How to test:** The property is pinned by a unit test.
- Run `pnpm test`.
- Start `pnpm test:e2e` twice at the same time.
- Confirm both runs bind their own port and pass.


JackDevAU:

PR #7370, opened 2026-07-30
feat(v4): keep the CLI out of the build pipeline; implement the framework adapters

JackDevAU:
**TLDR:** The bin only writes files a person commits — no dev, no build — and the four framework adapters do the serving instead.

**Feats:**
- `tinacms codegen --check` writes nothing and exits 1 on a stale lock, so CI guards drift without running the bin in the build
- Next, Astro and Hono adapters are pure wiring; Express converts node streams both ways
- Vite becomes an optional peer; react and react-dom become optional
- Unused dependencies dropped

**How to test:** The check mode is the new surface.
- Run `pnpm test`.
- Run `node bin/tinacms.mjs codegen --check --root playground`.
- Confirm the exit code is 0.


JackDevAU:

PR #7371, opened 2026-07-30
chore(v4): move shared UI deps to the pnpm catalog; align the v4 manifests

JackDevAU:
**TLDR:** The shared UI dependencies move into the pnpm catalog, and the v4 packages consume them from there.

**Feats:**
- @headlessui/react, @heroicons/react and mermaid live in the catalog
- @tinacms/rich-text drops prism-react-renderer; @tinacms/tinacms gains @tanstack/react-query and the React Compiler plugin
- The v4 vitest configs run through the React Compiler
- The v3 plate plugin arrays drop their any[] annotation workaround

**How to test:** The lockfile is the contract.
- Run `pnpm install`. Confirm the lockfile does not change.
- Run `pnpm test` in `packages/v4/@tinacms/tinacms`.
- Run `pnpm test` in `packages/v4/@tinacms/rich-text`.


JackDevAU:

PR #7373, opened 2026-07-30
feat(v4): a field declares its own equality, and the form store asks it

JackDevAU:
**TLDR:** The form store consults a per-field equality instead of reference identity, so a typed-and-deleted body reaches clean again.

**Feats:**
- core/form/compare.ts: sameValue structural test, FieldEquality per open form
- The field contract gains isEqual, bound to the schema node and the document context
- rich-text answers equality through its codec

**How to test:** The store tests pin the behaviour; the playground shows it.
- Run `pnpm test`.
- Run `pnpm dev` and open a document.
- Type one character in the body, then delete it.
- Confirm the document shows no unsaved badge.


JackDevAU:

PR #7374, opened 2026-07-30
feat(v4): datetime field plugin

JackDevAU:
**TLDR:** The datetime field plugin: ISO-string values with no zone math.

**Feats:**
- The schema/plugin/client/ui quartet, like the other primitive fields
- A YAML frontmatter Date folds back to its string form on parse and validate, so an untouched document digests back unchanged

**How to test:** The field suite covers parse, validate and render.
- Run `pnpm test`.


JackDevAU:

PR #7383, opened 2026-07-30
feat(v4): the lock carries the tinacms version; the runtime validates the plugin graph harder

JackDevAU:
**TLDR:** tina-lock.json stamps the installed tinacms version (LOCK_VERSION 5), and the runtime rejects more invalid plugin graphs at boot.

**Feats:**
- compileSchema stamps schema.version from the installed package
- An overrides entry without a matching provides is rejected; so is a plugin with two singleton slices
- The compile and the boot compose the field registry the same way, so they fail the same way
- The RPC handler answers 401 when no auth hooks are mounted; DEFAULT_ROLE_PERMISSIONS is frozen; a handler cannot be tagged twice
- Document extension matching is case-insensitive

**How to test:** The compile, resolve and RPC suites cover each rule.
- Run `pnpm test` and `pnpm types`.

### DIFF
--- a/packages/v4/@tinacms/tinacms/playground/tina/tina-lock.json
+++ b/packages/v4/@tinacms/tinacms/playground/tina/tina-lock.json
@@ -1,5 +1,5 @@
 {
-  "version": 4,
+  "version": 5,
   "schema": {
     "collections": [
       {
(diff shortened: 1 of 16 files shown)


JackDevAU:

PR #7384, opened 2026-07-30
feat(v4): TinaAdmin mounts its own provider and router

JackDevAU:
**TLDR:** TinaAdmin takes the config and mounts TinaProvider and HashRouter itself, so an app renders one component.

**Feats:**
- TinaAdmin takes config and queryClient; no hand-composed provider
- A string preview prop renders the preview iframe with the connection wired
- useAdminRoute reads react-router instead of raw hashchange events
- The save badge is a status landmark; the collapsed sidebar goes inert; document buttons carry accessible names
- Provider teardown chains on the previous lifecycle turn, so overlapping mounts cannot interleave

**How to test:** The admin suite runs against the new mount; the playground uses it.
- Run `pnpm test`.
- Run `pnpm dev` and open the admin.
- Confirm the admin boots and the preview renders.

### DIFF
--- a/packages/v4/@tinacms/tinacms/_docs/architecture.md
+++ b/packages/v4/@tinacms/tinacms/_docs/architecture.md
@@ -9,6 +9,8 @@ hooks → validation → digest.
 
 ## 1. Resolve the plugins into a registry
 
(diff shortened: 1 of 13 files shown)


JackDevAU:

PR #7385, opened 2026-07-30
feat(v4): tinacms init, and the barebones example that runs its output

JackDevAU:
**TLDR:** `tinacms init` scaffolds a project — config, admin route, first document — and the new barebones example is that output, running.

**Feats:**
- init writes the starter tina/config.ts, the admin route and a first document; it keeps every file that already exists
- codegen writes the admin files beside the lock, so a fresh project boots from two commands
- packages/v4/examples/barebones: one Vite app, one collection, a custom field, a preview, no framework
- The pnpm workspace gains packages/v4/examples/*; load-config imports vite optionally

**How to test:** Run the example, then run init against an empty folder.
- Run `pnpm install`.
- Go to `packages/v4/examples/barebones`.
- Run `pnpm dev`, open the admin, and edit the post.
- Run `node bin/tinacms.mjs init --root /tmp/tina-init-test`.
- Confirm it lists the files it wrote.


JackDevAU:

PR #7388, opened 2026-07-30
feat: add announcements banner

brookjeynes-ssw:
Add an announcement component linked to the /announcement endpoint

Issue: TinaCloud#3898 Post alerts to TinaCMS editor and TinaCloud dashboard
Assisted-by: OpenCode:big-pickle


18-th:
Do we want to share the dismissed state and the network call between two callouts?

18-th:
Can we get some tests at least for the core calls?

18-th:
We don't have any guards for the self-hosted setups anywhere. So the self-hosted will try to call into the void for the announcements.

packages/tinacms/src/admin/index.tsx
```diff
@@ -195,7 +199,15 @@ const PreviewInner = ({ preview, config }) => {
     }, 100);
   }, [ref.current]);
   const Preview = preview;
-  return <Preview url={url} iframeRef={ref} {...config} />;
+  return (
+    <div className='flex flex-col h-screen'>
+      <style>{`#tina-iframe { height: 100% !important; }`}</style>
```
18-th: Is manual styling left over from debugging?

brookjeynes-ssw: Resolved by 2eb64bc4070b023eb18a45a5a6374282abf4fb23

18-th (approved):
LGTM

Issue #6722, opened 2026-04-22
Media Library: allow editors to rename files


kulesy:
## Pain

Editors without direct git access have no way to rename a media file. The Media Library UI has no rename action, so the only path today is to delete the file and re-upload with the new name, or to rename it in GitHub — both of which require an editor who's comfortable working outside TinaCMS.

This becomes a particular problem when an upload lands with a bad filename (e.g. special characters that break downstream builds — see [tinacms#6359](https://github.com/tinacms/tinacms/issues/6359) for upload-time sanitisation). The content editor spots the issue but can't fix it themselves.

Reported via Discord (Koos, 2026-04-14) — editor uploaded an image with special characters, Astro build failed, no in-app way to rename.

> **Note:** The GitHub-push → Media Library sync does currently handle renames correctly (an earlier bug in this area has been fixed). This PBI is about closing the **UX gap** — giving editors a first-class rename action in the UI — not about fixing broken sync. The documented [delete-and-re-upload / GitHub-rename workarounds](https://tina.io/docs/reference/media/repo-based#renaming-media-files) work end-to-end today.

## Description

Add a **Rename** action to the Media Library UI that:

- Exposes a rename control (button or context menu item) on each file in the Media Library
- Prompts for a new filename with inline validation — sanitised against the same rules used on upload (see [tinacms#6359](https://github.com/tinacms/tinacms/issues/6359))
- Calls the backend rename endpoint (tracked separately — see **Depends On** below)
- Refreshes the Media Library listing and any open media picker to reflect the new path
- Surfaces error states clearly (name collision, invalid characters, backend failure)

Wires up the `rename?` stub on the `MediaStore` interface (already reserved per [tinacloud#3330](https://github.com/tinacms/tinacloud/issues/3330)).

## Out of scope

- Backend implementation — tracked in the companion tinacloud PBI (see **Depends On**)
- Rewriting content references (markdown/json) that point at the old path — cross-cutting concern, worth a separate PBI
- Filename sanitisation on upload — tracked in [tinacms#6359](https://github.com/tinacms/tinacms/issues/6359)

## Related

- [tinacms#6359](https://github.com/tinacms/tinacms/issues/6359) — filename sanitisation on upload (rename should apply the same rules)
- [tinacms#4465](https://github.com/tinacms/tinacms/issues/4465) — earlier Tina Cloud stale-paths bug (SSW-reported; now fixed)
[...]

PR #7391, opened 2026-08-02
feat(media): add a Rename action to the Media Manager (local media store)

Aibono1225:
Refs #6722 — implements the local file-system half. Tina Cloud rename stays blocked on tinacloud#3442 and is intentionally **not** part of this PR.

## What

Editors can now rename a media file from the Media Manager instead of deleting and re-uploading it, or going to GitHub.

Selecting a file in the preview panel offers **Rename** alongside Insert and Delete. The modal edits the basename, keeps the extension, previews the sanitised result using the same rules uploads apply, and reports collisions and missing files specifically rather than with a generic failure.

## How it works

`RenameModal` → `MediaManager.rename(from, to)` → `TinaMediaStore.rename` → `POST /media/rename` on the CLI dev server → `MediaModel.renameMedia`.

- `from` / `to` are media-root-relative `directory/filename` paths, the same shape `delete` already sends. The route also accepts (and ignores) `branch` for parity with the cloud contract.
- Both paths are resolved with `resolveStrictlyWithinBase`, so a rename can never land outside the configured media root.
- The move is `overwrite: false`; an existing destination comes back as a `NAME_COLLISION` instead of clobbering a file. Case-only renames (`photo.jpg` → `Photo.jpg`) hop through a staging name so they work on case-insensitive filesystems.
- `media:rename:success` refreshes every open media picker, and pickers previewing the renamed file follow it to its new path.

Screenshot 2026-08-03 at 11 44 40 am

**Figure: UI with Rename button**

Screenshot 2026-08-03 at 11 45 19 am

**Figure: UI of rename modal**

## Testing

- `packages/@tinacms/cli` — `MediaModel.renameMedia` (happy path, subdirectories, NOT_FOUND, NAME_COLLISION, folder source,
case-only rename, staging rollback, traversal ane route handler's status-code mapping, and thecross-origin gate on `/media/rename`.
- `packages/tinacms` — `MediaManager.rename` eveity detection and error mapping, `RenameModal`behaviour, and `MediaPicker` gating / list refresh / multi-picker sync.
- `playwright` — upload → rename → list → fetch l HTTP, plus collision, NOT_FOUND and traversalvectors.

### DIFF
--- a/packages/@tinacms/cli/src/next/commands/dev-command/server/media.ts
+++ b/packages/@tinacms/cli/src/next/commands/dev-command/server/media.ts
@@ -64,9 +64,7 @@ export const createMediaRouter = (config: PathConfig) => {
     // Body is populated by the bodyParser.json middleware registered ahead of
     // this router, the same way the /graphql route consumes it.
     const body = (req as unknown as { body?: unknown }).body;
-    const { from, to, ...rest } =
-      (body as { from?: unknown; to?: unknown }) || {};
-    void rest; // `branch` is accepted for parity with the cloud contract
+    const { from, to } = (body as { from?: unknown; to?: unknown }) || {};
 
     if (typeof from !== 'string' || typeof to !== 'string' || !from || !to) {
       res.statusCode = 400;
@@ -213,6 +211,17 @@ const RENAME_ERROR_STATUS: Record<RenameFailureCode, number> = {
   UNSUPPORTED: 400,
   BACKEND_FAILURE: 500,
 };
+
+/**
+ * Raised when a staged rename cannot put the file back where it started, so
+ * the file is left under the staging name. Carries that name so the response
+ * can tell the editor where to find it.
+ */
+class StagedRenameError extends Error {
+  constructor(public readonly stagingName: string) {
+    super(`Left the file as "${stagingName}" in the same folder.`);
+  }
+}
 
 /** fs-extra's move rejects an existing destination with a bare message. */
 const isDestinationExistsError = (error: unknown) =>
@@ -624,22 +633,33 @@ export class MediaModel {
       return {
         ok: false,
         code: 'BACKEND_FAILURE',
-        message: 'Failed to rename the file.',
+        message:
+          error instanceof StagedRenameError
+            ? `Failed to rename the file. ${error.message}`
+            : 'Failed to rename the file.',
       };
     }
   }
 
   /**
    * A case-insensitive filesystem can treat `a.jpg` -> `A.jpg` as a no-op, so
-   * hop through a unique sibling name. On failure the source is put back.
+   * hop through a unique sibling name. On failure the source is put back; if
+   * even that fails the file survives under the staging name, which
+   * StagedRenameError reports rather than leaving it to be found by accident.
    */
   private async renameViaStaging(source: string, destination: string) {
-    const staging = join(path.dirname(source), `.tina-rename-${randomUUID()}`);
+    const stagingName = `.tina-rename-${randomUUID()}`;
+    const staging = join(path.dirname(source), stagingName);
     await fs.move(source, staging, { overwrite: false });
     try {
       await fs.move(staging, destination, { overwrite: false });
     } catch (error) {
-      await fs.move(staging, source, { overwrite: false }).catch(() => {});
+      try {
+        await fs.move(staging, source, { overwrite: false });
+      } catch (restoreError) {
+        console.error(restoreError);
+        throw new StagedRenameError(stagingName);
+      }
       throw error;
     }
   }
--- a/packages/@tinacms/cli/src/next/commands/dev-command/server/media.ts
+++ b/packages/@tinacms/cli/src/next/commands/dev-command/server/media.ts
@@ -1,3 +1,4 @@
+import { randomUUID } from 'crypto';
 import type { ServerResponse } from 'http';
 import path, { join } from 'path';
 import busboy from 'busboy';
@@ -59,6 +60,48 @@ export const createMediaRouter = (config: PathConfig) => {
     }
   };
 
+  const handleRename = async (req: Connect.IncomingMessage, res) => {
+    // Body is populated by the bodyParser.json middleware registered ahead of
+    // this router, the same way the /graphql route consumes it.
+    const body = (req as unknown as { body?: unknown }).body;
+    const { from, to, ...rest } =
+      (body as { from?: unknown; to?: unknown }) || {};
+    void rest; // `branch` is accepted for parity with the cloud contract
+
+    if (typeof from !== 'string' || typeof to !== 'string' || !from || !to) {
+      res.statusCode = 400;
+      res.end(
+        JSON.stringify({
+          code: 'INVALID_FILENAME',
+          message: 'Both "from" and "to" are required.',
+        })
+      );
+      return;
+    }
+
+    try {
+      const result = await mediaModel.renameMedia({ from, to });
+      // `in` rather than `result.ok`: this repo compiles with `strict: false`,
+      // where boolean-literal discriminants do not narrow.
+      if ('code' in result) {
+        res.statusCode = RENAME_ERROR_STATUS[result.code];
+        res.end(JSON.stringify({ code: result.code, message: result.message }));
+        return;
+      }
+      res.statusCode = 200;
+      res.end(JSON.stringify({ success: true, from, to }));
+    } catch (error) {
+      if (error instanceof PathTraversalError) {
+        res.statusCode = 403;
+        res.end(
+          JSON.stringify({ code: 'INVALID_PATH', message: error.message })
+        );
+        return;
+      }
+      throw error;
+    }
+  };
+
   const handlePost = async function (
     req: Connect.IncomingMessage,
     res: ServerResponse
@@ -107,7 +150,7 @@ export const createMediaRouter = (config: PathConfig) => {
     req.pipe(bb);
   };
 
-  return { handleList, handleDelete, handlePost };
+  return { handleList, handleDelete, handlePost, handleRename };
 };
 
 export const parseMediaFolder = (str: string) => {
@@ -153,6 +196,28 @@ export interface PathConfig {
 }
 
 type SuccessRecord = { ok: true } | { ok: false; message: string };
+
+export type RenameFailureCode =
+  | 'NOT_FOUND'
+  | 'NAME_COLLISION'
+  | 'UNSUPPORTED'
+  | 'BACKEND_FAILURE';
+
+type RenameRecord =
+  | { ok: true }
+  | { ok: false; code: RenameFailureCode; message: string };
+
+const RENAME_ERROR_STATUS: Record<RenameFailureCode, number> = {
+  NOT_FOUND: 404,
+  NAME_COLLISION: 409,
+  UNSUPPORTED: 400,
+  BACKEND_FAILURE: 500,
+};
+
+/** fs-extra's move rejects an existing destination with a bare message. */
+const isDestinationExistsError = (error: unknown) =>
+  (error as { code?: string })?.code === 'EEXIST' ||
+  /dest already exists/i.test((error as Error)?.message || '');
 
 /**
  * Detects URL-encoded path-traversal sequences that should have been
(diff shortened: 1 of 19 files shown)


packages/tinacms/src/toolkit/components/media/media-manager.tsx
```diff
@@ -644,6 +689,10 @@ export function MediaPicker({
               deleteMediaItem={() => {
                 setDeleteModalOpen(true);
               }}
+              allowRename={canRename}
```
joshbermanssw:
```suggestion
              canRename={canRename}
```

Should we keep the naming consistent? (if yes apply throughout)

packages/tinacms/src/toolkit/components/media/modal.tsx
```diff
@@ -1,14 +1,17 @@
-import React from 'react';
 import { Input } from '@toolkit/fields';
+// Direct path rather than the '@tinacms/toolkit' barrel: the barrel re-exports
```
joshbermanssw: rm comments

packages/tinacms/src/toolkit/components/media/modal.tsx
```diff
@@ -20,6 +23,129 @@ interface NewFolderModalProps {
...
+  const sanitizedBase = sanitizeFilename(trimmedBase);
+  const sanitized = `${sanitizedBase}${ext}`;
+  const isEmpty = trimmedBase.length === 0;
+  // Nothing usable survived, so sanitizeFilename fell back to its placeholder
+  // instead of producing something derived from the input.
+  const isStripped =
+    !isEmpty && sanitizedBase === 'file' && trimmedBase !== 'file';
+  const isUnchanged = sanitized === filename;
+  const canSubmit = !processing && !isEmpty && !isStripped && !isUnchanged;
+  const showPreview = !isEmpty && !isStripped && sanitizedBase !== trimmedBase;
```
joshbermanssw: Not the number of consts, that part's fine. Every name here is used 3+ times, and inlining `canSubmit` into `disabled={...}` would read worse than naming it.

Two things I would change though:

1. Both comments explain *why*, which we'd rather keep in the PR body than inline.
2. This is a pure function of `(nextBase, ext, filename)` living inside a React component. To test it you have to render the modal, and it forces `sanitizeFilename`'s internal `'file'` fallback to leak out as a string literal on line 69.

Moving it next to `sanitizeFilename` in `utils.ts` solves both:

```ts
export const previewRename = (input: string, ext: string, current: string) => {
  const base = input.trim();
  const sanitizedBase = sanitizeFilename(base);
  const isEmpty = base.length === 0;
  const isStripped =
    !isEmpty && sanitizedBase === FALLBACK_NAME && base !== FALLBACK_NAME;
  const sanitized = `${sanitizedBase}${ext}`;

  return {
    sanitized,
    valid: !isEmpty && !isStripped && sanitized !== current,
    preview:
      !isEmpty && !isStripped && sanitizedBase !== base ? sanitized : null,
    hint: isEmpty
      ? 'Enter a file name.'
      : isStripped
        ? "That name isn't valid. Try using letters, numbers or hyphens."
        : null,
  };
};
```

The modal then becomes two lines:

```ts
const { sanitized, valid, preview, hint } = previewRename(nextBase, ext, filename);
const canSubmit = !processing && valid;
```

That keeps the `'file'` sentinel private to the module that defines it, makes the validation unit-testable without a DOM, and gives the "name reverted to the original, so the button is dead with no hint" case an obvious home if we decide to close it.

packages/tinacms/src/toolkit/core/media-store.default.ts
```diff
@@ -674,22 +683,53 @@ export class TinaMediaStore implements MediaStore {
     if (!folder.endsWith('/')) {
       folder = folder + '/';
     }
+    return folder;
+  }
+
+  /** Stripped directory does not have leading or trailing slashes */
+  private stripDirectorySlashes(directory: string): string {
```
joshbermanssw, line 690: 1. is this supposed to be different to the trimming done on `mediaItemPath` ? 

2. should this be moved out to a util

packages/tinacms/src/toolkit/components/media/modal.tsx
```diff
@@ -20,6 +23,129 @@ interface NewFolderModalProps {
...
+  const { base, ext } = splitFilename(filename);
+  const [nextBase, setNextBase] = React.useState(base);
+  const [processing, setProcessing] = React.useState(false);
+  const [error, setError] = React.useState<string | null>(null);
+
+  const trimmedBase = nextBase.trim();
+  // Sanitize the base on its own and re-attach the original extension, so a
+  // base containing slashes or dots can never rewrite or drop it.
+  const sanitizedBase = sanitizeFilename(trimmedBase);
+  const sanitized = `${sanitizedBase}${ext}`;
```
joshbermanssw: Is there a world where we need the `ext` to be sanitized too?

packages/tinacms/src/toolkit/components/media/modal.tsx
18-th: Disagree with 1, "why" comments are useful in the codebase. Agree with 2, suggested code reads cleaner

packages/tinacms/src/toolkit/components/media/utils.ts
```diff
@@ -57,6 +57,21 @@ const MAX_BASENAME_LENGTH = 200;
  * Example: `image-a\u0308.jpg` becomes `image-ä.jpg`,
  * so URLs use `%C3%A4` instead of the decomposed `%CC%88` sequence.
  */
+/**
```
18-th, line 56: This doc got appended to the one above it, so now sanitizeFilename below doesn't have a doco

packages/tinacms/src/toolkit/tina-cms.ts
```diff
@@ -110,6 +110,13 @@ export class TinaCMS extends CMS {
         level: 'error',
         message: 'Failed to delete file.',
       }),
+      'media:rename:failure': (event: CMSEvent & { error?: Error }) => ({
```
18-th: The modal itself already has an error message. Do we need this toast on top of it?

packages/@tinacms/cli/src/next/commands/dev-command/server/media.ts
```diff
@@ -497,6 +562,88 @@ export class MediaModel {
...
+    }
+  }
+
+  /**
+   * A case-insensitive filesystem can treat `a.jpg` -> `A.jpg` as a no-op, so
+   * hop through a unique sibling name. On failure the source is put back.
+   */
+  private async renameViaStaging(source: string, destination: string) {
+    const staging = join(path.dirname(source), `.tina-rename-${randomUUID()}`);
+    await fs.move(source, staging, { overwrite: false });
```
18-th, line 653: If the second move fails, this file stays in the repo, which is not ideal. Could we rethink this logic?

packages/@tinacms/cli/src/next/commands/dev-command/server/media.ts
```diff
@@ -59,6 +60,48 @@ export const createMediaRouter = (config: PathConfig) => {
     }
   };
 
+  const handleRename = async (req: Connect.IncomingMessage, res) => {
+    // Body is populated by the bodyParser.json middleware registered ahead of
+    // this router, the same way the /graphql route consumes it.
+    const body = (req as unknown as { body?: unknown }).body;
+    const { from, to, ...rest } =
+      (body as { from?: unknown; to?: unknown }) || {};
+    void rest; // `branch` is accepted for parity with the cloud contract
```
18-th: `...rest` and `void rest` aren't doing anything — destructuring `from` and `to` ignores extra keys either way. I removed both lines and the "ignores an extra branch field" test still passed, so that test isn't guarding anything the happy-path test doesn't already cover. Suggest dropping all three.

playwright/tina-playwright/tests/api/security.spec.ts
```diff
@@ -249,3 +254,44 @@ for (const route of MEDIA_ROUTES) {
...
+  },
+  {
+    label: "still-encoded traversal",
+    from: "..%2f..%2fetc%2fpasswd",
+    to: "stolen.txt",
+  },
+  {
+    label: "media root as destination",
+    from: "rename-security-probe.txt",
+    to: "",
```
18-th: This vector never reaches path resolution. `to: ""` fails the `!to` check in handleRename and returns 400 before resolveStrictlyWithinBase runs, which is what forces the `to === "" ? [400] : [403]` special case. `to: "."` would actually exercise the media-root rejection and let the status assertion collapse back to a plain 403.

packages/tinacms/src/toolkit/components/media/media-manager.tsx
Aibono1225: 👍 Changed to `allowRename` to keep consistent with `allowDelete`

packages/tinacms/src/toolkit/components/media/modal.tsx
Aibono1225: ✅

packages/tinacms/src/toolkit/components/media/modal.tsx
Aibono1225: >1. Both comments explain why, which we'd rather keep in the PR body than inline.

I kept the why comments but moved them to utils.ts where the logic now lives.

> 2. This is a pure function of (nextBase, ext, filename) living inside a React component. To test it you have to render the modal, and it forces sanitizeFilename's internal 'file' fallback to leak out as a string literal on line 69.

✅ Adopted the proposed function

packages/tinacms/src/toolkit/components/media/modal.tsx
Aibono1225: Checked - this would only affect files that bypass the normal upload flow, a little overengineered here. I've reverted the change ✅

packages/tinacms/src/toolkit/components/media/utils.ts
Aibono1225: ✅ Done - Fixed in [1c42112d2](https://github.com/tinacms/tinacms/pull/7391/changes/1c42112d219dc41544b0040033ca674ab3d37113#diff-741d987c452f46c7d137ee16cd8e4ac2e314a5ecd1253a9e8704e14e63fb1177)

packages/tinacms/src/toolkit/tina-cms.ts
Aibono1225: ✅ Done - dropped in https://github.com/tinacms/tinacms/pull/7391/changes/6e26bb27747536eba47ebaf7a877954ab9384159

packages/@tinacms/cli/src/next/commands/dev-command/server/media.ts
Aibono1225: ✅ Done - removed in https://github.com/tinacms/tinacms/pull/7391/changes/9dddb17ad74f29f4338aad627ba86c8ade193d81#diff-632a49d9fecfbe0476ede508b52cb2ca9fea8dcfa53d7d8d82da14db3353651b

playwright/tina-playwright/tests/api/security.spec.ts
Aibono1225: ✅ Done - Good catch ,fixed in https://github.com/tinacms/tinacms/pull/7391/changes/cb3dcde6a8927c99feaa90cd0903bbb9e6128c17#diff-a42744471362c1ede993abcfec000b70c11313a93a07cdd8c66690f10cf73366

packages/@tinacms/cli/src/next/commands/dev-command/server/media.ts
Aibono1225: ✅ Done - Updated in https://github.com/tinacms/tinacms/pull/7391/changes/9dddb17ad74f29f4338aad627ba86c8ade193d81

Before: if restoring the source file failed, the error was ignored and the file was left under a random name.

After: return a specific error that includes the exact filename

packages/tinacms/src/toolkit/core/media-store.default.ts
Aibono1225: >1. is this supposed to be different to the trimming done on mediaItemPath ?

Yes. `stripDirectorySlashes` only trims one slash from each end, while `mediaItemPath` normalizes the whole path and also maps `.` to `''`.


>2. should this be moved out to a util

Better left here for this PR. This was extracted from the existing upload logic so rename could reuse it. Moving it or merging it into a utility could change paths, so that should be a separate change

18-th (dismissed):
LGTM

18-th (approved):
LGTM

PR #7392, opened 2026-08-03
feat: add tina-markdown web component

brookjeynes-ssw:
There's no way to render markdown content when not using a framework.
This adds a `tina-markdown` web component which can be passed tina's markdown AST.

```js
const postsResponse = await client.queries.postConnection();
const posts = postsResponse.data.postConnection.edges.map((post) => {
    return {
        title: post.node.title,
        body: post.node.body,
    };
});

const postsContainer = document.getElementById("posts");
for (const post of posts) {
    const tinaMarkdown = document.createElement("tina-markdown");
    const markdownAst = JSON.stringify(post.body);
    tinaMarkdown.setAttribute("content", markdownAst)

    postsContainer.appendChild(tinaMarkdown);
}
```

See [examples/js/markdown-rendering/index.html](https://github.com/tinacms/tinacms/blob/bj/20260803/feat/tina-markdown-web-component/examples/js/markdown-rendering/index.html) for a more detailed example


packages/@tinacms/web-components/src/tina-markdown.js
```diff
@@ -0,0 +1,127 @@
...
+
+function renderNode(node) {
+  if (node.type === 'text') return renderText(node);
+
+  const tag = TAGS[node.type];
+  const el = document.createElement(tag);
+
+  if (node.type === 'html' || node.type === 'html_inline') {
+    const htmlContainer = document.createElement('div');
+    htmlContainer.innerHTML = node.value;
```
kulesy: **Major:** this injects content-authored HTML unsanitized, which diverges from every other Tina renderer: React returns `child.value` as escaped text unless the user opts in with an explicit component ([rich-text/index.tsx:520-526](https://github.com/tinacms/tinacms/blob/6acd1b0412cbe2ec141c8fbe359dc0bcc2231d83/packages/tinacms/src/rich-text/index.tsx#L520-L526)) and Astro does the same ([Node.astro:69-74](https://github.com/tinacms/tinacms/blob/6acd1b0412cbe2ec141c8fbe359dc0bcc2231d83/packages/%40tinacms/astro/src/Node.astro#L69-L74)), so raw HTML nodes are inert by default everywhere else. Here `innerHTML` builds live DOM instead: running `{type: 'html', value: ''}` through the component in happy-dom produces `` in the shadow root with the handler attribute intact. `innerHTML` will not run `<script>`, but `onerror` and `onload` fire normally in a real browser, so anyone who can edit content (or land a content PR) gets script execution on every visitor of the site. Worth making this opt-in via an `allow-html` attribute with escaping as the default, or at minimum calling out the divergence loudly in the README.

packages/@tinacms/web-components/src/tina-markdown.js
```diff
@@ -0,0 +1,127 @@
...
+  h2: 'h2',
+  h3: 'h3',
+  h4: 'h4',
+  h5: 'h5',
+  h6: 'h6',
+  p: 'p',
+  ol: 'ol',
+  ul: 'ul',
+  li: 'li',
+  // lic: "div", // TODO: is this needed? seems to render fine without.
```
kulesy: **Major:** `lic` is not optional, it is on the core list path. The parser wraps every list item's content in one (`listItem` returns `li > lic > content` at [remarkToPlate.ts:98-109](https://github.com/tinacms/tinacms/blob/6acd1b0412cbe2ec141c8fbe359dc0bcc2231d83/packages/%40tinacms/mdx/src/parse/remarkToPlate.ts#L98-L109)), which is why the React renderer carries an explicit `lic` case rendering a `div` ([rich-text/index.tsx:287-299](https://github.com/tinacms/tinacms/blob/6acd1b0412cbe2ec141c8fbe359dc0bcc2231d83/packages/tinacms/src/rich-text/index.tsx#L287-L299)). Left unmapped, `TAGS[node.type]` is `undefined` and `document.createElement(undefined)` coerces to the literal tag name "undefined": feeding the parser's exact list shape through the component in happy-dom gives `<div><ul><li><undefined>item one</undefined></li></ul></div>`. Children still render so it looks correct on screen, which is presumably where "seems to render fine without" came from, but every list ships an invalid element that CSS and assistive tech cannot target, and the same silent failure hits any other unmapped node type. Mapping `lic` to `div` plus a fallback in `renderNode` for unknown types would cover both, and a list test would lock it in.

brookjeynes-ssw: resolved by 0f7e8d56c5b289cdc870aad856d7e093bb0d7adc

packages/@tinacms/web-components/src/tina-markdown.js
brookjeynes-ssw: I don't believe this is our responsibility? React and Astro don't sanitise either, it's just by coincidence with the way the content is returned. TinaCMS expects a level of trust from it's authors as projects are invite-based.

Since we are doing raw DOM manipulations, there's no way to return "escaped html". Everything needs to be an element at some point. The only way to "escape" the HTML is by returning it as a text node but that just renders it as if it were text.

wicksipedia: Agreed that raw HTML should render, but I think the renderers should line up where it counts... see #7424.

`innerHTML` renders the markup and also runs `onerror`/`onload`, which isn't a great OOTB experience for people.

Neither the React nor Astro suite covered html nodes, so after adding tests for them first, it confirmed that they escape. So #7424 doesn't match them on that.

I did add a components map so a site developer can supply their own rendering logic in case they want to bypass DOMPurify.

URLs on a and img nodes stay as authored, so we're deliberately more permissive than React and Astro there. Worth knowing that DOMPurify does drop dangerous schemes inside raw HTML, so a link written as markdown and the same link written as raw HTML come out differently.

wicksipedia (dismissed):
aprroved via the pr off this that was merged

wicksipedia (approved):
aprroved via the pr off this that was merged

PR #7394, opened 2026-08-03
🔒 Escape the listing directory in the Cloudinary search expression

joshbermanssw:
## TL;DR

`listMedia` interpolated the caller-supplied `directory` straight into the Cloudinary Search API expression (`folder="<directory>"`). Cloudinary treats a quoted term literally *except* for `"` and `*`, so those two characters were parsed as expression syntax instead of being matched as part of the folder name — this routes the value through an escaper first.

## Why

Completes **SEC-5**, the last of the Cloudinary follow-ups listed under "Not in this PR" in #7088.

`resolveDirectory()` already runs on this path, but it only validates traversal and its normalised result is [deliberately discarded](https://github.com/tinacms/tinacms/blob/main/packages/next-tinacms-cloudinary/src/handlers.ts#L142) (Cloudinary derives `public_id` from folder + `use_filename`, so applying the resolved prefix would change the naming model). Nothing was escaping the value on the way into the query.

Per Cloudinary's [search expression syntax](https://cloudinary.com/documentation/search_expressions), a quoted term needs `"` and `*` backslash-escaped. The escaper handles the backslash itself in the same pass so it cannot consume the escape added for a following character.

## Tests

`tests/cloudinary-search-expression.test.ts` — 8 vectors: quote, asterisk, backslash, backslash-before-quote, repeated occurrences, empty string, and the composed `folder="…"` term.

**Not exercised against a live Cloudinary account.** The escaping follows the documented syntax and is pinned by unit test only.

## Follow-up

Still open from the same group: a real `mediaRoot` boundary for the Azure and Cloudinary adapters (SEC-2), and the `mediaRoot/*`-scoped credential policy docs (SEC-4). Both stay tracked on tinacms/tinacloud#3761, so this PR closes SEC-5 only — don't close the tracker on merge.

## Links

- PBI: tinacms/tinacloud#3761
- Follow-up from #7088


packages/next-tinacms-cloudinary/src/handlers.ts
```diff
@@ -151,7 +152,7 @@ async function listMedia(
 
     const query = useRootDirectory
       ? 'folder=""'
-      : `folder="${mediaListOptions.directory}"`;
+      : `folder="${escapeSearchValue(mediaListOptions.directory)}"`;
```
kulesy, line 159: `req.query.directory` is typed `string | string[] | undefined`, and the `as string` up at line 124 doesn't hold for a repeated param: `?directory=a&directory=b` arrives as an array, `resolveDirectory` returns `''` without throwing (its `typeof !== 'string'` guard), and then `escapeSearchValue` hits `value.replace is not a function`, so the outer catch turns it into a 500 where the old template literal produced `folder="a,b"` and a harmless empty 200. Not reachable from the media manager UI, but it is a behaviour change this PR introduces, and normalising once at line 124 (`Array.isArray(d) ? d[0] : d`) would make the cast honest rather than leaving the new `string` parameter fed by an unsound assertion.

packages/next-tinacms-cloudinary/src/handlers.ts
```diff
@@ -135,10 +136,10 @@ async function listMedia(
       try {
         // Validation only: reject upward traversal in the listing directory for
         // consistency with the other adapters. The normalised result is
-        // intentionally discarded; the raw directory is still interpolated into
-        // the search expression below, so this does NOT bound the listing the
-        // way resolveDirectory bounds the S3/DOS prefix. Search-expression
-        // escaping (SEC-5) is a separate follow-up.
+        // intentionally discarded; Cloudinary derives public_id from folder +
+        // use_filename, so applying the resolved prefix here would change the
+        // naming model. Bounding the listing to a mediaRoot is tracked
```
kulesy: The `public_id` rationale reads like it migrated over from the upload path: `listMedia` never derives a `public_id`, so that isn't actually what stops you from using the resolved value here. The real blocker is that `resolveDirectory` returns a trailing-slash prefix (`normalized + '/'`), which wouldn't match a Cloudinary `folder=` term. I'd also keep the old comment's sharpest sentence, that this does NOT bound the listing the way `resolveDirectory` bounds the S3/DOS prefix, since that's the bit a future reader most needs to know and "Bounding the listing to a mediaRoot is tracked separately" loses the point that it currently isn't bounded at all.

packages/next-tinacms-cloudinary/src/search-expression.ts
```diff
@@ -0,0 +1,13 @@
+/**
+ * Escape a value for interpolation into a quoted Cloudinary Search API term.
+ *
+ * Cloudinary treats a double-quoted term literally except for `"` and `*`,
+ * which keep their meaning and must be backslash-escaped. The backslash itself
+ * is escaped first (in the same pass) so it cannot consume the escape added
+ * for a following character.
```
kulesy: Live probing suggests this particular mechanism doesn't exist: `folder="demo\"` errors past end-of-string, which confirms `\"` is consumed as an escaped quote (and is exactly why the fix works), but `folder="demo\\"` errors the same way instead of parsing as a folder literally named `demo\`, so Cloudinary doesn't collapse `\\` into a literal backslash and escaping the backslash can't be "consuming the escape added for a following character". No behaviour impact either way, since `resolveDirectory` rejects any directory containing a backslash before this ever runs, so the branch is unreachable and I'd keep it as a conservative hedge in case that guard moves. Just worth rewording so the comment doesn't assert a parser behaviour the API doesn't appear to have.

kulesy:
Moved my two verification comments off this PR and onto the private SEC-5 tracker (tinacms/tinacloud#3761), per SECURITY.md, since the fix here hasn't shipped yet. Short version for anyone following along: I exercised this against a live Cloudinary account and it behaves as intended, both for the case the PR targets and for ordinary folder names, so the "not exercised against a live Cloudinary account" line in the description can come out whenever you next touch it. The three inline threads are addressed in 54d3b10da, which also narrows a query-param edge case that was surfacing as a 500 and corrects a couple of comments.

kulesy (approved):

PR #7395, opened 2026-08-03
feat(v4): local data layer, defineConfig, CLI and admin UI

JackDevAU:
tl;dr: the whole v4 line except rich-text, consolidated from the old stacked PRs into one commit (their branches keep the full history).

What's in it:

- **Local content plugin:** fs provider, markdown/JSON adapters, and the hosted GraphQL pipeline
- **defineConfig:** one composition root, compiled to `tina-lock.json` with a contract-version gate
- **CLI:** `tinacms codegen` and `tinacms init`, kept out of the build pipeline, no fixed ports
- **Admin UI:** the shell driven by the compiled schema, plugin-contributed screens, content reads through React Query
- **Preview:** the connection as a store, focus marks the active field, resizable form column
- **Field plugins:** string, boolean, number, datetime, each declaring its own equality
- **Barebones example:** runs the output of `tinacms init`

To test: `pnpm install`, then `pnpm types && pnpm test` in `packages/v4/@tinacms/tinacms`, or `pnpm dev` to poke the playground.


18-th (approved):
LGTM

PR #7396, opened 2026-08-03
feat(v4): rich-text renderer, Plate editor and field plugin

JackDevAU:
tl;dr: the rich-text half of v4, stacked on #7395, and the combined tree matches the tip of the original stack.

What's in it:

- **The renderer:** framework-free `src/rich-text`, with `TinaMarkdown` as its React binding
- **`@tinacms/rich-text`:** the Plate editor extracted into its own v4 package
- **The field plugin:** `t.richText`, per-document markdown/MDX codecs, and the field UI
- **Restores the body field** in the playground and barebones example, plus the Tailwind `@source` lines

To test: `pnpm types && pnpm test` in `packages/v4/@tinacms/tinacms`, or `pnpm dev` and edit a post body in the playground.

### DIFF
--- a/packages/v4/@tinacms/rich-text/package.json
+++ b/packages/v4/@tinacms/rich-text/package.json
@@ -1,0 +1,85 @@
+{
+  "name": "@tinacms/rich-text",
+  "version": "4.0.0-alpha.0",
+  "private": true,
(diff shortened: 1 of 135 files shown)


18-th (approved):
:shipit:

PR #7401, opened 2026-08-04
fix(admin): prevent long filenames from overflowing the collection table

### LINKED ISSUE #7399: Admin collection view pushes table & pagination buttons out of viewport when title & filename are too long
### The exact error message

```shell
I have markdown pages migrated from a wp site so they have long titles and filenames.
in the admin -> collections view for the pages, the table has this structure:
Title column:
'Welcome to the site'
'src/content/news/a-really-long-legacy-slug-path-with-no-spaces.mdx'

Then the Filename column has
'a-really-long-legacy-slug-path-with-no-spaces.mdx'

This stretches the table out of the right of the viewport, and even the horizontal table scrollbar doesn't scroll enough. But it also causes the pagination buttons underneath to be offscreen and unreachable.
```

### Steps to reproduce

1. visit `/admin/index.html#/collections/news/~`
2. Note that for my site a filename will be more than 60 characters long without spaces.
3. See the filename duplicated into the Filename column
4. Will not be able to see a next page button as the page is too wide but the page itself doesn't scroll horizontally (the table does scroll horizontally)

Expected the filepath in the 'title' column and filename in the 'filename' column to wrap at certain length of characters even though there's no space.
Actual: got a table too wide and pagination buttons offscreen

### Your environment

- "tinacms": "^3.11.0",
- "@tinacms/cli": "^2.5.6",
- @tinacms/astro": "^0.6.1",
- "@astrojs/vercel": "^10.0.8",
- "@astrojs/node": "^10.1.4",

Hosted on vercel, connected to tinacloud
Google Chrome for windows. No extensions.
Accessed through myapp.vercel.app/admin/index.html#/collections/news/~

### A way for us to reproduce

_No response_

```typescript

```

### Client ID

_No response_

### Validations

waterWang:
## Description

When documents have long filenames without spaces (e.g., legacy imported slugs like `a-really-long-legacy-slug-path-with-no-spaces.mdx`), the collection table columns expand beyond the viewport. This pushes the pagination buttons off-screen, making them unreachable.

## Fix

Added `break-all` (Tailwind's `word-break: break-all`) to:
1. The folder path cell (showing path segments)
2. The title/filename cell (showing document title or filename)
3. The filename cell (shown when a document has a title)

This ensures long unbroken strings wrap at character boundaries, keeping the table within its scrollable container.

## Before

The table expands beyond the viewport, and pagination buttons become unreachable.

## After

Long filenames wrap at character boundaries, and the table stays within the viewport.

Fixes #7399

### DIFF
--- a/packages/tinacms/src/admin/pages/CollectionListPage.tsx
+++ b/packages/tinacms/src/admin/pages/CollectionListPage.tsx
@@ -937,7 +937,7 @@ const CollectionListPage = () => {
                                                   </a>
                                                 </td>
                                                 <td
-                                                  className='px-3 py-3'
+                                                  className='pl-5 pr-3 py-3 break-all'
                                                   colSpan={4}
                                                 >
                                                   <span className='leading-5 block text-sm font-medium text-gray-400 truncate'>
@@ -976,7 +976,7 @@ const CollectionListPage = () => {
                                               key={`document-${document.node._sys.relativePath}`}
                                             >
                                               <td
-                                                className='pl-5 pr-3 py-3'
+                                                className='pl-5 pr-3 py-3 break-all'
                                                 colSpan={hasTitle ? 1 : 2}
                                               >
                                                 <a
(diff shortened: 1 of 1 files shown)


joshbermanssw:
Hi @waterWang  - could you please include a screenshot of the before/ after :)

PR #7403, opened 2026-08-04
fix(mdx): keep carriage returns out of parsed text, and keep task-list state

wicksipedia:
Two parser fixes in `@tinacms/mdx`, found while reviewing the v4 rich-text work. Both are v3, both ship on the v3 release train, and neither depends on v4 — so they are split out here rather than sitting behind the v4 stack.

### Carriage returns

A CRLF pair survives micromark into the `value` of a text node, so a document authored on Windows carries a `\r` into the editor. `parseMDX` now normalizes the source to line feeds before parsing, and passes the normalized source to `remarkToSlate`.

### GFM task lists

A ticked checkbox came back unticked. `parseMDX` now reads `checked` onto the `li` node, and both stringifiers (`stringify/index.ts` and `next/stringify/pre-processing.ts`) write it back. `ListItemElement` gains an optional `checked?: boolean`, set only on task-list items — additive, so no consumer breaks.

### Notes

- Changeset included: patch for `@tinacms/mdx`.
- `packages/@tinacms/mdx`: **77 test files, 119 tests passing.**
- The same two source changes also appear in #7405 (the v4 review-findings branch), because v4 tests depend on them. The files and the changeset are byte-identical, so whichever lands first, the other rebases away cleanly.

### DIFF
--- a/.changeset/mdx-line-endings-and-task-lists.md
+++ b/.changeset/mdx-line-endings-and-task-lists.md
@@ -1,0 +1,7 @@
+---
+"@tinacms/mdx": patch
+---
+
+`parseMDX` now normalizes CRLF to LF before parsing. A carriage return used to survive micromark into the value of a text node, so a document authored on Windows carried `\r` into the editor.
+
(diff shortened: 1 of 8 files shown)


18-th (approved):
LGTM

PR #7404, opened 2026-08-04
ci: move to Biome 2 and enforce two AGENTS.md rules as lint plugins

wicksipedia:
The `AGENTS.md` files give two coding rules. Biome 1.9.4 has no rule that checks them. As a result, code on `main` breaks the error-narrowing rule in seven places. This PR moves the repo to Biome 2.5. It adds the two rules as Biome plugins. It also corrects the seven places.

### The rules

- **Narrow an error in an `if`, not in an expression** (root `AGENTS.md`). The plugin reports `instanceof` on a class with a name that ends in `Error`. The plugin makes an exception for the condition of an `if` statement or a `while` statement. A ternary, an `&&` chain, or a ternary in an `if` body all fail. `if (ok && cause instanceof Error)` passes, because the narrowing stays in the condition.
- **Do not use `&&` for conditional rendering** (`packages/v4/AGENTS.md`). `{value && <X />}` renders the falsy value. React renders `0` as the text `0`. An empty string leaves an empty text node. The plugin reports an `&&` operator when its right operand is a JSX element or a fragment. This applies in `.tsx` files, with or without parentheses.

### Why plugins, and not the built-in rule

Biome 2.3.8 added the rule `suspicious/noLeakedRender`. That rule accepts `{count > 0 && <X />}` and `{!!items.length && <X />}`. `packages/v4/AGENTS.md` forbids both forms. So the built-in rule is weaker than the written rule. Biome has no rule for error narrowing.

Biome documents GritQL plugins as still under development. The documentation says: "Many things already work, but bugs are still expected and some features are still outright missing". It also says that syntax node names such as `JsIfStatement` "may change between versions" (https://biomejs.dev/reference/gritql/).

Both plugins use node names. A later Biome bump can need a change to them. If a plugin does not compile, `pnpm lint` fails to load it. `pnpm lint` does not skip the plugin.

### What changed

- `pnpm-workspace.yaml` moves `@biomejs/biome` to `^2.5.14`. `biome migrate` rewrote the six `biome.json` files. The nested example configurations extend the root with `"extends": "//"`. With a relative path, Biome resolves the plugin paths from the nested directory and does not load them.
- The option `css.parser.tailwindDirectives` is on. Biome then parses the Tailwind 4 stylesheets. The Hugo configuration excludes the `layouts/` directory. Biome 2 parses `.html` files, and those files are Go templates.
- The Biome 2 formatter changes 26 files. Ten `useHookAtTopLevel` suppressions in `GetCollection.tsx` and one in `alerts.test.ts` suppress nothing under Biome 2. This PR removes them.
- `scripts/biome/error-narrowing.grit` and `scripts/biome/conditional-jsx.grit` hold the two rules. `biome.json` limits both to `packages/v4`. It skips the shadcn and Plate registry directories, because their update commands overwrite local edits. A later positive glob keeps the hand-written `field-wrapper.tsx` in scope.
- The seven corrections are in five `@tinacms/tinacms` files. Six were already in this PR. The seventh is in `editor/preview-connection.ts`, and it arrived on `main` with the reference field. The corrections land before the plugins, so every commit in this PR passes `pnpm lint`.
- This PR removes `scripts/check-conventions.mjs` and `tests/check-conventions.test.ts`. `pnpm lint` runs `biome lint` again.

### Verification

- `pnpm lint` reports `Checked 2138 files` and no diagnostics.
- `pnpm exec biome format ./packages/ ./examples/` reports `Checked 1857 files. No fixes applied.`
- The plugins ran against a copy of `packages/v4` from `main`, with the old `FieldWrapper`. They report the seven narrowing lines and the one `label && <Label />` line. They report nothing in the registry directories.
- `pnpm build` passes 25 of 25 tasks. `pnpm test` passes 66 of 66 tasks.

This PR has no changeset. Inside the published packages, the only changes are formatter output and removed suppression comments. `@tinacms/tinacms` under `packages/v4` is private.

### DIFF
--- a/package.json
+++ b/package.json
@@ -30,7 +30,7 @@
 		"publish:beta": "pnpm publish -r --tag beta --no-git-checks --access public",
 		"publish:dev": "pnpm publish -r --tag dev --no-git-checks --access public",
 		"push-tags": "pnpm exec changeset tag && git pull && git push --follow-tags",
-		"lint": "biome lint ./packages ./examples",
+		"lint": "biome lint ./packages ./examples && node ./scripts/check-conventions.mjs",
 		"format": "biome format --write ./packages/ ./examples/"
 	},
 	"devDependencies": {
--- a/packages/v4/@tinacms/tinacms/src/admin/document-form.tsx
+++ b/packages/v4/@tinacms/tinacms/src/admin/document-form.tsx
@@ -48,11 +48,11 @@ function SaveButton() {
           setFailure(null);
           save().catch((cause) => {
             console.error('[tinacms] Save failed:', cause);
-            setFailure(
-              cause instanceof Error && cause.message
-                ? cause.message
-                : 'Save failed.'
-            );
+            if (cause instanceof Error && cause.message) {
+              setFailure(cause.message);
+            } else {
+              setFailure('Save failed.');
+            }
           });
         }}
       >
--- a/packages/v4/@tinacms/tinacms/src/cli/index.ts
+++ b/packages/v4/@tinacms/tinacms/src/cli/index.ts
@@ -121,7 +121,11 @@ export const runCli = async (
     }
     return await codegenCommand(values, { ...context, cwd, log, logError });
   } catch (cause) {
-    logError(cause instanceof Error ? cause.message : String(cause));
+    if (cause instanceof Error) {
+      logError(cause.message);
+    } else {
+      logError(String(cause));
+    }
     return 1;
   }
 };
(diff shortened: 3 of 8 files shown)


wicksipedia:
Moved the v4 Playwright workflow out of this PR and into #7405. The e2e suite cannot pass on `main`: `visual-editing.spec.ts` needs the rich-text body field, and `@tinacms/rich-text` is still in #7396, while the five `form-column-resize` specs need the column-width fix that lives in #7405. Verified on #7405: 11 passed.

PR #7405, opened 2026-08-04
fix(v4): review findings on the v4 core and rich-text stack

wicksipedia:
Stacked on #7396. 47 commits from a review pass over the v4 core and rich-text work, each one its own fix with tests.

Two sibling PRs carry the parts that do not belong on this stack:
- **#7403** — the two `@tinacms/mdx` (v3) parser fixes, which ship on the v3 release train
- **#7404** — the convention checker and the v4 Playwright workflow, which merge to `main` independently

### What is in here

| Area | n | Examples |
|---|---|---|
| `fix` | 22 | document path through a link with no target; bound the Express request body; form reseed and save races; datetime instant; codec round-trips; accessible names |
| `docs` | 6 | datetime field spec, field-plugin manifest shape, codec limits, accessibility rules |
| `test` | 7 | DOM harness, boundary enforcement, coverage gaps in both packages |
| `refactor` | 3 | drop dead files, keep the editor off the storage-format package, ternary rendering |
| `ci` / `chore` / `perf` / `style` / `feat` | 9 | coverage gate, biome no-any, mermaid on demand, list without bodies, task-list checkbox |

### Notable

- **`list` no longer carries document content.** Opening one document in a 50-document collection went from 84,206 B to 3,114 B.
- **A block in a list item destroyed the line.** Code Block / Horizontal Rule with the caret in a list item emptied the item's text and inserted nothing — Plate's list normaliser rejects the shape. All three entry points now end the list first.
- **Toolbar buttons had no accessible name.** 27 of 30 were unnamed or announced their own SVG slug (`code-block`, `format bold`). The tooltip text now names the control; the three that show text keep it inside the name (WCAG 2.5.3).
- **Coverage is gated.** `@tinacms/tinacms` 97.5%, `@tinacms/rich-text` 61.1%, both above their thresholds, and the gate is proven to fail when coverage drops.

### Verification

- `@tinacms/tinacms`: **709 tests passing**
- `@tinacms/rich-text`: **181 tests passing**
- `pnpm lint` clean, `pnpm format` leaves the tree clean, `pnpm types` clean under `packages/v4`
- Each fix was mutation-proved: break the guard, watch the *named* test go red, restore, verify the restore by content hash

### Note on overlap with #7403

This branch carries the same two `@tinacms/mdx` source commits and the same changeset as #7403, because the v4 rich-text tests depend on them. The files are byte-identical, so whichever lands first the other rebases away cleanly.


kulesy (dismissed):

18-th (dismissed):
:shipit:

18-th (approved):
:shipit:

PR #7412, opened 2026-08-05
🐛 Support soft returns in the standard markdown editor

### LINKED ISSUE #7408: ✨ Soft Returns - Support soft return behavior in standard markdown editor
Requested by: @isaaclombardssw via [YakShaver.ai](https://YakShaver.ai) 🦬
cc: @isaaclombardssw, @adamcogan, @kulesy, @18-th

Hi Team!

&nbsp;&nbsp;[🟥&nbsp;&nbsp;Watch the video (1 min 30 sec)](https://www.youtube.com/watch?v=lYtgwzaROHQ)

&nbsp;&nbsp;Url from screen share: [https://www.ssw.com.au/rules/admin/#/rules/reply-done](https://www.ssw.com.au/rules/admin/#/rules/reply-done)

See https://tina.io/docs/reference/types/rich-text#line-breaks

### Pain
Investigate and resolve inconsistent soft return behavior in ⭐ TinaCMS markdown editing. According to the TinaCMS documentation, soft returns should be supported, however they currently only work inside quote blocks and not within the standard non-nested markdown editor. Determine whether this is a documentation issue or missing functionality in the editor implementation. If feasible, implement soft return support in the main markdown editor so line breaks can be inserted without creating new paragraphs.

### Acceptance Criteria
1. The behavior of soft returns in the standard markdown editor is investigated and documented.
2. A decision is made whether to update documentation or implement missing functionality.
3. If implemented, soft returns work consistently in the standard markdown editor.
4. Soft return behavior matches documented TinaCMS functionality.
5. Existing quote block soft return behavior continues to function correctly.
6. Any limitations or unsupported scenarios are documented.

### Reproduce Steps
1. Open the ⭐ TinaCMS rich text editor using the standard markdown field.
2. Enter multiline text in the main markdown section.
3. Attempt to create soft returns within the text editor.
4. Observe that the generated markdown does not reflect soft returns.
5. Create a quote block and attempt the same action.
6. Observe that soft returns work correctly inside quote blocks.

### Screenshot
Screenshot
**Figure: Soft returns not working in standard markdown editor**

isaaclombardssw:
Closes #7408

## Investigation

Soft returns worked in blockquotes and nowhere else. Two different code paths were responsible:

- **Blockquotes** — `createBlockquoteEnterBreakPlugin` intercepted Enter and inserted a `break` element, which `@tinacms/mdx` serializes to a `\` hard break.
- **Everywhere else** — `shift+Enter` fell through to Plate's `SoftBreakPlugin`, which does `editor.tf.insertText('\n')`. Markdown treats a bare newline inside a paragraph as a soft wrap, so the break was re-flowed into a space on save.

Confirmed against the serializer:

| Editor node | Serialized markdown |
|---|---|
| text `"line one\nline two"` | `line one⏎line two` — soft wrap, renders as one line |
| `break` element | `line one\`⏎`line two` — hard break, renders as `<br>` |

So it's a functionality gap, not a docs problem. `@tinacms/mdx` already parses and stringifies `break` in plain paragraphs (`markdown-basic-break` fixture) — only the editor never produced one.

## Fix

Generalised the blockquote plugin into `createSoftBreakPlugin`: `shift+Enter` inserts a `break` element everywhere, and plain Enter inside a blockquote keeps doing what it already did. `SoftBreakPlugin`'s config is untouched — it only ever runs now where the new plugin bails.

Also stopped the handler swallowing `mod+Enter` / `mod+shift+Enter` inside blockquotes. It preventDefault'd on any Enter, and `ExitBreakPlugin` skips events that are already handled, so cmd+Enter couldn't exit a quote. Required anyway, since `mod+shift+Enter` would otherwise hit the new shift branch.

## Limitations

- **Code blocks** keep a real `\n` via `SoftBreakPlugin` — a hard break there would be wrong.
- **Table cells** are skipped. GFM cells can't hold a hard break: a `break` element serializes away to a space, and the current `\n` at least survives as `&#xA;`. A literal `<br>` in the cell is the workaround. Unchanged from today's behaviour either way.

## Tests

`create-html-block/index.test.ts` drives the handler headlessly over six cases. Verified the two new assertions fail against `main`:

```
× inserts a break element on shift+Enter in a paragraph
× leaves mod+Enter to ExitBreakPlugin inside a blockquote
```

The other four (plain Enter in a paragraph, plain Enter in a blockquote, code blocks, table cells) pass before and after — they're there to pin the behaviour that shouldn't move.

Note: `src/rich-text/index.test.tsx` fails on this branch and on a clean `main` alike — it needs `@tinacms/mdx` built first. Unrelated.

https://claude.ai/code/session_01XQ3RmyQZZeuaQFqBXPQm5m

### DIFF
--- a/.changeset/soft-returns-standard-markdown.md
+++ b/.changeset/soft-returns-standard-markdown.md
@@ -1,0 +1,19 @@
+---
+"tinacms": patch
+---
+
+Fix soft returns (`shift+Enter`) being dropped outside blockquotes in the rich-text editor.
+
+Plate's `SoftBreakPlugin` inserts a literal `\n` into the text node. In markdown a bare newline inside a paragraph is a soft wrap, so serialization re-flowed it into a space and the line break vanished on save. Blockquotes were unaffected because a separate plugin already inserted a `break` element there — hence the inconsistency.
+
+`shift+Enter` now inserts the same `break` element everywhere, which serializes to a `\` hard break and round-trips as a `<br>`:
+
+```md
+123 Abc Street\
+Town Central, CA\
+90210
+```
+
+Two places keep the old behaviour: code blocks (which want a real newline) and GFM table cells (which cannot represent a hard break — put a literal `<br>` in the cell instead).
+
+`mod+Enter` / `mod+shift+enter` inside a blockquote also work again; the blockquote handler used to swallow them before `ExitBreakPlugin` could exit the quote.
--- a/packages/tinacms/src/toolkit/fields/plugins/mdx-field-plugin/plate/plugins/create-html-block/index.test.ts
+++ b/packages/tinacms/src/toolkit/fields/plugins/mdx-field-plugin/plate/plugins/create-html-block/index.test.ts
@@ -1,0 +1,134 @@
+import { BlockquotePlugin } from '@udecode/plate-block-quote/react';
+import { CodeBlockPlugin } from '@udecode/plate-code-block/react';
+import { TablePlugin } from '@udecode/plate-table/react';
+import { createPlateEditor } from '@udecode/plate/react';
+import { describe, expect, it } from 'vitest';
+import {
+  ELEMENT_BREAK,
+  createBreakPlugin,
+  createSoftBreakPlugin,
+} from './index';
+
+const pressEnter = (
+  value: any[],
+  path: number[],
+  modifiers: { shiftKey?: boolean; metaKey?: boolean } = {}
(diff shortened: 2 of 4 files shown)


isaaclombardssw:
Closing — superseded by #7452, which landed the shift+Enter hard-break fix for #7408 (along with the related work in #7448–#7451).

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #7418, opened 2026-08-06
chore(v4): fix import ordering in local-data-layer test

### LINKED ISSUE #7416: chore: biome check fails on local-data-layer.test.ts; CI lint never runs organizeImports
### What's happening

`biome check` fails on a v4 test file with an import-ordering error. From the repo root:

```
node_modules/.bin/biome check packages/v4/@tinacms/tinacms/src/plugins/content/local/server/local-data-layer.test.ts
```

```
× Import statements could be sorted:

    5 │ - import·{·dispatchContentRequest·}·from·'./content-request';
    6 │ - import·{·createGraphQLPipeline·}·from·'../graphql/graphql-pipeline';
    5 │ + import·{·createGraphQLPipeline·}·from·'../graphql/graphql-pipeline';
    6 │ + import·{·dispatchContentRequest·}·from·'./content-request';
```

### Why CI is green anyway

The lint job's **path scope is fine** — `biome lint ./packages` walks 1876 files and does include `packages/v4/**`. The gap is the **command**:

| CI step | Runs | Covers `organizeImports`? |
|---|---|---|
| `pnpm lint` → `biome lint ./packages ./examples` | lint rules | ❌ |
| `pnpm format` → `biome format --write` + `git diff --exit-code` | formatting | ❌ |

`organizeImports` only runs under `biome check` (= lint + format + organizeImports), and there is no `check` script in `package.json`. So `main` is genuinely green and always has been — this isn't a regression, and it isn't v4-specific.

### Size of the underlying backlog

Repo-wide, `biome check ./packages ./examples` reports **478 errors — all 478 are `organizeImports`**. Zero lint errors, zero format errors:

| Package | Violations |
|---|---|
| `packages/tinacms/src` | 147 |
| `packages/@tinacms/graphql` | 62 |
| `packages/@tinacms/mdx` | 57 |
| `packages/@tinacms/cli` | 35 |
| `packages/@tinacms/search` | 9 |
| `packages/@tinacms/schema-tools` | 7 |
| `packages/v4/@tinacms` | 2 |
| others (≤5 each) | ~159 |

### Scope of this issue

Fix the single file that fails today, so anyone running Biome locally (or via an editor integration that calls `check` rather than `lint`) stops getting a spurious error on a file they didn't touch.

Sweeping all 478 and switching the CI step to `biome check` is a separate call — it wants one mechanical `biome check --write` commit across the repo followed by the CI change, and should be its own issue if we want to go there.

kulesy:
Closes #7416

**TL;DR** Reorder two imports in `local-data-layer.test.ts` so the file passes `biome check`.

**Pain:** the file trips `organizeImports` under `biome check`, so anyone running Biome locally — or through an editor integration that calls `check` rather than `lint` — gets a spurious error on a file they didn't touch. CI never caught it because `pnpm lint` and `pnpm format` don't run `organizeImports` at all, so this sat on `main` unnoticed and trains people to ignore Biome output.

**Solution:** swaps `../graphql/graphql-pipeline` above `./content-request`, exactly what `biome check --write` produces. Behaviour is unchanged: `graphql-pipeline` is `vi.mock`'d (hoisted above all imports) and `content-request` imports it only as a type, so there's no runtime edge between the two.

### DIFF
--- a/packages/v4/@tinacms/tinacms/src/plugins/content/local/server/local-data-layer.test.ts
+++ b/packages/v4/@tinacms/tinacms/src/plugins/content/local/server/local-data-layer.test.ts
@@ -2,8 +2,8 @@ import { promises as fs } from 'node:fs';
 import { tmpdir } from 'node:os';
 import path from 'node:path';
 import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
-import { dispatchContentRequest } from './content-request';
 import { createGraphQLPipeline } from '../graphql/graphql-pipeline';
+import { dispatchContentRequest } from './content-request';
 import { type LocalDataLayer, createLocalDataLayer } from './local-data-layer';
 
 vi.mock('../graphql/graphql-pipeline', () => ({


kulesy:
### Verification

- `biome check` on the file — **clean** (was 1 error)
- `pnpm lint` — clean, 1876 files
- `pnpm format` then `git diff` — no diff beyond the one line in this PR
- `vitest run local-data-layer.test.ts` — **37/37 pass**

The test run matters more than the lint result here: reordering imports changes module evaluation order, so it's worth confirming it's inert rather than assuming. It is, for two independent reasons — `vi.mock('../graphql/graphql-pipeline')` is hoisted above all imports by Vitest, and `content-request.ts` pulls in `graphql-pipeline` only via `import type { GraphQLResult }`, which is erased at compile time.

### Follow-up, deliberately not in this PR

Flipping the CI step from `biome lint` to `biome check` would fail immediately on **478 pre-existing `organizeImports` violations** across the repo (breakdown in #7416). They're all mechanically fixable — `biome check --write`, no `--unsafe` needed — but that's one large sweep commit plus a CI change, and it shouldn't ride along on a one-line fix. Worth its own issue if we want to close the gap properly.

The two remaining v4 ones, if someone wants v4 clean on its own: `src/plugins/fields/datetime/datetime-field.test.tsx` and `playground/tina/config.ts`.

kulesy:
Done in https://github.com/tinacms/tinacms/pull/7405

PR #7154, opened 2026-07-06
feat(v4): @tinacms/ui shadcn primitives package

JackDevAU:
> **Stack 5/5** (top) · prev: #7153
> `feat/v4-form-store` → `feat/v4-playground` → `feat/v4-preview-adapter` → `feat/v4-form-continuity` → `feat/v4-shadcn-ui`

## TL;DR

New `@tinacms/ui` workspace package with shadcn-generated primitives; the string and boolean field UIs adopt them, replacing raw markup.

## Feats

- `@tinacms/ui` package: Button, Checkbox, Input, Label (shadcn CLI output) plus a `FieldWrapper` that owns error rendering, with Tailwind 4 CSS-first globals.
- String/boolean field UIs swap their raw inputs and `TODO(shadcn)` comments for the shared primitives.
- Playground gains the `@tailwindcss/vite` plugin and an `index.css` that `@source`-scans the package sources Tailwind's root-scan misses.

## Notes

The `pnpm-lock.yaml` entry for the new package couldn't ride this lane (lockfile is dependency-locked to another in-flight branch) — frozen-lockfile CI may fail on this PR until that lands.

## Testing

Full v4 package suite passes against the new primitives (121 tests, 16 files), including the string/boolean field suites.

## Status

WIP top of stack — only the two existing fields are converted; remaining field types adopt `@tinacms/ui` as they land.

PR #7419, opened 2026-08-06
test(@tinacms/tinacms): warm plugin client segments in v4 provider tests

### LINKED ISSUE #7417: Flaky: v4 tests that render <TinaProvider> race a lazy plugin import on loaded CI runners
`packages/v4/@tinacms/tinacms/src/plugins/fields/string/string-field.test.tsx` > "StringField rendering" > "renders the ingested value into a text input" fails intermittently on the `build (macOS-latest)` job. The same pattern exists in every v4 test that renders `<TinaProvider>`, so the flake is not isolated to this one case.

### The exact error message

```
TestingLibraryElementError: Unable to find a label with the text of: title

Ignored nodes: comments, script, style
<body>
  <div />
</body>
 ❯ waitForWrapper node_modules/@testing-library/dom/dist/wait-for.js:163:27
 ❯ findByLabelText node_modules/@testing-library/dom/dist/query-helpers.js:86:33
 ❯ src/plugins/fields/string/string-field.test.tsx:54:33
```

### Steps to reproduce

1. Open any PR that touches nothing in `packages/v4/`.
2. Wait for the `build (macOS-latest)` job.
3. The job intermittently fails on this single test (1 failed / 475 passed) while `build (ubuntu-latest)` passes on the identical tree.
4. Re-run the failed job with no code change — it passes.

Observed on PR #7394, commit `caaf016bcc76e17e8c1e5080d47c45d9034fbb62`: attempt 1 failed, attempt 2 passed on the same `head_sha`. Failing job: https://github.com/tinacms/tinacms/actions/runs/31062504940/job/92493212232

Expected: the test passes deterministically, or fails only when the code is actually broken.

Actual: the assertion times out against an empty DOM (`<body><div /></body>`), so `TinaProvider` never rendered its children.

`TinaProvider` renders `null` until its async boot resolves each plugin's `client()` thunk. That thunk is a dynamic `import()`, which Vite is not able to preload because it is not in the test's static module graph. Vite therefore transforms it on demand during the test — including the `@tinacms/ui` (shadcn/Radix) graph that `string-field.ui.tsx` pulls in since #7154 — through a single transform server shared by ~40 concurrently running test files.

That cost lands inside the `findBy*` budget. Measured locally on an idle machine, one cold segment import costs **270 ms** against **0.015 ms** warm. On the failing CI run the test burned **5038 ms** against the configured `asyncUtilTimeout: 5000`, and the next test in the file then took 1030 ms — the import landed roughly 6 s after the first render.

#7154 already raised `asyncUtilTimeout` from 1 s to 5 s for this exact reason ("too tight for TinaProvider's async boot (plugin graph + lazy segment imports) on loaded CI runners"). macOS has now drifted past 5 s, so raising the ceiling again only moves the flake rather than removing it.

### Your environment

- Affected job: `build (macOS-latest)` in `.github/workflows` (ubuntu passes on the same tree)
- Package: `packages/v4/@tinacms/tinacms`, vitest 2.1.9, happy-dom, `@testing-library/dom` 10.4.1
- Non-default setup: `vitest.config.ts` sets `testTimeout: 30_000` and runs `@vitejs/plugin-react` with `babel-plugin-react-compiler`; `src/test/setup.ts` sets `configure({ asyncUtilTimeout: 5000 })`
- 10 test files render `<TinaProvider>` and share the pattern: the four field tests (`string`, `boolean`, `datetime`, `number`) and six `src/editor/*` tests

### A way for us to reproduce

Deterministically, without waiting for a bad CI day — render `<TinaProvider>` under a budget shorter than a cold segment import:

```ts
configure({ asyncUtilTimeout: 50 });
```

A test file with no warm-up fails with the error above. The same file passes when the plugin's client segment is loaded in `beforeAll` first.

### Impact

False red builds on unrelated PRs, which trains reviewers to ignore CI failures.

kulesy:
Closes #7417

**TL;DR** Make the v4 tests that render `<TinaProvider>` deterministic instead of racing a lazy plugin import.

**Pain:** these tests flake red on macOS CI on unrelated PRs, which trains reviewers to ignore CI failures. `TinaProvider` renders `null` until its async boot resolves each plugin's `client()` thunk — a dynamic import Vite transforms on demand, including the `@tinacms/ui` graph behind it. In a test body that cost falls inside the `findBy*` budget, so a loaded runner blows the timeout and asserts against an empty DOM. #7154 already raised `asyncUtilTimeout` to 5s for exactly this reason; macOS has now drifted past 5s too, so raising the ceiling again only moves the flake.

**Solution:** loads the core plugin client segments once in the shared test setup, so the transform happens before any test body runs and every boot reads from the module cache. Eight lines in `src/test/setup.ts`, nothing for a test author to remember — a new test that renders `<TinaProvider>` is covered automatically.


kulesy:
### Evidence

On PR #7394 (`caaf016bcc76e17e8c1e5080d47c45d9034fbb62`), attempt 1's `build (macOS-latest)` job [failed](https://github.com/tinacms/tinacms/actions/runs/31062504940/job/92493212232) on this single test while `build (ubuntu-latest)` passed on the identical tree; attempt 2 passed on the same `head_sha` with no code change.

The failing log shows the shape of the race directly:

```
❯ src/plugins/fields/string/string-field.test.tsx (10 tests | 1 failed) 6235ms
  × StringField rendering > renders the ingested value into a text input 5038ms
    → Unable to find a label with the text of: title
    <body>
      <div />
    </body>
  ✓ StringField rendering > falls back to the descriptor default value when absent 1030ms
```

Three things pin the cause:

- **5038 ms against `asyncUtilTimeout: 5000`** — it timed out, it did not assert something wrong.
- **`<body><div /></body>`** — the provider was still rendering `null`, so boot never completed.
- **the *next* test took 1030 ms** — the import finally landed ~6 s after the first render. It is the first test in the file that pays the cold cost, which is why only one test fails.

Measured cost of a single cold segment import, on an idle machine with a warm Vite cache:

```
cold=270.8ms  warm=0.015ms  ratio=17758x
```

### Why not just raise the timeout again

#7154 — the PR that swapped the raw `<input>` for shadcn primitives, making the lazy graph heavy — added `configure({ asyncUtilTimeout: 5000 })` in the same commit, with the comment *"findBy\*/waitFor default to 1s — too tight for TinaProvider's async boot (plugin graph + lazy segment imports) on loaded CI runners."* The ceiling has already been raised once and CI has drifted past it. The unbounded work has to leave the timed window, not get a bigger window.

### Test plan

Rather than asserting the fix works, I ran a controlled experiment — two identical tests under a deliberately tight 50 ms budget:

| | result |
|---|---|
| **cold** (no warm-up) | fails with the *exact* CI error, `Unable to find a label with the text of: title` |
| **warm** (`beforeAll`) | passes |

- [x] Full v4 suite: 469 tests pass
- [x] The 10 affected files: **10/10 consecutive runs** green
- [x] `pnpm types` clean (`tsc` + `tsconfig.test.json`)
- [x] `biome check` clean on all 13 changed files

### Trade-off the reviewer should weigh

I first tried the tidier approach — warm once globally in `src/test/setup.ts` — and rejected it on measurement. It loads the React UI graph into all ~40 test files including node-only ones, and **more than doubled cumulative setup: 9.1s → 20.4s**. The targeted `beforeAll` costs +2.7s. The cost of the targeted approach is that a *new* test file rendering `<TinaProvider>` must remember to call the helper; the helper's doc comment explains when it is needed.

### Out of scope, worth a follow-up

`resolveClientSegments` awaits client thunks **sequentially** in a `for` loop, so real browser boot latency is the *sum* of every plugin's import — four for a default config. `Promise.all` over the mapped array would parallelize while preserving result order (order matters for the field-capability conflict check). That is a production win, not a test fix, so I left it out of this PR.

kulesy:
### Approach changed — corrects a measurement error in the comment above

The first version added a `warmPluginClients` helper and called it from `beforeAll` in the 10 files that render `<TinaProvider>`. I rejected the simpler "warm once in `setup.ts`" on the grounds that it took cumulative setup from 9.1s to 20.4s.

**That metric was wrong to decide on.** Cumulative setup sums across parallel workers; nobody waits that long. On wall clock (median of 3 runs of the full v4 suite):

| Approach | Wall clock | Protection |
|---|---|---|
| Per-file `beforeAll` | 2.96s | opt-in — a new test can silently reintroduce the flake |
| Warm once in `setup.ts` | 3.41s | automatic |

The real price of "impossible to forget" is **~0.45s (+15%)** on a 3-second suite, not the 2x I implied. For a bug whose entire cost is that it fails *silently, months later, on somebody else's PR*, that is worth paying.

So this now does the simple thing. The diff dropped from 13 files / 61 insertions to **8 lines in `src/test/setup.ts`**, and `warm-plugins.ts` is gone.

### Verification of the automatic property

The point of the change is that a test author never has to know the rule exists, so I tested exactly that — a fresh file with no warm-up call anywhere in it, under a deliberately tight 50 ms budget:

| `setup.ts` warm | Result |
|---|---|
| enabled | **passes in 19ms** |
| disabled | fails with `Unable to find a label with the text of: title` — the exact CI error |

- [x] Full v4 suite: 469 tests pass
- [x] Field + editor tests: **8/8 consecutive runs** green
- [x] `pnpm types` clean, `biome check` clean

### Note for reviewers

The suite gets ~0.45s slower, and that is deliberate. If the v4 suite grows enough for it to matter, the cheap next step is splitting the vitest config into projects so only `.tsx` tests load the warm — all 10 `TinaProvider` files are `.tsx` and the node-only tests are `.ts`, which would cut it to roughly +0.15s. Not worth the config surface today.

KahaMason (approved):
LGTM

PR #7424, opened 2026-08-10
fix(@tinacms/web-components): sanitise raw HTML in tina-markdown

wicksipedia:
Branches off `bj/20260803/feat/tina-markdown-web-component` and targets it, not `main`.

`tina-markdown` passed `html` / `html_inline` node values to `innerHTML`, which builds live DOM from them. This routes them through DOMPurify, so authored markup still renders and executable content does not.

Commit 1 is tests only. It records what the React and Astro renderers already do, so the baseline is visible before anything changes:

| Case | React | Astro | web-components (before → after) |
|---|---|---|---|
| block `html` | escaped | escaped | renders → renders sanitised |
| inline `html_inline` in `<p>` | inline | inline | wrapped in a `div` → inline |
| `` | no element | no element | handler intact → handler stripped |
| `<script>` | n/a | n/a | present → dropped |
| consumer opt-in | `components.html` | `components.html` | none → `TinaMarkdown.components.html` |

Raw HTML still renders, which is the behaviour you argued for in the thread. `TinaMarkdown.components` is how a site developer opts back into unsanitised output, matching the `components` prop on the other two renderers.

URLs on `a` and `img` nodes are deliberately left as authored, so they stay permissive where React and Astro sanitise. Note that DOMPurify does drop dangerous schemes inside raw HTML, so an `<a>` written as markdown and the same `<a>` written as raw HTML are not treated alike.

```bash
pnpm build --filter=@tinacms/web-components...
pnpm --filter @tinacms/web-components test           # 25
pnpm --filter tinacms exec vitest run src/rich-text  # 7
pnpm --filter @tinacms/astro test                    # 80
```


brookjeynes-ssw:
image
image

Custom components via `tina-markdown`

wicksipedia:
@brookjeynes-ssw your changes lgtm

brookjeynes-ssw (approved):

PR #7429, opened 2026-08-11
fix(@tinacms/app): stop a failed raw markdown editor blanking the whole admin

wicksipedia:
https://youtu.be/S9WrOnLfq5M

## What breaks

Clicking the raw markdown button can blank the entire admin.

It fails two different ways, and both used to cost you the page.

**The chunk fails to fetch.** The raw editor loads lazily:

```tsx
const RawEditor = React.lazy(() => import('./fields/rich-text'));
```

`App.tsx` wrapped it in `Suspense` and nothing else. `Suspense` handles a pending promise. It does nothing for a rejected one. A stale cache after a deploy or a flaky connection is enough.

**The field fails to serialize.** `RawEditor` calls `serializeMDX` in a `useMemo` during render, so content it cannot represent throws with the chunk already loaded. Apply inline code to a phrase, bold one word inside it, click Raw Markdown:

```
Marks inside inline code are not supported
```

Reproduced in the kitchen-sink admin on `main`: `document.body.innerText` empty, zero children under the root. React unmounts the tree and the editing session is gone, taking unsaved work in the other fields with it.

## The change

A boundary around the lazy editor. The failure now stays inside the field. The rest of the form keeps working, and a button returns you to the rich-text editor through the `setRawMode(false)` already threaded through.

The message names the outcome rather than a cause, since the boundary cannot tell the two apart, and shows the thrown message underneath. For the formatting case that tells you which formatting to undo. It is also the string a bug report needs verbatim, which `CLAUDE.md` makes the first required field of one.

I scoped it to this field rather than reusing the full-screen boundary in `tina-cms.tsx`. A field that failed to open is a local problem, and taking over the screen to announce it repeats the behaviour this fixes.

One layout fix came with it: the field wrapper sets `white-space: nowrap`, which the message inherited and which clipped it mid-sentence.

## Tests

Five: children render, a failed import stays contained, a failed serialize stays contained, the thrown message reaches the screen, and the dismiss button leaves raw mode.

This is the package's first `.tsx` test, so `vitest.config.ts` picks up `src/**/*.test.tsx` and `@testing-library/react` joins devDependencies from the existing catalog entry.

## Scope

This does not fix either underlying failure. `serializeMDX` still cannot represent a mark inside inline code, which is worth its own issue. This fixes losing the admin when something in that field goes wrong.

### DIFF
--- a/.changeset/app-raw-editor-error-boundary.md
+++ b/.changeset/app-raw-editor-error-boundary.md
@@ -1,0 +1,9 @@
+---
+"@tinacms/app": patch
+---
+
+Switching a rich-text field to raw markdown no longer risks blanking the whole admin.
+
+The raw editor arrives as a lazy chunk. `Suspense` covered the wait, but nothing caught a rejected import, so a failed fetch tore down the entire editing session: no message, no way back, and any unsaved work in the other fields gone with it. A flaky network or a stale cache after a deploy was enough to trigger it.
+
+The failure now stays inside the field it came from. You get a message saying your content is untouched, and a button back to the rich-text editor.
--- a/.changeset/app-raw-editor-error-boundary.md
+++ b/.changeset/app-raw-editor-error-boundary.md
@@ -2,8 +2,8 @@
 "@tinacms/app": patch
 ---
 
-Switching a rich-text field to raw markdown no longer risks blanking the whole admin.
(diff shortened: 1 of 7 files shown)


isaaclombardssw (dismissed):
Image

Nice ✅

isaaclombardssw (approved):
✅

PR #7427, opened 2026-08-11
fix(@tinacms/mdx): keep bold and italic when the selection starts or ends with a space

wicksipedia:
Fixes #7426

Demo - https://youtu.be/OX4HWD5bhmc

## What was broken

Select a word *and its trailing space*, then hit bold. TinaCMS saved `**word **`. CommonMark will not close emphasis that sits against a space, so the published page rendered literal asterisks and the bold was gone. The editor kept showing it as bold, so nothing looked wrong until publish.

Plate is behaving correctly here: it stores `{ text: 'word ', bold: true }`, and `mdast-util-to-markdown` writes `**` + children + `**` verbatim without normalising.

## The fix

`normalizeMarkWhitespace` runs on the mdast tree after it is built from the editor value and before markdown is written, on both serializer paths. It hoists edge whitespace out of `strong` / `emphasis` / `delete`, drops marks left with nothing inside, and trims whitespace that lands at a block boundary.

```
Some **word **more   →   Some **word** more
```

Both v3 and v4 serialize through `serializeMDX`, so both get the fix.

## Related cases, same root cause

Found while writing the tests:

| Input | Saved before |
|---|---|
| `{text:' word', bold}` | `** word**` |
| `{text:'word ', italic}` | `*word *` |
| `{text:'word ', strikethrough}` | `~~word ~~` |
| `{text:'word ', bold, italic}` | `***word ***` |
| `{text:' ', bold}` | `** **` |
| `{text:'', bold}` | `****` |
| multi-node marked run | `**ab **` |
| mark inside a multi-child link | `[**w **tail](u)` |
| 4+ leading spaces in a mark | became an indented code block |

Left alone on purpose: `inlineCode` (`to-markdown` already pads it correctly), `highlight` (serializes to `<mark>`), and interior whitespace such as `**Hello *world*, again**`, which has to survive.

## Testing

Tests came first and failed before the fix landed. 32 new tests across both parser configurations, including round trips that reparse the output.

Checked in a real editor against `examples/next/kitchen-sink`: keyboard and toolbar produce byte-identical output, and the rendered page emits `<strong>`. An untouched document still saves byte-identical. No churn in the 79 existing fixture files.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #7430, opened 2026-08-11
fix(@tinacms/mdx): keep author whitespace at block edges

isaaclombardssw:
The block-edge trim in #7427 also deletes NBSP indentation and spacer paragraphs nobody touched - `\s` matches U+00A0, and the trim runs whether or not anything was hoisted.

Now clears only the whitespace the hoist moved to the edge. 8 tests added, 159/159 green.

Repro, main vs this branch: https://youtu.be/o9Y_pLkSfs4

### DIFF
--- a/packages/@tinacms/mdx/src/stringify/mark-whitespace.test.ts
+++ b/packages/@tinacms/mdx/src/stringify/mark-whitespace.test.ts
@@ -6,6 +6,8 @@ import { serializeMDX } from './index';
 
 const passthrough = (value: string) => value;
 
+const NBSP = ' ';
+
 const fields: [string, RichTextField][] = [
   ['mdx', { name: 'body', type: 'rich-text' }],
   [
@@ -19,16 +21,21 @@ const paragraph = (children: Plate.InlineElement[]): Plate.RootElement => ({
   children: [{ type: 'p', children }],
 });
 
-const serialize = (
-  children: Plate.InlineElement[],
+const serializeRoot = (
+  value: Plate.RootElement,
   field: RichTextField
 ): string => {
-  const result = serializeMDX(paragraph(children), field, passthrough);
+  const result = serializeMDX(value, field, passthrough);
   if (typeof result !== 'string') {
     throw new Error(`Expected a string, received ${typeof result}`);
   }
   return result;
 };
+
+const serialize = (
+  children: Plate.InlineElement[],
+  field: RichTextField
+): string => serializeRoot(paragraph(children), field);
 
 const boldTextsOf = (markdown: string, field: RichTextField): string[] => {
   const bolds: string[] = [];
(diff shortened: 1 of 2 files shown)


wicksipedia (commented):
The diagnosis here is right, and I think the model is right too. `\s` matching U+00A0 was destroying author content, and trimming whitespace the pass never moved was out of scope for a function called `hoistFromMarks`. Tracking provenance and clearing only what the hoist put at the edge is the correct fix. Moving `mergeText` after the edge-clearing is load-bearing, since merging destroys the node identity the `fromHoist` set depends on, and it is easy to reorder those two lines later without noticing. Worth a comment on that line if you agree.

One thing needs resolving before this lands.

## Author indentation now round-trips into a code block on the markdown parser

The block-edge trim was also protecting against four leading spaces reaching the file. Removing it for author whitespace brings that back, and it is not only a hazard for downstream tooling — it breaks inside Tina, on the `markdown` parser:

```ts
const field = { name: 'body', type: 'rich-text', parser: { type: 'markdown' } };
serialize([{ type: 'text', text: '    word' }], field);
// this branch: "    word\n"   -> parseMDX gives a code_block
// base:        "word\n"       -> parseMDX gives a paragraph
```

The `mdx` parser is fine, because indented code is disabled there, so a test written against `mdx` alone passes and hides this. The same happens when author indentation sits next to whitespace the hoist moved, which needs only three typed spaces to cross the threshold:

```ts
serialize(
  [
    { type: 'text', text: '   ' },
    { type: 'text', text: ' bold', bold: true },
  ],
  field
); // "    **bold**\n"
```

An author who indents a line and bolds a word starting with a space gets a code block on reload. That is the gesture #7427 is about, so the two changes interact.

I do not think this should send you back to the unconditional trim, which traded this bug for a worse one. But the split should be recorded rather than discovered later. These pass as written and can go in as-is:

```ts
  it('writes author indentation verbatim', () => {
    expect(serialize([{ type: 'text', text: '    word' }], field)).toBe(
      '    word\n'
    );
  });

  it('leaves author indentation next to whitespace the hoist moved', () => {
    expect(
      serialize(
        [
          { type: 'text', text: '   ' },
          { type: 'text', text: ' bold', bold: true },
        ],
        field
      )
    ).toBe('    **bold**\n');
  });
```

and, outside the `describe.each` so each parser is named:

```ts
/**
 * Four spaces of author indentation survive serialization, so what the file
 * means depends on who reads it back. These pin the split rather than endorse
 * it: the markdown parser round trip is a defect, not a decision.
 */
describe('author indentation round trip', () => {
  const [[, mdxField], [, markdownField]] = fields;
  const indented = [{ type: 'text' as const, text: '    word' }];

  it('stays a paragraph through the mdx parser', () => {
    const markdown = serialize(indented, mdxField);
    expect(
      (parseMDX(markdown, mdxField, passthrough) as Plate.RootElement).children
    ).toEqual([{ type: 'p', children: [{ type: 'text', text: 'word' }] }]);
  });

  it('becomes a code block through the markdown parser', () => {
    const markdown = serialize(indented, markdownField);
    expect(
      (parseMDX(markdown, markdownField, passthrough) as Plate.RootElement)
        .children
    ).toEqual([
      {
        type: 'code_block',
        value: 'word',
        children: [{ type: 'code_line', children: [{ text: 'word' }] }],
      },
    ]);
  });
});
```

That is 42 green in `mark-whitespace.test.ts`, up from 36.

The actual fix belongs in its own PR rather than this one. The root cause is that `handlers['text']` in `packages/@tinacms/mdx/src/stringify/index.ts:106` drops the space escape for all phrasing content unconditionally, so a leading space at a block start is never escaped. Keeping that escape at a block edge would preserve author whitespace without the ambiguity, and would make the `BLOCK_BOUNDARIES` special case unnecessary.

## Smaller things

**The `NBSP` const holds a literal U+00A0 byte.** The name is good, but the value is invisible in the diff and in most editors, and it does not survive a copy-paste or an editor that normalizes whitespace on save. `const NBSP = '\u00a0'` reads identically, shows a reviewer what it is, and can be grepped.

**Trailing whitespace now reaches the file.** `[bold("bold "), text("    ")]` serializes to `"**bold**     \n"`. Markdown ignores it at the end of a paragraph, but format-on-save and most CI whitespace checks strip it, so files will pick up diffs nobody made.

**The heading test passes for a different reason than the others.** `## &#x20;Title` comes from `mdast-util-to-markdown`'s own heading handler, which escapes a leading space to avoid a setext ambiguity, not from anything in this pass. It is a good test to have, but it does not exercise the change, and it is the reason paragraphs and headings behave differently here.

**Typing.** `fromHoist` as `Set<Md.Text>` instead of `Set<Md.PhrasingContent>` drops the `(edge as Md.Text)` cast, since `edge?.type === 'text'` already narrows.

wicksipedia:
Correction to my earlier comment. I called the code block behaviour a regression and said it needed resolving before this lands. Both were wrong, and the framing was unfair to this PR.

I measured against #7427's branch rather than against main. On main today, before either PR, `    word` already serializes verbatim and reads back as a `code_block` on the `markdown` parser. The unconditional trim in #7427 hid that by deleting the whitespace, so comparing this branch to #7427 made a restoration look like a break.

What this branch actually does is put main's behaviour back for author whitespace while fixing the content loss #7427 would otherwise have shipped. It does not make anything worse than what is released, and nothing here blocks it.

The `code_block` handling is still a real bug, but it predates both PRs and belongs in its own change against main. The cause is that `handlers['text']` drops mdast-util-to-markdown's leading-space rule, `{character: ' ', before: '[\r\n]'}`, which is what would otherwise escape the indentation as `&#x20;`. The same filter is duplicated in `src/stringify/index.ts` and `src/next/stringify/to-markdown.ts`, which is why the two parsers disagree. Keeping that one rule in both places preserves the indentation and round trips it intact, and the existing suite stays green.

The pinning tests are still worth having, but as a record of main's behaviour rather than of anything this PR changed.

The rest of my earlier comment stands: the `NBSP` const holding a literal U+00A0 byte, the `mergeText` ordering that `fromHoist` depends on, and `tableCell` writing whitespace that GFM discards on read.

Apologies for the misdirection.

PR #7431, opened 2026-08-11
fix(@tinacms/mdx): write serializer errors for the person who has to fix them

wicksipedia:
## What breaks

Apply inline code to a phrase, bold one word inside it, and saving fails with:

```
Marks inside inline code are not supported
```

"Mark" is a Slate term. A content editor has never met it, and the sentence does not say which formatting to remove.

This matters more since #7429, which prints the thrown message verbatim inside the field. At that point these strings stopped being developer-facing exceptions and became UI copy, without being rewritten as UI copy.

## The change

Six messages reworded, all reachable by editing content:

| Before | After |
|---|---|
| `Marks inside inline code are not supported` | `Inline code can't have other formatting on it. Remove the formatting from the code text.` |
| `Marks inside highlight are not supported` | `Highlighted text can't have other formatting on it. Remove the formatting from the highlighted text.` |
| `BlockElement: X is not yet supported` | `This block can't be saved as markdown ("X"). Remove it from the field to continue.` |
| `BlockContentElement: X is not yet supported` | same as above |
| `InlineElement: X is not supported` | `This content can't be saved as markdown ("X"). Remove it from the field to continue.` |
| `Stringify props: X not yet supported` | `A "X" field can't be saved as markdown yet. Remove this block, or ask a developer to change the field.` |

Each keeps the offending type name, so a bug report still carries it. Both the `mdx` and `markdown` serializer paths raise these, so both were updated.

## Left precise on purpose

Schema and template errors are unchanged: `Global templates not supported`, `Rich-text list is not supported`, `Unable to find template for JSX element X`. These fire from schema configuration, so a developer hits them on the first run and the exact term is what they need.

## Tests

Six, across both parser configurations: the message raised for the code-plus-bold combination, the message raised for an unwritable block, and an assertion that neither contains any form of the word "marks". The last exists because that is the specific failure being fixed, and a plain equality check would not catch a reworded-but-still-jargony replacement.

## Not in scope

Neither combination is made to work. `serializeMDX` still cannot represent formatting inside inline code, which is worth its own issue. This changes what the editor is told when it happens.

### DIFF
--- a/.changeset/mdx-editor-facing-mark-errors.md
+++ b/.changeset/mdx-editor-facing-mark-errors.md
@@ -1,0 +1,7 @@
+---
+"@tinacms/mdx": patch
+---
+
+The errors raised for formatting combinations that cannot be saved now say what to change.
+
+Applying inline code and then bolding a word inside it used to fail with "Marks inside inline code are not supported". The raw markdown editor shows that message verbatim in the field, so a content editor was handed a Slate term and no indication of which formatting to undo. It now reads "Inline code can't have other formatting on it. Remove the formatting from the code text." The equivalent message for highlighted text was reworded the same way.
--- a/.changeset/mdx-editor-facing-mark-errors.md
+++ b/.changeset/mdx-editor-facing-mark-errors.md
@@ -2,6 +2,8 @@
 "@tinacms/mdx": patch
 ---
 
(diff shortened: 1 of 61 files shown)


isaaclombardssw (approved):
✅ Jumpscare 59 lints, error changes are good

PR #7433, opened 2026-08-15
fix(@tinacms/app): move graphiql off the 2023 pre-release pin

fetzu:
## The pin

The catalog pins `graphiql` to `3.0.0-alpha.1` — published 2023-06-20, four days before stable 3.0.0, and never revisited. Current stable is 5.2.4.

Because the pin is exact, no downstream consumer can override it. Every install of `@tinacms/app` gets:

```
@tinacms/app -> graphiql@3.0.0-alpha.1 -> @graphiql/react@0.18.0
                |- markdown-it@12.3.2 -> linkify-it@3.0.3
                `- codemirror@5.65.20
```

`markdown-it@12.3.2` and `linkify-it@3.0.3` carry GHSA-6v5v-wf23-fmfq, GHSA-22p9-wv53-3rq4 and GHSA-v245-v573-v5vm.

## Bumping graphiql is not enough

`@graphiql/react` declares `markdown-it: ^14.1.0`, which looks patched — `14.1.2` is the fix. But `@tinacms/mdx` depends on `typedoc`, which declares the same `^14.1.0`, and pnpm deduped both to the `14.1.0` already in the lockfile. One version below the fix, and the graphiql bump alone leaves all three advisories exactly where they were.

So this adds a `markdown-it: ^14.3.0` override. It lifts both chains at once — `markdown-it@14.3.0`, and `linkify-it@5.0.2` behind it — and the repo ends up with a single copy of each rather than the two it had.

That override is the part that actually closes the advisories. Without it the rest is cosmetic.

## Why 4 and not 5

I went to 5 first. Two things sent me back.

**It drops the controlled `query` and `variables` props.** In 5 they become `initialQuery` / `initialVariables`, applied once when the store is created. The playground's "Queries" sidebar is built entirely on the old behaviour — clicking a collection sets React state that flows back in as `query={query}`, and the editor replaces its contents. On 5 that sidebar keeps rendering, keeps responding to clicks, and silently stops updating the editor. Nothing fails at build time. Restoring it means rewriting the sidebar against the new zustand store and `updateActiveTabValues`.

**It swaps CodeMirror for a bundled Monaco.** That undoes the `monaco-editor` removal currently sitting in `.changeset/` — the one that stopped shipping 73 MB nobody executed. It also needs `import 'graphiql/setup-workers/vite'`, whose `?worker` specifiers are a Vite feature. `@tinacms/app` ships source (`main: src/main.tsx`) and is built by whatever `@tinacms/cli` the user has installed, so that import makes Monaco chunking and worker support a build requirement for every consumer, across the caret ranges `CONTRIBUTING`-published packages float on.

4.1.2 keeps CodeMirror and keeps the prop contract, and its `@graphiql/react@0.34.1` reaches the same patched `markdown-it`. Same advisories closed, none of that surface area. 5 is still worth doing — it just wants its own PR, with the sidebar rewrite and the Monaco decision made deliberately rather than as fallout from a security bump.

## The crash the build did not catch

The playground compiled, typechecked and rendered fine on 4. Executing any query blew up the whole admin:

```
TypeError: Cannot read properties of undefined (reading 'query')
  at HistoryContextProvider
```

`defaultTabs={[]}` was the cause. An empty array is truthy, so `getDefaultTabState` returns it verbatim and you get `{ activeTabIndex: 0, tabs: [] }` — zero tabs, with the active index pointing at one. 4 dereferences `tabs[activeTabIndex]` when recording history, and 3.0.0-alpha.1 only ever survived it because a `??` short-circuit on a non-empty `query` prop skipped the same lookup.

Zero tabs was never a state the playground wanted. Dropping the prop lets GraphiQL build its single default tab seeded from `query`/`variables`, which is what it was reaching for.

## Verification

`pnpm audit`, before and after:

```
BEFORE  GHSA-6v5v-wf23-fmfq, GHSA-22p9-wv53-3rq4, GHSA-v245-v573-v5vm
AFTER   none of the three present
        high 110 -> 108, moderate 104 -> 102
```

`pnpm why` now reports one patched copy of each: `markdown-it@14.3.0`, `linkify-it@5.0.2`.

A green build proves nothing here, so the playground was exercised in a browser against `examples/next/kitchen-sink` on a local dev server:

- schema loads — the Docs explorer lists the full generated type set
- `globalConnection` executes and returns `"name": "Tina Kitchen Sink"` from `content/global/index.json`
- `post` executes **with variables** (`{"relativePath": "large-file.md"}`) and returns the document — this is the path that would have gone quiet on 5
- Docs, History and the Tina "Queries" pane all work; History is specifically the one that was crashing
- no graphiql-related console errors

`codemirror@5.65.20` is still in the tree. 4 uses it, and it is not one of the three advisories.

## Scope

The `markdown-it` override reaches past graphiql into `typedoc`'s chain. That is deliberate — it is the only way the advisories actually clear — but it is the one change here that is not strictly about the playground.

Not fixed, noticed in passing: `Playground.tsx` puts its `key` on the `<button>` instead of the `<li>`, so React warns on every render of the Queries list. Pre-existing and unrelated to the bump.


fetzu:
Note: I am unable to sign the CLA, both https://cla-assistant.io/tinacms/tinacms and https://cla-assistant.io/tinacms/tinacms?pullRequest=7433 serve a page with a blank/unsettable "Version" field.

joshbermanssw (dismissed):
LGTM

fetzu:
@joshbermanssw sorry, it seems like merging upstream/main dismissed the review, didn't mean to create more work...

joshbermanssw (dismissed):

joshbermanssw (approved):

PR #7436, opened 2026-08-16
chore: add missing changeset for the multer 2.2.0 bump

### LINKED ISSUE #7435: multer 2.2.0 bump (#7079) landed without a changeset
**Pain**

#7079 bumped `multer` from 1.4.5-lts.1 to 2.2.0 in the `catalog:` but merged without a changeset.

`multer` is consumed by `next-tinacms-cloudinary` and `next-tinacms-dos`, both published and neither in the changesets `ignore` list. As it stands:

- `next-tinacms-cloudinary` will ship the multer 1.x to 2.x major on the back of an unrelated patch entry (`cloudinary-escape-search-expression`), so its changelog describes a search-expression fix while a breaking runtime dependency change rides along.
- `next-tinacms-dos` gets no release at all, so its published dependency stays on multer 1.4.5-lts.1 while `main` says 2.2.0.

multer 1.x is end of life and 2.x changes file-handling behaviour, so this is not a cosmetic omission.

Note the same gap was caught by hand on #7076, where the esbuild changeset was added manually. Nothing in CI requires a changeset, so this depends on whoever reviews noticing.

**Approach**

Add a changeset covering both packages before Version Packages #7287 merges, otherwise `next-tinacms-dos` waits for the following release. Separately, consider a CI check that fails a PR touching `pnpm-workspace.yaml` or a published package manifest with no changeset present.

**Acceptance Criteria**

- [ ] A changeset names both `next-tinacms-cloudinary` and `next-tinacms-dos` and describes the multer bump
- [ ] Version Packages #7287 shows both packages queued for release
- [ ] A decision is recorded on whether to enforce changesets in CI

kulesy:
Closes #7435

**TL;DR** #7079 bumped multer 1.4.5-lts.1 to 2.2.0 without a changeset. This adds one, so both affected packages record the change.

**Pain:** multer is consumed via `catalog:` by `next-tinacms-cloudinary` and `next-tinacms-dos`, both published, and neither is in the changesets `ignore` list. #7079 landed the bump with no changeset, and nothing in CI requires one. As it stands, `next-tinacms-cloudinary` will ship the multer 1.x to 2.x major on the back of an unrelated patch entry (`cloudinary-escape-search-expression`), so the changelog describes a search-expression fix while a breaking runtime dependency change rides along. `next-tinacms-dos` gets no release at all, leaving its published dependency on multer 1.4.5-lts.1 while `main` says 2.2.0.

**Solution:** Added a changeset marking both packages `patch` and naming the actual change, including that multer 1.x is end of life and 2.x alters file-handling behaviour. Version Packages #7287 is still open, so this is picked up in the current release rather than the one after.

### DIFF
--- a/.changeset/bump-multer-off-eol-1x.md
+++ b/.changeset/bump-multer-off-eol-1x.md
@@ -1,0 +1,6 @@
+---
+'next-tinacms-cloudinary': patch
+'next-tinacms-dos': patch
+---
+
+Bump `multer` to 2.2.0. multer 1.x is end of life; 2.x changes file-handling behaviour, so review upload paths when upgrading.


kulesy:
Validated with `changeset status`, which resolves the file to exactly the two intended packages:

```
- next-tinacms-cloudinary
- next-tinacms-dos
```

`patch` follows the convention used on #7076. If you would rather signal the breaking runtime dependency more strongly, `minor` on both is a one-line change here.

wicksipedia (approved):

PR #7437, opened 2026-08-17
fix(tinacms-authjs): unwrap CommonJS default imports of next-auth

### LINKED ISSUE #7434: tinacms-authjs: ESM default-imports of CommonJS next-auth break outside webpack (Netlify Functions, Astro, plain Node ESM)
`tinacms-authjs` ships ESM that default-imports CommonJS `next-auth`, so it only works under a webpack-style `__esModule` interop. Under real Node ESM, or an esbuild `platform=node` bundle (Netlify Functions, Astro, any non-Next self-host), the default imports resolve to the module object instead of the function and every auth route 500s.

Reported in Discord `#ask-for-help` by a user on Astro + Netlify Functions.

### The exact error message

```shell
Function tina has returned an error: (0 , import_credentials.default) is not a function
```

### Steps to reproduce

Minimal, no Netlify needed:

1. `npm i tinacms-authjs@24.0.2 next-auth@4.24.15 esbuild`
2. `entry.js`:
   ```js
   import { TinaCredentialsProvider } from 'tinacms-authjs'
   export const handler = async () => TinaCredentialsProvider({ databaseClient: {} })
   ```
3. `esbuild entry.js --bundle --platform=node --target=node22 --format=cjs --outfile=out.cjs`
   (the same shape Netlify's zip-it-and-ship-it produces)
4. `node -e "require('./out.cjs').handler()"`

Expected: the provider is constructed and `/api/tina/auth/session` returns a session.

Actual: `(0 , import_credentials.default) is not a function`, and `/api/tina/auth/session` returns 500.

### Root cause

`packages/tinacms-authjs/src/index.ts` does:

```ts
import NextAuth from 'next-auth'
import CredentialsProvider from 'next-auth/providers/credentials'
```

`@tinacms/scripts` builds this with `platform: 'node'`, `format: 'esm'` and peer deps external (`packages/@tinacms/scripts/src/index.ts:383-391`), so both default imports survive verbatim into `dist/index.js`.

next-auth v4 is CommonJS. `providers/credentials.js` is:

```js
Object.defineProperty(exports, "__esModule", { value: true });
exports.default = Credentials;
```

Node ESM binds a CJS default import to `module.exports`, so the imported binding is `{ default: [Function] }`, not the function:

```
$ node -e "import('next-auth/providers/credentials').then(m => console.log(typeof m.default))"
object
```

esbuild with `platform=node` emits `__toESM(require_credentials(), 1)`, which deliberately reproduces that same Node semantic, so the bundled output fails identically. This is why switching `node_bundler` between `esbuild` and `zisi` changes nothing.

webpack (and therefore Next.js) unwraps `__esModule` on a default import, which is why every self-hosted path we test passes and only the non-Next ones break.

`import NextAuth from 'next-auth'` has the identical problem, so `NextAuth(authOptions)(req, res)` inside `AuthJsBackendAuthProvider.extraRoutes.auth.handler` throws `NextAuth is not a function` as soon as the first error is worked around. `getServerSession` is a named import and is fine, as is `dist/tinacms.js`, which only uses named imports from `next-auth/react`.

### Suggested fix

Interop-safe unwrapping at the two call sites in `packages/tinacms-authjs/src/index.ts`:

```ts
import NextAuthImport from 'next-auth'
import CredentialsImport from 'next-auth/providers/credentials'

const NextAuth = (NextAuthImport as any).default ?? NextAuthImport
const CredentialsProvider = (CredentialsImport as any).default ?? CredentialsImport
```

This is correct under both webpack and Node ESM.

### Workaround for users today

Rebuild the provider and patch the auth route handler in the Netlify function:

```js
import { TinaAuthJSOptions, AuthJsBackendAuthProvider } from 'tinacms-authjs'
import NextAuthImport from 'next-auth'
import CredentialsImport from 'next-auth/providers/credentials'

const NextAuth = NextAuthImport.default ?? NextAuthImport
const Credentials = CredentialsImport.default ?? CredentialsImport

const provider = Credentials({
  credentials: {
    username: { label: 'Username', type: 'text' },
    password: { label: 'Password', type: 'password' },
  },
  authorize: async ({ username, password }) => {
    const r = await databaseClient.authenticate({ username, password })
    return r.data?.authenticate || null
  },
})
provider.name = 'TinaCredentials'

const authOptions = TinaAuthJSOptions({
  databaseClient,
  secret: process.env.NEXTAUTH_SECRET,
  providers: [provider],
})

const authProvider = AuthJsBackendAuthProvider({ authOptions })
authProvider.extraRoutes.auth.handler = async (req, res, opts) => {
  const url = new URL(req.url, `http://${req.headers?.host || 'localhost'}`)
  req.query.nextauth = url.pathname?.replace(`${opts.basePath}auth/`, '')?.split('/')
  await NextAuth(authOptions)(req, res)
}
```

Verified through the same esbuild config: a simulated `GET /api/tina/auth/session` returns 200 with an empty session instead of the 500.

### Your environment

- `tinacms-authjs` 24.0.2 (latest)
- `next-auth` 4.24.15
- `tinacms` 3.11.0, `@tinacms/datalayer` 2.0.29, `@tinacms/astro` 0.6.1
- Astro 7.2.1, React 19.2.8
- Node 22, Netlify Functions (`netlify dev`), `node_bundler = "esbuild"`
- Self-hosted backend, MongoDB + GitHub git provider

Also affects the docs: https://tina.io/docs/reference/self-hosted/tina-backend/netlify-functions/ and https://tina.io/docs/reference/self-hosted/auth-provider/authjs both show a config that cannot work outside Next.js.

kulesy:
Closes #7434

**TL;DR** Make `tinacms-authjs` work outside webpack: unwrap next-auth's CommonJS default export, and stop the node entry pulling in the browser `tinacms` package.

**Pain:** two independent bugs, both invisible to Next.js users because webpack resolves what Node's ESM loader will not. First, `next-auth` v4 is CommonJS with a Babel `__esModule` marker: webpack unwraps `exports.default`, while Node and esbuild bind the default import to `module.exports`, so `CredentialsProvider` was an object rather than a function and the module threw at load with `(0 , import_credentials.default) is not a function`. Second, the node entry imported a single string constant from `./tinacms`, dragging the whole browser package in; published `tinacms@3.11.0` has a bare directory import of `@heroicons/react/solid` and a named import from `color-string`, neither of which Node's ESM loader will resolve. Netlify Functions, Vercel Functions and plain Node ESM all failed, and the docs for the first two have been shipping a config that cannot run.

**Solution:** an `interopDefault` helper unwraps both default imports, which is correct under webpack and Node ESM alike. The constant moves to its own leaf module so the node bundle imports only `next-auth`, and `./tinacms` re-exports it so consumers are unchanged. `tests/authjs-node-entry.test.ts` guards both regressions, asserting the built entry's import specifiers and constructing a provider in a real Node ESM subprocess. Verified end to end against the docs' own Netlify Functions sample: sign in, session carrying the expected role, and guarded routes returning 401 without a cookie. Dropping the `tinacms` import also takes that bundled function from 21.0 MB to 5.4 MB.


kulesy:
**How this was verified**

Rebuilt the package locally, dropped the built `dist/` over `tinacms-authjs@24.0.2` in a scratch project, and ran the **unmodified** Netlify Functions sample from the docs (express 4 + cors + cookie-parser + `serverless-http` + `TinaNodeBackend` + `AuthJsBackendAuthProvider`) bundled with `esbuild --bundle --platform=node --format=cjs`, the way zip-it-and-ship-it bundles functions. Real API Gateway events, real cookie jar. Only `databaseClient` was stubbed, implementing the two operations `tinacms-authjs` actually calls.

Before, at module load:

```
TypeError: (0 , import_credentials.default) is not a function
    at TinaCredentialsProvider
    at TinaAuthJSOptions
```

After, same file, nothing else changed:

| Request | Result |
| --- | --- |
| `GET /auth/session` logged out | `200 {}` |
| `GET /auth/csrf` | `200` + token |
| `POST /auth/callback/credentials` good creds | `200`, session cookie set |
| `GET /auth/session` after login | `200` with `role: "user"` |
| `POST /api/tina/gql` with cookie | `400 {"error":"no variables"}` |
| `POST /api/tina/gql` no cookie | `401 {"error":"Unauthorized"}` |
| `POST /auth/callback/credentials` wrong password | `401 CredentialsSignin` |
| `GET /auth/session` after bad login | `200 {}` |

The `400` vs `401` pair is the one that matters: with a session cookie the request clears `isAuthorized` and reaches the GraphQL handler, which rejects my deliberately empty body; without a cookie it is rejected at the gate. So authorization still works, it is not merely that the module now loads.

**Why `.default ?? X` and not a bundler flag**

The failure is not Netlify-specific. Node's own ESM loader shows it directly:

```
$ node -e "import('next-auth/providers/credentials').then(m => console.log(typeof m.default))"
object
```

`__esModule` is a Babel/webpack convention, not a spec feature, so Node ignores it and esbuild's `platform=node` interop deliberately reproduces Node's behaviour. `.default ?? X` is correct under both conventions: under Node the module object has `.default`, and under webpack the value is already the function and `.default` is `undefined`. I checked there is no third layer (`Credentials.default.default` is `undefined`).

`getServerSession` needed no change: named imports bind by reading properties off `module.exports` under both conventions. `dist/tinacms.js` is likewise unaffected, using only named imports from `next-auth/react`.

**Checks**

- `biome check` clean on the file. It was reporting an import-order error on `main` already; since I was editing those exact lines I sorted them, which is the small reorder in the diff.
- `tsc --noEmit` error count is unchanged at 284, all of them pre-existing and in the sibling `tinacms` package.

**Related**

Docs currently carry a workaround for this in tinacms/tina.io#4775. Once this ships, that section should be deleted rather than left to rot.

packages/tinacms-authjs/src/index.ts
```diff
@@ -1,9 +1,15 @@
...
+import NextAuthImport, { AuthOptions } from 'next-auth';
+import { getServerSession } from 'next-auth/next';
+import CredentialsImport from 'next-auth/providers/credentials';
 import { TINA_CREDENTIALS_PROVIDER_NAME } from './tinacms';
 
+// next-auth v4 is CommonJS: webpack unwraps its `__esModule` default but Node's
+// ESM loader and esbuild do not, breaking every non-Next.js backend. See #7434.
+const NextAuth = (NextAuthImport as any).default ?? NextAuthImport;
+const CredentialsProvider =
+  (CredentialsImport as any).default ?? CredentialsImport;
```
18-th: Could we cast to an actual type instead of any here?

18-th:
I don't believe this fixes the plain Node ESM imports, so this only addresses half of the #7434

packages/tinacms-authjs/src/index.ts
kulesy: Done in d4b10d7 - swapped the casts for a typed helper:

```ts
const interopDefault = <T>(mod: T & { default?: T }): T => mod.default ?? mod;
```

No cast needed at all, since `default` being optional lets a plain function satisfy the parameter type. I checked both bindings keep their real next-auth types rather than degrading to `any`, using an `IsAny` probe compiled against the package's own tsconfig.

kulesy:
@18-th on the plain Node ESM point: you are right that it still does not work, but the remaining blocker is not in this package, and this PR does fix the interop under Node ESM. Evidence, all on Node 22 with `"type": "module"`:

**1. The interop itself is fixed under Node ESM.** Taking the built `dist/index.js` from this branch and the published 24.0.2, stubbing only the `tinacms` import (see point 2 for why), and constructing the provider:

```
published 24.0.2 -> TypeError: CredentialsProvider is not a function
this branch      -> {"providerId":"credentials","providerName":"TinaCredentials",
                     "providerType":"credentials","defaultProviders":1,"authHandler":"function"}
```

And directly, with no bundler involved:

```
$ node --input-type=module -e "
import NA from 'next-auth';
import C from 'next-auth/providers/credentials';
console.log('raw      :', typeof NA, typeof C);
console.log('unwrapped:', typeof (NA.default ?? NA), typeof (C.default ?? C));"
raw      : object object
unwrapped: function function
```

**2. What actually blocks plain Node ESM is `tinacms`, not `tinacms-authjs`.** `import('tinacms-authjs')` never reaches our code:

```
$ node --input-type=module -e "import('tinacms-authjs').catch(e => console.log(e.message))"
Directory import '.../node_modules/@heroicons/react/solid' is not supported resolving ES modules
imported from .../node_modules/tinacms/dist/index.js
Did you mean to import "@heroicons/react/solid/index.js"?
```

`import('tinacms')` on its own fails identically with `ERR_UNSUPPORTED_DIR_IMPORT`. `tinacms/dist/index.js` imports `@heroicons/react/solid` as a bare directory, which webpack and bundlers resolve but Node's ESM loader will not. That is a separate bug in a different package, and fixing it here is not possible.

So the split is: this PR fixes the half that is `tinacms-authjs`'s to fix, and it is the half that unblocks every bundled backend today (Netlify, Vercel, and anything else going through esbuild, which is the actual reported failure in #7434). Fully unbundled Node ESM additionally needs the heroicons directory import in `tinacms` resolved. Happy to raise that as its own issue, or fold it in here if you would rather this PR own both, though it would then be touching an unrelated package.

Worth noting #7434 was filed from the esbuild/Netlify symptom, so if you want the issue to stay open until the `tinacms` side is also fixed, I can drop the `Closes` from this PR and link it instead.

18-th:
Thanks, the typed helper is what I was after. I emitted declarations from d4b10d7 and `TinaCredentialsProvider` is back to `CredentialsConfig<...>` rather than `any`.

Your heroicons finding is real, and so is the one I hit. Published `tinacms@3.11.0` has a named import from `color-string` at `dist/index.js:48` and `import "@heroicons/react/solid"` at line 82. You saw heroicons first because a directory import fails during resolution, before the linking step where color-string fails. My local build has no heroicons import at all, which is why I got the other one.

I wasn't asking you to fix `tinacms` though, only to stop importing it. `src/index.ts:5` pulls the whole browser package into the node entry to get one string constant. Move that constant into its own file and the node build never touches `tinacms`, so neither of its bugs is reachable.

I tried it against d4b10d7 and it's five lines: a one-line `src/constants.ts`, one changed import in `index.ts`, and `tinacms.ts` importing and re-exporting it so nothing changes for consumers. The node bundle drops to three imports, all `next-auth`, and plain Node ESM constructs the provider. It also takes a bundled function from 16.8 MB to 840 KB, which is worth having on its own.

It's a workaround rather than a real fix, since those two `tinacms` imports still need sorting out for anyone importing that package directly. But it does mean this PR can honestly claim plain Node ESM.

So: fold it in here, or drop the `Closes` and link the issue instead?

*— Ivan's clanker*

kulesy:
Folded in as e09344b, thanks, this is a much better angle than fixing `tinacms` from the outside.

Confirmed your `color-string` finding: published 3.11.0 has the named import at `dist/index.js:48` and `import "@heroicons/react/solid"` at line 82, and heroicons wins the race because a directory import fails at resolution before linking.

The node entry is now three imports, all `next-auth`. Plain Node ESM constructs the provider with nothing stubbed, and the minimal bundled function goes 19,237,168 -> 842,617 bytes, close to your 840 KB. The Netlify path is unchanged: login 200, session with `role: "user"`, `/api/tina/gql` 400 with a cookie and 401 without.

Agreed it's a workaround rather than a real fix, since those two `tinacms` imports still bite anyone importing that package directly. Happy to raise that separately. I've left `Closes #7434` on since the reported symptom is now fixed on every backend, but say the word and I'll swap it for a link.

18-th (dismissed):
LGTM

18-th (approved):
LGTM

PR #7442, opened 2026-08-17
fix(deps): override qs to ^6.15.2 to clear three advisories

### LINKED ISSUE #7441: Four vulnerable qs copies in the lockfile, only visible when a PR adds a new version string
**Pain**

The lockfile carries four separate `qs` copies (6.5.3, 6.5.5, 6.11.0, 6.14.0), and every one is below the 6.14.1 patch line for GHSA-6rw7-vpxm-498p (moderate, arrayLimit bypass allowing DoS via memory exhaustion). Two further advisories apply: GHSA-w7fw-mjwx-w883 (patched 6.14.2) and GHSA-q8mj-m7cp-5q26 (patched 6.15.2).

`main` reports green because `dependency-review` reports what a diff *introduces*, not what is already present. It only surfaced when Dependabot regenerated lockfiles and pulled qs 6.5.3 up to 6.5.5, adding a version string the base did not have. Three unrelated PRs (#7425 js-yaml, #7421 ip-address, #7321 next-auth) then failed the same check for a vulnerability that has been sitting on `main` all along.

Worth noting the trap: pinning qs back to 6.5.3 would turn the check green while fixing nothing, since 6.5.3 is equally affected.

The chain is test-only: `vitest@0.32.4` -> `jsdom@15.2.1` -> `request-promise-native@1.0.9` -> `request@2.88.2` -> `qs`. Both `request` and `request-promise-native` are deprecated upstream, and jsdom 15 is from 2019.

**Approach**

Short term, override `qs` to `^6.15.2` so all copies resolve past every open advisory. Longer term, retire `vitest@0.32.4`: the repo already runs `vitest@2.1.9` elsewhere, and it is the old copy alone that drags in jsdom 15 and the deprecated `request` chain.

**Acceptance Criteria**

- [ ] Only one `qs` resolution remains in `pnpm-lock.yaml`, at 6.15.2 or later
- [ ] `dependency-review` passes on #7425, #7421 and #7321 without further changes
- [ ] A decision is recorded on retiring `vitest@0.32.4` and the deprecated `request` chain

kulesy:
Closes #7441

**TL;DR** One line in `pnpm-workspace.yaml` collapses four vulnerable `qs` copies into one patched version, and unblocks three Dependabot PRs as a side effect.

**Pain:** The lockfile carried four separate `qs` copies (6.5.3, 6.5.5, 6.11.0, 6.14.0), every one below the 6.14.1 patch line for GHSA-6rw7-vpxm-498p, with two further advisories applying up to 6.15.2. `main` looked green only because `dependency-review` reports what a diff introduces rather than what is already present, so the problem stayed invisible until Dependabot regenerated some lockfiles and pulled qs 6.5.3 up to 6.5.5. Three unrelated PRs (#7425, #7421, #7321) then failed the same check for a vulnerability that had been on `main` all along. The tempting fix, pinning qs back to 6.5.3, would turn the check green while fixing nothing, because 6.5.3 is equally affected.

**Solution:** Added `qs: ^6.15.2` to the existing `overrides:` block, which collapses all four copies to a single 6.15.3, past the patch line for all three open advisories. The affected chain is test-only (`vitest@0.32.4` to `jsdom@15.2.1` to `request-promise-native` to `request` to `qs`), so nothing user-facing changes. Retiring `vitest@0.32.4` and the deprecated `request` chain is the real fix and is tracked separately on #7441.

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -656,6 +656,7 @@ catalogs:
       version: 3.25.76
 
 overrides:
+  qs: ^6.15.2
   '@clerk/clerk-js>react': 19.2.3
   '@clerk/clerk-js>react-dom': 19.2.3
   '@clerk/backend>react': 19.2.3
@@ -14043,16 +14044,8 @@ packages:
     peerDependencies:
       react: ^16.8.0 || ^17.0.0 || ^18.0.0
 
-  qs@6.11.0:
-    resolution: {integrity: sha512-MvjoMCJwEarSbUYk5O+nmoSzSutSsTwF85zcHPQ9OrlFoZOYIjaqBAJIqIXjptyD5vThxGq52Xu/MaJzRkIk4Q==}
-    engines: {node: '>=0.6'}
-
-  qs@6.14.0:
-    resolution: {integrity: sha512-YWWTjgABSKcvs/nWBi9PycY/JiPJqOD4JA6o9Sej2AtvSGarXxKC3OQSk4pAarbdQlKAh5D4FCQkJNkW+GAn3w==}
-    engines: {node: '>=0.6'}
-
-  qs@6.5.3:
-    resolution: {integrity: sha512-qxXIEh4pCGfHICj1mAJQ2/2XVZkjCDTcEgfoSQxc/fYivUZxTkk7L3bDBJSoNrEzXI17oUO5Dp07ktqE5KzczA==}
+  qs@6.15.3:
+    resolution: {integrity: sha512-O9gl3zCl5h5blw1KGUzQKhA5oUXSl8rwUIM5o0S3nCXMliSvy5Dzx7/DJcI+SwgICv+IneSZwhBh1oSyEHA71A==}
     engines: {node: '>=0.6'}
 
   quansync@0.2.11:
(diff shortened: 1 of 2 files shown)


kulesy:
Verified locally before pushing:

| Check | Result |
|---|---|
| `qs` resolutions | 4 copies (6.5.3, 6.5.5, 6.11.0, 6.14.0) collapse to a single **6.15.3** |
| `pnpm install --frozen-lockfile` | passes, exit 0, including the supply-chain policy check |
| `pnpm run build` | exit 0, 25/25 turbo tasks successful |
| `pnpm run test` | exit 0, 66/66 turbo tasks, **2345 tests passed, 0 failed** |

The test run is the one that matters, since `qs` is only reachable through `vitest@0.32.4` -> `jsdom@15.2.1` -> `request@2.88.2`. Forcing qs from the 6.5.x line to 6.15.3 does not disturb `request`.

Also confirmed the regenerated lockfile does not pull in `@codemirror/view@6.43.9`, which was tripping the `minimumReleaseAge` gate on other branches earlier today.

Advisories cleared: GHSA-6rw7-vpxm-498p (moderate, patched 6.14.1), GHSA-w7fw-mjwx-w883 (low, patched 6.14.2), GHSA-q8mj-m7cp-5q26 (moderate, patched 6.15.2).

KahaMason (approved):
LGTM 🚀

PR #7445, opened 2026-08-18
fix(tinacms): skip TinaCloud identity requests when logged out

### LINKED ISSUE #7444: Logged-out /admin floods the console with misleading 401/CORS errors from identity.tinajs.io
### The exact error message

`Access to fetch at 'https://identity.tinajs.io/v2/apps/<clientId>/currentUser' from origin '<site>' has been blocked by CORS policy: No 'Access-Control-Allow-Origin' header is present on the requested resource.`

### Steps to reproduce

1. Open the /admin of any TinaCloud-connected site while logged out
2. Open the browser console before clicking log in

Expected: no failed requests before a login is attempted, or a clear "not logged in" message. Actual: the editor fires `currentUser` and project requests with no Authorization header; they 401 without CORS headers and the console fills with "Blocked by CORS policy" errors that look like a broken Site URL config. Support threads regularly chase this red herring while the real login failure is only visible in the login popup's own console.

### Your environment

Any TinaCloud-connected site, all browsers. Not reproducible in local mode.

kulesy:
Closes #7444

**TL;DR** Stop logged-out /admin loads from firing anonymous TinaCloud identity requests that always 401 and surface as CORS errors.

**Pain:** Every logged-out visit to a TinaCloud-connected /admin fires currentUser and project requests with no Authorization header. They 401 at the gateway without CORS headers, so the console fills with "Blocked by CORS policy" errors that look like a broken Site URL config. Support threads regularly start from this red herring while the real login failure sits unseen in the login popup's own console.

**Solution:** getUser() and the editorial-workflow getProject() call return early when no token is stored, so a logged-out load no longer fires the failing identity requests. The session-check failure log now includes the response status and points at the login popup console, GetDocument's auth guard is actually awaited, and the previously unhandled getProject() rejection is caught. A CORS error in the console now genuinely indicates an origin/Site URL mismatch.

### DIFF
--- a/.changeset/lazy-donkeys-shake.md
+++ b/.changeset/lazy-donkeys-shake.md
@@ -1,0 +1,5 @@
+---
+'tinacms': patch
+---
+
+Skip TinaCloud identity requests when no auth token is stored. Logged-out admin loads no longer produce misleading 401/CORS console errors; a clear console message now points at the login popup console instead. Also fixes an unawaited auth guard in GetDocument and an unhandled promise rejection when the project settings request fails.
--- a/.changeset/lazy-donkeys-shake.md
+++ b/.changeset/lazy-donkeys-shake.md
@@ -2,4 +2,4 @@
 'tinacms': patch
 ---
 
-Skip TinaCloud identity requests when no auth token is stored. Logged-out admin loads no longer produce misleading 401/CORS console errors; a clear console message now points at the login popup console instead. Also fixes an unawaited auth guard in GetDocument and an unhandled promise rejection when the project settings request fails.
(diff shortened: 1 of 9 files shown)


wicksipedia:
Approach lgtm. Pushed three commits.

The `isAuthenticated()` check in `GetDocument` was never false before, because an unawaited promise is always truthy, so the fetch always ran. Awaiting it turns it into a real check, and nobody had finished the false path: it skips the fetch and never calls `setLoading(false)`, so the document view spins until you reload. `ef160b4b2` settles the loading state on every outcome, with a test that fails on your previous commit.

`ec58bba05` pulls the `access_token ?? id_token` rule into `AbstractAuthProvider.getAccessToken()` so `getUser()` can go back through `fetchWithToken` instead of building its own header. The third copy in `TinaCloudProvider.tsx:529` stays put, since putting the method on the public `AuthProvider` interface would break hand-written providers.

`974579ac4` is the same hang in `useGetCollection` and `useSearchCollection`. Not yours, it is on main today: enter edit mode, `localStorage.removeItem('tina.local.isLogedIn')`, click a collection. Separate changeset for it.

608 tests pass. The `getProject()` skip has no test, it needs a live client ID and a full provider render.

wicksipedia (approved):

PR #7447, opened 2026-08-18
fix(deps): move @tinacms/cli off the vulnerable js-yaml 4.1.1

### LINKED ISSUE #7446: @tinacms/cli ships a vulnerable js-yaml 4.1.1 (GHSA-5p4m-2wfm-xmqj)
**Pain**

`@tinacms/cli` on `main` resolves `js-yaml` to **4.1.1**, which is inside the vulnerable range of GHSA-5p4m-2wfm-xmqj (high). It is a runtime `dependencies` entry and the package is published, so `npm install @tinacms/cli` currently pulls a vulnerable copy.

The advisory covers both major lines, which is what makes this easy to miss:

| Range | First patched |
|---|---|
| `>= 3.0.0, < 3.15.1` | 3.15.1 |
| `>= 4.0.0, < 4.3.1` | 4.3.1 |

#7425 closed the 3.x side by moving `@tinacms/graphql` to 3.15.1, so the repo looks addressed at a glance. The 4.x side was left open. Dependabot had originally proposed rewriting the cli specifier to `^3.15.1`, which would have been a cross-major downgrade rather than a fix, and correcting that downgrade is what left the cli on 4.1.1.

**Approach**

Set `@tinacms/cli` to `^4.3.1`, the patch on its own major line. After that no workspace package sits on a vulnerable js-yaml. A transitive 4.1.1 remains via `astro`, `cosmiconfig` and `eslint`, which is upstream of this repo and not fixable here.

Worth reviewing separately whether a security advisory spanning two major lines needs a check that each consumer moves to the patch on *its* line, since neither CI nor `dependency-review` flagged the 4.x gap.

**Acceptance Criteria**

- [ ] `@tinacms/cli` resolves js-yaml to 4.3.1 or later
- [ ] No workspace package resolves js-yaml inside `< 3.15.1` or `>= 4.0.0 < 4.3.1`
- [ ] A published `@tinacms/cli` no longer carries the advisory

kulesy:
Closes #7446

**TL;DR** `@tinacms/cli` currently ships a vulnerable js-yaml. One specifier change moves it to the patch on its own major line.

**Pain:** GHSA-5p4m-2wfm-xmqj (high) covers both major lines, `>= 3.0.0 < 3.15.1` patched by 3.15.1 and `>= 4.0.0 < 4.3.1` patched by 4.3.1. #7425 closed the 3.x side by moving `@tinacms/graphql` to 3.15.1, which makes the repo look addressed, but `@tinacms/cli` was left on 4.1.1, inside the vulnerable 4.x range. js-yaml is a runtime `dependencies` entry there and the package is published, so `npm install @tinacms/cli` pulls a vulnerable copy today. Neither CI nor `dependency-review` flagged it, since the 4.x version was already present on `main` rather than introduced by a diff.

**Solution:** Sets the specifier to `^4.3.1`, resolving 4.3.1. Afterwards no workspace package sits on a vulnerable js-yaml. A transitive 4.1.1 remains via `astro`, `cosmiconfig` and `eslint`, which is upstream of this repo and out of scope here.

### DIFF
--- a/packages/@tinacms/cli/package.json
+++ b/packages/@tinacms/cli/package.json
@@ -91,7 +91,7 @@
         "esbuild": "catalog:",
         "fs-extra": "catalog:",
         "graphql": "15.8.0",
-        "js-yaml": "^4.1.0",
+        "js-yaml": "^4.3.1",
         "many-level": "catalog:",
         "memory-level": "catalog:",
         "minimatch": "catalog:",
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -1438,8 +1438,8 @@ importers:
         specifier: 15.8.0
         version: 15.8.0
       js-yaml:
-        specifier: ^4.1.0
-        version: 4.1.1
+        specifier: ^4.3.1
+        version: 4.3.1
       many-level:
         specifier: 'catalog:'
         version: 2.0.0
@@ -12172,6 +12172,10 @@ packages:
 
   js-yaml@4.1.1:
     resolution: {integrity: sha512-qQKT4zQxXl8lLwBtHMWwaTcGfFOZviOJet3Oy/xmGk2gZH677CJM9EvtfdSkgWcATZhj/55JZ0rmy3myCT5lsA==}
+    hasBin: true
+
+  js-yaml@4.3.1:
+    resolution: {integrity: sha512-CY6crGq313MX8GkwvB7tzgp99vjQxY1++5y10/BKN/GUfHqWaOGQMNZkBvqSzsZKWk/ijwHlWzzkLulsGHhjWQ==}
     hasBin: true
 
   jsbn@0.1.1:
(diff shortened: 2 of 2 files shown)


kulesy:
Verified locally:

| Check | Result |
|---|---|
| `@tinacms/cli` resolution | `^4.1.0` → 4.1.1 becomes `^4.3.1` → **4.3.1** |
| Workspace packages on a vulnerable js-yaml | **none** (was 1) |
| `pnpm install --frozen-lockfile` | passes, including the supply-chain policy check |
| `pnpm run build` | exit 0, 25/25 turbo tasks |
| Lockfile scope | one entry added (`js-yaml@4.3.1`), nothing removed |

For transparency on how this was missed: Dependabot originally proposed rewriting the cli specifier from `^4.1.0` to `^3.15.1` on #7425. That is a cross-major downgrade rather than a fix, so it was reverted to `^4.1.0` before merge, on the incorrect assumption that 4.1.1 was already safe. It is not, because the same advisory covers `>= 4.0.0 < 4.3.1`. This PR applies the correct target.

The remaining transitive `js-yaml@4.1.1` comes from `astro`, `cosmiconfig` and `eslint`. Those are upstream and cannot be resolved from this repo without an override; raising that separately is probably the right call if it matters.

isaaclombardssw (approved):
Do we need a changeset for this? Change itself looks good

Issue #7415, opened 2026-08-06
🐛 Hard breaks are silently dropped or corrupted by mdast-util-to-markdown (root cause for #5426, #6555, #7408)


isaaclombardssw:
Root cause for the long-running soft/hard return bugs is a single upstream handler, `mdast-util-to-markdown`'s `hardBreak`. It produces two distinct, opposite failure modes, which is why the last three fix attempts each fixed one and regressed the other.

Filing this as a consolidated root-cause writeup for #5426, #6555 and #7408, and as the blocker analysis for the two stalled PRs (#5451 closed, #6766 open).

### The exact error message

Two of the three failure modes are **silent** — no error, wrong output. That is the core of the problem.

The third throws, on reading back markdown the serializer itself just wrote:

```
StaticPhrasingContent: break is not yet supported
```

Thrown from `packages/@tinacms/mdx/src/parse/remarkToPlate.ts:358`, which collapses the entire rich-text field to an `invalid_markdown` blob.

### Steps to reproduce

1. Any `rich-text` field, any schema, either `parser: { type: 'markdown' }` or `'mdx'`.
2. Produce a Plate tree containing a `{ type: 'break' }` element in each of the positions below.
3. Run it through `serializeMDX`, then `parseMDX` on the result.

Three positions fail:

- Inside an `h3`-`h6` heading
- As the last child of a block
- Inside a link's children

### What you expected vs. what actually happened

Verified against the real serializer and parser. `KEPT stable` means the break survives a `stringify -> parse -> stringify` round trip unchanged.

| Container | Result | Markdown emitted |
|---|---|---|
| paragraph | KEPT stable | `one\` + newline + `two` |
| h1, h2 | KEPT stable | rewrites heading to setext form |
| h3 - h6 | **LOST** | `### one two` |
| blockquote | KEPT stable | `> one\` |
| list item | KEPT stable | `* one\` |
| bold / marked text | KEPT stable | `**one**\` |
| block-initial break | KEPT stable | `\` + newline + `one` |
| consecutive breaks | KEPT stable | `one\` + `\` |
| **block-final break** | **LOST, UNSTABLE** | `one\` -> reparses as literal text `one\` |
| **inside a link** | **PARSE THROWS** | `[a\` + newline + `b](/x)` |
| table cell | LOST | collapses to a space |
| code block | n/a | wants a literal newline |

### Your environment

- Reproduced on `main` at `ece366bd5`, TinaCMS monorepo.
[...]

PR #5504, opened 2025-02-18
Feat - Add strikethrough

Ben0189:
Add strikethrough feature into RTE (from @JackDevAU )
Add test case for strikethrough

PR #7449, opened 2026-08-19
🐛 Parse a hard break, and strikethrough, inside a link (#7415)

isaaclombardssw:
Part 2 of 5 against #7415. Stacked on #7448.

## What breaks

| Do this in the editor | Saves as | On reopen |
|---|---|---|
| Put the cursor inside a link, Shift+Enter, save | `[a\`⏎`b](/x)` — correct markdown | **whole field** collapses to an `invalid_markdown` blob |
| Strikethrough some text inside a link, save | `[~~a~~](/x)` — correct markdown | **whole field** collapses |

Neither needs the editor at all — both are valid CommonMark, so a hand-authored file does it too. The markdown written is fine; it is reading it back that throws.

## The fix

Link children go through `staticPhrasingContent` in `parse/remarkToPlate.ts`, a narrower switch than the `phrasingContent` one used everywhere else, and anything missing from it throws — taking the whole field with it, not just the link. It was missing `break`. It was also missing `delete`, one line away, with the identical failure.

`delete` has been missing since strikethrough shipped: #5504 (`754161452`) added `case 'delete'` to `phrasingContent` and not to its sibling. So strikethrough has worked everywhere except inside a link ever since, and there it threw rather than degrading.

That divergence is the real defect — two near-duplicate switches where one gets each new case and the other silently does not. `staticPhrasingContent` will keep collecting these until it is derived from `phrasingContent` rather than hand-maintained beside it.

#7415 only names the `break` case. Fixing that alone would leave the same crash in the same switch, so the test covers every mark that can appear in link children, through both parsers, asserting the field parses and the markdown reaches a fixed point.

Both parse paths converge on `remarkToSlate`, so one case covers `markdown` and `mdx` alike.

## Evidence

The matrix from #7448 flips only the `a` rows:

```
-a (inline)  mid  "[one\\\ntwo](/x)\n"  1  0  INVALID UNSTABLE RESAVE THREW
+a (inline)  mid  "[one\\\ntwo](/x)\n"  1  1  parses
```

`RESAVE THREW` disappearing is a side effect: it was only set because the field had already collapsed, and the `markdown` parser has no `invalid_markdown` case on the way back out. That gap still exists for any unparseable content — it is just no longer reachable this way.

## Testing

`pnpm --filter @tinacms/mdx test` — 204 passing. Use `pnpm test`, not `npx vitest` (this package pins vitest `^0.32.4`, its siblings `^2.1.9`).


joshbermanssw (dismissed):

joshbermanssw (dismissed):

joshbermanssw (dismissed):

KahaMason (approved):
Re-approving after rebase from main

Issue #5426, opened 2025-01-14
🐛 TinaCMS | Hard returns can render as `\`


wicksipedia:

kldavis4:
@JackDevAU I believe this [PR](https://github.com/tinacms/tinacms/pull/5451) will fix the issue

ncn-ssw:
Next PR - https://github.com/tinacms/tinacms/pull/5451

ncn-ssw:
@JackDevAU has recommended that this is delayed until the completion of https://github.com/tinacms/tinacms/issues/5257 due to the difficulties encountered with fixing the problem without causing other bugs.

gustxyz:
Any update on this?  @ncn-ssw

isaaclombardssw:
Closing in favour of #7415, which documents the shared root cause for this and the other reports of the same behaviour.

Both symptoms (a hard break silently degrading to a space, and a block-final break writing a dangling `\`) come from the same handler: `mdast-util-to-markdown`'s `hardBreak`. #7415 has the verified representability matrix showing exactly which containers can and cannot hold a hard break, why the previous attempts (#5451, #6766) each stalled, and what any fix has to cover.

Not fixed yet — tracking continues in #7415.

PR #7450, opened 2026-08-19
🐛 Stop writing a hard break markdown cannot represent (#5426)

isaaclombardssw:
Part 3 of 5 against #7415, and the fix for #5426 (open since January 2025). Stacked on #7449.

## What breaks

Every row here writes a character the author never typed.

| Do this in the editor | Saves as | Next save |
|---|---|---|
| Shift+Enter at the end of a paragraph | `one\` | `one\\` — backslash is now content |
| Shift+Enter twice at the end | `one\` | `one\\` |
| Shift+Enter at the end of a blockquote or list item | `> one\` / `* one\` | same |
| Shift+Enter with the cursor just before inline HTML | `one\ <em>x</em>` — literal backslash | stable at the corrupted form |
| Shift+Enter just before an inline template | `one\`⏎`<DateTime />` — embed promoted to its own block, paragraph torn in two | `one\\` |
| **Open and save a file containing** `one\`⏎`<em>x</em>` **— no edit at all** | `one\ <em>x</em>` | stable at the corrupted form |

That last row is the one that needs no keystroke.

## The fix

Two positions have no representable form.

**Nothing after it.** A trailing backslash is a hard break in CommonMark only when another line follows it in the same block. With nothing after it, `one\` reads back as a literal backslash — the break is lost and the backslash enters the content as data.

**A line-start-sensitive neighbour after it.** A break puts its neighbour at the start of a line. `toMarkdown` will not let `<` open a line, so it rewrote the break as `\` plus a space; and an inline element at line start parses as flow, so it is promoted out of the paragraph — or writes nothing at all when no template matches it, leaving the `\` dangling.

Both are handled by a walk over the mdast tree before `toMarkdown`, called by the `markdown` and `mdx` serializers alike. Representability here is *positional*, not type-based, which is why no list of container types could express it. A break before raw HTML becomes a **space** (HTML always writes, so the word separation is worth keeping); before an inline element it is **dropped** (an element that writes nothing would leave only trailing whitespace).

Three shapes the walk has to get right, each of which was wrong in an earlier revision and each of which is now a test: the break is never the last child (Slate keeps a spacer); a repeated Shift+Enter emits two; the recursive descent must skip the spacer to reach a break at the end of a link.

**Trade-off worth naming:** in the HTML case the author's line break becomes a space. That is a downgrade, chosen over a literal backslash in their prose. Preserving it means patching upstream's `unsafe` table to let `<` open a line, which risks `<` starting lines everywhere else.

## Testing

`pnpm --filter @tinacms/mdx test` — 228 passing. `break-neighbours.test.ts` asserts on verbatim `editor.children`, not hand-built trees. Use `pnpm test`, not `npx vitest` (this package pins vitest `^0.32.4`, its siblings `^2.1.9`).

### DIFF
--- a/packages/@tinacms/mdx/src/dangling-breaks.ts
+++ b/packages/@tinacms/mdx/src/dangling-breaks.ts
@@ -1,0 +1,116 @@
+import type * as Md from 'mdast';
+
+/**
+ * A trailing backslash is a hard break in CommonMark only when another line
+ * follows it in the same block. With nothing after it, `one\` reads back as a
+ * literal backslash, so the break is lost and a character the author never
+ * typed enters the content as data (#5426). Reshapes in place.
+ */
+export const dropDanglingBreaks = <T extends Md.Root>(tree: T): T => {
+  walk(tree);
+  return tree;
+};
+
+/** Blockquotes and list items hold paragraphs, which the walk reaches anyway. */
+const BLOCKS = new Set(['paragraph', 'heading', 'tableCell']);
+
+type Parent = { type?: string; children: unknown[] };
+
+const isParent = (node: unknown): node is Parent =>
+  Array.isArray((node as Parent | undefined)?.children);
+
+const walk = (node: Parent) => {
+  // What sits after a break is a local question, so it holds at any inline
+  // depth. Only "trailing" needs the block gate — `**one\** two` is not dangling.
+  dropBreaksBeforeLineStartSensitive(node.children);
+  if (BLOCKS.has(node.type ?? '')) {
+    trimTrailingBreaks(node);
+  }
+  for (const child of node.children) {
+    if (isParent(child)) {
+      walk(child);
+    }
+  }
+};
+
+/**
+ * A break puts whatever follows it at the start of a line, and these two read
+ * differently there:
+ *
+ * - raw HTML — `toMarkdown` will not let `<` open a line, so it rewrites the
+ *   break to `\` plus a space, which reads back as a literal backslash;
+ * - an inline MDX element — at line start it parses as flow, so the element is
+ *   promoted out of the paragraph, or writes nothing at all when no template
+ *   matches it, leaving the `\` dangling.
+ *
+ * Either way the break has no representable form in front of one, so drop it
+ * rather than write a character the author never typed.
+ */
+const dropBreaksBeforeLineStartSensitive = (children: unknown[]) => {
+  for (let index = children.length - 1; index >= 0; index--) {
+    if (!isBreak(children[index])) {
+      continue;
+    }
+    const next = children
+      .slice(index + 1)
+      .find((sibling) => !writesNothing(sibling));
+    const type = (next as Parent | undefined)?.type;
+    if (type === 'html') {
+      // Raw HTML always writes, so keep the word separation the break gave —
+      // unless what precedes it already ends in whitespace, where a second
+      // space would only churn the author's file.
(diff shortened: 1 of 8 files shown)


joshbermanssw (commented):
Reviewed part 3. Logic is sound — the two unrepresentable positions are real and the space-vs-drop split is a considered call. One scoping gap below.

packages/@tinacms/mdx/src/dangling-breaks.ts
```diff
@@ -0,0 +1,114 @@
...
+type Parent = { type?: string; children: unknown[] };
+
+const isParent = (node: unknown): node is Parent =>
+  Array.isArray((node as Parent | undefined)?.children);
+
+const walk = (node: Parent) => {
+  if (BLOCKS.has(node.type ?? '')) {
+    dropBreaksBeforeLineStartSensitive(node.children);
+    trimTrailingBreaks(node);
+  }
```
joshbermanssw, line 28: The neighbour pass only sees the direct children of `paragraph`/`heading`/`tableCell`, so a break before inline HTML nested inside a link keeps the backslash this PR exists to remove:

```
save 1: "[one\⏎<em>x</em>](/x)"
save 2: "[one\ <em>x</em>](/x)"   ← literal backslash, then stable at the corrupted form
```

`trimTrailingBreaks` recurses through the last writing child, but `dropBreaksBeforeLineStartSensitive` doesn't recurse at all.

"What sits after this break" is a purely local question, so it's valid at any inline depth — links, `strong`, `emphasis`, `delete`. Only `trimTrailingBreaks` needs the block gate, since "trailing" is meaningless except relative to a whole block (`**one\** two` isn't dangling).

Applied locally: the link case becomes `[one <em>x</em>](/x)` — no backslash, stable — and all 228 tests still pass with no snapshot churn.

```suggestion
const walk = (node: Parent) => {
  dropBreaksBeforeLineStartSensitive(node.children);
  if (BLOCKS.has(node.type ?? '')) {
    trimTrailingBreaks(node);
  }
```

isaaclombardssw: Good catch — reproduced it exactly as you described, `[one\ <em>x</em>](/x)` and stable at the corrupted form.

Applied your suggestion. Your reasoning is the right cut: "what sits after this break" is local so it holds at any depth, while "trailing" only means anything relative to a whole block. The block gate now guards `trimTrailingBreaks` alone.

Added a regression test for the nested case in `break-neighbours.test.ts`, and confirmed a real break inside a link still survives (`[one\\⏎two](/x)`). 229 passing.

joshbermanssw (dismissed):
Scoping gap is closed, and the regression test pins it.

Swept the nested containers against the updated branch — all backslash-free and stable across a re-save:

| Case | Writes |
|---|---|
| break → html in a link | `[one <em>x</em>](/x)` |
| break → html in bold | `**one** <em>x</em>` |
| …in a link in a blockquote | `> [one <em>x</em>](/x)` |
| …in a link in a list item | `* [one <em>x</em>](/x)` |
| …in a table cell | `\| one <em>x</em> \|` |

Real breaks are untouched, and a trailing break inside a link mid-paragraph correctly keeps its backslash, since that position can represent one.

229 passing, Biome clean. The HTML-becomes-a-space downgrade is the right trade against writing a character the author never typed.

KahaMason (approved):
re-approving after trying to resolve the PR stack

PR #7451, opened 2026-08-19
🐛 Split an h3–h6 on a hard break instead of losing it (#7415)

isaaclombardssw:
Part 4 of 5 against #7415. Stacked on #7450.

This one carries a product decision, not just a technical fix.

## What breaks

| Do this in the editor | Saves as | Should be |
|---|---|---|
| Shift+Enter mid-`h3`–`h6` | `### one two` — break silently gone | break preserved somehow |
| Shift+Enter inside a link inside an `h3` | `### [one two](/x)` — break silently gone | same |
| Shift+Enter at the end of an `h3` | `### one`⏎⏎`###` — a bare heading appears | nothing extra |
| Shift+Enter mid-`h1`/`h2` | setext rewrite, break survives | **unchanged — see below** |

## The fix

`###` and deeper have no setext form, so `mdast-util-to-markdown` degrades the break to a space and the line break vanishes with no error. A break in an `h3`–`h6` now splits it into headings of the same level:

```
### one⏎two   ->   ### one
                   
                   ### two
```

Content survives, the line break survives as a visual break, the file keeps its ATX style, and the second save is a fixed point. What the author loses is that it is now two headings — two TOC entries, two `<h3>` elements, and reopening shows two heading blocks. That is the trade: a visible structural change instead of silent content loss.

**Known wart, worth a decision:** the split happens at save time, so the editor still shows one heading with a `<br>` until reload. Mirroring it in a Plate normalizer would fix that and is not in this PR.

## `#` and `##` are deliberately left alone

They *do* have a setext form, so the break already survives as a real break in one heading — and that is already `main`'s behaviour, it just produced a soft wrap that rendered as a space. The stack adds the `\` that makes it work. Splitting them would be new behaviour replacing something about to start working correctly, so the split starts at depth 3. Pinned from both sides: `it.each([1, 2])` asserts h1/h2 keep their break in one node.

## Two ordered passes, and two guards

Trim the whole tree first, then split. One walk doing both split a heading that still held a trailing break, and the second segment then held only what followed it — an empty copy of the link, written as a visible `### [](/x)`.

The split is recursive, because the editor puts the break *inside* the link when the cursor is in link text. And it needs two guards, each of which was a real content-loss bug before it existed:

- **A childless parent is not empty.** `[].every()` is vacuously true, so "do all children write nothing?" answers yes for a node with none — and a self-closing embed reaches mdast with exactly that shape. `### one<br /><Cta url="/x" />` deleted the embed.
- **An emptied node must still prune**, or a break ending a link writes a visible `[](/x)` nobody typed.

## Testing

`pnpm --filter @tinacms/mdx test` — 238 passing. Use `pnpm test`, not `npx vitest` (this package pins vitest `^0.32.4`, its siblings `^2.1.9`).


joshbermanssw (dismissed):
Swept the edge cases the description names, all fixed points:

| Case | Writes | Back |
|---|---|---|
| `h3` mid | `### one` / `### two` | `[h3,h3]` |
| break inside a link in `h3` | `### [one](/x)` / `### [two](/x)` | `[h3,h3]`, no empty `[](/x)` |
| `h1`/`h2` mid | `one\` / `two` / `===` | `[h1]`, untouched |
| `h3` trailing break | `### one` | no bare `###` |
| `h3` double break | `### one` / `### two` | empty segment dropped |
| `h3` + self-closing embed | `### one<Cta url="/x" />` | embed survives |

239 passing, Biome clean, and the matrix flips exactly the `h3`-`h6` mid rows with `blocks` going 1 to 2. (`###` inside a blockquote throws, but identically on the base — pre-existing, not this PR.)

One thing for a follow-up rather than a blocker: in that last row the embed survives but the break is silently dropped, because `trimTree` removes a break before an inline element before `splitTree` can see it. Splitting is representable there — `### one` / `### <Cta url="/x" />` parses as `[h3,h3]` and is a fixed point — so the drop rule from part 3 may be too broad once the container is a heading. Happy for that to be its own issue.

The product trade (two headings, two TOC entries) and the save-time-only split are both called out in the description, so no surprises there.

KahaMason (approved):
re-approving after trying to resolve PR stack

PR #7452, opened 2026-08-19
🐛 Make shift+Enter produce a line break that survives a save (#6555, #7408)

### LINKED ISSUE #6555: Shift+Enter in rich-text editor serializes as \n (soft break) instead of \ (hard break), causing line breaks to render as spaces
**TinaCMS version:** 3.7.0  
 **Next.js:** 16.2.0 / React 19.2.0  
 **Field type:** `rich-text`
 
 ---
 
 ### Description
 
 Pressing `Shift+Enter` in the rich-text WYSIWYG editor does not produce a visible line break on the rendered page. According to the 
[docs](https://tina.io/docs/reference/types/rich-text#line-breaks), Shift+Enter should produce a hard line break (`\` at end of line in markdown), which TinaMarkdown renders as a `{ 
type: "break" }` AST node.
 
 ### Steps to Reproduce
 
 1. Create a `rich-text` field (e.g. `body`)
 2. In the TinaCMS editor, type two lines and press `Shift+Enter` between them
 3. Save and inspect the stored JSON/markdown
 4. Observe the rendered page
 
 ### Expected Behaviour
 
 The markdown stored in the content file should contain a **hard break** (`\` at end of line):
 
 ```
 Line one\
 Line two
 ```
 
 (Note the trailing `\` after "Line one" — this is the CommonMark hard line break syntax.)
 
 TinaCloud's GraphQL parser should produce a `{ type: "break" }` AST node → TinaMarkdown renders `<br />` → visible line break on page. ✅
 
 ### Actual Behaviour
 
 The Slate editor serializes Shift+Enter as a plain `\n` (soft break):
 
 ```
 Line one\nLine two
 ```
 
 TinaCloud's GraphQL parser preserves the `\n` inside the text node value:
 
 ```json
 { "type": "text", "text": "Line one\nLine two" }
 ```
 
 No `{ type: "break" }` node is produced. TinaMarkdown renders the string as-is → HTML collapses `\n` to a space → **no visible line break**. ❌
 
 ### Workaround
 
 Pre-process the AST before passing to `<TinaMarkdown>` to split text nodes on `\n` and insert `{ type: "break" }` nodes manually:
 
 ```tsx
 function injectBreaks(node: any): any {
   if (!node || typeof node !== "object") return node;
   if (Array.isArray(node)) return node.map(injectBreaks);
   if (node.type === "text" && typeof node.text === "string" && node.text.includes("\n")) {
     const parts = node.text.split("\n");
     return parts.flatMap((part: string, i: number) => {
       const textNode = { ...node, text: part };
       return i < parts.length - 1 ? [textNode, { type: "break" }] : [textNode];
     });
   }
   if (node.children) {
     return { ...node, children: node.children.flatMap(injectBreaks) };
   }
   return node;
 }
 ```
 
 ### Additional Context
 
 - The docs at https://tina.io/docs/reference/types/rich-text#line-breaks claim Shift+Enter produces a `\` hard break — this does not match actual behaviour in v3.7.0
 - The bug is in the **Slate → markdown serializer**: it should write `\` + newline for soft break nodes but writes only `\n`
 - Confirmed via server-side AST logging that TinaCloud preserves `\n` in text node values (does not normalize to space), so the fix belongs in the serializer

### LINKED ISSUE #7408: ✨ Soft Returns - Support soft return behavior in standard markdown editor

isaaclombardssw:
Part 5 of 5 against #7415, and the last. Stacked on #7451.

**Supersedes #7412** — that is the first commit here, unchanged, plus the v4 port and a corrected changeset.

## What breaks

| Do this in the editor | Saves as | On the page |
|---|---|---|
| Shift+Enter in a paragraph | `one`⏎`two` — a soft wrap | `one two` — line break gone |
| Shift+Enter in a list item or table cell | same | same |
| Shift+Enter in an `h1`/`h2` | setext rewrite, still a soft wrap | `one two` |
| Shift+Enter in a blockquote | `> one\` — **correct** | `<br>` |
| Cmd+Enter inside a blockquote | swallowed — cannot exit the quote | — |

The blockquote row is why this looked so inconsistent: a separate plugin already inserted a `break` element there, so one keystroke took two different code paths.

## The fix

Plate's `SoftBreakPlugin` inserts a literal `\n`. In markdown a bare newline inside a paragraph is a soft wrap, so serialization re-flowed it into a space. `createHardBreakPlugin` inserts a `break` element instead, which serializes to a `\` hard break and round-trips as a `<br>`:

```md
123 Abc Street\
Town Central, CA\
90210
```

Two exceptions stay on `SoftBreakPlugin`: code blocks want a real newline, and GFM table cells cannot represent a hard break at all.

The handler also claims the event before it mutates. It used to delete an expanded selection and *then* return without `preventDefault` if the focus was missing, and `SoftBreakPlugin` only skips an event already handled — so it added a literal newline on top of the deletion.

`mod+Enter` / `mod+shift+Enter` inside a blockquote work again; the old handler `preventDefault`'d on any Enter, so `ExitBreakPlugin` never ran.

## Why this is safe now and was not in #7412

#7412 stayed a draft because inserting a `break` everywhere regressed three containers — it turned a recoverable `&#xA;` in an `h3` into deleted content, wrote a dangling `\` at the end of a block, and made links throw. All fixed earlier in this stack (#7449, #7450, #7451), with the matrix in #7448 as the check that they stay fixed. That is the whole reason for the stack order: this commit on its own is the change that has failed review three times.

## Two editors now

v4 shipped `packages/v4/@tinacms/rich-text` with its own copy of these plugins and an identical config. Ported there too, in the v4 comment register. Both suites pin the invariant the design rests on: this plugin must be registered before `SoftBreakPlugin`, because the only thing preventing a double insert is `preventDefault` reaching it first.

## Correction to #7412's changeset

It offered a literal `<br>` as the table-cell workaround. That is wrong — `TinaMarkdown` renders `html_inline` escaped, so `| c<br>c2 |` shows on the page as the text `<br>`. Verified end to end.

## Testing

`pnpm --filter @tinacms/mdx test` 238, `--filter @tinacms/rich-text` 186, `--filter tinacms` 615. Use `pnpm test`, not `npx vitest`.

## Closes

Fixes #6555, fixes #7408. Closes out #7415 with #7448–#7451.


joshbermanssw (dismissed):
All three suites match the numbers claimed: mdx 239, rich-text 186, tinacms 615. (They failed for me at first purely because workspace deps were not built — green after `turbo run build`.)

Verified:

- The payoff round-trips. `123 Abc Street` / `Town Central, CA` / `90210` writes real `\` hard breaks and is a fixed point.
- Table cells degrade to a space with content preserved and stable, which is the documented GFM limitation rather than silent loss.
- v3 and v4 handlers are byte-identical apart from one comment word, so the port has not drifted.
- The ordering invariant is pinned in both suites — `KEY_HARD_BREAK` must index before `SoftBreakPlugin`, and that is the only thing preventing a double insert.
- Changeset scoping is right: only `tinacms` needs the patch, since `@tinacms/rich-text` is `private: true`.

The `preventDefault`-before-mutate change is a good catch independent of the rest — the old handler deleted an expanded selection and then bailed, letting `SoftBreakPlugin` add a literal newline on top of the deletion.

Stack order is well justified: this commit alone would regress the three containers #7449 to #7451 just fixed, with the #7448 matrix as the check that they stay fixed.

KahaMason (approved):
re-approving after trying to resolve PR stack

PR #7459, opened 2026-08-19
ci: flag a PR that changes a published package's deps with no changeset

### LINKED ISSUE #7435: multer 2.2.0 bump (#7079) landed without a changeset

kulesy:
Closes #7435

**TL;DR** Six dependency PRs merged in four days with no changeset, so none of them ever reached npm. This adds a CI check that flags it.

**Pain:** A PR that changes a published package's dependencies without a changeset merges cleanly, produces no version bump, no changelog entry and no release, so the change never reaches npm users. Nothing catches it: CI is green, `dependency-review` only reports what a diff introduces, and review has missed it six times in four days (#7079, #7320, #7422, #7425, #7321, #7447). Two of those were security fixes, which meant the advisory stayed open for anyone installing the package while `main` looked fixed. A naive rule is not enough either, since plenty of manifest changes genuinely need no changeset.

**Solution:** Adds `.github/scripts/check-changeset.mjs` and a `require-changeset` workflow that resolves which *published* packages a PR actually affects, either through a changed manifest or through a moved `catalog:` entry that a published package consumes, and skips anything private or listed in the changesets `ignore` config. The check goes red only when such a package is affected and no `.changeset/*.md` is present. It is **not** in the required-status-checks list (`license/cla`, `build (macOS-latest)`, `build (ubuntu-latest)`), so a failure is a visible red X rather than a merge blocker, and it fails open when it cannot determine an answer. Dependabot is deliberately not exempt, since its PRs are the ones that keep merging unreleased; a `skip-changeset` label is the escape hatch for a change that genuinely ships nothing.


kulesy:
Validated against every real case from the last four days before opening, rather than only on synthetic input:

| PR | What it is | Expected | Result |
|---|---|---|---|
| #7425 | `@tinacms/graphql` js-yaml, no changeset | fail | fail |
| #7321 | `tinacms-authjs` next-auth, no changeset | fail | fail |
| #7422 | mermaid, has a changeset | pass | pass |
| #7438 | tar, has a changeset | pass | pass |
| #7402 | postcss, only `@examples/*` consume it | pass | pass |
| #7442 | qs override, no published manifest touched | pass | pass |

The last two are the ones that matter for false positives. A simple "a manifest changed, demand a changeset" rule fails both: #7402 changes the `catalog:` block but only private example packages consume postcss, and #7442 changes `pnpm-workspace.yaml` overrides without touching any package manifest. The script resolves the affected published packages instead of pattern-matching paths, so both correctly pass.

This PR also passes its own check, since it touches no package manifests.

Two things worth a reviewer's opinion:

1. **Dependabot is not exempt.** Its PRs will go red until someone adds a changeset. That is the intent, since four of the six misses were Dependabot PRs and each needed a hand-written changeset anyway, but it does mean more red on that queue. Exempting it is a one-line change if you would rather it warn.
2. **Catalog resolution is a heuristic.** It matches a published package to a moved catalog key via a `catalog:` specifier in its deps. It will not catch a package that consumes a catalog entry indirectly, so it is a floor rather than a proof.

kulesy:
**Demoed on this PR against real CI.** Three consecutive runs on the branch:

| Commit | What it contained | Result |
|---|---|---|
| `f19627c2` | the two new files, no manifest change | success |
| `d2016505` | a js-yaml bump in `@tinacms/graphql`, deliberately no changeset | **failure** |
| `e3fa9cbc` | revert of the above | success |

Checks tab

Figure: `Require Changeset` appears alongside the other checks on the PR. It is deliberately not in the required-checks list (`license/cla`, `build (macOS-latest)`, `build (ubuntu-latest)`), so a failure shows red without gating the merge button.

Failing run

Figure: the failing run on `d2016505`, which changed a published package's dependency with no changeset.

The log a contributor sees:

```
No changeset found, but this PR changes dependencies of published packages:
  - @tinacms/graphql

Without a changeset these packages get no version bump and no release, so the
change never reaches npm users. Run "pnpm changeset" and commit the result.
If the change genuinely needs no release, add an empty changeset ("pnpm changeset --empty").
```

Worth noting the demo caught a real bug: the first version pinned `setup-node` v5, which resolves `packageManager` from package.json and fails when pnpm is absent. Switched to the v6.4.0 SHA every other workflow here already uses.

brookjeynes-ssw (approved):

PR #7467, opened 2026-08-20
perf(tinacms): the rich-text editor downloads mermaid even when nothing uses it

### LINKED ISSUE #7245: ⚡ Lazy-load mermaid: rich-text editor stops shipping 83 MB-closure diagrams to every code block

wicksipedia:
Closes #7245
Part of #7237

The rich-text editor imported `mermaid` statically, so every editing session downloaded a 636 KB diagram renderer whether or not any document contained a diagram. The import is now deferred until a mermaid code block is rendered or parsed.

## Browser payload

Measured with `tinacms build --local` on the kitchen-sink example.

| | Entry chunk | Gzipped |
|---|---:|---:|
| Before | 6,339,116 | 1,974,875 |
| After | 5,698,025 | 1,823,566 |

That is 641 KB off the entry chunk, or 10.1%. `mermaid` moves into its own `mermaid.core-*.js` chunk of 635,594 bytes (150,677 gzipped), fetched on demand. Chunk count goes from 97 to 98.

The split was confirmed rather than assumed: `flowchart-elk`, a string that only exists in mermaid's source, is present in the before-entry chunk, absent from the after-entry chunk, and present in the new chunk.

`mermaid` stays in `dependencies`. This changes what the browser downloads, not what a user installs.

## Implementation

`mermaid-element.tsx` and `code-block/code-block-element.tsx` now use `await import('mermaid')` inside the effects that already handle rendering and validation. The diagram element gains `loading | ready | failed` states, a placeholder while the chunk resolves, and a cancellation guard on unmount. A load failure surfaces through the existing `ErrorMessage` rather than being swallowed.

Both files mirror the v4 implementation already on `main` from #7405, so the two trees stay in step.

## Tests

Two commits add regression coverage, one per tree. Each was verified by restoring the static import, confirming the test fails, then restoring the fix.

### v3: `playwright/prebuilt-admin/e2e/mermaid-lazy.spec.ts`

Runs a real browser against a real production build, using the suite added in #7239.

Matching request URLs against the word "mermaid" would not have worked. With a static import the bundler inlines mermaid into the entry chunk, so no mermaid-named file is ever requested and a name-based assertion stays green on a broken build. The spec instead reads the built chunks in `public/admin/assets/`, identifies which files carry mermaid's own source by requiring three distinct string literals from it, and asserts on those filenames. That catches both failure shapes: inlined into the entry chunk, or eagerly imported beside it.

Two cases:

- A document with a non-mermaid code block downloads no chunk containing mermaid.
- A document with a diagram downloads one and renders an `<svg>`. Without this second case, deleting mermaid outright would satisfy the first.

If mermaid stops shipping those three literals, the marker set comes back empty and the spec fails with a message saying the markers are stale, rather than passing while checking nothing.

Failure output with the static import restored:

```
✘ editing a document with a non-mermaid code block loads no mermaid
Error: mermaid reached the browser for a document with no diagram.
It lives in: index-BFoWqzia.js
```

### v4: `packages/v4/@tinacms/rich-text/src/mermaid-lazy-load.test.ts`

v4 already had this fix from #7405 but nothing guarded it. A browser test is not possible there: `@tinacms/rich-text` is private, its `exports` point at raw `src`, and its built `dist/index.js` is 876 bytes containing no mermaid. There is no chunk graph to assert against.

The test reads the source tree instead. It rejects every eager reference to mermaid and requires at least one deferred reference, so the check cannot be satisfied by removing mermaid. Three assertions stop it passing vacuously: the walk must find more than 50 source files, the scan more than 200 module specifiers, and a fixture proves the pattern distinguishes `await import()` from the five eager forms including `export ... from` and `require()`.

A `TODO` in the file says to replace it with a browser spec modelled on the v3 one once v4 has a production build, and names the gap it cannot cover: a static mermaid import reaching the same bundle from a different package.

This follows the source-analysis precedent already in that package (`src/boundary.test.ts`). No changeset, since it is test-only in a private package.

## Verification

- `turbo run build --filter="./packages/**"`: 25 of 25 pass.
- `vitest run` in `packages/tinacms`: 608 passed, 5 skipped.
- `vitest run` in `packages/v4/@tinacms/rich-text`: 186 of 186 across 17 files.
- `tsc --noEmit` in `packages/tinacms`: 145 errors before and after, byte-identical. All pre-existing, all in `@tinacms/mdx`.
- `biome check`: clean on added files. One pre-existing import-ordering error in the `@udecode/plate-code-block` block is unchanged and untouched.

## Limits

The Playwright spec has only been run locally on macOS and Chromium. It has not run in CI, and not under the config's retry setting.


packages/tinacms/src/toolkit/fields/plugins/mdx-field-plugin/plate/components/plate-ui/mermaid-element.tsx
```diff
@@ -1,26 +1,68 @@
...
+    const node = mermaidRef.current?.querySelector<HTMLElement>('.mermaid');
+    if (!node) return;
+
+    let isCancelled = false;
+    setStatus('loading');
+
+    const render = async () => {
+      try {
+        const { default: mermaid } = await import('mermaid');
+        await mermaid.run({ nodes: [node] });
```
joshbermanssw: If `node` is ever null, the effect returns with `status` stuck on `'loading'`, leaving the `<pre>` at `opacity-0` behind a permanent placeholder. Folding the lookup into the `try` routes it through the existing failure path, and it still runs in the same tick since an async body is synchronous up to the first `await`.

```suggestion
    let isCancelled = false;
    setStatus('loading');

    const render = async () => {
      try {
        const node = mermaidRef.current?.querySelector<HTMLElement>('.mermaid');
        if (!node) throw new Error(RENDERER_FAILED);

        const { default: mermaid } = await import('mermaid');
        await mermaid.run({ nodes: [node] });
```

packages/tinacms/src/toolkit/fields/plugins/mdx-field-plugin/plate/components/plate-ui/mermaid-element.tsx
```diff
@@ -1,26 +1,68 @@
...
+        >
+          <pre className='mermaid not-tina-prose'>{config}</pre>
+        </div>
+        {status === 'loading' ? (
+          <div className='absolute inset-0 flex items-center justify-center text-sm text-muted-foreground'>
+            Loading diagram…
+          </div>
+        ) : null}
       </div>
+      {status === 'failed' ? <ErrorMessage error={error} /> : null}
```
joshbermanssw: `plate/components/` uses `cond && (…)` for conditional rendering in 13 places and `? … : null` in none, including the `lang === 'mermaid'` guard in `code-block-element` that renders this component. Both conditions here are strict boolean comparisons, so there is no falsy-leak risk.

```suggestion
        {status === 'loading' && (
          <div className='absolute inset-0 flex items-center justify-center text-sm text-muted-foreground'>
            Loading diagram…
          </div>
        )}
      </div>
      {status === 'failed' && <ErrorMessage error={error} />}
```

packages/tinacms/src/toolkit/fields/plugins/mdx-field-plugin/plate/components/plate-ui/mermaid-element.tsx
wicksipedia: Applied in 0400f4ad7

packages/tinacms/src/toolkit/fields/plugins/mdx-field-plugin/plate/components/plate-ui/mermaid-element.tsx
wicksipedia: Applied in 0400f4ad7

joshbermanssw (approved):
LGTM

Issue #5465, opened 2025-01-31
⚠️  Peer Dependency Warnings and Deprecated Packages


JackDevAU:
CC: @wicksipedia 

### Description  

When installing Tina CMS with the Hugo starter template (also seeing these on all other starters) (using Node.js LTS v22.13.0), multiple peer dependency warnings and deprecated package notices appear. The warnings indicate conflicts with `yup` versions and outdated dependencies such as `@babel/plugin-proposal-class-properties`, `@babel/plugin-proposal-object-rest-spread`, `q`, and `react-beautiful-dnd`. Additionally, the installation results in 9 security vulnerabilities (8 high, 1 critical), raising concerns about the stability and security of the package.  

### Acceptance Criteria  

- [ ] Investigate the peer dependency conflicts, especially related to `yup` versions.  
- [ ] Update or replace deprecated packages where possible.  
- [ ] Assess the security vulnerabilities and determine if they can be mitigated through dependency updates.  
- [ ] Test the installation process to ensure minimal warnings and a stable setup.  
- [ ] Provide documentation or guidance for users encountering similar issues.  

🤖

joshbermanssw:
Resolved across five PRs: #6187 (yup → v1), #7468 (dropped typedoc, fs-extra, crypto-js; bumped cloudinary for GHSA-g4mf-96x5-5m2c), #7469 (final-form family for the React 19 peer), #7466 (uvu out of @tinacms/mdx), and #7484 (orphaned react-beautiful-dnd and @types/crypto-js declarations).

PR #7456, opened 2026-08-19
chore: add missing changesets for two merged dependency fixes

kulesy:
Closes #7455

**TL;DR** #7425 and #7321 merged without changesets, so both fixes are stranded on `main`. This adds them.

**Pain:** `@tinacms/graphql` and `tinacms-authjs` are both published and neither is in the changesets `ignore` list, so a dependency change in either needs a changeset to produce a version bump and a release. #7425 moved `@tinacms/graphql` to js-yaml 3.15.1 and #7321 moved `tinacms-authjs` to next-auth 4.24.15, both without one. The js-yaml case matters most: it closed GHSA-5p4m-2wfm-xmqj on the 3.x line, and `@tinacms/graphql` is the package that parses content YAML, so the published package still resolves the vulnerable 3.14.2 until a release ships.

**Solution:** Adds a changeset for each, both `patch`, naming the bump and its advisory where there is one. Version Packages #7443 is still open, so these are picked up in the current release rather than the one after.

PR #7454, opened 2026-08-19
chore: add missing changeset for the js-yaml 4.3.1 bump

kulesy:
Closes #7453

**TL;DR** #7447 fixed the vulnerable js-yaml but merged without a changeset, so the fix is sitting on `main` unreleased. This adds one.

**Pain:** `@tinacms/cli` is published and not in the changesets `ignore` list, so a dependency change there needs a changeset to produce a version bump and a release. #7447 moved it from the vulnerable js-yaml 4.1.1 to 4.3.1, closing GHSA-5p4m-2wfm-xmqj on `main`, but with no changeset there is no release. The published package keeps resolving 4.1.1, so `npm install @tinacms/cli` still pulls the vulnerable copy and the advisory stays open for users.

**Solution:** Adds a changeset marking `@tinacms/cli` `patch` and naming the js-yaml bump with its advisory. Version Packages #7443 is still open, so this is picked up in the current release rather than the one after.

PR #7468, opened 2026-08-20
🧹 Drop three unused deps and bump cloudinary to clear install warnings

### LINKED ISSUE #7475: 🧹 Unused and deprecated deps ship to every user and warn on install
**Pain**

Four dependencies ship to every user without earning their place, and each one shows up in the install output of any starter:

| Dep | Where | Problem |
|---|---|---|
| `typedoc` | `@tinacms/mdx` runtime deps | Nothing imports it. Landed by copy-paste in Dec 2022 and has shipped ever since, dragging an unsatisfiable `typescript` peer range with it. |
| `fs-extra` | `@tinacms/metrics` peer | Declared `^9.0.1` while the repo runs `^11.3.0`, so the range is never satisfied and the warning fires on every install. |
| `crypto-js` | `@tinacms/cli`, `tinacms` | Deprecated. One call site in the CLI; `tinacms` never imports it at all. |
| `cloudinary` | catalog, `^1.41.3` | Carries GHSA-g4mf-96x5-5m2c, a high-severity argument injection, patched in 2.7.0. |

That is 2 deprecation notices and 2 of the 6 unmet-peer rows a Hugo starter prints on install.

This is the tractable slice of #5465, which also covers the `yup` conflicts, `react-final-form` and a documentation AC. Those are out of scope here: `react-final-form` needs three catalog bumps moving together and the Playwright form suite as a gate, which is not comparable blast radius to removing code nothing calls.

**Approach**

Replace each single call site with a node builtin, or delete the declaration where there is no call site. Bump the `cloudinary` catalog entry to the advisory's first patched version, keeping it a caret so consumers can override. Clean up whatever the `typedoc` removal strands. Every published package touched needs its own changeset, or the fix never reaches npm.

**Acceptance Criteria**

- [ ] `typedoc`, `fs-extra` and `crypto-js` are gone from the packages above, with the behaviour of each replaced call site unchanged
- [ ] `cloudinary` resolves to a version not affected by GHSA-g4mf-96x5-5m2c
- [ ] A Hugo starter install prints 1 deprecation notice and 4 unmet-peer rows
- [ ] Each published package touched is named by a changeset and appears in `changeset status`

joshbermanssw:
Closes #7475

**TL;DR** Drops three dependencies nothing uses and moves a fourth off a high-severity advisory, taking a Hugo starter install from 2 deprecation notices and 6 unmet-peer rows down to 1 and 4. Part of #5465, which stays open for the `yup` conflicts, `react-final-form` and the documentation AC.

**Pain:** Four dependencies reach every user's `node_modules` without earning their place. `@tinacms/mdx` carries `typedoc` in its runtime `dependencies`, where it landed by copy-paste in Dec 2022, dragging an unsatisfiable `typescript` peer range behind it. `@tinacms/metrics` declares a `fs-extra@^9.0.1` peer while the repo itself runs `^11.3.0`, so the range is never satisfied and the warning fires on every install. `crypto-js` is deprecated and sits in two packages, one of which never imports it. And the `cloudinary` catalog entry is pinned to a 1.x line carrying GHSA-g4mf-96x5-5m2c, a high-severity argument injection patched in 2.7.0.

**Solution:** Each removal replaces its single call site with a node builtin, or deletes a declaration that had no call site. `fs-extra`'s one use was `readFileSync`, which it re-exports unchanged from `node:fs`. `crypto-js` generated the default `NEXTAUTH_SECRET` during `tinacms init`, and since `WordArray.toString()` defaults to hex, `randomBytes(16).toString('hex')` is byte-identical. Nothing imports `typedoc` at all, so it goes along with the `docs` script that was its only user, plus everything that stranded: three orphaned catalog entries, the `docs` scripts in `@tinacms/graphql` and `@tinacms/search` that never declared `typedoc` themselves, both `typedoc.json` files, and the generated `spec.md`. `cloudinary` moves to `^2.7.0`, the advisory's first patched version, kept as a caret so consumers can still override; every method this package calls is unchanged across the major. Each of the five published packages touched carries its own changeset, including `@tinacms/metrics`, which no other bump would have reached because its only workspace dep is a devDependency and changesets skips the version bump in that case.


kulesy (dismissed):

kulesy:
## Evidence and reviewer notes

Moved out of the description to keep it to the house shape.

### Why each removal is safe

| Dep | Change | Why it was safe |
|---|---|---|
| `fs-extra` | Removed from `@tinacms/metrics` | Never actually used. The one call was `readFileSync`, which fs-extra re-exports unchanged from `node:fs`. The `^9.0.1` peer was never satisfied anyway. |
| `typedoc` | Removed from `@tinacms/mdx` runtime deps | Nothing in `src/` imports it. It landed in `dependencies` by copy-paste in Dec 2022 and shipped to every user since. Only the `docs` script used it, so the script goes too. |
| `crypto-js` | Removed from `@tinacms/cli` and `tinacms` | One call site, generating the default `NEXTAUTH_SECRET` at init. `WordArray.toString()` defaults to hex, so `randomBytes(16).toString('hex')` is byte-identical: both give 32 lowercase hex characters. `crypto-js@4.2.0`'s `random()` already used `cryptoSecureRandomInt()`, so this is a dependency removal, not a security fix. `tinacms` declared it without ever importing it. |
| `cloudinary` | `^1.41.3` to `^2.7.0` | Two majors, but v2.0.0 has only three breaking changes and none reach this package: `secure` now defaults true (already set explicitly, and `CloudinaryConfig` has no `secure` key so the `Object.assign` cannot override it), URL analytics applies only to SDK-generated URLs while this package reads `secure_url` off Search API responses, and dropped Node 6/8 is long past. Every method used is unchanged; `uploader.destroy`'s signature is identical between the v1.41.3 and v2.7.0 type declarations. |

### Before / after

Verified with `scripts/try-starter-local.sh`, which packs the workspace and overrides the starter's whole Tina graph so transitive deps resolve locally too.

**Before**
```
WARN  3 deprecated subdependencies found: crypto-js@4.2.0, prebuild-install@7.1.3, q@1.5.1
WARN  Issues with peer dependencies found
├─┬ tinacms 3.12.0
│ ├─┬ react-final-form 6.5.9
│ │ └── ✕ unmet peer react@"^16.8.0 || ^17.0.0 || ^18.0.0": found 19.2.8
│ └─┬ @tinacms/search → @tinacms/graphql → @tinacms/mdx → typedoc 0.26.11
│     └── ✕ unmet peer typescript@"4.6.x || … || 5.6.x": found 5.9.3
└─┬ @tinacms/cli 2.6.0
  ├─┬ @tinacms/app → graphiql 3.0.0-alpha.1
  │ ├── ✕ unmet peer react / react-dom: found 19.2.8
  │ └─┬ @graphiql/react 0.18.0
  │   ├── ✕ unmet peer react / react-dom: found 19.2.8
  │   └─┬ @headlessui/react 1.7.19
  │     └── ✕ unmet peer react / react-dom: found 19.2.8
  └─┬ @tinacms/metrics 2.1.1
    └── ✕ unmet peer fs-extra@^9.0.1: found 11.4.0
```

**After**
```
WARN  1 deprecated subdependencies found: prebuild-install@7.1.3
WARN  Issues with peer dependencies found
├─┬ tinacms 3.12.0
│ └─┬ react-final-form 6.5.9
│   └── ✕ unmet peer react@"^16.8.0 || ^17.0.0 || ^18.0.0": found 19.2.8
└─┬ @tinacms/cli 2.6.0
  └─┬ @tinacms/app → graphiql 3.0.0-alpha.1
    ├── ✕ unmet peer react / react-dom: found 19.2.8
    └─┬ @graphiql/react 0.18.0 → @headlessui/react 1.7.19
      └── ✕ unmet peer react / react-dom: found 19.2.8
```

One caveat on reading that. Because the harness overrides the **whole** Tina graph it installs `next-tinacms-cloudinary` too, and `cloudinary@1.41.3` is the only path to `q@1.5.1` in the lockfile. A real starter does not install that package, so `q` was never in its tree. The honest starter-facing numbers are **2 deprecations to 1, and 6 peer rows to 4**. Both peer rows that disappear do reach a real starter, via `tinacms → … → @tinacms/mdx` and `@tinacms/cli → @tinacms/metrics`.

### Changesets

Five published packages are touched and none is in the changesets `ignore` list, so each needs a changeset to produce a version bump:

- `next-tinacms-cloudinary`, for the cloudinary bump
- `@tinacms/cli` and `tinacms`, for the crypto-js removal
- `@tinacms/metrics`, for the fs-extra peer removal
- `@tinacms/mdx`, for the typedoc removal
- `@tinacms/graphql` and `@tinacms/search`, for the docs tooling removal

`@tinacms/metrics` is the one worth understanding. It is not a dependent of any other bumped package, since its only workspace dep is `@tinacms/scripts` as a **devDependency**, and changesets skips the version bump when the changed dependency is devDependency-only. Without its own changeset the fs-extra fix would have sat on `main` unreleased and the peer warning would have survived in every install, which is the same failure mode #7454 and #7456 already had to patch.

### Deliberately not here

**`react-final-form`.** The remaining peer row needs three catalog bumps moving together (`final-form` 4 to 5, `final-form-arrays` 3 to 4, `react-final-form` 6 to 7). All three are one coordinated Flow to TypeScript rewrite published the same day and labelled "no breaking changes", with React 19 added to the peer range.

It is kept separate because the blast radius is not comparable to anything here. The three removals above touch nothing, and cloudinary is confined to one optional media package, while final-form underpins every form field in the admin through ~10 files in `packages/tinacms/src/toolkit/`. The real risk is the ~30 behaviour fixes in the `.0.1` patches: dirty-state handling (`modified` cleared after `reset()`, `keepDirtyOnReinitialize` with arrays), the checkbox/radio `parse`-vs-`format` fix, and `final-form-arrays`' `removeBatch` numeric-sort fix against group-list fields. That wants the Playwright form suite as a gate, not a dependency-hygiene review.

**`prebuild-install`** is transitive via `better-sqlite3` and not fixable here.

**Two known follow-ups**, both left out because they move dependency edges and pull `lodash` from 4.17.21 to 4.18.1 across roughly a dozen of them. That deserves to be a headline rather than a side effect:

- Raising the cloudinary floor to `^2.9.0`. 2.7.0 and 2.8.0 both still declare `q@^1.5.1`; 2.9.0 is the first release that drops it. The lockfile resolves 2.10.1 so this repo is already clean, but a consumer pinned to an early 2.7.x keeps `q`.
- Removing `@types/crypto-js` from `next-tinacms-azure`, `-cloudinary`, `-dos` and `-s3`. None imports crypto-js, and all four pin v3 types against what was a v4 library.

### Tests

Every package builds and typechecks clean and the full suite passes: **66 turbo tasks** across `build`, `types` and `test`. `pnpm install --frozen-lockfile` is clean. The docs-tooling cleanup leaves `pnpm-lock.yaml` **byte-identical**, because none of it touches a resolved dependency edge. Starter output verified via `scripts/try-starter-local.sh` above.

**Not verified:** cloudinary upload, browse and folder listing against a live account. CI cannot cover it, so it is worth a smoke test before merge.

kulesy (dismissed):

kulesy (approved):

PR #7472, opened 2026-08-21
fix(tinacms): Button and IconButton never passed disabled to the DOM

### LINKED ISSUE #7471: Bug Report 🐞 Disabled buttons are still focusable and keyboard-activatable
### The exact error message

```shell
No error message — the failure is silent. A button rendered with `disabled`
carries no `disabled` attribute in the DOM:

<button class="... pointer-events-none opacity-30 cursor-not-allowed">Create Branch</button>
```

### Steps to reproduce

1. Open the TinaCMS admin and go to any form with a button that has a disabled state — e.g. the branch switcher's **Create Branch** screen, or the media library's **Rename** modal.
2. Leave the input empty (or invalid) so the button renders in its disabled state — greyed out at 30% opacity.
3. Click it with the mouse: nothing happens, as expected.
4. Now press `Tab` until focus lands on that same greyed-out button.
5. Press `Enter` (or `Space`).

Expected: a button shown as disabled cannot be focused or activated, by mouse or keyboard.

Actual: the button is still in the tab order, still takes focus, and `Enter`/`Space` fires its `onClick`. `Button` and `IconButton` in `packages/tinacms/src/toolkit/styles/button.tsx` destructure `disabled` and use it only to pick styling classes (`pointer-events-none opacity-30 cursor-not-allowed`) before spreading the remaining props, so the attribute never reaches the element. `pointer-events-none` blocks the pointer but not the keyboard.

The `busy` prop has the same shape — it also only sets `pointer-events-none` — so a button already mid-flight stays keyboard-activatable, which is a double-submit path.

This affects every disabled button in the CMS: there are 18 `<Button>`/`<IconButton>` call sites carrying a `disabled` prop, across 11 files in `packages/tinacms/src/toolkit`. (An earlier revision of this issue said ~56, which was a raw grep of every `disabled=` in the toolkit and swept in native inputs and Radix primitives that were never affected.)

### Your environment

- `tinacms` v3.12.0 (reproduced on `main`)
- `@tinacms/cli` v2.6.0
- Framework-independent — this is in the shared admin toolkit, so it reproduces on any framework and in both local and TinaCloud mode
- No non-default setup required: stock admin, no custom MediaStore, no custom auth

### A way for us to reproduce

Reproduces against the repo's own `playwright/tina-playwright` fixture — the **Create** button on the "Add Folder" dialog is one of the affected buttons.

Originally spotted during review of #6763.

kulesy:
Closes #7471

**TL;DR** Make the toolkit `Button` and `IconButton` actually render the `disabled` attribute.

**Pain:** both components destructured `disabled` and used it only to pick styling classes, so the attribute never reached the element. `pointer-events-none` blocks the pointer but not the keyboard, which left every disabled button in the CMS focusable and activatable with Enter or Space — 18 `<Button>`/`<IconButton>` call sites carrying a `disabled` prop, across 11 files. The same hole applied to `busy`, so a button already mid-flight could be re-fired from the keyboard into a double submit.

**Solution:** passes the attribute through on both components, and folds `busy` into it to match the `disabled={disabled || busy}` that `DropdownButton` already ships. `Button` is polymorphic via `as`, so the attribute is omitted for tags that cannot carry it. Three test suites and a rename-modal click guard existed only to work around the hole and are removed.


kulesy:
### Verification

Rebased onto `main` at `ce8851035` and re-run there:

| Suite | Result |
|---|---|
| `tinacms` unit (`CI=true vitest run`) | 55 files, 615 passed, 5 skipped |
| `pnpm test` (monorepo) | 66/66 tasks |
| `pnpm types` | 50/50 tasks |
| `playwright/tina-playwright` | 55 passed, 3 skipped |
| `playwright/prebuilt-admin` (production build) | 12 passed |

The 5 unit and 3 e2e skips are pre-existing `test.skip`.

The two e2e suites are the load-bearing ones here, since jsdom can't prove the real behaviour. `prebuilt-admin`'s custom-field save round-trip exercises a `busy` button against the minified production bundle, and `create-folder.spec.ts` now asserts `toBeDisabled()` where it previously asserted `toHaveClass(/pointer-events-none/)` — that assertion passing in Chromium is the direct evidence the attribute lands.

### The tests are real regression guards

Reverting `button.tsx` with the new tests in place fails all 6 new cases plus the 3 updated suites, so they fail for the right reason rather than passing vacuously.

Worth flagging one detail for whoever reviews the test file: I first wrote these as focus assertions (`button.focus()` → expect focus elsewhere) and they failed. The tests were wrong, not the fix — **jsdom gates `.click()` on `disabled` but does not implement the focus gate**. They're now written as handler-activation assertions, which jsdom does enforce. Anyone extending this file should avoid focus-based assertions for the same reason.

### On folding `busy` into `disabled`

This is the one genuine behaviour change and worth a deliberate look. `DropdownButton` already ships `disabled={disabled || busy}` ([dropdown-button.tsx#L203](https://github.com/tinacms/tinacms/blob/main/packages/tinacms/src/toolkit/styles/dropdown-button.tsx#L203)), so this makes `Button` consistent with its own sibling rather than introducing a new opinion. The live double-submit paths it closes are `busy={submitting}` in `form-builder.tsx` and `busy={uploading}` in `media-manager.tsx`.

Regression risk is low: `busy` already set `pointer-events-none`, so no caller could have depended on clicking a busy button. Only keyboard activation and focusability change. Tooltips on disabled buttons were already suppressed by `pointer-events-none`, so that behaviour is unchanged.

### Removed workarounds

The rename modal's `if (!canSubmit) return;` guard is dead because `canSubmit = !processing && valid` feeds `disabled={!canSubmit}`, which now genuinely blocks activation. `onSubmit` preventDefault handlers and other form-level guards are untouched — those fix separate bugs.

kulesy:
## Button states in the admin

Captured from the local admin (`playwright/tina-playwright`) on this branch.

**This PR changes no styling.** The class bundles are untouched, so a disabled button looks pixel-identical before and after. I re-ran the same capture against `main` and all four frames are byte-identical by SHA-256. The screenshots show *which* states exist; the table underneath is where the change actually is.

| Disabled (invalid input) | Enabled (valid input) |
|---|---|
|  |  |

Busy, mid-save, with the mutation held open so the state is visible. This is the double-submit path that folding `busy` into `disabled` closes:



Disabled pagination in a collection list, a second real instance:



### What actually changed

Same admin, same interactions, reading each button off the live DOM. `kbd` is `tabIndex >= 0 && !disabled`, i.e. can a keyboard user still reach and fire it.

| Button | State | `disabled` attr before | after | Keyboard-reachable before | after |
|---|---|---|---|---|---|
| Previous / Next | disabled | absent | **present** | **yes** | no |
| Create (folder) | empty name | absent | **present** | **yes** | no |
| Create (folder) | invalid name | absent | **present** | **yes** | no |
| Create (folder) | valid name | absent | absent | yes | yes |
| Save | busy (in-flight) | absent | **present** | **yes** | no |

Every row marked **yes** under "keyboard-reachable before" is a button the CMS rendered as visibly unavailable that a keyboard user could still tab to and fire with Enter or Space. The last row is the double-submit path.

Screenshots are hosted on `fix/button-disabled-attribute-screenshots` and are not part of this diff.

kulesy:
## Before / after, recorded

The screenshots above can't show this bug, because it is behavioural: the styling is identical either way. Here it is as a recording. Same Playwright script run twice, once against `main` and once against this branch, driving the real admin in `playwright/tina-playwright`.

The steps are identical in both runs: open **Add Folder**, type `a b c d` (invalid), then `Tab` `Tab` `Enter` from the input. No mouse.

before and after

**Before** (`main`) - two tabs land focus on the greyed-out **Create** button, and `Enter` fires its handler. The folder creation is attempted with the name validation just rejected, and the editor gets *"There was an error creating the folder. Please try again. If the problem continues, please contact the developer"*. That dialog is the bug: a control the UI presents as unavailable ran anyway, and surfaced a developer-facing error to an editor who did nothing wrong.

**After** (this PR) - identical keystrokes. `Tab` skips the button, `Enter` does nothing, and the modal just sits there with the validation message.

Worth noting the handler at [`CollectionListPage.tsx`](https://github.com/tinacms/tinacms/blob/8fe93ceb7c665400d0c380875fd02d1f57afefd3/packages/tinacms/src/admin/pages/CollectionListPage.tsx) has no internal `isFolderNameValid` re-check - it calls `createFunc()` directly. It was relying entirely on `disabled` to gate it, which is exactly the assumption that did not hold.

Full 37s version with titles, including the run-up: [button-disabled-demo.mp4](https://raw.githubusercontent.com/tinacms/tinacms/8fe93ceb7c665400d0c380875fd02d1f57afefd3/button-disabled-demo.mp4)

<details>
<summary>How it was recorded</summary>

Playwright `recordVideo` at 1000x640, same script both runs. Between runs: `git checkout main -- packages/tinacms/src/toolkit/styles/button.tsx`, `pnpm --filter tinacms build`, restart the dev server, so the admin bundle genuinely differs. Verified each build by grepping the emitted `dist/index.js` for the guard. Title cards and badges composited with ffmpeg; nothing in the admin frames is edited.

</details>

Recording and screenshots are hosted on `fix/button-disabled-attribute-screenshots` and are not part of this diff.

joshbermanssw (commented):
Good catch on the underlying bug — `pointer-events-none` blocking the pointer but not the keyboard is exactly the kind of hole that hides for years.

Two things on the non-`button` fallback, though: as written it still allows activation, so disabled link-buttons remain clickable. Suggestions inline.

Nothing to change on `IconButton` — it always renders a real `<button>`, so `disabled={disabled || busy}` is right there.

packages/tinacms/src/toolkit/styles/button.tsx
```diff
@@ -63,6 +63,15 @@ export const Button = ({
...
   };
+  const inert = disabled || busy;
+  // `as` can render a tag that has no `disabled` attribute, so those fall back
+  // to the aria/tabIndex pair used for disabled link-buttons elsewhere.
+  const inertProps =
+    typeof Tag !== 'string' || Tag === 'button'
+      ? { disabled: inert }
+      : inert
+        ? { 'aria-disabled': true, tabIndex: -1 }
+        : {};
```
joshbermanssw: `aria-disabled` + `tabIndex={-1}` leaves the control activatable. `tabIndex` only removes it from the *tab order* — the element can still be clicked, still be focused programmatically, and still fire on Enter once focused. That's the same advisory-only problem as the `pointer-events-none` this PR is replacing.

Dropping `href` does all three at once: an `<a>` without one isn't focusable and isn't activatable, so tab-order removal comes for free rather than being asserted separately.

I'd also narrow the first branch. `typeof Tag !== 'string'` sends `disabled` to any component passed via `as`, which assumes it accepts one — if it doesn't, it either swallows the prop or forwards it to a DOM node that shouldn't carry it.

And a naming point: `inert` is a real HTML attribute with real semantics, and it's arguably the right tool here (it removes the subtree from focus, hit-testing and the a11y tree, and is baseline now). Using the name for something else means we can't reach for it later without confusion. The trade-off if you do want it: `inert` drops the element from the accessibility tree entirely, so a screen reader stops announcing it, versus `aria-disabled` announcing it as present-but-unavailable.

```suggestion
  const isDisabled = disabled || busy;
  const inertProps =
    Tag === 'button'
      ? { disabled: isDisabled }
      : isDisabled && { 'aria-disabled': true, href: undefined };
```

(When `isDisabled` is false the second branch evaluates to `false`, and spreading `false` applies nothing.)

packages/tinacms/src/toolkit/styles/button.tsx
```diff
@@ -75,6 +84,7 @@ export const Button = ({
         className
       )}
       {...props}
+      {...inertProps}
```
joshbermanssw, line 86: Whatever the markup says, the handler is the only part that can't be bypassed — attributes are advisory, a missing `onClick` isn't. This also covers tags nobody anticipated being passed to `as`.

```suggestion
      {...inertProps}
      onClick={isDisabled ? undefined : props.onClick}
```

packages/tinacms/src/toolkit/styles/button.tsx
kulesy: You're right, and I verified it rather than taking it on faith: rendering `<Button as='a' href disabled onClick>`, the anchor is programmatically focusable and `.click()` fires the handler. Advisory-only, exactly the weakness this PR set out to remove. Fixed in 9ffcbfb.

Taken:
- **`href: undefined`** — this is the real fix, agreed.
- **`inert` rename** — fair, and I hadn't thought about it. It's a real attribute and squatting on the name blocks reaching for it later. Now `isDisabled`. Not adopting the attribute itself, for the reason you gave: dropping the control out of the a11y tree is worse here than `aria-disabled` announcing it as present-but-unavailable.
- **`onClick` guard** — taken, see the other thread.

One deviation and one pushback:

**Kept `tabIndex: -1` alongside `href: undefined`.** Redundant for an anchor, as you say. It earns its place for any other string tag someone routes through `as` (`span`, `div`), where there's no `href` to drop and the handler guard is the only other line of defence. Cheap, and it fails safe.

**Kept `typeof Tag !== 'string'` on the first branch.** Your concern is real, but narrowing it to `Tag === 'button'` means a component-valued `as` gets `href: undefined` — a prop it very likely doesn't understand — and never gets native `disabled`, which is what actually blocks click, Enter and Space in the browser. Most components that render a button accept `disabled`; one that doesn't either ignores it or forwards it to a host node, and if that node isn't a `<button>` React drops it with a warning rather than doing damage. With the `onClick` guard now in place the activation risk you're pointing at is covered either way, so I'd rather keep the stronger native semantics for the common case. Happy to narrow it if you still disagree - it's a one-line change and no call site exercises it today.

Also worth flagging since it shapes the tests: dropping `href` changes the element's implicit role, so it's no longer `role="link"`. Test queries moved to `container.querySelector('a')`.

packages/tinacms/src/toolkit/styles/button.tsx
kulesy: Agreed, and taken verbatim in 9ffcbfb. This is the part that doesn't depend on getting the markup right for a tag we haven't thought of yet.

It also happens to be the only half of the fix I can prove in this test suite. jsdom doesn't implement the focusability side — it will happily `.focus()` an href-less anchor, and it never gated focus on `disabled` either — so a test asserting "not focusable" would be asserting jsdom's behaviour, not ours. The handler guard is directly observable, so the new tests assert `href` is gone and the handler doesn't fire, and leave the focus claim to the browser.

Covered now: disabled link doesn't fire, busy link doesn't fire, enabled link keeps its `href` and still fires normally. Verified the three real `as='a'` call sites are untouched — none passes `disabled` or `busy`, so they render exactly as before, `href` and all.

joshbermanssw (approved):
🚀

joshbermanssw (approved):

PR #6751, opened 2026-04-29
chore(tinacms-pkgs): remove unused devDeps across @tinacms/* packages

JackDevAU:
## Summary

Consolidates four small package-level cleanup commits into one PR (~5 file changes total).

- **`@tinacms/auth`** — remove unused `react`/`react-dom` devDeps (no React imports anywhere in the package).
- **`tinacms-authjs`** — remove unused `react-dom` devDep (no `react-dom` imports; `react` retained for typing).
- **`@tinacms/datalayer`** — remove unused `jest-diff`, `jest-file-snapshot`, `jest-matcher-utils` devDeps.
- **`@tinacms/search`** — remove unused `jest-diff`, `jest-file-snapshot`, `jest-matcher-utils` devDeps.

Plus a single lockfile sync commit reflecting the removals.

Each removal was verified via grep: no source/test/build file in the affected package imports the removed dep.

### Side-effect to note

Removing `react-dom` from `tinacms-authjs/devDependencies` causes pnpm to resolve `react-dom` via the `peerDependencies` range (`^17.0.2 || ^18 || ^19`) instead of the previous `^19.2.3` devDep pin. The lockfile reflects the resolution moving from `19.2.3` to `18.3.1` for this package's local dev/test context. Functionally equivalent for consumers (peer dep contract unchanged); only affects workspace-internal resolution.

## Test plan

- [ ] `pnpm install --frozen-lockfile` succeeds
- [ ] `pnpm build` succeeds across affected packages
- [ ] `pnpm test` passes in `@tinacms/datalayer`, `@tinacms/search`, `@tinacms/auth`, `tinacms-authjs`

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #7476, opened 2026-08-21
chore(tinacms-pkgs): point repository.directory at each package's own folder

kulesy:
No linked issue.

**TL;DR** Eight published packages pointed their npm "repository" link at the wrong folder; this fixes all eight and drops a dead `generate:schema` script from three packages.

**Pain:** Each of these eight packages declares a `repository.directory` copy-pasted from whichever package it was forked from, so the "repository" link on its npm page resolves to unrelated source. Two of the referenced folders (`packages/tina-graphql`, `packages/next-tinacms-dos`) do not exist any more at all. Separately, `@tinacms/metrics`, `@tinacms/cli` and `@tinacms/schema-tools` each declare `"generate:schema": "pnpm node scripts/generateSchema.js"` - none of the three has a `scripts/` directory, `generateSchema.js` exists nowhere in the repo, `git log --all` shows it has never existed on any ref, and nothing invokes it.

**Solution:** Point each `repository.directory` at the package's own folder and delete the three dead `generate:schema` lines. A sweep of every `package.json` under `packages/` confirms these eight were the only mismatches and that all are published, so the changeset names all nine touched packages at `patch` - the eight above plus `@tinacms/cli`, touched by the script removal only. Shape follows 556a1625 (#6751). Nothing here touches a dependency edge, and `pnpm-lock.yaml` is byte-identical to `main`.


joshbermanssw (approved):
LGTM

Issue #7474, opened 2026-08-21
✨ Media Manager - filter by file type


joshbermanssw:
### Pain

The media manager has search and an All/Folders/Files toggle, but no way to narrow by file type. In a library with thousands of assets, an editor looking for a PDF still scrolls past every PNG.

Slice of #6996. Search from that issue has already shipped; sorting is tracked separately in #4600.

### Approach

Filtering happens server-side. The client sends the selected type and renders what comes back, rather than filtering a page it has already loaded. Client-side filtering only ever narrows the current page, so on any paginated store it would show a near-empty grid while the real matches sit further down the list.

**TinaCloud, assets-api** (lands first)

Add an `ext` query param to the v2 list endpoint, parsed alongside `search` in `koa-admin.ts` and applied in the same pass over the index, before the search branch and before folder derivation so folders are not derived from excluded files. Folders follow the precedent search already set: a folder appears only when it actually contains a match.

The endpoint currently has no declared contract - `search`, `limit`, `cursor` and `expiresIn` are each read ad-hoc from `ctx.request.query`. Rather than adding a fourth positional optional, fold them into one typed object so there is a single place that answers "what does this endpoint accept":

```ts
type ListQuery = {
  search?: string;
  exts?: Set<MediaExtension>;
  limit?: number;
  cursor?: string;
};
```

**TinaCMS**

Widen the existing `mediaFilter` state (`media-manager.tsx`, currently `'all' | 'folders' | 'files'`) to carry an extension or category, so `MediaFilterToggle` and any future per-field `accept` share one mechanism instead of two parallel ones. Add the param to `MediaListOptions` and to the URL built in `media-store.default.ts`.

One shared source of truth for the list, so the UI, the schema types and the endpoint cannot drift:

```ts
export const MEDIA_EXTENSIONS = ['jpg', 'png', 'svg', 'webp', 'pdf', 'mp4'] as const;
export type MediaExtension = (typeof MEDIA_EXTENSIONS)[number];
export type MediaCategory = 'image' | 'video' | 'audio' | 'document';
```

Derive the dropdown options from the extensions actually present in the library rather than the full list, so we never offer SVG to someone who has none.

### Acceptance criteria

- [ ] `ext` is accepted by the v2 list endpoint and filtering is done server-side
- [ ] The endpoint's accepted params are declared in one typed place
- [ ] Media manager has a file-type control alongside search and the folders/files toggle
- [ ] Options reflect the file types actually present in the library
[...]

Issue #6996, opened 2026-05-29
Media Manager: add sort, search, and file-type filtering controls


joshbermanssw:
[🟥 Watch the video (39 sec)](https://www.youtube.com/watch?v=MlGknuNy_ao)

![Media Manager view showing mixed asset types (folders, SVG/PNG) with no obvious sort/filter/search controls](https://sayakshaverproduction.blob.core.windows.net/images/desktop-screenshots/fa38aca5-30b6-4df1-9765-4bd466a8e5ac.png?sv=2025-05-05&se=2046-05-29T06%3A32%3A41Z&sr=b&sp=rd&sig=5vvqNJp0nFnFiMm%2BGvQ5FePUDP%2BI%2Bq5JRzDIxtAXXls%3D)

### Pain
The TinaCMS **Media Manager** is difficult to use when there are many assets because there is currently no clear way to:
- sort items
- search assets
- filter by file type (e.g., folders vs SVG vs PNG)

This makes it time-consuming to locate the correct media item and increases the chance of selecting/using the wrong asset.

### Suggested Solution
Add basic discoverability and organization controls to the Media Manager:
- Default sorting (with user-selectable sort options)
- Search input for asset name/path
- File-type filtering (e.g., Folders, Images, SVG, PNG, etc.)

### Acceptance Criteria
- Media Manager provides a visible control to change sorting.
- Media Manager has a search input that filters the displayed items by name.
- Media Manager provides a file type filter that can show only a selected type (e.g., only PNGs).
- Sorting, searching, and filtering can be cleared to return to the full list.

### Tasks
- [ ] Add a sorting control to the Media Manager UI.
- [ ] Implement a default sort order for assets.
- [ ] Add a search input to the Media Manager UI.
- [ ] Implement client-side filtering of items based on the search query.
- [ ] Add a file-type filter control to the Media Manager UI.
- [ ] Implement client-side filtering of items by file type.
- [ ] Add a clear/reset action to remove search and filters.

/cc @kulesy @JackDevAU

PR #7477, opened 2026-08-21
✨ Filter the media manager by file type

joshbermanssw:
Test it out yourself - https://tina-io-git-jb-show-off-filter-tinacms.vercel.app/admin/index.html#/~/

## TL;DR

An image field could declare `accept` and still be handed a mismatched file, and there was no way to narrow a large library to just the PDFs. This adds `accept` to the image field, a type filter to the media manager toolbar, and refuses a file outside the field's types before it is uploaded.

## Reviewer notes

- **A field `accept` replaces the toolbar rather than seeding it.** The control renders as a locked chip when the field constrains types, so there is nothing to combine.
- **Both dropzones enforce it.** The field's own dropzone used to fall back to the global `media.accept`, so a dropped file was uploaded and only then refused by the insert guard, leaving it in the library.
- **MIME keys, not bare extensions.** react-dropzone keys `accept` by MIME type and, with `useFsAccessApi` on by default, drops extension-only keys from the native picker — which then offers every file and rejects the choice afterwards.
- **`jpg` and `jpeg` resolve to each other**, so `accept: 'jpeg'` does not reject `photo.jpg`.
- **Works without TinaCloud.** The v2 endpoint filters server-side; local dev and `staticMedia` are filtered per page, so `extensionFilterable` is honest for every mode of the store. Local filtering happens after pagination, so a page can come back near-empty while matches sit further on — cloud filters before paginating and does not.
- **Extension-less filenames fall back to `src`.** Cloudinary's `original_filename` drops the extension, so a filename-only check would refuse every file that store returns.
- **Insert guard is separate from validation.** It runs in `onChange`, not on render, so content saved before an `accept` existed doesn't start failing for editors who didn't cause it.

Screenshot 2026-08-24 at 4 49 16 pm

**Figure: Filter inside of the media manager by category**

Screenshot 2026-08-21 at 4 48 00 pm

**Figure: autocomplete in type def**

## Links

- [tinacms#7481](https://github.com/tinacms/tinacms/issues/7481) - the `accept` option
- [tinacms#7474](https://github.com/tinacms/tinacms/issues/7474) - the toolbar filter
- [tinacms#6996](https://github.com/tinacms/tinacms/issues/6996)
- [tinacloud#4013](https://github.com/tinacms/tinacloud/pull/4013) - server-side `ext` filtering, merged and deployed

---

Written by Claude Opus 5 in Claude Code.


packages/@tinacms/schema-tools/src/types/index.ts
```diff
@@ -288,6 +407,19 @@ export type ImageField = (
...
+     * overriding the global `media.accept`.
+     *
+     * @example
+     * ```ts
+     * accept: 'pdf'
+     * accept: ['png', 'svg']
+     * accept: 'image'
+     * ```
+     */
+    accept?: MediaAccept | MediaAccept[];
```
kulesy, line 422: **Major:** `accept` on a `list: true` image field is silently unenforced. The type allows it, but `resolveField`'s image-list branch builds the item field as a bare `{component: 'image'}` ([resolveField.ts#L41-L49](https://github.com/tinacms/tinacms/blob/06b550ada87c49f642b3d7b894832e2b1847526d/packages/%40tinacms/schema-tools/src/schema/resolveField.ts#L41-L49)) and the list plugin spreads only `field.field` into each item ([list-field-plugin.tsx#L161-L170](https://github.com/tinacms/tinacms/blob/06b550ada87c49f642b3d7b894832e2b1847526d/packages/tinacms/src/toolkit/fields/plugins/list-field-plugin.tsx#L161-L170)), so per-item inputs get no dropzone restriction, no locked chip, and no insert guard. `uploadDir` has the same pre-existing hole, but `accept` is new and the type surface promises it works on galleries - either carry `accept` into the inner field in `resolveField`, or exclude it from the list variant and document that.

packages/tinacms/src/toolkit/core/media-store.default.ts
```diff
@@ -191,6 +192,10 @@ export class TinaMediaStore implements MediaStore {
 
   searchable = true;
 
+  // The v2 list endpoint filters server-side; local and `staticMedia` are
+  // filtered per page in `list`.
+  extensionFilterable = true;
```
kulesy: **Major:** one `extensionFilterable = true` covers three modes, but only the cloud path filters before paginating - local and `staticMedia` filter each page after the fetch, which is the near-empty-grid behaviour #7474 ruled out. This file already has the per-mode pattern (`rename` is set only in local mode, with a comment explaining why), and search gates harder at the call site (`canSearch = !isStatic && !!searchable`, [media-manager.tsx#L546](https://github.com/tinacms/tinacms/blob/06b550ada87c49f642b3d7b894832e2b1847526d/packages/tinacms/src/toolkit/components/media/media-manager.tsx#L546)), so a static store shows this dropdown where it hides search. Either set the flag per mode and hide the control where filtering is per-page, or teach the local dev server the `ext` param - it already handles `search` server-side.

packages/@tinacms/schema-tools/src/types/index.ts
joshbermanssw: Confirmed and fixed in f34374eaa. `resolveField` now carries `accept` into the item field of the list variant, so the per-item input gets the dropzone restriction, the locked chip and the insert guard. Chose that over excluding it from the list variant because a gallery is one of the motivating examples in #7481.

`uploadDir` still has the hole you noted — left alone as pre-existing and out of scope here.

Covered by `resolveField.image.test.ts`: accept on the item field for a list, absent when undeclared, and still on the field itself when not a list.

packages/tinacms/src/toolkit/core/media-store.default.ts
joshbermanssw: You're right, and #7474 says it outright — client-side filtering "would show a near-empty grid while the real matches sit further down the list" is the stated reason filtering belongs on the server. The per-page pass reintroduced exactly that. Fixed in f34374eaa by taking the second option you offered.

- The local dev server now accepts `ext` on `/media/list` and applies it before paginating, in the same pass `search` already uses. Client-side `filterByExtension` is gone.
- `extensionFilterable` is derived rather than a flat `true`: `get extensionFilterable() { return !this.isStatic }`. `staticMedia` is a build-time snapshot with no filtering pass, so it reports no support and the control stays hidden — same gating shape as `canSearch`.

Tests: six in the dev server covering filter-before-paginate, case-insensitivity, blank input as no-filter, dotfiles, and combining with `search`; plus store-side coverage that `ext` reaches both endpoints and that the client no longer re-filters the response.

kulesy (approved):

PR #7478, opened 2026-08-21
fix: Boolean field's checkbox has no id, so its label is not associated with ...

### LINKED ISSUE #7473: Boolean field's checkbox has no id, so its label is not associated with it
The `boolean` field renders a visually-hidden `<input type="checkbox">` behind a styled switch. The input's `id` and the switch label's `htmlFor` are both meant to come from a `name` prop that never arrives, so both render empty and the label is not associated with any control.

### The exact error message

There isn't one — nothing is logged and nothing throws. The defect is visible only in the DOM:

```json
{ "type": "checkbox", "name": "enabled", "id": "" }
```

and the corresponding label:

```html
<label class="bg-none p-0 outline-none w-12 h-7" role="switch">
```

with no `for` attribute.

### Steps to reproduce

1. Add a `boolean` field to any collection.
2. Open that document in the admin.
3. Inspect the toggle in devtools.
4. Observe the `<input type="checkbox">` has `id=""`, and the sibling `<label role="switch">` has no `for`.

Clicking the visible switch still works, because the hidden input sits on top of it with `z-20` and receives the click directly rather than via label association.

**Expected:** the checkbox carries `id="<field name>"` and the label carries the matching `for`, so assistive technology announces the label with the control and clicking the label text toggles it.

**Actually:** both attributes are empty. Screen readers get an unlabelled checkbox, and the label is inert.

### Root cause

`Toggle` destructures a top-level `name` prop:

```tsx
// packages/tinacms/src/toolkit/fields/components/toggle.tsx
export const Toggle: FC<ToggleProps> = ({ input, field, name, disabled = false }) => {
  // ...
  <ToggleInput id={name} type='checkbox' {...input} />
  <label htmlFor={name} role='switch'>
```

but `wrapFieldsWithMeta` only passes `name` to the surrounding `FieldMeta`, and renders the component with `<Field {...props} />`:

```tsx
// packages/tinacms/src/toolkit/fields/plugins/wrap-field-with-meta.tsx
<FieldMeta name={props.input.name} ...>
  <Field {...props} />
</FieldMeta>
```

`props` carries `input`, `field`, `meta`, `tinaForm` and `index`, but no top-level `name`, so `name` is `undefined` inside `Toggle`. `input.name` is populated correctly, which is why the `name` attribute is right while `id` is blank.

Likely fix is to use `input.name` for both `id` and `htmlFor` rather than the unpassed prop.

### Your environment

- `tinacms` 3.12.0 (reproduced on `main`)
- Reproduced in `examples/next/kitchen-sink` against the `announcement` collection's `enabled` field, and independently in a production admin build
- Nothing non-default: no custom field component, no custom auth, local mode

```ts
{
  type: 'boolean',
  label: 'Enabled',
  name: 'enabled',
}
```

Roshan931:
Fixes #7473

## Root cause

Toggle reads an unpassed `name` prop for the checkbox id

## Changes

- Derive the checkbox id from `input.name` when no explicit `name` prop is supplied, and use that same value for the switch label's `htmlFor`. Since `FieldMeta`'s label already points at `input.name`, the visible field label now names the checkbox. Added a focused unit test covering both the wrapped-field path (no `name` prop) and the password plugin path (explicit `name`).

### DIFF
--- a/.changeset/toggle-field-checkbox-id.md
+++ b/.changeset/toggle-field-checkbox-id.md
@@ -1,0 +1,5 @@
+---
+'tinacms': patch
+---
+
+Associate the toggle field's label with its checkbox. The input now falls back to the form field name for its `id`, so clicking the label toggles the value and screen readers announce the field name.
--- a/packages/tinacms/src/toolkit/fields/components/toggle.tsx
+++ b/packages/tinacms/src/toolkit/fields/components/toggle.tsx
@@ -3,7 +3,7 @@ import * as React from 'react';
 import { Field } from '@toolkit/forms';
 
 export interface ToggleProps {
-  name: string;
+  name?: string;
   input: any;
   field: ToggleFieldDefinition;
   disabled?: boolean;
@@ -26,6 +26,7 @@ export const Toggle: FC<ToggleProps> = ({
   disabled = false,
 }) => {
   const checked = !!(input.value || input.checked);
+  const inputId = name ?? input.name;
   let labels: null | FieldLabels = null;
 
   if (field.toggleLabels) {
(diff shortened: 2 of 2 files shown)


joshbermanssw:
G'day @Roshan931 - Thanks for the PR! Could you please fix the CI and add a changeset?

Roshan931:
Exported `ToggleFieldDefinition`, which is what the failing types build could not name once `name` became optional, and added a `tinacms` patch changeset.

joshbermanssw (approved):
LGTM

joshbermanssw (approved):

PR #7480, opened 2026-08-23
⏪ Revert the media accept commits that bypassed review

joshbermanssw:
## TL;DR

Four commits (`46fd5d0`, `780ab54`, `57ea7f2`, `a711b88`) landed on `main` directly instead of through a pull request. This reverts them; the work itself is unchanged and lands via #7477.

The resulting tree is identical to `d6a0a2f`, the commit before they were applied.

## Reviewer notes

- **Nothing is lost.** Every line reverted here exists on the `jb/media-field-accept` branch behind #7477.
- **#7477 needs a re-apply after this merges.** Its branch shares those four commits as ancestors, so a plain merge would keep them reverted.

## Links

- [#7477](https://github.com/tinacms/tinacms/pull/7477)

---

Written by Claude Opus 5 in Claude Code.


joshbermanssw:
Superseded - the revert was applied directly to main as 74ae2c5.

PR #7483, opened 2026-08-24
fix(tinacms): guard the admin against an undefined collection or document

### LINKED ISSUE #7482: Admin shows another collection's documents, or throws, when the session check starts failing
**Pain**

[#7445](https://github.com/tinacms/tinacms/pull/7445) fixed the collection list spinning forever when the TinaCloud session check says the user is not signed in, by moving `setLoading(false)` out of the `isAuthenticated()` branch. Two problems took its place.

On a navigation the hooks never clear the previous fetch, so `collection` is stale rather than undefined. Reproduced in the kitchen-sink admin: sign in, load Posts, drop the session, click Authors, and the panel renders the Posts documents under the Authors route with no indication anything is wrong. Against 3.12.0 the same sequence gives the infinite spinner instead, so this is new.

On a fresh mount there is nothing to be stale, so `collection` is undefined with no error set, `GetCollection` runs its auto-open effect against it and reads `collection.name`, and the top-level `ErrorBoundary` catches the throw and replaces the admin with a raw `Cannot read properties of undefined (reading 'name')` card. The fetch-failure path reaches the same line and predates #7445.

**Approach**

Clear the collection and document in the hooks when the session check fails, and guard the undefined case in both components so they render the error screen rather than handing `undefined` to their children.

**Acceptance Criteria**

- Switching collections after the session goes away shows an error, never another collection's documents
- A fresh mount with a failing session check renders an error screen, not a raw TypeError card
- A failed `fetchCollection` renders an error screen rather than throwing
- Tests cover each case and fail without the fix

kulesy:
Closes #7482

**TL;DR** When the TinaCloud session check starts failing, the admin now returns the user to the login modal with a "Your session has ended" message, instead of showing another collection's documents or crashing.

**Pain:** [#7445](https://github.com/tinacms/tinacms/pull/7445) fixed the collection list spinning forever when the session check says the user is not signed in, by moving `setLoading(false)` out of the `isAuthenticated()` branch. That introduced two problems in its place. On a navigation the hooks never clear the previous fetch, so `collection` is stale rather than undefined: the list renders the last collection's documents under the new collection's heading, with no indication anything is wrong. On a fresh mount there is nothing to be stale, so `collection` is undefined with no error set, `GetCollection` runs its auto-open effect against it and reads `collection.name`, and the top-level `ErrorBoundary` catches the throw and replaces the admin with a raw TypeError card. The hook tests added with that PR assert exactly the undefined state and pass, because they exercise the hooks rather than the components that consume them.

**Solution:** a failed session check now dispatches `cms:session-expired`; the auth wall subscribes, un-latches, and shows the login modal with a "Your session has ended. Please sign in again to continue editing." error line, and signing back in lands on the route the user was already on. The hooks clear their state so nothing stale survives, `GetCollection` and `GetDocument` render an "Unable to load" popup (the admin's standard modal chrome) rather than handing `undefined` to their children for non-auth load failures, and `GetCollection`'s auto-open effect bails out early, which also clears the same crash on a failed `fetchCollection`. Tests cover the stale-collection path (including the event dispatch), the logged-out component renders, and the fetch-failure path, each verified to fail without the fix.


kulesy:
Verified in the kitchen-sink admin (local mode, packages built from this branch). Steps for every screenshot: sign in, load the Posts collection, drop the session, then click the Authors collection.

## Before

tinacms 3.12.0: infinite spinner

Figure: tinacms 3.12.0 (latest release) - the collection list spins forever

main: stale Posts under the Authors route

Figure: main (unreleased, after #7445) - the sidebar and route say Authors, but the panel silently renders the Posts documents

## After (this PR)

login modal with session-ended message

Figure: the user goes straight back to the login modal, with a "Your session has ended" message explaining why

Authors renders correctly after signing back in

Figure: signing back in lands on the Authors route the user was heading to, with the correct documents and the message cleared

For load failures that are not auth-related (e.g. an unknown collection slug), an "Unable to load" popup in the admin's standard modal chrome renders instead of the previous full-screen takeover.

joshbermanssw (approved):

PR #7486, opened 2026-08-24
fix(tinacms): detect session expiry at the content API client

### LINKED ISSUE #7485: Session expiry during a save shows a generic error instead of returning to login
**Pain**

PR https://github.com/tinacms/tinacms/pull/7483 returns the user to the login modal when the TinaCloud session check fails, but the detection lives in the admin's collection/document data hooks, so only loads trigger it. Every other authenticated call goes through `Client.request` or `fetchWithToken`, which don't participate: a session that expires while a document is open for editing is caught by `CollectionUpdatePage`'s `onSubmit` and shown as a generic "There was a problem saving your document" dialog. The same underlying failure gets two different user experiences depending on which operation trips over it.

Two mechanical causes underneath: `Client.request` throws the same generic error for any non-200 without distinguishing a 401, and the `registerApi` event bridge forwards `api.events` to the global bus with an unbound `dispatch`, so the api-to-cms direction documented on `Client.events` has never actually delivered an event.

**Approach**

Dispatch `cms:session-expired` and throw a typed `SessionExpiredError` from `Client.request` on a 401, fix the `registerApi` bridge (bound forwarders with re-entrancy guards), and have the admin's error handlers skip the new error so nothing paints over the login modal.

**Acceptance Criteria**

- A save, delete, rename or folder creation against a dead session returns the user to the login modal with the session-ended message, not a generic error dialog
- Non-401 failures keep their existing error messages and never trigger the re-login flow
- An event dispatched on a registered api's own bus reaches global subscribers, and the two-way bridge cannot loop

kulesy:
Closes #7485

**TL;DR** A session that expires mid-edit returns the user to the login modal on every path - GraphQL and REST transports, saves, deletes, renames, media, search - with an alerts suppression window guaranteeing nothing paints over the login modal, and the event bridge this rides on fixed after being a silent no-op since it shipped.

**Pain:** #7483 put session-expiry detection in the admin's collection/document data hooks, so only loads triggered the re-login flow. Every other authenticated call fell through to its own ad-hoc failure: generic "problem saving" dialogs over the login modal, a "Document was not deleted" toast, the media manager's misleading static Unauthorized panel, broken branch/billing panels, and a success toast for a save that never ran. Wiring the client in exposed two deeper faults: `registerApi` bridged `api.events` to the global bus with an unbound `dispatch` whose `this.listeners` guard silently dropped every api-to-cms event since the api-events RFC landed, and `getUser` swallowed any network failure as "logged out", so one transient identity-API blip could eject a validly signed-in user.

**Solution:** the session-expired contract (typed `SessionExpiredError`, a type-predicate guard, the dispatch helper) lives in `toolkit/core` and is shared by all dispatch sites. `Client.request` dispatches and throws on a GraphQL 401 (draining the body, and logging a clientId/branch diagnostic for custom content APIs whose misconfiguration 401s indistinguishably); `fetchWithToken` notifies the same flow when a tokened REST request 401s, covering branches, billing, editorial polling, search indexing and media. The auth wall suppresses new alerts between expiry and re-login as an enforcement net over the per-catch-site guards, `getUser` retries once and then surfaces network failures as errors rather than expiry, the save helpers signal their submit handlers so no success toast fires for an unsaved document, session expiries are no longer recorded as save-failure analytics, and the bridge forwards both directions with guards scoped to the in-flight event so nested dispatches still bridge - with the long-skipped `registerApi` marker test revived to pin it. Every behavioral fix is red-checked; suite is 58 files / 709 tests green, plus a live kitchen-sink walkthrough (screenshots in the comment below).


kulesy:
Verified in the kitchen-sink admin (local mode). Steps for every screenshot: sign in, then the content API starts returning 401 (fetch patched, standing in for a server-side session kill) while the local session key is still present. Every screenshot is the real UI; the Before set is built from main at the 3.12.1 release commit, the After set from this branch (behavior re-verified by DOM assertion against the final head 71549e8f2, which adds the transport-wide coverage and alert suppression on top). examples/shared shows zero content changes afterwards - the 401 genuinely blocked every mutation.

## Before (tinacms 3.12.1)

raw GetCollection error over the broken error screen

Figure: switching collections - a raw "[Error] GetCollection failed: Unable to complete request, Unauthorized" dialog over the generic error screen; nothing mentions the session and there is no way back to login

Document was not deleted toast

Figure: deleting a document - "Document was not deleted, ask a developer for help or check the console" with no re-login path

There was a problem saving your document dialog

Figure: saving in visual editing - "There was a problem saving your document" with the raw Unauthorized error; the user is told to check an Error FAQ rather than to sign in

## After (this PR)

login modal with session-ended message

Figure: all three flows above now land here - the login modal with the reason, and no error dialogs painted over it (DOM assertions on each flow: warnToast false, problemDialog false, loginModal true)

Authors after re-login

Figure: signing back in lands on the route the user was heading to

Not demonstrable in local mode, resting on code-level verification instead: the editorial-workflow delete guard (needs a TinaCloud protected branch) and the CheckSchema guard (runs at mount, before an in-page fetch patch can exist) - both are one-line guards on the same mechanism shown above.

harley-j (approved):
🚀 LGTM

Issue #6924, opened 2026-05-19
Field Type Implementation - Select


18-th:
## Pain

Select fields are simple on the surface, but option handling, defaults and validation need to be consistent. If select behaviour stays tied to form internals, option rendering and stored values can drift from the shared field contract.

## Approach

Implement `SelectField` as a field plugin against the field contract. The plugin owns option rendering, selected value handling, ingest, digest and validation while preserving existing schema and storage behaviour.

## Expected outcomes

- Select fields work through the v4 field plugin system.
- Options, defaults and selected values are handled consistently.
- Select field behaviour is isolated and testable.
- Custom select configuration can use the metadata wrapping pattern.

## Acceptance Criteria

- `SelectField` exists as a field plugin that conforms to the field contract.
- Ingest and digest behaviour preserves the existing select schema and stored value behaviour.
- Rendering supports the expected editor control for selecting from configured options.
- Empty, default and invalid option values are handled clearly.
- Validation runs through the shared field validation path.
- Metadata wrapping works for custom select configuration.
- Tests cover rendering, option selection, defaults, invalid values, ingest, digest and validation.

kulesy:
✅ Done

PR #7489, opened 2026-08-26
feat(v4): add select field

brookjeynes-ssw:
This PR adds support for a `select` tag.

.. code:: tsx

	const collection = {
		name: 'post',
		fields: [
			t.select({
				name: 'status',
				label: 'Status',
				required: true,
				options: [
					{ value: 'draft', label: 'Draft' },
					{ value: 'published', label: 'Published' },
				],
			}),
		],
	};

Issue: https://github.com/tinacms/tinacms/issues/6924 Field Type Implementation - Select


.changeset/fifty-geckos-open.md
```diff
@@ -0,0 +1,7 @@
+---
+"@tinacms/tinacms": patch
+"@tinacms/ui": patch
+"@examples/v4-barebones": patch
```
kulesy: **Major:** This changeset lists `@examples/v4-barebones`, but `.changeset/config.json` ignores `@examples/*`, and changesets rejects mixed changesets at version time (`assemble-release-plan@6.0.9` throws "Found mixed changeset ... Mixed changesets that contain both ignored and not ignored packages are not allowed"), so the changesets/action step in `publish.yml` will start failing on main after this merges. Dropping this line fixes it - though prior v4 feature PRs shipped no changeset at all since `@tinacms/tinacms` and `@tinacms/ui` are both `private: true` alphas, so the whole file may be unnecessary.

packages/v4/@tinacms/tinacms/src/plugins/fields/select/select-field.ui.tsx
```diff
@@ -0,0 +1,53 @@
...
+        items={items}
+        value={value ?? null}
+        onValueChange={(newValue) => setValue(newValue ?? undefined)}
+      >
+        <SelectTrigger ref={triggerRef} id={address}>
+          <SelectValue placeholder='Select...' />
+        </SelectTrigger>
+        <SelectContent>
+          {field.options.map((option) => (
+            <SelectItem key={option.value} value={option.value}>
```
kulesy, line 52: **Major:** Only `field.options` are rendered and base-ui's Select has no built-in clear, so once an author picks a value on a non-required field there is no way to return it to absent - the schema side anticipates exactly this (`selectSchema`'s optional branch preprocesses `''`/`null` to `undefined`) but the UI never offers an empty item, and v3 handles it by prepending `{ label: 'Choose an option', value: '' }` to every select ([resolveField.ts:87](https://github.com/tinacms/tinacms/blob/50eb230caa823939a9cf4f05f79410c15dbb9696/packages/%40tinacms/schema-tools/src/schema/resolveField.ts#L87)). Suggest rendering a null item (e.g. "None") when `field.required` is not set, mapped to `undefined` through the existing `onValueChange`.

brookjeynes-ssw: Resolved by 8141939198edd3a02c5d6643e75f47ea4ca9dc34

kulesy (approved):

PR #7491, opened 2026-08-26
fix(@tinacms/cli): surface remote GraphQL errors in schema check

### LINKED ISSUE #7493: tinacms build reports "remote GraphQL schema does not exist" when TinaCloud returns a GraphQL error
### The exact error message

```
The remote GraphQL schema does not exist. Check indexing for this branch.

Additional info: Branch: <branch>, Client ID: <client id>
```

Reported by `tinacms build` with error code `ERR_CLOUD_CHECK_FAILED`.

### Steps to reproduce

1. Create a schema that TinaCloud's content API rejects at query time. One real example: a rich-text block template whose `fields` is an empty array, which makes the server respond to every request with `Input Object type <Template>Filter must define one or more fields.`
2. Push the branch and let TinaCloud index it (indexing succeeds).
3. Run `tinacms build`.

Expected: the build fails with the error the content API actually returned, pointing at the schema problem.

What actually happened: `fetchRemoteGraphqlSchema` in `packages/@tinacms/cli/src/next/commands/build-command/index.ts` returns `data?.data` without checking `res.status` or inspecting the body for `errors`. A GraphQL error response has an `errors` array and no `data`, so `checkGraphqlSchema` treats the schema as missing and tells the user to check indexing, while the TinaCloud dashboard shows indexing succeeded. In a real support case the user spent days investigating indexing when the server was returning the exact cause on every request.

### Your environment

@tinacms/cli 2.6.1 (also current main), building against TinaCloud. No non-default setup required beyond a schema the content API rejects.

kulesy:
Closes #7493

**TL;DR** `fetchRemoteGraphqlSchema` now throws the server's actual GraphQL error instead of swallowing it. The generic "check indexing" message is kept only for a genuinely empty schema, and the public API surface is unchanged.

**Pain:** when the remote schema check failed, `tinacms build` always printed "The remote GraphQL schema does not exist. Check indexing for this branch." Two production cases cost customers days, because the real error was sitting in the response body the whole time.

**Solution:** surfaces the server's error message and HTTP status, and adds a stale `tina/tina-lock.json` hint when the schema reports zero collections.


kulesy:
Before/after of the error `tinacms build` prints when the content API responds with a GraphQL errors payload. Reproduced by running the real `fetchRemoteGraphqlSchema` from main and from this branch against a local mock of the content API returning the exact payload from the reported case (HTTP 200 with an `errors` array), with the caller's message handling applied verbatim; not captured from a live TinaCloud project.

**Before (main):**

```
The remote GraphQL schema does not exist. Check indexing for this branch.

Additional info: Branch: main, Client ID: demo-client-id
```

**After (this branch):**

```
The remote GraphQL API returned an error: Input Object type PageBodySponsorshipTiersFilter must define one or more fields.

Additional info: Branch: main, Client ID: demo-client-id
```

kulesy:
Before and after, running the real `fetchRemoteGraphqlSchema` against a local stub server replaying the exact payload from the second production case (`DocumentFilter` and `DocumentMutation` reported empty on an HTTP 200):

```
=== BEFORE (main) ===
The remote GraphQL schema does not exist. Check indexing for this branch.

Additional info: Branch: main, Client ID: <clientId>

=== AFTER (this PR) ===
The remote GraphQL API returned an error: Input Object type DocumentFilter must define one or more fields.
Input Object type DocumentMutation must define one or more fields.

The remote schema has no collections, which usually means tina/tina-lock.json is out of date. Only `tinacms dev` regenerates that file, so a project that only runs `tinacms build` keeps publishing a stale lock file. Run `tinacms dev` locally, then commit and push tina/tina-lock.json.

Additional info: Branch: main, Client ID: <clientId>
```

Captured locally against a stub server, not against production TinaCloud. This is CLI output, so the terminal text above is the whole of the user-visible surface; there is no UI to screenshot.

**Test plan**

- [x] 9 unit tests in `fetch-remote-graphql-schema.test.ts` covering: success, GraphQL errors on a 200, GraphQL errors on a non-2xx, non-2xx with a non-GraphQL body, non-2xx with an unparseable body, 200 with an unparseable body, 200 with no schema, the stale lock hint firing, and the stale lock hint *not* firing for a per-template error.
- [x] `jest` in `@tinacms/cli`: 338/338 pass.
- [x] The negative test was mutation checked: adding `PageBodySponsorshipTiersFilter` to the hint list makes it fail, so it is a real guard rather than a vacuous assertion.
- [x] `biome check` clean on the changed files. `tsc` output is identical to main's, which is pre-existing unbuilt-sibling-package errors only.
- [x] Rebased onto `7c2190666` with no conflicts.
- [x] Patch changeset for `@tinacms/cli`.

packages/@tinacms/cli/src/next/commands/build-command/fetch-remote-graphql-schema.ts
```diff
@@ -0,0 +1,81 @@
...
+
+  let data;
+  try {
+    data = await res.json();
+  } catch {
+    data = undefined;
+  }
+
+  if (data?.errors?.length) {
+    const messages = data.errors.map((error) => error.message);
```
joshbermanssw: Tiny thing, but I think this one can bite.

We only reach this branch because the server misbehaved, and `error.message` assumes it misbehaved politely. If an entry in `errors` isn't a `{ message: string }`, `messages` picks up an `undefined`, and `staleLockFileHint` then calls `.includes()` on it:

```
{ errors: ['Rate limited'] }  ->  TypeError: Cannot read properties of undefined (reading 'includes')
{ errors: [{ code: 'X' }] }   ->  TypeError: Cannot read properties of undefined (reading 'includes')
```

A gateway or CDN sitting in front of TinaCloud answering with `{"errors":["..."]}` is a pretty ordinary thing to hit, and it has no reason to follow the GraphQL spec.

What makes it worth fixing is where it lands: the `TypeError` escapes and hides the server's real error, which is the same swallowing this PR exists to stop. You already wrap `res.json()` for exactly this reason, so it's really just extending the same suspicion one line further.

```suggestion
    const messages = data.errors.map((error) =>
      typeof error?.message === 'string' ? error.message : JSON.stringify(error)
    );
```

Falling back to `JSON.stringify` instead of dropping the entry keeps something useful in the message, and everything stays a string so `staleLockFileHint` is happy either way.

packages/@tinacms/cli/src/next/commands/build-command/fetch-remote-graphql-schema.ts
```diff
@@ -0,0 +1,81 @@
...
+  });
+
+  let data;
+  try {
+    data = await res.json();
+  } catch {
+    data = undefined;
+  }
+
+  if (data?.errors?.length) {
```
joshbermanssw: Genuine question rather than a request.

This throws whenever `errors` is present at all, before looking at whether a schema actually came back. GraphQL does allow a partial response carrying both `data` and `errors`, so a build that used to pass would now fail hard on one of those.

For introspection specifically I suspect you're fine, since a schema that fails to build returns no data at all, and all three production cases in your description are that shape. So this may well be deliberate.

If you'd rather be conservative given it's going out as a patch:

```suggestion
  if (data?.errors?.length && !data?.data) {
```

That keeps the loud failure for every case you've actually hit, and leaves a partial response working. Happy either way, just flagging it because it's a behaviour change rather than a pure message change.

packages/@tinacms/cli/src/next/commands/build-command/index.ts
```diff
@@ -705,11 +704,22 @@ export class BuildCommand extends BaseCommand {
...
+      ({ remoteSchema, remoteProjectVersion } = await fetchRemoteGraphqlSchema({
         url: apiURL,
         token,
+      }));
+    } catch (e) {
+      bar.tick({
+        prog: '❌',
       });
+      if (e instanceof Error && config?.branch) {
+        e.message += `\n\nAdditional info: Branch: ${config.branch}, Client ID: ${config.clientId} `;
```
joshbermanssw: Nit, very much take it or leave it.

This is the same two lines as the block just below at L730, trailing space and all. Matching the existing shape was the right instinct, but since there are now two copies three lines apart, might be worth a small local helper while you're in here:

```ts
const withProjectInfo = (message: string) =>
  config?.branch
    ? `${message}\n\nAdditional info: Branch: ${config.branch}, Client ID: ${config.clientId}`
    : message;
```

Then this becomes `e.message = withProjectInfo(e.message)`, and the one below becomes `throw new Error(withProjectInfo('The remote GraphQL schema does not exist. Check indexing for this branch.'))`.

Definitely not worth a round trip on its own.

packages/@tinacms/cli/src/next/commands/build-command/fetch-remote-graphql-schema.ts
kulesy: Good catch, took your suggestion as is. If something sitting in front of TinaCloud sends back errors that aren't shaped the way GraphQL says they should be, the old code would crash on them and we'd lose the server's real message, which is the exact thing this PR is meant to fix. Added three tests for it: a plain string in the errors list, one with no message on it, and a broken one sitting next to a good one so the lock file hint still shows up.

packages/@tinacms/cli/src/next/commands/build-command/fetch-remote-graphql-schema.ts
kulesy: Took this, with one small change. It now checks that the request actually succeeded as well as that a schema came back. So if the server sends a schema and some errors together, we use the schema like we always did, but if the request itself failed we still show what the server said instead of just a status code. You're right that we've never seen a half answer from the real API, but this is a patch release, so I'd rather not break a build that was working fine. There's a test each way, and both fail if I take the matching half back out.

packages/@tinacms/cli/src/next/commands/build-command/index.ts
kulesy: Done, went a bit further than you suggested. The helper now covers all four places, including the two in the Tina schema check further down, since that one picked up the same duplication in this PR. Kept the trailing space so the message reads exactly the same as it did before.

packages/@tinacms/cli/src/next/commands/build-command/fetch-remote-graphql-schema.ts
kulesy: Good catch, and confirmed. A body like `{"errors":["Rate limited"]}` throws a TypeError inside `staleLockFileHint` and the server's real message is lost, which is the exact swallowing this PR is meant to stop. I've taken your suggestion as-is. I also applied the same normalisation in the new `fetch-schema-sha.ts`, which has the same map. Tests cover a plain string entry, an entry with no `message`, and a null entry sitting next to a real one so the stale-lock hint still fires.

isaaclombardssw (approved):
## 🤖 Agent Review

### 🚨 Must Fix / Address
- ✌️ **None** - CI is green on the head SHA and the 24 new tests pass locally. Every candidate blocker I raised was demoted on verification.

### ⚠️ Notable Changes
- 🎯 **Real root-cause fix at the right layer** - the error TinaCloud was already returning now reaches the user instead of being replaced by a misleading "Check indexing for this branch".
- **Spec-correct GraphQL handling** - `hasUsableSchema = res.ok && Boolean(data?.data)` keeps partial-data-plus-errors responses working, and only treats `errors` as fatal when no schema came back.
- **The stale-lock hint is factually right** - only `dev-command/index.ts:131` writes `tina/tina-lock.json`, so "only `tinacms dev` regenerates that file" holds. That is the hint that closes the support ticket outright.
- **Good extraction** - both helpers move into testable sibling modules matching the existing `server.ts` / `waitForDB.ts` pattern, with 24 fast network-free tests.

### 📞 Callouts
- **`request()` still has the same hazard** - it does `await res.json()` before the `!res.ok` check, so an HTML 502 from `checkClientInfo` still surfaces as a raw SyntaxError in the same build. Pre-existing and untouched by this PR, but the file now carries two opposite patterns for the same problem three functions apart. Worth a follow-up.
- **Deeper cause worth a separate look** - `fields: z.array(TinaFieldZod)` in `@tinacms/schema-tools` has no `.nonempty()`, so an empty-`fields` template passes local validation and is only caught by a TinaCloud round-trip. Unverified, separate change.

### 🤏 Small Feedback
- **Non-array `errors`** - `data?.errors?.length` is truthy for a string, so `{"errors": "..."}` would hit `.map` and throw a TypeError. No plausible producer (GraphQL mandates a list), but `Array.isArray(data?.errors) && data.errors.length` is a free one-liner in both files.
- **Dead re-export** - `export { fetchRemoteGraphqlSchema, fetchSchemaSha }` at `build-command/index.ts:45` has no importers, and `src/index.ts` exports only types plus `default cli`, so it was never published surface.
- **~35 duplicated lines** - the error-message construction is identical across the two new modules, plus `stubResponse` copied byte-for-byte between the test files. A small local helper would collapse both.
- **The twins disagree on `errors` precedence** - `fetchSchemaSha` throws on `errors` even with a valid `tinaSchema`, while `fetchRemoteGraphqlSchema` deliberately tolerates data-plus-errors.
- **`let remoteSchema;`** - implicit evolving `any`, so `buildClientSchema(remoteSchema)` is no longer type-checked.
- **"could not be parsed as JSON"** - a literal `null` or `""` body parses fine and falls through the `!data` guard, misdiagnosing a valid-JSON 200.

Issue #6918, opened 2026-05-19
Field Type Implementation - Array


18-th:
## Pain

The existing `List` property sits outside the normal field model, which makes collection-style editing feel like a special case. That special casing complicates serialization, validation, nested editing and custom field support because arrays do not behave like every other field type.

## Approach

Implement `ArrayField` as the first-class replacement for `List`. The field plugin owns array ingest, digest, rendering and validation, and supports item schemas through the same metadata and field plugin patterns used elsewhere.

## Expected outcomes

- `List` is replaced by a normal field plugin.
- Collections behave like the rest of the field system.
- Array items can use the same field contract, metadata wrapping and validation path as other fields.
- Array field behaviour is isolated and testable.

## Acceptance Criteria

- `ArrayField` exists as a field plugin that conforms to the field contract.
- Existing `List` use cases are represented through `ArrayField`.
- Ingest and digest behaviour preserves the expected collection schema and stored value behaviour.
- Rendering supports adding, removing, reordering and editing array items.
- Validation supports array-level rules and item-level rules.
- Nested item fields integrate with the shared field plugin system.
- Tests cover rendering, add/remove/reorder, nested item editing, ingest, digest and validation.

brookjeynes-ssw:
Done - Closed by #7492

<img width="1365" height="1003" alt="Image" src="https://github.com/user-attachments/assets/e3f0ee9c-d079-41e0-b613-715137c55bae" />

PR #7492, opened 2026-08-26
feat(v4): add array field

brookjeynes-ssw:
This PR adds support for `ArrayField`.

.. code:: tsx

	const collection = {
		name: 'post',
		fields: [
			t.array({
				name: 'authors',
				label: 'Authors',
				required: true,
				fields: [
					t.string({ name: 'name', label: 'Name', required: true }),
					t.string({ name: 'role', label: 'Role' }),
				],
			}),
		],
	};

image

Issue: https://github.com/tinacms/tinacms/issues/6918 Field Type Implementation - Array
Signed-off-by: brookjeynes-ssw brookjeynes@ssw.com.au


packages/v4/@tinacms/tinacms/src/core/field/contract.ts
```diff
@@ -11,6 +12,9 @@ export interface FieldMetadata {
 
 export interface FieldTransformContext {
   documentPath?: string;
+  // The form provider and the save path set this. A compound field uses it
+  // to parse, serialize, and validate its item fields.
+  registry?: FieldRegistry;
```
joshbermanssw, line 17:
```suggestion
  // Optional because only a compound field needs it: to parse, serialize and
  // validate its children it must look up their descriptors by type.
  registry?: FieldRegistry;
```

`registry?: FieldRegistry` already says "a registry, sometimes". The part worth a comment is when and why it's there, rather than who sets it.

packages/v4/@tinacms/tinacms/src/editor/field-errors.ts
```diff
@@ -19,3 +21,67 @@ export const fieldErrorMessages = (
   if (entry.types) return Object.values(entry.types).filter(Boolean);
   return entry.message ? [entry.message] : [];
 };
+
+const ENTRY_KEYS = new Set(['type', 'message', 'types']);
```
joshbermanssw, line 25: Worth stating the two shapes up front. Reading this cold, the "why does this file exist" only clicks once you see them side by side.

```suggestion

// Errors live in two shapes:
//   Tina, flat:              "authors.0.name"
//   react-hook-form, nested:  authors -> 0 -> name
// `nestFieldErrors` converts flat -> nested. `flattenFieldErrors` goes back.

const ENTRY_KEYS = new Set(['type', 'message', 'types']);
```

packages/v4/@tinacms/tinacms/src/editor/hooks.ts
```diff
@@ -108,6 +113,7 @@ export function useFormSave(): () => Promise<void> {
     const values = getValues();
     const digested = digestDocument(values, collection.fields, registry, {
       documentPath: path,
+      registry,
```
joshbermanssw, line 116: isnt registry aleady passed in?

packages/v4/@tinacms/tinacms/src/editor/resolver.ts
```diff
@@ -1,24 +1,24 @@
...
+import { nestFieldErrors } from './field-errors';
 
 export const buildFormResolver =
   (
     collection: CollectionSchema,
     registry: FieldRegistry
   ): Resolver<TinaDocument> =>
   (values) => {
-    const errors: Record<string, FieldErrorEntry> = {};
+    const flat: Record<string, string[]> = {};
```
joshbermanssw, line 13: `flat` is a bit of an empty name for this - can we bring back the errors part n call it 

`flatErrors` 

perhap salso just a comment above 

```
// Record key is the node address and the value is an array of error messages
```

packages/v4/@tinacms/tinacms/src/editor/provider.tsx
```diff
@@ -239,14 +237,20 @@ export function FormProvider({
...
+          // field. A nested field name collapses to its top-level address.
+          const topLevel = name.split('.')[0];
+          store.setFieldValue(
+            formId,
+            toFieldAddress(topLevel),
+            values[topLevel]
+          );
         }
+        const flat: Record<string, string[]> = {};
+        flattenFieldErrors(errors ?? {}, '', flat);
```
joshbermanssw, line 250: Cab we squash these into one line 

`const flat = flattenFieldErrors(errors ?? {})`

Also updating name to `flatErrors` if the other case has been updated 🙏

Note you'll need to change the sig if changed

packages/v4/@tinacms/tinacms/src/editor/provider.tsx
```diff
@@ -239,14 +237,20 @@ export function FormProvider({
...
         const mirrored: FieldErrors = {};
-        for (const [field, entry] of Object.entries(
-          (errors ?? {}) as Record<string, FieldErrorEntry | undefined>
-        )) {
-          const messages = fieldErrorMessages(entry);
-          if (messages.length > 0) mirrored[toFieldAddress(field)] = messages;
+        for (const [field, messages] of Object.entries(flat)) {
```
joshbermanssw, line 252:
```suggestion
        for (const [address, messages] of Object.entries(flat)) {
```

packages/v4/@tinacms/tinacms/src/plugins/fields/array/array-field.schema.ts
```diff
@@ -0,0 +1,40 @@
...
+export const array = (
+  config: Omit<ArrayFieldSchema, 'type'>
+): ArrayFieldSchema => ({ ...config, type: ARRAY_FIELD_TYPE });
+
+const labelOf = (node: ArrayFieldSchema): string => node.label ?? node.name;
+
+const plural = (count: number): string => (count === 1 ? 'item' : 'items');
+
+export const arraySchema = (node: FieldSchema): ZodType => {
+  const field = node as ArrayFieldSchema;
```
joshbermanssw, line 22: Can we check this casting instead? 
Like
```
const isArrayField = (node: FieldSchema): node is ArrayFieldSchema =>
  node.type === ARRAY_FIELD_TYPE && Array.isArray((node as ArrayFieldSchema).fields);
```

then we check on this line instead 

an array node right now with 0 fields will go past the cast and explode later when the `undefine` fields is accessed

packages/v4/@tinacms/tinacms/src/plugins/fields/array/array-field.schema.ts
```diff
@@ -0,0 +1,40 @@
...
+export interface ArrayFieldSchema extends BaseFieldSchema {
+  type: typeof ARRAY_FIELD_TYPE;
+  fields: FieldSchema[];
+  min?: number;
+  max?: number;
+}
+
+export const array = (
+  config: Omit<ArrayFieldSchema, 'type'>
+): ArrayFieldSchema => ({ ...config, type: ARRAY_FIELD_TYPE });
```
joshbermanssw, line 15: `required: true` with `min: 0` is a contradiction, and right now `??` silently lets `min` win, so `required` is quietly dropped... this may need to be expanded out to a config file to be re-used on other fields 

```suggestion
export const array = (
  config: Omit<ArrayFieldSchema, 'type'>
): ArrayFieldSchema => {
  if (config.required && config.min === 0) {
    throw new Error(
      `Array field "${config.name}" sets both \`required: true\` and \`min: 0\`. Remove one.`
    );
  }
  return { ...config, type: ARRAY_FIELD_TYPE };
};
```

Putting it here rather than in `arraySchema` keeps it out of the resolver, which runs on every keystroke.

packages/v4/@tinacms/tinacms/src/plugins/fields/array/array-field.client.tsx
```diff
@@ -0,0 +1,61 @@
...
+  return context.registry;
+};
+
+export default defineClientPlugin({
+  field: {
+    Component: ArrayField,
+    metadata: { layout: 'block', labelable: false },
+    schema: arraySchema,
+    parse: (stored: unknown, node, context) => {
+      const field = node as ArrayFieldSchema;
```
joshbermanssw, line 25: same node casting issue as mentioned in `array-field.schema.ts`

joshbermanssw (commented):
Couple things

packages/v4/@tinacms/tinacms/src/editor/provider.tsx
```diff
@@ -239,14 +237,20 @@ export function FormProvider({
...
         const store = useFormStore.getState();
         if (name !== undefined) {
-          store.setFieldValue(formId, toFieldAddress(name), values[name]);
+          // The store's live-values mirror is flat, one entry per top-level
+          // field. A nested field name collapses to its top-level address.
+          const topLevel = name.split('.')[0];
+          store.setFieldValue(
+            formId,
+            toFieldAddress(topLevel),
+            values[topLevel]
```
kulesy, line 245: After a save, editing a field inside an item doesn't re-enable the Save button, because the store and the form end up sharing the same array so nothing looks changed.

packages/v4/@tinacms/tinacms/src/plugins/fields/array/array-field.client.tsx
```diff
@@ -0,0 +1,61 @@
...
+      const registry = requireRegistry(context);
+      const items = Array.isArray(stored) ? stored : [];
```
kulesy: Content that isn't already an array of objects gets blanked out here, and the blank version is what gets written back on the next save.

packages/v4/examples/barebones/tina/config.ts
```diff
@@ -19,9 +19,25 @@ export const postCollection = {
   fields: [
     t.string({ name: 'title', label: 'Title', required: true }),
     t.boolean({ name: 'featured', label: 'Featured' }),
+    t.number({
+      name: 'flat-validation-test',
```
kulesy: This one looks left over from testing, and it makes the example open with a validation error.

packages/v4/@tinacms/tinacms/src/core/field/contract.ts
kulesy: The registry is optional here but the array field can't work without it, so a missing one blows up at runtime instead of being caught by the compiler.

packages/v4/@tinacms/tinacms/_docs/field-plugins.md
```diff
@@ -218,3 +235,57 @@ Do no more steps. You do not change the registry.
...
+  `type`/`message` — sitting alongside it. `useFieldErrors(address)`
+  (`editor/hooks.ts`) gets a compound field's own message from its children
+  instead, by reading: `collectFieldErrorMessages`
+  (`editor/field-errors.ts`) walks every node under `address` in whatever
+  tree react-hook-form actually kept, and collects every message it finds.
+  So `useFieldErrors('authors')` sees an item's message the same way
+  `useFieldErrors('authors.0.name')` does, at any depth, without the resolver
+  needing to place it there.
+
+This makes every ordinary field an item field for free — an `array`'s `fields`
```
kulesy, line 288: Not quite true for isEqual, which only gets applied to top-level fields, so a rich-text inside an array loses its equality check.

packages/v4/@tinacms/tinacms/src/editor/hooks.ts
kulesy: Yeah it is, which is why the optional copy on the context is worth a second look - a missing one gets past the compiler and only turns up at runtime.

kulesy (approved):

Issue #5432, opened 2025-01-18
TinaMarkdown component renders table and td elements with unnecessary inline styling by default


geotrev:
### Describe the issue that you're seeing. Any Loom videos or screenshots usually help a lot!

In the latest build (2.6.0), `TinaMarkdown` will render tables by default:

- With solid borders as inline styles. The inline styles are on the `table` and `td` elements.
- The `type` prop leaks into the element attributes

<img width="507" alt="Image" src="https://github.com/user-attachments/assets/105593c4-ef20-4cc7-82fb-ecd53473a3e8" />
<img width="278" alt="Image" src="https://github.com/user-attachments/assets/0d87ad00-a929-47f2-bb53-b6b796107d3d" />

Deploy from the repro repository: https://tinamarkdowntablebug.netlify.app/

I've observed this independently in two projects. Only way to fix is to provide a custom `components` prop that overrides the elements:

```
{
  table: ({ children }) => <table>{children}</table>,
  td: ({ children }) => <td>{children}</td>,
}
```

### Reproduction

https://github.com/geotrev/tinacms-markdown-tables-bug

### Steps to reproduce

1. Create a Next JS app and init Tina with a schema containing a `rich-text` field type
2. Add a table to the rich text field, then render it to a page

Expected: Table renders with no inline styles

Actual: Table renders with inline styles

### System Info

```shell
System:
    OS: Linux 5.15 Ubuntu 22.04.4 LTS 22.04.4 LTS (Jammy Jellyfish)
    CPU: (32) x64 Intel(R) Core(TM) i9-14900K
    Memory: 29.98 GB / 31.23 GB
    Container: Yes
    Shell: 5.8.1 - /usr/bin/zsh
  Binaries:
    Node: 20.12.1 - ~/.nvm/versions/node/v20.12.1/bin/node
    Yarn: 1.22.22 - ~/.nvm/versions/node/v20.12.1/bin/yarn
    npm: 10.5.0 - ~/.nvm/versions/node/v20.12.1/bin/npm
  Browsers:
    Chrome: 124.0.6367.155
  npmPackages:
    @tinacms/cli: 1.8.0 => 1.8.0
    tinacms: 2.6.0 => 2.6.0
```

### Validations

- [x] Follow our [Code of Conduct](https://github.com/tinacms/tinacms/blob/main/CODE_OF_CONDUCT.md)
- [x] Read the [docs](https://tina.io/docs).
[...]

PR #7494, opened 2026-08-26
fix(rich-text): render semantic thead/th for markdown tables

### LINKED ISSUE #7169: ✨ Rich-text: render semantic <thead>/<th> for markdown tables in TinaMarkdown
## Pain

`TinaMarkdown` renders a markdown (GFM) table with every row inside a single `<tbody>` as `<td>` - it never emits `<thead>` or `<th>`. This diverges from standard GFM (GitHub, remark, markdown-it all render the header row as `<thead><th>`), and it means:

- Header cells can't be targeted or styled - there is no `<th>`/`<thead>`, and the `th` component override never fires for markdown tables.
- The output is not semantic or accessible - assistive tech relies on `<th>` (and scope) to associate headers with their cells.

Reported twice in Discord #ask-for-help (once as table header styling, once as "make the table component semantic").

Source: `packages/tinacms/src/rich-text/index.tsx`, the `case 'table':` block. The editor-authored MDX `<table>` path a few lines above already emits `<thead>/<th>` when a header row is set; markdown tables just do not.

## Approach

In `case 'table'`, treat the first row as the header (GFM guarantees row 0 is the header): render it as `<thead>` with `<th>` cells through the existing `th` component slot, and the remaining rows as `<tbody>`/`<td>`. This mirrors the existing `firstRowHeader` branch, so it stays consistent with code already in the file, and it makes the `th`/`tr`/`td` overrides work for markdown tables.

While in the same block, fold in the sibling cleanup from #5432 (drop the default inline border styles, stop leaking `type`/`align` as DOM attributes).

Notes:
- This changes rendered DOM for existing sites (anyone styling the header via `td` or `tbody tr:first-child`), so it needs a changeset with a migration note, and possibly an opt-out.
- `<tfoot>` is intentionally out of scope: GFM has no footer syntax, so there is nothing to map to one. A footer would be a separate opt-in convention, not standards work.

## Acceptance criteria

- A markdown table renders as `<table><thead><tr><th>...</th></tr></thead><tbody>...<td>...</td>...</tbody></table>`.
- Column alignment is preserved on both `<th>` and `<td>`.
- The `th`, `tr`, and `td` component overrides apply to markdown tables.
- Covered by a test, and a changeset with a migration note.

Related: #5432

lastenvoy30:
Closes #7169

## What changed
Markdown (GFM) tables previously rendered every row as `<tbody><td>`, with no
`<thead>` or `<th>` — diverging from standard GFM output and breaking
accessibility (assistive tech relies on `<th>` to associate headers with data
cells) and the `th` component override never firing for markdown tables.

The first row of a markdown table is now rendered as `<thead><th>`, mirroring
the existing `firstRowHeader` behavior for editor-authored MDX tables.
Remaining rows render as `<tbody><td>` as before. Column alignment is
preserved on both `<th>` and `<td>`. Also drops the default inline
border/padding styles and stops leaking `align` as a raw DOM attribute,
per the cleanup requested in #5432.

## Breaking change
Sites styling table headers via `td` or `tbody tr:first-child` selectors
will need to update those selectors to target `th`/`thead` instead.
Included in the changeset.

## Testing
Added 4 new tests covering thead/th rendering, tbody/td rendering, header
isolation from body, and alignment preservation. All existing tests
still pass.


packages/tinacms/src/rich-text/index.tsx
```diff
@@ -465,52 +465,58 @@ const Node = ({ components, child }) => {
...
-    case 'table':
+    case 'table': {
       const rows = child.children || [];
+      const [headerRow, ...bodyRows] = rows;
       const TableComponent =
-        components['table'] ||
-        ((props) => (
-          <table style={{ border: '1px solid #EDECF3' }} {...props} />
-        ));
+        components['table'] || ((props) => <table {...props} />);
```
kulesy, line 472: **Dropping the default border and padding is a second break, and it is not in the changeset**

This also removes the default table border and cell padding, which every existing site sees, but the changeset only warns about the header selectors. v4 kept those on purpose: `packages/v4/@tinacms/tinacms/src/rich-text/render.ts:147` keeps border and padding for gfm tables and drops them only for mdx ones, and `packages/@tinacms/astro/src/TableNode.astro:34` keeps them too. The thead change does not need them gone, so either leave them in or call it out in the changeset.

packages/tinacms/src/rich-text/index.tsx
```diff
@@ -465,52 +465,58 @@ const Node = ({ components, child }) => {
...
-            {...props}
-          />
+        (({ align, ...props }) => (
+          <td style={{ textAlign: align || 'auto' }} {...props} />
         ));
       const align = child.props?.align || [];
       return (
         <TableComponent>
+          {headerRow && (
+            <thead>
```
kulesy, line 488: **The Astro renderer mirrors this branch on purpose and gets left behind**

`packages/@tinacms/astro/src/TableNode.astro` renders the same node and its comment says it copies this branch, first row not promoted to a header because that is what React does. Its test asserts that at `packages/@tinacms/astro/src/__tests__/TinaMarkdown.test.ts:172`. Astro is published at 0.6.1, so after this lands the same content renders differently in the two packages. Either update Astro and its snapshot here, or change that comment and test to say the split is deliberate.

kulesy (commented):
Two things worth sorting before this goes in. The core change looks right to me: I checked that a `table` node round-trips through stringify into a GFM pipe table, so row 0 really is always the header, and the MDX path is untouched. Both comments are about the mirrors this stops short of rather than the split itself.

packages/tinacms/src/rich-text/index.tsx
joshbermanssw: Went with calling it out rather than restoring the borders, since #7169's Approach section explicitly asks to fold in the #5432 cleanup in this same block.

`.changeset/tiny-toes-teach.md` now carries a second **Breaking change:** block for the removed `border`/`padding` and the `align` attribute leak, so both breaks are in the changelog rather than just the header selectors.

Worth noting the v3/v4 split this leaves: v4 keeps the GFM border and padding, v3 drops them. If you'd rather v3 match v4 and leave #5432 to be decided once, say so and I'll restore them instead.

packages/tinacms/src/rich-text/index.tsx
joshbermanssw: Updated Astro rather than documenting the split, so the two packages keep rendering the same content the same way.

`TableNode.astro` now promotes row 0 to `<thead>`/`<th>`, honours the `th` override, and drops `TABLE_STYLE` and the cell border/padding so it mirrors the React branch exactly. Rewrote the doc comment that claimed the first row is deliberately not promoted.

The test is renamed and inverted: it asserts `<thead>`/`<th>`, per-column alignment on both tags, and that row 0 does not leak into `<tbody>`. Snapshot regenerated and the stale one removed.

Added a separate `@tinacms/astro` minor changeset so 0.6.1 consumers get their own note instead of the markup changing silently.

`pnpm --filter @tinacms/astro test`: 80 passed, no type errors.

kulesy (approved):

PR #7495, opened 2026-08-26
Show configured Git author in the editorial workflow save modal

Aibono1225:
## Summary

Updates the existing protected-branch save modal for https://github.com/tinacms/tinacms/issues/6984 to show:

```text
Committing as
<configured author>
```

* Personal authoring shows the user's `fullName`, falling back to email
* Bot authoring shows `TinaCloud bot`
* the Change link opens the TinaCloud Git settings page
* normal/non-protected saves are unchanged

## Implementation notes

TinaCMS already fetches the authenticated user during initialization. This change retains that response on the existing `Client` instance and reads `gitAuthoring.mode` when rendering the modal, avoiding an additional request.

If `gitAuthoring` is absent, the new row is not rendered, so the change remains backward-compatible with older TinaCloud deployments and custom auth providers.

## Testing

* full `packages/tinacms` test suite
* formatting checks
* manually verified Personal and Bot modes against a TinaCloud-connected protected-branch project

## Deployment

The UI can ship independently, but the new row remains hidden until TinaCloud deploys the `currentUser.gitAuthoring` field.

## Screenshot
tina2


packages/tinacms/src/internalClient/index.ts
```diff
@@ -124,6 +124,8 @@ export class Client {
   events = new EventBus(); // automatically hooked into global event bus when attached via cms.
   protectedBranches: string[] = [];
   usingEditorialWorkflow: boolean = false;
+  // Whatever the auth provider returned from getUser(); shape varies by provider.
+  user?: any;
```
joshbermanssw: Can we perhaps make this a bit more strict? 

i.e. create the type or use `AuthenticatedUser | boolean`

packages/tinacms/src/toolkit/form-builder/create-branch-modal.tsx
```diff
@@ -178,6 +178,53 @@ export const CreateBranchModal = ({
   );
 };
 
+const getInitials = (name: string) =>
+  name
+    .split(/[\s@._-]+/)
```
joshbermanssw: `@` is in the separator class, so for the `user.email` fallback the domain leaks into the avatar as if it were a surname:

| Input | Current | Expected |
|---|---|---|
| `chloe@ssw.com.au` | `CS` — the **S** is from *ssw* | `C` |
| `ada@example.com` | `AE` — the **E** is from *example* | `A` |
| `josh.berman@ssw.com.au` | `JB` ✓ (correct by luck — the local part has a dot) | `JB` |
| `Ada Lovelace` | `AL` ✓ | `AL` |

Only reachable on the email fallback, so it's cosmetic — but it shows in the modal for anyone without a `fullName` set.

```suggestion
    .split('@')[0]
    .split(/[\s._-]+/)
```

Separate and not worth a suggestion: `part[0]` indexes a UTF-16 code unit, so a display name starting outside the BMP yields a lone surrogate — `'👩‍🔬 Ada'` → `"\ud83dA"`, which renders as a tofu box. `[...part][0]` covers it, if display names can carry emoji.

packages/tinacms/src/toolkit/form-builder/create-branch-modal.tsx
```diff
@@ -178,6 +178,53 @@ export const CreateBranchModal = ({
...
+    .filter(Boolean)
+    .slice(0, 2)
+    .map((part) => part[0].toUpperCase())
+    .join('');
+
+export const CommittingAs = () => {
```
joshbermanssw: This `export` puts `CommittingAs` on the package's public API surface — it rides the barrel chain straight out:

```
create-branch-modal.tsx
  → form-builder/index.ts:10   export * from './create-branch-modal'
  → toolkit/index.ts:28        export * from '@toolkit/form-builder'
  → src/index.ts:15            export * from './toolkit'
```

So `import { CommittingAs } from 'tinacms'` resolves, and the component is semver-bound from here on — for something rendered in exactly one place, in this same file.

`CreateBranchModal` and `CreateBranchPromptModal` are exported the same way, but those are consumed across files (`form-builder.tsx`, `CollectionListPage.tsx`, `media-workflow-overlay.tsx`), so they've earned it. This one is exported so the test can import it.

Not a one-line fix, since dropping the `export` breaks `create-branch-modal.test.tsx`'s import — it'd mean exercising the row through `CreateBranchPromptModal` instead. Worth doing, or happy for it to be public?

packages/tinacms/src/internalClient/index.ts
Aibono1225: ✅ Done - `Client.user` is now typed as `AuthenticatedUser | boolean`

packages/tinacms/src/toolkit/form-builder/create-branch-modal.tsx
Aibono1225: ✅ Fixed - email fallbacks now build initials from the local part only, so `chloe@ssw.com.au` becomes `C` while dotted local parts like `josh.berman@ssw.com.au` still become `JB`. I also made the first-character extraction Unicode-safe.

packages/tinacms/src/toolkit/form-builder/create-branch-modal.tsx
Aibono1225: ✅ Done - Good catch, `CommittingAs` is now internal to `create-branch-modal.tsx` and no longer flows through the barrel exports. The tests now exercise the row through `CreateBranchPromptModal` instead of importing the component directly.

joshbermanssw (approved):
LGTM

PR #7497, opened 2026-08-27
fix(tinacms): media manager stops refreshing on events after paging

### LINKED ISSUE #7496: Media Manager stops refreshing on rename/delete/pageSize events after scrolling past page one
### The exact error message

No error is thrown — the failure is silent. The media listing simply stops updating in response to rename/delete/page-size events.

### Steps to reproduce

1. Open the Media Manager on a library with more than one page of items (any store that returns `nextOffset`).
2. Scroll down so the infinite-scroll loader fetches page two (the paging offset is now set).
3. Rename or delete any file (or change the page size). The listing refreshes correctly this one time.
4. Rename or delete another file.

Expected: every rename/delete/page-size change refreshes the listing, as it does before scrolling.

Actually: after step 3, subsequent renames, deletes, and page-size changes no longer refresh the listing — the grid shows stale entries until the user changes directory, search, or filter (any change to the listing effect's deps re-registers the subscription).

Mechanism (verified on `main`): in `packages/tinacms/src/toolkit/components/media/media-manager.tsx`, the listing `useEffect` both loads media and returns `cms.events.subscribe(['media:delete:success', 'media:rename:success', 'media:pageSize'], …)`. The handler calls `setRefreshing(true)` + `resetOffset()` + `resetList()`. When the offset was non-undefined, `resetOffset()` changes the effect's deps, the effect re-runs, its cleanup unsubscribes, and the effect body hits the early `if (refreshing) return;` guard **before** resubscribing — so no subscription is registered, and `refreshing` flipping back to false doesn't re-run this effect.

### Your environment

- tinacms `main` (reproduced at 7c2190666 via a unit test; pre-existing, predates #7477)
- Default repo-based media store — nothing non-default required; any `MediaStore` whose `list` pages via `nextOffset` is affected

### A way for us to reproduce

The regression test in the linked PR pages to page two via a stubbed `IntersectionObserver`, dispatches `media:rename:success` twice, and asserts the second event still triggers a re-list. It fails on `main` and passes with the fix.

kulesy:
Closes #7496

**TL;DR** Keep the Media Manager's refresh subscription alive after paging past page one.

**Pain:** after an editor scrolls past page one, renames, deletes, and page-size changes silently stop refreshing the media listing — the grid shows stale files until they change folder, search, or filter.

**Solution:** subscribe to the media events in a dedicated effect so the listing effect's re-runs can no longer tear the subscription down, with a regression test covering the paging sequence.


kulesy:
**Verification**

- New test `keeps refreshing on rename events after paging past page one` fails on the pre-fix code (times out waiting for the second rename to re-list — confirmed by stashing the source change and re-running) and passes with the fix.
- Full file: 10/10 pass. All media component suites: 53/53 pass (`vitest run src/toolkit/components/media/`).
- Biome clean on both changed files.

**Reviewer note:** the event handler body is unchanged — it only uses functional updates and fixed-value setters (`setActiveItem` callback, `setRefreshing(true)`, `resetOffset`, `resetList`), which is what makes the narrow `[cms.media.isConfigured]` dep array safe. The listing effect keeps its existing `refreshing` guard, so the load-once-per-refresh behaviour is unchanged too.

No screenshots: the bug is a *missing* refresh over time — before/after stills of the listing are pixel-identical and can't show the difference, and there's no other visual surface to this change.

Aibono1225 (approved):
LGTM

PR #7500, opened 2026-09-01
fix(v4): match the v3 gray-matter engine configuration in the v4 markdown adapter

wicksipedia:
Thanks to @utkusen, @hash3liZer and @eros938 for spotting this one.

The v4 markdown adapter called `gray-matter` without an `engines` option, so it used the library defaults. The v3 stack does not: it passes a replacement engine map in `packages/@tinacms/graphql/src/database/util.ts`.

This brings the v4 adapter in line with v3.

The adapter now passes the same options object to all three `gray-matter` calls: `parse`, the re-parse inside `serialize`, and `stringify`. `matter.stringify()` re-parses the string it is given, so it needs the options as well.

The options object also replaces the empty cache object the adapter passed before, which keeps the previous behaviour of not sharing the `gray-matter` parse cache between calls.

## Testing

`format-adapters.test.ts` covers the new behaviour.

### DIFF
--- a/packages/v4/@tinacms/tinacms/src/plugins/content/local/adapters/format-adapters.test.ts
+++ b/packages/v4/@tinacms/tinacms/src/plugins/content/local/adapters/format-adapters.test.ts
@@ -129,6 +129,18 @@ describe('markdown adapter', () => {
       ).toThrow(/Expected a string for body field/);
     }
   });
+
+  it('refuses to run code in frontmatter', () => {
+    const globals = globalThis as Record<string, unknown>;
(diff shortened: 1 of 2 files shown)


KahaMason (approved):
LGTM🚀

eros938:
Hey @wicksipedia we have submitted advisories for tina cms and then we added pull requests for its fix can we know when can it be published and cve be assigned also we have sent them via your mail

wicksipedia:
Hey @eros938, this is not getting a CVE as `@tinacms/tinacms` is not a published npm package.

We appreciate the effort that goes into submitting advisories, so (since there are not changelogs for this package) we can credit on the PR

PR #7501, opened 2026-09-01
fix(@tinacms/graphql): preserve underlying PUT error messages

Aibono1225:
Fixes https://github.com/tinacms/tinacloud/issues/2108

## Summary
Preserve the underlying storage bridge error message when a document write fails.

Previously, `Database.put` wrapped bridge errors with a generic message such as `Error in PUT for src/pages/example.md`, discarding the actionable reason from the user-facing GraphQL error. For example, GitHub may return `Could not create file: Changes must be made through a pull request.`, but the CMS would only show the generic PUT failure.

With this change, the underlying message is preserved, e.g. `Error in PUT for src/pages/example.md: Could not create file: Changes must be made through a pull request.`

## Changes
- Preserve `error.message` when wrapping failed PUT operations in `TinaFetchError`
- Add regression coverage through the GraphQL resolver path
- Add a patch changeset for `@tinacms/graphql`

## Testing
- `npx vitest run tests/bridge-failure-mid-batch`
- `npx changeset status --since=origin/main`


harley-j (approved):
🚀 LGTM

PR #7507, opened 2026-09-02
Fix validation for templates with empty fields

### LINKED ISSUE #7490: 🐛 Schema validation allows a template with empty `fields`, producing an invalid GraphQL schema
### Pain

```
```

### Approach

### Acceptance Criteria

lprnmns:
## Problem

Templates currently accept an empty `fields` array. A rich-text block using that template can then produce an invalid GraphQL input object with no fields.

Fixes #7490.

## Fix

Apply the existing non-empty `fields` validation used by sibling schema definitions to `Template.fields`. The regression test verifies both rejection and the expected validation message.

## Tests

- `pnpm --dir packages/@tinacms/schema-tools exec jest --config jest.config.js --runInBand` — 116 passed
- `pnpm --filter @tinacms/schema-tools types` — passed
- `pnpm --filter @tinacms/schema-tools build` — passed
- `pnpm biome lint packages/@tinacms/schema-tools/src/validate/schema.ts packages/@tinacms/schema-tools/src/validate/index.spec.ts` — passed
- `pnpm biome format packages/@tinacms/schema-tools/src/validate/schema.ts packages/@tinacms/schema-tools/src/validate/index.spec.ts` — passed

The required patch changeset for `@tinacms/schema-tools` is included.

### DIFF
--- a/.changeset/tidy-templates-validate.md
+++ b/.changeset/tidy-templates-validate.md
@@ -1,0 +1,5 @@
+---
+"@tinacms/schema-tools": patch
+---
+
+Reject templates with empty `fields` arrays during schema validation.
--- a/packages/@tinacms/schema-tools/src/validate/index.spec.ts
+++ b/packages/@tinacms/schema-tools/src/validate/index.spec.ts
@@ -345,6 +345,23 @@ const schemaWithEmptyTemplates: Schema = {
     },
   ],
 };
+
+const schemaWithTemplateWithEmptyFields: Schema = {
+  collections: [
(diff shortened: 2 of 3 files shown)


joshbermanssw:
Thanks @lprnmns, your diagnosis and fix were spot on! #7548 landed first with the same `Template.fields` guard, plus the equivalent check for rich-text templates in `fields.ts`, so I'm closing this one as superseded. Appreciate the contribution, and we'd love to see more from you.

PR #7516, opened 2026-09-04
Fix inconsistent document IDs on Windows across query paths

### LINKED ISSUE #7508: 🐛 Bug - Title, sidebar, and breadcrumb are not synchronized when editing document title
Requested by: @Aibono1225 via [YakShaver.ai](https://YakShaver.ai) 🦬
cc: @Aibono1225, @kulesy, @18-th

Hi Team!

&nbsp;&nbsp;[🟥&nbsp;&nbsp;Watch the video (1 min 48 sec)](https://www.youtube.com/watch?v=4RVs1ypn_mA)

&nbsp;&nbsp;Url from screen share: [https://localhost:3000/admin/index.html#/~/docs](https://localhost:3000/admin/index.html#/~/docs)

### Pain
When editing a document title in TinaCMS, the title displayed in the page header, sidebar, and breadcrumb become out of sync. Updating the title field initially updates the sidebar and breadcrumb but not the page title. After interacting with the title field directly and making another change, the page title updates but the sidebar and breadcrumb no longer update. This behavior was reproduced locally in Tina Docs and appears to be a TinaCMS synchronization/state update issue.

Reproduced on Windows

### Acceptance Criteria
1. Editing a document title updates the page title, sidebar, and breadcrumb consistently and in real time.
2. All UI elements reflecting the document title remain synchronized after every edit.
3. The issue is reproducible and resolved in the Tina Docs project.
4. No additional field interaction is required for title updates to propagate across the UI.
5. Changes to the title field behave consistently across supported templates where applicable.

### Reproduce Steps
1. Run the Tina Docs project locally.
2. Open a document with a visible page title, sidebar entry, and breadcrumb.
3. Edit the document title field by adding text or an emoji.
4. Observe that the sidebar and breadcrumb update, but the page title does not.
5. Click into the page title field directly and make another change.
6. Observe that the page title updates, but the sidebar and breadcrumb no longer update.

### Screenshot
Screenshot
**Figure: Title update out of sync with sidebar and breadcrumb**

Aibono1225:
Fixes: https://github.com/tinacms/tinacms/issues/7508

### What

Normalize document IDs and `_sys.path` on Windows so the same document has a consistent identity regardless of how it is resolved.

Previously, direct document lookups could produce Windows-style paths such as:

`content\docs\index.mdx`

while reference or connection queries returned:

`content/docs/index.mdx`

This caused TinaCMS Visual Editing to treat the same document as separate forms, which could make different parts of the preview update independently.

### Fix

Two places produced the Windows-style paths:

1. `transformDocumentIntoPayload` now normalizes `fullPath` with the existing `normalizePath` helper, and uses it for the document `id`, `_sys.path`, `relativePath`, and path metadata parsing.
2. `getValidatedPath` now returns a POSIX `realPath`, so every mutation resolver built on it compares against the same identity the refs index and content files store. Without this, `_sys.hasReferences` was always `false` on Windows, deleting a referenced document skipped the dangling-reference cleanup, and renaming one left the referring documents pointing at the old path.

### Tests

Regression tests cover Windows-style document paths, identical IDs across Windows and POSIX inputs, `_sys` path metadata, reference lookups from a Windows-style path, and a rename that rewrites a POSIX-stored reference when path segments are joined with Windows separators.

The existing `@tinacms/graphql` resolver tests pass with the fix.

### Screenshot

Before:
image

After:
image


joshbermanssw (commented):
Fix is correct and low-risk. `normalizePath` is just `.replace(/\\/g, '/')`, so `relativePath` is byte-identical and `path.parse` already handled `\` on Windows — only `id` and `_sys.path` change, which is the point. It also now matches what `database.get()` keys on (`database/index.ts:388`), so POSIX is the right direction.

One gap left, same class of bug — see the inline comment.

Also worth confirming before this closes #7508: that issue never states an OS, and this only changes behaviour on Windows.

packages/@tinacms/graphql/src/resolver/index.ts
```diff
@@ -199,6 +203,7 @@ export const transformDocumentIntoPayload = async (
   hasReferences?: boolean
 ) => {
   const collection = tinaSchema.getCollection(rawData._collection);
+  const normalizedPath = normalizePath(fullPath);
```
joshbermanssw, line 206: The backslashes come from `getValidatedPath` (`:768`, `path.join(collection.path, relativePath)`), and that raw path goes to more than just this function:

- `:429` — `hasReferences(fullPath, …)` for `_sys.hasReferences`
- `:985` — `hasReferences(realPath, …)` in `resolveDeleteDocument`

Both do an exact `eq: id` against the refs index, whose keys are the raw reference values read out of content files (`datalayer.ts:851`) and so are always POSIX. A backslash path can never match, which on Windows means `_sys.hasReferences` is always `false` **and deleting a referenced document silently skips the dangling-reference cleanup**.

Normalizing once in `getValidatedPath` instead would fix that too and make this line unnecessary. (Read from the code path — not run on Windows.)

packages/@tinacms/graphql/src/resolver/index.test.ts
```diff
@@ -1900,4 +1901,51 @@ describe('index', () => {
...
+        windowsPath,
+        rawData,
+        tinaSchema
+      );
+
+      expect(payload.id).toBe(posixPath);
+      expect(payload._sys.path).toBe(posixPath);
+      expect(payload._sys.basename).toBe('index.mdx');
+      expect(payload._sys.filename).toBe('index');
+      expect(payload._sys.extension).toBe('.mdx');
```
joshbermanssw, line 2016: `relativePath`/`breadcrumbs` are only asserted equal in the next test, never asserted correct:

```suggestion
      expect(payload._sys.extension).toBe('.mdx');
      expect(payload._sys.relativePath).toBe('index.mdx');
```

Aibono1225: ✅ Good catch - Added an explicit `_sys.relativePath === "index.mdx"` assertion so the test now verifies the expected value, not just Windows/POSIX equality.

packages/@tinacms/graphql/src/resolver/index.ts
Aibono1225: Nice point - I confirmed this also affects `_sys.hasReferences` and delete cleanup on Windows.

I left the ID normalization in `transformDocumentIntoPayload` since some document paths bypass `getValidatedPath`, and fixed the reference comparisons where they happen instead.

Aibono1225:
Hi @joshbermanssw 

> Also worth confirming before this closes #7508: that issue never states an OS, and this only changes behaviour on Windows.

✅ Done - documented the Windows reproduction on #7508

packages/@tinacms/graphql/src/resolver/index.ts
```diff
@@ -997,7 +1001,7 @@ export class Resolver {
             const { object, updated } = updateObjectWithJsonPath(
               refDoc,
               path,
-              realPath,
+              normalizePath(realPath),
```
joshbermanssw: This fix is right, but its twin 85 lines up didn't get it. `resolveUpdateDocument` (`:919-924`) still passes the raw values:

```ts
updateObjectWithJsonPath(docWithRef, path, realPath, newRealPath)
```

`updateObjectWithJsonPath` matches with strict `===` (`:329`) against values read out of content files, which are always POSIX. So on Windows `findReferences` now *finds* the referrers — thanks to your `eq: normalizePath(id)` fix — but rewrites none of them: **renaming a referenced document silently leaves the references pointing at the old path.** And if it did match, `newRealPath` would write a backslash path into the content file.

`normalizePath` on both args fixes it, symmetric with this hunk.

The rename test (`index.test.ts:1557`) still uses `path.join` as the stored reference value, so it passes on POSIX either way — worth re-anchoring to a POSIX literal the way you did for the delete test.

Still think normalizing once in `getValidatedPath` (`:772`) would be cheaper than tracking each consumer — `sanitizePath` already strips the separators and `path.join` puts them straight back — but either way this site needs it. (Read from the code path, not run on Windows.)

joshbermanssw: Taken over and applied in 4c3fdd8.

Went with the root-cause version you suggested: `getValidatedPath` (`:774`) now returns `normalizePath(path.join(...))`, so `realPath`/`newRealPath` reach `updateObjectWithJsonPath` already POSIX and the rename rewrite matches. `validatePath` does the same normalization as its first step, so the traversal check is unchanged. That also made the `normalizePath(realPath)` on the delete site redundant, so it's gone; `normalizePath(id)` stays inside `hasReferences`/`findReferences` for callers that bypass `getValidatedPath` (`getDocument(fullPath)`).

Re-anchored the rename reference test to POSIX literals, and added one that stubs `path.join` to emit backslashes: it fails on the old code with the reference left at `posts/hello.md`, passes now.

kulesy (approved):

Issue #6951, opened 2026-05-23
React 19 peer dependency warning caused by locked `@headlessui/react@2.1.8`


Crownpack07:
### Summary

When using `tinacms` with React 19, npm produces peer dependency resolution warnings because TinaCMS currently depends on `@headlessui/react@2.1.8`, which still declares a React 18 peer dependency.

Newer versions of `@headlessui/react`, specifically `2.2.10`, already support React 19.

This appears to cause unnecessary installation friction and warning noise for projects using React 19.

---

### Current Environment

* React: `19.2.4`
* React DOM: `19.2.4`
* Next.js: `16.2.4`
* TinaCMS: `3.8.1`
* npm: `10.9.0`
* Node.js: `22.x`

---

### Warning Output

```txt
npm warn ERESOLVE overriding peer dependency
npm warn While resolving: @headlessui/react@2.1.8
npm warn Found: react@19.2.4

npm warn Could not resolve dependency:
npm warn peer react@"^18" from @headlessui/react@2.1.8
npm warn node_modules/tinacms/node_modules/@headlessui/react
npm warn   @headlessui/react@"2.1.8" from tinacms@3.8.1
```

The same warning also occurs for `react-dom`.

---

### Additional Context

The issue appears to specifically come from TinaCMS resolving:

```txt
@headlessui/react@2.1.8
```

which still has:

```json
"peerDependencies": {
  "react": "^18",
  "react-dom": "^18"
}
```

However newer versions such as:

```txt
@headlessui/react@2.2.10
```

already support React 19.

---

### Expected Behavior

Ideally TinaCMS would either:

* upgrade to a newer `@headlessui/react` version with React 19 peer support
* widen the dependency range if compatible
* or otherwise avoid pinning to a React 18-only peer dependency version

This would reduce npm install warnings and improve the React 19 developer experience.

---

### Notes

The application itself appears to function correctly at runtime, this is primarily an install/dependency ecosystem issue surfaced by newer npm versions.

PR #7518, opened 2026-09-04
fix: bump @headlessui/react to 2.2.10 for React 19 peer support

wakqasahmed:
`@tinacms/rich-text` pulls in `@headlessui/react` for the MDX inline-component toolbar (`Popover`/`PopoverButton`/`PopoverPanel`/`Transition`), and the version is pinned in the workspace catalog at `2.1.8`. That release only declares `react`/`react-dom` `^18` as peers, so any project consuming `tinacms` with React 19 gets an `ERESOLVE overriding peer dependency` warning on install (issue #6951).

Bumped the catalog pin to `2.2.10`, the latest release, which declares `^18 || ^19 || ^19.0.0-rc` and still supports React 18. The `Popover`/`PopoverButton`/`PopoverPanel`/`Transition` API used in `create-mdx-plugins/component.tsx` is unchanged between the two versions, and the full `@tinacms/rich-text` test suite (191 tests, including `component.test.tsx` which exercises that component) passes unmodified against the bumped version. Lockfile updated accordingly.

Added a changeset per the contributing guide.


wakqasahmed:
Hi @kulesy @joshbermanssw — noticed this PR doesn't have a reviewer assigned yet — it's been sitting for a while, CI is green and it's mergeable. Would you (or whoever's best placed) be able to take a look, or point me to who should? Thanks!

PR #7520, opened 2026-09-07
fix(tinacms): revert the final-form family to v4 so nested panel edits survive navigating back

### LINKED ISSUE #7519: 🐛 Edits inside nested panels are lost when breadcrumbing back (3.12.1)
### 🩸 Pain

Since 3.12.1, an edit made inside a nested panel (a group, an object-list item or a block) is thrown away the moment you breadcrumb back to the parent. Save goes back to disabled and the field shows its old value, so editors have to save inside every list item before backing out.

3.12.1 bumped `react-final-form` to 7.0.1, which resets a field to its initial value when it mounts at a path with no registered field. The sidebar swaps its whole field set on every breadcrumb, so the parent field remounts and is reset. Upstream: https://github.com/final-form/react-final-form/issues/1095, fix pending in https://github.com/final-form/react-final-form/pull/1096.

### 🛠️ Approach

- [ ] Revert `final-form`, `final-form-arrays` and `react-final-form` to the v4 family until the upstream fix ships
- [ ] Add an admin e2e test that edits inside a nested panel, breadcrumbs back, and checks the edit is still there

### ✅ Acceptance Criteria

- [ ] Editing several list items and saving once at the parent works again in the admin and in visual editing
- [ ] The new e2e test fails on 3.12.1 and passes on the fix

kulesy:
Closes #7519

**TL;DR** Revert final-form, final-form-arrays and react-final-form to the v4 family. Edits made inside a nested panel survive breadcrumbing back again, and nothing else about the form UI changes.

**Pain:** since 3.12.1, editing inside a group, an object-list item or a block and then breadcrumbing back to the parent silently throws the edit away, so editors have to save inside every item.

**Solution:** pins the three packages back to the versions 3.12.0 shipped with until the upstream fix in react-final-form lands, and adds an admin e2e test that fails on 3.12.1.


harley-j (approved):

kulesy:
Before and after, same steps both times: edit a nav label inside Header > Nav Links, click the Header breadcrumb, reopen the item. Recorded headless in the kitchen-sink admin in local mode.

https://github.com/user-attachments/assets/ad53a0eb-87e0-4e86-b975-89140d73ad29

Figure: first half is 3.12.1 with react-final-form 7.0.1, where the edit is thrown away and Save goes back to disabled. Second half is this branch, where the edit survives and Save stays enabled.

PR #7522, opened 2026-09-07
Scope the admin preview to the site's own origin

wicksipedia:
## What

The admin preview loads a page in an iframe and opens a postMessage channel to it. This scopes both to the admin's own origin. The preview path is resolved against that origin, and when it points elsewhere the preview stays on the site root and shows a warning instead of opening it.

## Why resolve, not filter

A character filter does not hold here: the URL parser treats a backslash as a slash and drops leading tabs, so those slip past a filter and still resolve off-origin. Resolving the path first and rejecting anything that leaves the origin catches every form.

## Also

The active-field lookup now guards against a preview whose document the admin cannot read (a cross-origin or sandboxed frame), where it previously threw and unmounted the admin.

## Tests

`preview-url.test.ts` covers ordinary paths, the query and fragment, an empty splat, and the off-origin forms (leading slash, double slash, backslashes, leading tab, encoded).


packages/tinacms/src/admin/preview-url.ts
```diff
@@ -0,0 +1,32 @@
...
+  if (!baseOrigin || !splat) {
+    return { path: '/', offOrigin: false };
+  }
+  try {
+    const resolved = new URL(`/${splat}`, baseOrigin);
+    if (resolved.origin !== baseOrigin) {
+      return { path: '/', offOrigin: true };
+    }
+    return {
+      path: `${resolved.pathname}${resolved.search}${resolved.hash}`,
```
kulesy: A link like `#/~/..//evil.example` still opens the other site, because the check passed on the absolute URL but the path returned here starts with `//` and the browser resolves that again.

packages/tinacms/src/admin/preview-url.ts
```diff
@@ -0,0 +1,32 @@
...
+  splat: string | undefined,
+  baseOrigin: string = typeof window !== 'undefined'
+    ? window.location.origin
+    : ''
+): { path: string; offOrigin: boolean } {
+  if (!baseOrigin || !splat) {
+    return { path: '/', offOrigin: false };
+  }
+  try {
+    const resolved = new URL(`/${splat}`, baseOrigin);
```
kulesy, line 21: `getExpectedPreviewOrigin` in `@tinacms/app` resolves this same string a few frames later and already disagrees with this one about what a parse failure means, so should they share a helper?

packages/tinacms/src/admin/preview-url.ts
wicksipedia: Confirmed, and thank you — this was a real bypass. I reproduced it before fixing:

```
splat                    returned path      iframe lands on
..//evil.example         //evil.example     https://evil.example   <-- escaped
../..//evil.example      //evil.example     https://evil.example   <-- escaped
a/..//evil.example       //evil.example     https://evil.example   <-- escaped
.//evil.example          //evil.example     https://evil.example   <-- escaped
```

Exactly your reasoning: the origin check ran against the absolute URL, but the `pathname` it produced could itself begin with two slashes, and the caller resolves that a second time as an iframe src.

Fixed in ba5a2e0 by resolving the returned path again and rejecting it on the same rule:

```ts
const path = `${resolved.pathname}${resolved.search}${resolved.hash}`;
if (new URL(path, baseOrigin).origin !== baseOrigin) {
  return { path: '/', offOrigin: true };
}
```

Added your four shapes as cases. Removing the new guard fails exactly those four and passes the other eleven, so they pin the behaviour rather than just passing.

packages/tinacms/src/admin/preview-url.ts
wicksipedia: Good catch on the divergence, though I do not think they should share a helper as-is, because they are answering different questions.

`resolvePreviewPath` asks "is this router splat a safe same-origin path?" and must fail closed. `getExpectedPreviewOrigin` asks "given a URL we have already decided to load, whose messages do we trust?" — and it deliberately supports a preview served from another origin. That is a tested behaviour:

```ts
it('keeps the origin of an absolute preview URL', () => {
  expect(getExpectedPreviewOrigin(`${PREVIEW}/posts/hello`, ADMIN)).toBe(PREVIEW);
});
```

So locking it to `window.location.origin` would remove cross-origin previews, which is a product decision rather than a refactor.

On the disagreement itself: after this change `PreviewInner` only ever hands it a same-origin path, so from that call site the two cannot reach different answers. The latent trap is real, but it is a trap for a future caller, not a live bug.

Happy to raise a separate issue for unifying them, with "do we still support a preview on another origin?" as the question that decides the shape. I would rather not answer that inside a security fix.

wicksipedia: Correcting myself: my previous reply was wrong, and your suggestion is stronger than I allowed.

I claimed cross-origin previews are a supported feature. They are not. The only cross-origin preview concept is the `ui.previewUrl` config option, and it never reaches the iframe — it is used in `branch-button.tsx:82` as `window.open(previewUrl, '_blank')`, opening a branch deploy in a new tab.

The chain into the iframe has a single producer:

```
params['*'] -> resolvePreviewPath -> setURL   (one call site)
            -> <Preview url={url}> -> iframe src
                                   -> useGraphQLReducer(url) -> getExpectedPreviewOrigin(url)
```

`setURL` is called in exactly one place, from `resolvePreviewPath`. So after this PR `url` is always a same-origin path, and the absolute-URL branch of `getExpectedPreviewOrigin` has no caller that can reach it. The test I quoted asserts the helper's behaviour in isolation; it is not evidence of a product capability. I should have traced the callers before citing it.

Which changes what I would do. If a cross-origin preview URL cannot be produced legitimately, then `getExpectedPreviewOrigin` returning a foreign origin is only ever a bug or an attack, and deriving it from the URL buys nothing. The simplification is not to share a resolver but to delete the second one:

```ts
const expectedOrigin = window.location.origin;
```

Then the two cannot disagree, because only one place decides. That is also the second layer the original report asked for, which I had skipped.

It is internal to `@tinacms/app` — no package export, no consumer elsewhere in the repo — so the change is contained. I will put it up separately from this PR so the origin-scoping fix stays reviewable on its own, and so the "is an absolute preview URL still allowed?" question gets its own diff rather than being decided in a footnote here.

wicksipedia: Rolled into this PR in 3f8eeda rather than deferred, so the two layers land together.

`getExpectedPreviewOrigin` is gone. The channel now uses the admin's own origin directly:

```ts
export const getPreviewOrigin = (): string =>
  typeof window !== 'undefined' ? window.location.origin : '';
```

There is no longer a second place that decides, so the two cannot disagree — which was your point. Dropping the parameter also means no caller can pass a URL that nominates an origin, so the failure mode is removed by construction rather than guarded against.

The test that asserted the old absolute-URL behaviour is replaced by one asserting the opposite, with a note on why the branch is gone.

Contained: no package export, no consumer outside `@tinacms/app`. `@tinacms/app` suite passes (14) and the package builds.

The changeset now covers both packages.

kulesy (approved):

PR #7523, opened 2026-09-07
Sanitize link and image urls in the tina-markdown element

wicksipedia:
## What

`tina-markdown` renders a rich-text node's URL straight into an anchor's `href` and an image's `src`. This routes both through `sanitizeUrl`, so a URL whose scheme is not on the allow list (`http`, `https`, `mailto`, `tel`, `xref`) renders empty instead.

```js
if (node.url && node.type === 'a') el.href = sanitizeUrl(node.url);
if (node.url && node.type === 'img') el.src = sanitizeUrl(node.url);
```

## Why this guard

`sanitizeUrl` is the same helper the React (`tinacms/rich-text`) and Astro renderers already use, so all of them now treat a given URL the same way. It ships as the dependency-free `@tinacms/mdx/sanitize-url` subpath, which adds 0.65 kB rather than the markdown toolchain, so `@tinacms/mdx` is added as a workspace dependency.

`img.src` is included even though a script URL does not execute from it, because that line had no validation either.

## Tests

`tina-markdown.test.ts` adds cases for a link and an image whose scheme is not allowed, and one asserting `mailto:` and relative URLs still pass through unchanged. 29 pass.

### DIFF
--- a/packages/@tinacms/web-components/package.json
+++ b/packages/@tinacms/web-components/package.json
@@ -22,10 +22,10 @@
   },
   "dependencies": {
     "@tinacms/bridge": "workspace:^",
-    "@tinacms/mdx": "workspace:^",
     "dompurify": "catalog:"
   },
   "devDependencies": {
+    "@tinacms/mdx": "workspace:*",
     "@tinacms/scripts": "workspace:*",
     "happy-dom": "catalog:",
     "vite": "^5.4.14",
--- a/packages/@tinacms/web-components/package.json
+++ b/packages/@tinacms/web-components/package.json
@@ -22,6 +22,7 @@
   },
   "dependencies": {
     "@tinacms/bridge": "workspace:^",
+    "@tinacms/mdx": "workspace:^",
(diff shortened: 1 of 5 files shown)


packages/@tinacms/web-components/package.json
```diff
@@ -22,6 +22,7 @@
   },
   "dependencies": {
     "@tinacms/bridge": "workspace:^",
+    "@tinacms/mdx": "workspace:^",
```
kulesy: Putting `@tinacms/mdx` in dependencies means anyone installing this package pulls prettier, acorn and the whole remark tree on disk, even though the build already inlines the one small file we use.

kulesy (approved):

packages/@tinacms/web-components/package.json
wicksipedia: moved to dev dependencies.
realistically they'd pull it to disk anyways since they'd pull down tinacms to their site.

PR #7525, opened 2026-09-07
Reject malformed URLs that name a scheme, instead of returning them unchecked

wicksipedia:
## What

`sanitizeUrl` returned the value it was given when `new URL()` threw:

```ts
try {
  parsedUrl = new URL(url);
} catch (error) {
  return url;
}
```

That branch exists for relative URLs. They have no scheme, so `new URL()` cannot parse them alone. But the branch also returned any other value that failed to parse, with no check.

The function now tells the two cases apart. When it looks for a scheme, it ignores characters that a scheme cannot contain. If the value names a scheme, the function returns an empty string. If the value names no scheme, it is relative, and the function returns it unchanged.

## Scope

- A value that names a scheme and **parses** was already rejected by the allow list. That is unchanged.
- A value that names a scheme and does **not** parse was returned unchanged. That is what this fixes.
- Relative URLs (`/blog/post`, `./x`, `#a`, `?q=1`) do not change. Tests cover them.

## Tests

This helper had no tests, although six renderers depend on it. This PR adds 10. They cover the allow list, relative URLs, empty input, disallowed schemes, and malformed values that name a scheme. The full `@tinacms/mdx` suite (251 tests) and `tinacms` suite (766 tests) pass unchanged.

### DIFF
--- a/packages/@tinacms/mdx/src/sanitize-url.ts
+++ b/packages/@tinacms/mdx/src/sanitize-url.ts
@@ -6,6 +6,19 @@
  * `tinacms`'s rich-text renderer) don't have to pull in the full remark/mdast/micromark
  * markdown-parsing toolchain that the rest of `@tinacms/mdx` bundles.
  */
+
+/**
+ * Characters that carry no meaning inside a scheme, so they cannot be used to
+ * disguise one. Stripped before looking for a scheme, never from the URL itself.
+ */
+const IGNORED_IN_SCHEME = /[\u0000-\u001f\u007f\u200b-\u200d\ufeff]/g;
+
+/** A scheme, per RFC 3986: a letter, then letters, digits, `+`, `-` or `.`. */
+const SCHEME_PREFIX = /^[a-zA-Z][a-zA-Z0-9+.\-]*:/;
+
+const namesAScheme = (url: string) =>
+  SCHEME_PREFIX.test(url.replace(IGNORED_IN_SCHEME, '').trim());
+
 export const sanitizeUrl = (url: string | undefined) => {
   const allowedSchemes = ['http', 'https', 'mailto', 'tel', 'xref'];
   if (!url) return '';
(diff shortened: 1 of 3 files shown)


packages/@tinacms/mdx/src/sanitize-url.ts
```diff
@@ -6,6 +6,19 @@
  * `tinacms`'s rich-text renderer) don't have to pull in the full remark/mdast/micromark
  * markdown-parsing toolchain that the rest of `@tinacms/mdx` bundles.
  */
+
+/**
+ * Characters that carry no meaning inside a scheme, so they cannot be used to
+ * disguise one. Stripped before looking for a scheme, never from the URL itself.
+ */
+const IGNORED_IN_SCHEME = /[\u0000-\u001f\u007f\u200b-\u200d\ufeff]/g;
```
brookjeynes-ssw: Can we link a doc / website to where these are from?

wicksipedia: Good question, and answering it turned up a gap rather than just needing a link.

The hand-picked list had no source because I chose it, which is the problem you spotted. It is now the two Unicode general categories that actually describe the intent, with citations:

```ts
const IGNORED_IN_SCHEME = /[\p{Cc}\p{Cf}]/gu;
```

- **`Cc`** (control) — the URL parser removes ASCII tab and newline from anywhere in its input, so a tab inside `javascript` still reaches a browser as `javascript:`. Verified: `new URL('java\tscript:x').protocol` returns `javascript:`. https://url.spec.whatwg.org/#url-parsing
- **`Cf`** (format) — soft hyphen, zero-width joiners, byte order mark. These render as nothing, so they hide a scheme from whoever is reading the link, even though the parser itself rejects them. https://www.unicode.org/reports/tr44/#General_Category_Values

`SCHEME_PREFIX` now cites RFC 3986 section 3.1 for the grammar it implements.

The gap this found: my list was `U+0000-U+001F`, `U+007F`, `U+200B-U+200D` and `U+FEFF`. `Cf` is wider than that — it also covers the soft hyphen (U+00AD), the left-to-right mark (U+200E) and the word joiner (U+2060), which I had missed. Those three are now tested, and putting the old list back fails exactly those three. So this is a real fix rather than only a comment.

15 tests pass, the full `@tinacms/mdx` suite (256) passes, and the property escape survives the build (`target: esnext`).

kulesy (approved):

Issue #6921, opened 2026-05-19
Field Type Implementation - Reference


18-th:
## Pain

Reference fields connect content across collections, so they carry more behaviour than a simple scalar field. Today that behaviour is tied to editor and lookup internals, which makes references harder to test, harder to customize and harder to expose through the same field plugin model as other fields.

## Approach

Implement `ReferenceField` as a field plugin against the field contract. The plugin owns reference rendering, lookup integration, ingest, digest and validation while preserving existing reference schema and stored value behaviour.

## Expected outcomes

- Reference fields work through the v4 field plugin system.
- Collection lookup and selected-reference behaviour are owned by the reference field plugin.
- References can use the same validation and metadata wrapping path as other fields.
- Reference field behaviour is isolated and testable.
- Reverse reference queries are supported, letting a document resolve the documents that reference it. That work is split out into #7532 and is not delivered here.

## Acceptance Criteria

- `ReferenceField` exists as a field plugin that conforms to the field contract.
- Ingest and digest behaviour preserves existing reference schema and stored value behaviour.
- Rendering supports searching, selecting and clearing referenced content.
- Reference lookup integrates with the editor's collection/content APIs through a typed boundary.
- Missing or invalid references are handled clearly.
- Validation runs through the shared field validation path.
- The preview receives a reference as the document it points at, not as the stored path, so a site reads one shape whether it renders statically or through the editor.
- Tests cover rendering, lookup, select, clear, missing references, ingest, digest and validation.

## Out of scope

Reverse reference queries move to #7532. They need an index in the content layer rather than anything in the field plugin, so they ship separately. Close this issue when the field itself is done; #7532 tracks the rest.

joshbermanssw:
✅ Done 

<img width="1536" height="996" alt="Screenshot 2026-09-11 at 9 27 39 am" src="https://github.com/user-attachments/assets/a574678f-014f-4675-9027-af92781ba6c9" />

**Figure: UI**

PR #7533, opened 2026-09-09
✨ Add the reference field to v4

joshbermanssw:
cc: —
Related PBI: https://github.com/tinacms/tinacms/issues/6921

## TL;DR

v4 had no `reference` field, so a document could not point at another document. This adds `ReferenceField` as a field plugin, and it resolves the reference for the preview so a site reads `post.author.title` while editing, not the raw path string.

**Files to review (23, +1562 / -16):**

| File | Why |
|---|---|
| `plugins/fields/reference/*` *(start here)* | The plugin: schema helper, validator, descriptor, combobox UI, 18 tests. |
| `core/form/references.ts` | Walks the schema against form values and swaps reference paths for documents. Pure, no React. |
| `editor/preview-connection.ts` | Resolves on the way out, warms the cache separately so posting stays synchronous. |
| `editor/hooks.ts` | Three accessors: the form's collection, the schema's collections, and a nullable content slice. |
| `graphql/graphql-pipeline.ts` | One line: `collections` now survives the map into v3 schema shape. |
| `@tinacms/ui/components/combobox.tsx` | shadcn combobox, plus the input-group and textarea it depends on. |

## Why

A reference stores a path (`author: content/authors/ada.md`), but a site renders the document that path names. v3 got this for free because its admin re-ran the site's GraphQL query on every edit and substituted resolved documents before streaming ([`graphql-reducer.ts:353`](https://github.com/tinacms/tinacms/blob/main/packages/%40tinacms/app/src/lib/graphql-reducer.ts)).

v4's preview posts raw form values instead, so without this the site sees two different shapes:

| | Page load | Editing |
|---|---|---|
| Before | `{ title: 'Test Title' }` | `'content/pages/test.mdx'` |
| After | `{ title: 'Test Title' }` | `{ title: 'Test Title' }` |

## How

The field itself is an ordinary plugin. The preview half is the part worth reading:

```mermaid
graph LR
    A[form values] --> B{resolveReferences}
    B -->|reads| C[react-query cache]
    B --> D[postMessage]
    A --> E[collectReferences]
    E -->|fetches what is missing| C
    C -.->|re-post on arrival| D
```

Posting reads the cache only, so a keystroke never waits on the network. A reference that has not arrived keeps its path for a frame, which the site already handles as the missing case.

## Reviewer notes

- **Options come from content, not config.** `collections` names collections; the documents come from `content.list()` at open time. A field naming several collections needs `useQueries`, since hooks cannot loop.
- **Content slice is now nullable.** `useOptionalContentSlice` exists so a preview with no content plugin degrades to bare paths instead of throwing.
- **Options are labelled by path.** `DocumentSummary` deliberately carries no content, so there is no title to show. This is where v3's `ui.optionComponent` would go.
- **shadcn generated a bad import.** `add combobox` wrote `import { cn } from "cn"` in all three files and added a junk `cn` package. Repointed at `@tinacms/ui/lib/utils`, dep removed.
- **A referenced document is not ingested.** `resolveReferences` substitutes the raw stored document, so a referenced rich-text body would arrive as markdown source, not the AST. Fine for scalars; a gap if someone renders a referenced body.

## Tests

`reference-field.test.tsx` (18) covers rendering, search, lookup across multiple collections, select, clear, missing references, validation, and the ingest/digest round trip, reading through a stub content capability.

`references.test.ts` (8) covers the walk, including references nested in array items.

`graphql-pipeline.test.ts` gains a reference resolution query. It fails with `Cannot read properties of undefined (reading 'includes')` if the `collections` passthrough is removed.

776 tests pass; typecheck clean.

Screenshot 2026-09-11 at 9 27 39 am

**Figure: UI**

---
Changes made by Claude Opus 5 (1M context) via Claude Code.

### DIFF
--- a/packages/v4/@tinacms/tinacms/src/core/form/references.ts
+++ b/packages/v4/@tinacms/tinacms/src/core/form/references.ts
@@ -1,0 +1,105 @@
+import type {
+  CollectionSchema,
+  FieldSchema,
+  TinaDocument,
+} from '../schema/types';
+
+const REFERENCE_FIELD_TYPE = 'reference';
+
+export interface ReferenceTarget {
+  collection: string;
+  path: string;
+}
+
+const isDocument = (value: unknown): value is TinaDocument =>
+  typeof value === 'object' && value !== null && !Array.isArray(value);
+
+const childFieldsOf = (node: FieldSchema): FieldSchema[] | undefined => {
+  const { fields } = node as { fields?: FieldSchema[] };
+  return Array.isArray(fields) ? fields : undefined;
+};
+
+// A reference stores the path of the document it points at. `collections`
+// narrows the candidates, and the one whose `path` contains that document owns
+// it — `content.get` needs a collection name, not a path.
+const targetOf = (
+  node: FieldSchema,
+  value: unknown,
+  collections: CollectionSchema[]
+): ReferenceTarget | null => {
+  if (typeof value !== 'string' || value.length === 0) return null;
+  const candidates = node.collections;
+  if (!candidates) return null;
+  const owner = collections.find(
+    (collection) =>
+      candidates.includes(collection.name) &&
+      collection.path != null &&
+      value.startsWith(`${collection.path}/`)
+  );
+  return owner ? { collection: owner.name, path: value } : null;
+};
+
+export const collectReferences = (
+  values: TinaDocument | undefined,
+  fields: FieldSchema[],
+  collections: CollectionSchema[]
+): ReferenceTarget[] => {
+  const targets: ReferenceTarget[] = [];
+  const walk = (document: TinaDocument, nodes: FieldSchema[]): void => {
+    for (const node of nodes) {
+      const value = document[node.name];
+      if (node.type === REFERENCE_FIELD_TYPE) {
+        const target = targetOf(node, value, collections);
+        if (target) targets.push(target);
+        continue;
+      }
+      const children = childFieldsOf(node);
+      if (!children) continue;
+      if (Array.isArray(value)) {
+        for (const item of value) if (isDocument(item)) walk(item, children);
+      } else if (isDocument(value)) {
+        walk(value, children);
+      }
+    }
+  };
+  walk(values ?? {}, fields);
+  return targets;
+};
+
+export type ReferenceResolver = (
+  target: ReferenceTarget
+) => TinaDocument | undefined;
+
+// The preview renders the site's own shape, where a reference reads as the
+// document it points at. An unresolved target keeps its path, so the site sees
+// the same value it would for a reference that no longer exists.
+export const resolveReferences = (
+  values: TinaDocument | undefined,
+  fields: FieldSchema[],
+  collections: CollectionSchema[],
+  resolve: ReferenceResolver
+): TinaDocument => {
+  const walk = (document: TinaDocument, nodes: FieldSchema[]): TinaDocument => {
+    const resolved: TinaDocument = { ...document };
+    for (const node of nodes) {
+      const value = document[node.name];
+      if (node.type === REFERENCE_FIELD_TYPE) {
+        const target = targetOf(node, value, collections);
+        const referenced = target ? resolve(target) : undefined;
+        if (referenced) resolved[node.name] = referenced;
+        continue;
+      }
+      const children = childFieldsOf(node);
+      if (!children) continue;
+      if (Array.isArray(value)) {
+        resolved[node.name] = value.map((item) =>
+          isDocument(item) ? walk(item, children) : item
+        );
+      } else if (isDocument(value)) {
+        resolved[node.name] = walk(value, children);
+      }
+    }
+    return resolved;
+  };
+  return walk(values ?? {}, fields);
+};
--- a/packages/v4/@tinacms/tinacms/src/core/form/references.ts
+++ b/packages/v4/@tinacms/tinacms/src/core/form/references.ts
@@ -1,8 +1,7 @@
-import {
-  type CollectionSchema,
-  type FieldSchema,
-  REFERENCE_FIELD_TYPE,
-  type TinaDocument,
+import type {
+  CollectionSchema,
+  FieldSchema,
+  TinaDocument,
 } from '../schema/types';
 
 export interface ReferenceTarget {
@@ -18,9 +17,11 @@ const childFieldsOf = (node: FieldSchema): FieldSchema[] | undefined => {
   return Array.isArray(fields) ? fields : undefined;
 };
 
-// A reference stores the path of the document it points at. `collections`
-// narrows the candidates, and the one whose `path` contains that document owns
-// it — `content.get` needs a collection name, not a path.
+// A reference stores the path of the document it points at, and it is the only
+// node that declares `collections`. Reading that rather than the type key keeps
+// core from naming a field type. `collections` narrows the candidates, and the
+// one whose `path` contains that document owns it, because `content.get` needs
+// a collection name and not a path.
 const targetOf = (
   node: FieldSchema,
   value: unknown,
@@ -38,9 +39,9 @@ const targetOf = (
   return owner ? { collection: owner.name, path: value } : null;
 };
 
-// TODO(#7534): this walk reads `node.fields` and matches on the type key
-// itself, so it misses any other way a plugin nests children. `templates` is
-// the live gap. Fold it into the registry-driven path that `ingest.ts` uses.
+// TODO(#7534): this walk reads `node.fields`, so it misses any other way a
+// plugin nests children. `templates` is the live gap. Fold it into the
+// registry-driven path that `ingest.ts` uses.
 export const collectReferences = (
   values: TinaDocument | undefined,
   fields: FieldSchema[],
@@ -50,9 +51,9 @@ export const collectReferences = (
   const walk = (document: TinaDocument, nodes: FieldSchema[]): void => {
     for (const node of nodes) {
       const value = document[node.name];
-      if (node.type === REFERENCE_FIELD_TYPE) {
-        const target = targetOf(node, value, collections);
-        if (target) targets.push(target);
+      const target = targetOf(node, value, collections);
+      if (target) {
+        targets.push(target);
         continue;
       }
       const children = childFieldsOf(node);
@@ -85,9 +86,9 @@ export const resolveReferences = (
     const resolved: TinaDocument = { ...document };
     for (const node of nodes) {
       const value = document[node.name];
-      if (node.type === REFERENCE_FIELD_TYPE) {
-        const target = targetOf(node, value, collections);
-        const referenced = target ? resolve(target) : undefined;
+      const target = targetOf(node, value, collections);
+      if (target) {
+        const referenced = resolve(target);
         if (referenced) resolved[node.name] = referenced;
         continue;
       }
--- a/packages/v4/@tinacms/tinacms/src/core/form/references.ts
+++ b/packages/v4/@tinacms/tinacms/src/core/form/references.ts
@@ -1,10 +1,9 @@
-import type {
-  CollectionSchema,
-  FieldSchema,
-  TinaDocument,
+import {
+  type CollectionSchema,
+  type FieldSchema,
+  REFERENCE_FIELD_TYPE,
+  type TinaDocument,
 } from '../schema/types';
-
-const REFERENCE_FIELD_TYPE = 'reference';
 
 export interface ReferenceTarget {
   collection: string;
@@ -39,6 +38,9 @@ const targetOf = (
   return owner ? { collection: owner.name, path: value } : null;
 };
 
+// TODO(#7534): this walk reads `node.fields` and matches on the type key
+// itself, so it misses any other way a plugin nests children. `templates` is
+// the live gap. Fold it into the registry-driven path that `ingest.ts` uses.
 export const collectReferences = (
   values: TinaDocument | undefined,
   fields: FieldSchema[],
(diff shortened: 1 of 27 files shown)


joshbermanssw (commented):
Review against #6921. Every AC lands; the notes below are the five worth acting on before merge. Full context in the thread on each line.

packages/v4/@tinacms/ui/src/components/combobox.tsx
```diff
@@ -0,0 +1,300 @@
...
+function ComboboxClear({ className, ...props }: ComboboxPrimitive.Clear.Props) {
+  return (
+    <ComboboxPrimitive.Clear
+      data-slot='combobox-clear'
+      render={<InputGroupButton variant='ghost' size='icon-xs' />}
+      className={cn(className)}
+      {...props}
+    >
+      <XIcon className='pointer-events-none' />
+    </ComboboxPrimitive.Clear>
```
joshbermanssw, line 49: **a11y: the clear button has no accessible name.**

`AGENTS.md` (v4): *"An icon button takes `aria-label`... Make each icon in the button decorative with `aria-hidden='true'`."* This renders an icon-only button with neither.

The test proves the gap. `reference-field.test.tsx` reaches for `container.querySelector('[data-slot="combobox-clear"]')` because there is no role plus name to query by:

```tsx
const clear = container.querySelector('[data-slot="combobox-clear"]');
```

Suggested:

```tsx
<ComboboxPrimitive.Clear
  data-slot='combobox-clear'
  aria-label='Clear'
  ...
>
  <XIcon aria-hidden='true' className='pointer-events-none' />
```

Then the assertion becomes `getByRole('button', { name: 'Clear' })`, which is what the accessibility section asks for.

I know this file is shadcn-generated and `AGENTS.md` says not to hand-write these primitives, but the `cn` import is already patched here, so one more line is consistent.

packages/v4/examples/barebones/tina/config.ts
```diff
@@ -38,7 +39,20 @@ export const postCollection = {
...
+  label: 'Pages',
+  path: 'content/pages',
+  format: 'mdx',
+  fields: [
+    t.string({ name: 'title', label: 'Title', required: true }),
+    t.boolean({ name: 'featured', label: 'Featured' }),
+    // The custom field of this project. tina/rating-field.tsx is the whole plugin.
+  ],
+} satisfies CollectionSchema;
+
```
joshbermanssw, line 56: **Stale comment.** This looks copy-pasted from `postCollection`. `pageCollection` has no custom field, and `rating-field.tsx` is not referenced here, so the comment describes something that is not on this collection.

Suggest deleting it.

While in this file: `biome check` currently fails on it. The `t.reference({ ... })` call above wants to be multi-line, and there is a double blank line before `pageCollection`. CI will catch it.

packages/v4/@tinacms/tinacms/src/core/form/references.ts
```diff
@@ -0,0 +1,105 @@
+import type {
+  CollectionSchema,
+  FieldSchema,
+  TinaDocument,
+} from '../schema/types';
+
+const REFERENCE_FIELD_TYPE = 'reference';
```
joshbermanssw: **Duplicated constant, and it fails silently.**

`REFERENCE_FIELD_TYPE` is also declared in `plugins/fields/reference/reference-field.schema.ts:4`. The two are independent string literals, so renaming the plugin's constant leaves this one pointing at the old value and preview resolution quietly stops matching. No type error, no failing test that I can see.

Importing the exported one from the plugin would couple core to a plugin, which is presumably why it was retyped. If that coupling is unwanted, the alternative is to stop matching on the type string at all; see my note on `childFieldsOf` below.

packages/v4/@tinacms/tinacms/src/core/form/references.ts
```diff
@@ -0,0 +1,105 @@
...
+  path: string;
+}
+
+const isDocument = (value: unknown): value is TinaDocument =>
+  typeof value === 'object' && value !== null && !Array.isArray(value);
+
+const childFieldsOf = (node: FieldSchema): FieldSchema[] | undefined => {
+  const { fields } = node as { fields?: FieldSchema[] };
+  return Array.isArray(fields) ? fields : undefined;
+};
```
joshbermanssw, line 18: **Second schema walk to keep in sync.**

`core/form/ingest.ts` already walks a document against its fields, and it dispatches per node through the registry so each field plugin owns its own recursion. This file hand-rolls a parallel walk: `childFieldsOf` reads `node.fields` directly, with its own array and object handling, and branches on a hardcoded `'reference'` type string.

Two consequences:

1. Any field plugin that nests children some other way is invisible here. `templates` is the live example: `FieldSchema.templates` holds `TemplateSchema.fields`, which this walk never enters. Not a bug today (only `rich-text` uses `templates`, and it holds no reference fields), but it becomes one the moment array templates land.
2. It cuts against the #6921 AC *"Reference field behaviour is isolated and testable"*. The resolution logic lives in core keyed off a literal, not in the plugin.

Not asking for a rewrite in this PR. Worth a `TODO` with an ADR pointer, or folding the traversal into the registry-driven path later.

Separately: this preview resolution work (this file plus the `preview-connection.ts` rework, roughly 190 lines) is not in #6921 at all. It is defensible, the field is not much use in preview without it, but it is the riskiest code in the diff and it sits on the hot path of every preview post for every field type. Either add a line to #6921 or split it out so it is tracked.

packages/v4/@tinacms/tinacms/src/editor/preview-connection.ts
```diff
@@ -17,26 +26,74 @@ export function usePreviewConnection(
...
+        if (cachedDocument(collection, path)) continue;
+        queryClient
+          .fetchQuery({
+            queryKey: contentKeys.document(collection, path),
+            queryFn: () => content.get(collection, path),
+            staleTime: CONTENT_STALE_TIME,
+          })
+          .then(repost)
+          .catch(() => {});
+      }
```
joshbermanssw, line 91: **Bug: a deleted reference re-fetches and re-posts on every keystroke.**

`cachedDocument` unwraps `?.document`. When `content.get` resolves `null` (the document is gone), react-query caches `null`, so `cachedDocument` returns `undefined` and the `if (cachedDocument(...)) continue;` guard never trips.

So for a broken reference, every `publish` re-enters `fetchQuery`. Inside `CONTENT_STALE_TIME` that is a cached resolve plus a redundant `repost()`, meaning a duplicate `postMessage` per keystroke. After 30s it is a real `content.get` per keystroke.

This is exactly the missing-reference case the field handles so carefully in the UI.

Fix is to test cache presence rather than the unwrapped document:

```ts
if (queryClient.getQueryState(contentKeys.document(collection, path))) continue;
```

Also on the `.catch(() => {})`: this swallows the failure with nothing logged. `AGENTS.md` asks for narrowed handling on caught values. Falling back to the bare path is the right behaviour, but a failed content fetch leaving no trace will be painful to debug.

joshbermanssw: Confirmed, and it was worse than a duplicate fetch inside the stale window. Reproduced it: one keystroke, two `postMessage` calls.

Guard now tests the cache entry rather than the unwrapped document:

```ts
const key = contentKeys.document(collection, path);
if (queryClient.getQueryState(key)) continue;
```

Regression test in `preview-connection.test.tsx` asserts one post per edit for a reference whose document is gone. It fails with `expected "spy" to be called 1 times, but got 2 times` against the old guard.

My first attempt at that test asserted the `content.get` call count and passed against the broken code, because `fetchQuery` serves from cache inside `CONTENT_STALE_TIME`. Your description was more precise than my test; the post count is the symptom that shows inside the window.

`.catch` now narrows the cause and logs which document failed and which collection it was read from.

packages/v4/@tinacms/ui/src/components/combobox.tsx
joshbermanssw: Fixed. `aria-label='Clear'` on the button, `aria-hidden='true'` on the icon, and both tests now use `getByRole('button', { name: 'Clear' })` instead of reaching for the data-slot.

Agree on the shadcn tension. `AGENTS.md` says not to hand-write primitives shadcn provides, not that generated output is untouchable, and the `cn` import already had to be corrected in the same file.

packages/v4/@tinacms/tinacms/src/core/form/references.ts
joshbermanssw: Fixed, in the direction you did not list: core owns the key and the plugin imports it.

Core never imports from plugins anywhere (checked), so I did not want to be the first. `REFERENCE_FIELD_TYPE` now sits in `core/schema/types.ts` beside `FieldSchema.collections`, which core already carries for this field alone. `reference-field.schema.ts` re-exports it so the plugin keeps the same surface every other field has, and `reference-field.plugin.ts` is unchanged.

This is a smaller version of what you raise below. #7534 covers removing it from core entirely.

packages/v4/examples/barebones/tina/config.ts
joshbermanssw: Both fixed. Comment deleted, and the formatting was CI's `pnpm format && git diff --exit-code` step, caught in bc5f42f.

packages/v4/@tinacms/tinacms/src/core/form/references.ts
joshbermanssw: Both halves actioned, neither by rewriting here.

**The walk.** #7534 tracks folding it into the registry-driven path, with the `templates` gap and the array-templates trigger you named. `core/form/references.ts` carries a `TODO(#7534)`. Worth adding that this file is the only place in core that matches on a field type key at all; everything else dispatches through the registry, which supports your read that it is the odd one out.

**The scope.** You are right that it was not in #6921. I have added an AC there rather than splitting it, because the field is close to useless in the preview without it and shipping them apart would leave a release where a reference renders as a raw path:

> The preview receives a reference as the document it points at, not as the stored path, so a site reads one shape whether it renders statically or through the editor.

Say the word if you would rather it were a separate PBI and I will split it.

One more thing found while acting on your note below: that hot path had a real bug for a deleted reference, now fixed with a regression test.

packages/v4/@tinacms/tinacms/src/plugins/fields/reference/reference-field.schema.ts
```diff
@@ -0,0 +1,40 @@
...
+import {
+  type BaseFieldSchema,
+  type FieldSchema,
+  REFERENCE_FIELD_TYPE,
+} from '../../../core/schema/types';
+
+// The key lives in core, which reads it to resolve references for the
+// preview. Re-exported so this file stays the plugin's schema surface, as
+// every other field's does.
+export { REFERENCE_FIELD_TYPE };
```
brookjeynes-ssw: Should this not be defined inside this file and exported for consumers?

packages/v4/@tinacms/tinacms/src/plugins/fields/reference/reference-field.ui.tsx
```diff
@@ -0,0 +1,118 @@
...
+      ? { value, label: `${value} (missing)` }
+      : null;
+
+  const items = missing ? [missing, ...options] : options;
+  const selected = items.find((item) => item.value === value) ?? null;
+  const unusable = isLoading || error !== null;
+
+  return (
+    <FieldWrapper errors={error ? [...errors, error.message] : errors}>
+      <Combobox
```
brookjeynes-ssw, line 91: Any reason we're using a combobox over the native select element?

packages/v4/@tinacms/tinacms/src/plugins/fields/reference/reference-field.schema.ts
joshbermanssw: Yes, and it is back there now in e060da29.

You were right that it was the odd one out. Every other field declares its own `<TYPE>_FIELD_TYPE` in its schema file; this one reached into core because `core/form/references.ts` matched `node.type` against the literal, and core imports from plugins nowhere else.

The check turned out to be redundant. `targetOf` already required `node.collections`, which only a reference declares, and already rejected a value that is not a path, so the type comparison never decided anything:

```ts
// before
if (node.type === REFERENCE_FIELD_TYPE) {
  const target = targetOf(node, value, collections);
  ...
}

// after
const target = targetOf(node, value, collections);
if (target) { ... }
```

So core now names no field type at all, `REFERENCE_FIELD_TYPE` is declared and exported from `reference-field.schema.ts` like its siblings, and the re-export is gone. 779 tests pass.

This is also the direction the earlier thread on `references.ts` floated as the alternative to coupling core to a plugin, so it closes both.

packages/v4/@tinacms/tinacms/src/plugins/fields/reference/reference-field.ui.tsx
joshbermanssw: Search. A collection can hold far more documents than a select can usefully show, and the native element has no typeahead.

It was a `Select` in the first draft, same as the select field. #6921 asks for it explicitly:

> Rendering supports **searching**, selecting and clearing referenced content.

The labelling makes it matter more than it otherwise would. `content.list` returns a `DocumentSummary`, which deliberately carries a path and no content, so every option reads as `content/pages/about.mdx` rather than a title. Scanning that list by eye is worse than scanning titles, so typing to filter is doing real work.

It is still a Base UI primitive from the same shadcn family as `select.tsx`, so the dependency is unchanged; `@base-ui/react` was already here.

brookjeynes-ssw (approved):

Issue #6922, opened 2026-05-19
Field Type Implementation - Object


18-th:
## Pain

Object fields group nested values, but their behaviour often bleeds into form navigation, nested editing and serialization logic. Without a plugin implementation, object editing remains a special case instead of using the same field contract as scalar and array fields.

## Approach

Implement `ObjectField` as a field plugin against the field contract. The plugin owns object rendering, nested field composition, ingest, digest and validation, while supporting the form system's flat-mode editing where applicable.

## Expected outcomes

- Object fields work through the v4 field plugin system.
- Nested object editing composes with other field plugins.
- Object serialization and validation are owned in one predictable place.
- Object fields can participate in flat-mode editing rather than forcing deep navigation.

## Acceptance Criteria

- `ObjectField` exists as a field plugin that conforms to the field contract.
- Ingest and digest behaviour preserves the existing object schema and stored value behaviour.
- Rendering supports nested object fields through the shared field plugin system.
- Object fields work with dirty state, validation and persistence tracking.
- Flat-mode editing is supported where the form system enables it.
- Tests cover rendering, nested value updates, ingest, digest, validation and flat-mode behaviour.

brookjeynes-ssw:
Done as of https://github.com/tinacms/tinacms/pull/7535

<img width="1496" height="1372" alt="Image" src="https://github.com/user-attachments/assets/5cd2b017-1c47-402a-bf93-e01c1175ea59" />

PR #7535, opened 2026-09-09
feat(v4): add object field

brookjeynes-ssw:
This PR adds support for `Object`.

.. code:: tsx

	const collection = {
		name: 'post',
		fields: [
			t.object({
				name: 'seo',
				label: 'SEO',
				fields: [
					t.string({ name: 'title', label: 'Title', required: true }),
					t.string({ name: 'description', label: 'Description' }),
				],
			}),

		],
	};

image

Issue: https://github.com/tinacms/tinacms/issues/6922 Field Type Implementation - Object

### DIFF
--- a/packages/v4/@tinacms/tinacms/_docs/field-plugins.md
+++ b/packages/v4/@tinacms/tinacms/_docs/field-plugins.md
@@ -45,6 +45,9 @@ v4 supplies four more examples:
   of item fields. Its items reuse the ordinary field contract through
   `<FieldNode>` and `validateField` — see
   [Compound fields](#compound-fields) below.
+- The `object` field ([`object-field.md`](./object-field.md)) groups a fixed
+  set of nested fields under one name. It is the `array` field without the
+  repetition, and it uses the same compound-field mechanism.
 - The `select` field ([`select-field.md`](./select-field.md)) picks one value
   from a fixed `options` list. Its Zod validator is a `z.enum` with a custom
   `errorMap`, and it has no `defaultValue`, `parse`, or `serialize`.
--- a/packages/v4/@tinacms/tinacms/_docs/field-plugins.md
+++ b/packages/v4/@tinacms/tinacms/_docs/field-plugins.md
@@ -254,9 +254,11 @@ need three extra pieces. Each one has a plain, ordinary counterpart; refer to
 
 - **Rendering** — `<FieldNode address node>` (`editor/field.tsx`) renders one
   resolved node at an address, the same way `<Field>` does after it resolves a
-  node by name. A compound field's component calls `<FieldNode>` directly,
-  once for each item field, with a nested address such as `items.0.title`.
-  Export it from your own component the same way `array-field.ui.tsx` does.
+  node by name. A compound field's component renders one child field per row
+  with `<NestedFieldRow address node>` (`editor/field.tsx`) — the label and
+  `<FieldNode>` pair, with a nested address such as `items.0.title`. `array`
+  and `object` both use it; a new compound field imports it, it does not copy
+  it.
 - **Parse and serialize** — `context.registry` (`FieldTransformContext`,
   `core/field/contract.ts`) carries the registry into `parse` and `serialize`.
   A compound field's `parse`/`serialize` calls `ingestDocument`/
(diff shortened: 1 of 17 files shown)


joshbermanssw (commented):
Withdrawn — posted in error. Ignore.

packages/v4/@tinacms/tinacms/src/plugins/fields/object/object-field.ui.tsx
```diff
@@ -0,0 +1,59 @@
...
+import {
+  FieldNode,
+  useFieldAddress,
+  useFieldErrors,
+  useFieldRegistry,
+  useFieldSchema,
+} from '../../../editor';
+import { asObjectFieldSchema } from './object-field.schema';
+
+function ObjectFieldRow({
```
joshbermanssw: Is the overlap with `ItemFieldRow` deliberate?

This is byte-for-byte identical to `array-field.ui.tsx:18-37` apart from the component name: same `labelable` lookup, same `<Label>` with the same classes, same `<FieldNode>`. I diffed them to be sure.

If that is intentional (keep each compound field self-contained, accept the copy), fair enough, and worth a line in `field-plugins.md` under Compound fields so the next one knows to copy rather than import.

If not, a shared row exported alongside `FieldNode` would mean the third compound field gets it free.

Either way, one thing both copies drop: `document-form.tsx:23-38` renders a dirty indicator on its rows and these do not, so a nested field never shows the unsaved dot that a top-level field does. Pre-existing and not this PR's to fix, but a shared row is where it would get fixed once.

packages/v4/@tinacms/tinacms/src/plugins/fields/object/object-field.test.tsx
```diff
@@ -0,0 +1,238 @@
...
+    const title = await screen.findByLabelText('Title');
+    await userEvent.clear(title);
+    await waitFor(() => expect(screen.getAllByRole('alert')).toHaveLength(2));
+    for (const alert of screen.getAllByRole('alert')) {
+      expect(alert).toHaveTextContent('Title is required');
+    }
+  });
+});
+
+describe('ObjectField ingest and digest', () => {
```
joshbermanssw, line 211: Could this describe block gain the `array` composition cases?

Object-inside-array is the blocks shape, and it is the main reason both compound fields exist, but it is the one combination with no coverage here. The nested `number` case is the right instinct; one more level would lock in the property that actually matters, which is that neither field knows the other exists.

I ran a throwaway probe against this branch and all three pass, so this is a request for tests rather than a fix:

```ts
// object inside array
ingestDocument({ blocks: [{ hero: { height: 10 } }] }, [blocks], { registry })
// -> { blocks: [{ hero: { height: '10' } }] }, digests back to 10

// array inside object: round-trips clean

// validation through both
validateFieldTree(seo, registry.get('object'), { tags: [{ label: '' }] }, 'seo', registry)
// -> { 'seo.tags.0.label': ['label is required'] }
```

packages/v4/@tinacms/tinacms/_docs/field-plugins.md
```diff
@@ -45,6 +45,9 @@ v4 supplies four more examples:
   of item fields. Its items reuse the ordinary field contract through
   `<FieldNode>` and `validateField` — see
   [Compound fields](#compound-fields) below.
+- The `object` field ([`object-field.md`](./object-field.md)) groups a fixed
+  set of nested fields under one name. It is the `array` field without the
+  repetition, and it uses the same compound-field mechanism.
```
joshbermanssw, line 50: Could `_docs/plugins.md` get the matching entry?

`plugins.md` carries the index of field docs under "More data", and every other field v4 ships is listed there. `object-field.md` is not, so it is unreachable from the plugin docs entry point even though this file links it.

```md
  - [The `object` field](./object-field.md) — the field group that v4 supplies
```

brookjeynes-ssw: Resolved by 428aec280a400dc2a9773ed308e2ab071fe04004

packages/v4/@tinacms/tinacms/src/plugins/fields/object/object-field.test.tsx
brookjeynes-ssw: Resolved by b329215f25b0439eea1a29cb1dcbb09f1e80b69b

packages/v4/@tinacms/tinacms/src/plugins/fields/object/object-field.ui.tsx
brookjeynes-ssw: Resolved by 7b12ef3cebb29db7a12494a143ab58075a6b7d4b

joshbermanssw (dismissed):
LGTM ✅

All three comments addressed. Checked the branch out: 768 tests pass, `pnpm types` and `pnpm format` clean.

joshbermanssw (approved):

PR #7541, opened 2026-09-09
feat(tinacms): let the CMS write the copy for editorial-workflow indexing failures

### LINKED ISSUE #7540: ✨ CMS should write the copy for editorial-workflow indexing failures, not print TinaCloud's string
### 🩸 Pain

When an editorial-workflow save fails during indexing, the sentence the editor reads is written in TinaCloud and sent down as an opaque string. The CMS just prints it. That is backwards: the backend should say what went wrong, and the CMS should decide how to say it to a person. It already works that way for every other editorial-workflow failure, where `getEditorialWorkflowErrorMessage` switches on an error code and writes the copy itself. Indexing failures are the exception.

The create-branch modal also renders the message as plain text, so a docs link prints as `[text](url)` and newlines collapse into spaces.

### 🛠️ Approach

- [ ] Add an indexing-failure code alongside the existing `EDITORIAL_WORKFLOW_ERROR` codes and map it to friendly copy in the CMS
- [ ] Name the file that failed to index in that copy
- [ ] Render the event-log docs link in the create-branch modal as a real anchor, and let the copy break across lines
- [ ] Delete the temporary server-side message from TinaCloud once this ships

This depends on TinaCloud emitting an error code and the offending file alongside the existing plain message. Keep falling back to that message, because `tinacms` is published and editors run old versions.

### ✅ Acceptance Criteria

- [ ] An indexing failure shows CMS-authored copy that names the file that failed
- [ ] The docs link is clickable in the modal, not printed as markdown
- [ ] A CMS build that does not recognise the new code still shows TinaCloud's plain message

### 🔗 Related

- https://tina.io/docs/tinacloud/dashboard/event-log#common-causes
- https://github.com/tinacms/tinacloud/pull/4069
- https://github.com/tinacms/tinacloud/issues/4129
- https://github.com/tinacms/tinacloud/issues/3821

kulesy:
Closes #7540

**TL;DR** Move the wording of editorial-workflow indexing failures out of TinaCloud and into the CMS, keyed on TinaCloud's `INDEXING_FAILED` code. The branch modal now names the file that failed and links the docs as a real anchor, and any error the CMS does not recognise still shows TinaCloud's plain message.

**Pain:** when a save fails during indexing, the editor reads a sentence written server-side and handed over as an opaque string. It cannot even carry a working link, because the modal prints plain text, so one shows up literally as `[text](url)`.

**Solution:** the CMS now owns that copy: it maps the code to its own wording, names the file TinaCloud reports, and links the event-log docs. TinaCloud has sent the code and the file since https://github.com/tinacms/tinacloud/pull/4069.


kulesy:
Before and after on the editorial-workflow indexing failure, both from my TinaCloud dev stage running this branch, not production.

The repro pushes a content file whose block template doesn't exist, then saves a page from protected main. TinaCloud sends errorCode INDEXING_FAILED with the failing file, and the CMS now writes its own copy instead of printing the server sentence. It names the page and the collection it sits in, resolved from the failing file's own path rather than the collection being edited.

Indexing failure modal without this change

Figure: main at the branch point, printing TinaCloud's raw sentence as plain text.

Indexing failure modal with this change

Figure: this branch emphasises the page and its collection, and links the docs as a real anchor.

Both shots are tinacms 3.14.0 built from the same commit, so the only difference is this change.

packages/tinacms/src/toolkit/form-builder/editorial-workflow-utils.ts
```diff
@@ -109,3 +116,113 @@ export const checkBranchGuard = async (
...
+}
+
+// NOTE: [17 Sep 2026] EK - Always resolve from the failing file's own path, never
+// from the collection being edited: indexing covers the whole branch, so the file
+// that failed can belong to a different collection.
+export const collectionLabelResolver =
+  (schema?: CollectionLookup) =>
+  (file: string): string | undefined => {
+    const collection = schema?.getCollectionByFullPath?.(file);
+    return collection?.label || collection?.name;
```
joshbermanssw: `TinaSchema.getCollectionByFullPath` throws `Unable to find collection for file at ...` when nothing matches (`schema-tools/src/schema/TinaSchema.ts:164`); it never returns `undefined`. The test stubs it to return `undefined`, so it doesn't cover the real class.

This runs inside the `catch` in `use-editorial-workflow.ts:255`, before `setIsExecuting(false)`, so a file outside any collection leaves the progress modal stuck. `sidebar-body.tsx:216` guards the same call.

```suggestion
    try {
      const collection = schema?.getCollectionByFullPath?.(file);
      return collection?.label || collection?.name;
    } catch {
      return undefined;
    }
```

Typing `schema` as `Pick<TinaSchema, 'getCollectionByFullPath'>` instead of the hand-rolled `CollectionLookup` would have surfaced this through the return type.

packages/tinacms/src/toolkit/components/media/media-workflow-overlay.tsx
```diff
@@ -177,7 +184,10 @@ export const MediaWorkflowOverlay = () => {
         ...confirmState,
         branchName,
         isChecking: false,
-        errorMessage: getEditorialWorkflowErrorMessage(e),
+        errorMessageParts: getEditorialWorkflowError(
+          e,
+          collectionLabelResolver(cms.api.tina.schema)
+        ).messageParts,
```
joshbermanssw: Two things here.

1. `.messageParts` drops `.link`, so the media flow never renders the docs anchor. Keep the whole `EditorialWorkflowErrorCopy` in state and pass `errorLink` through to `CreateBranchPromptModal` on line 209 (it already knows how to render it).

2. An indexing failure doesn't reach this `catch` at all. `onConfirm` only runs `prepareMediaBranch`; the status poll happens later in `finalizeMediaWorkflow` (`media-store.default.ts:406`), whose `catch` goes through `dispatchMediaWorkflowError` as `err.message` (line 453). The overlay's `phase: 'error'` branch then prints `{state.message}` verbatim on line 253, so `[text](url)` still shows literally for media, which is the headline case in #7540.

Have `dispatchMediaWorkflowError` run the error through `getEditorialWorkflowError` (or carry the raw error and let the overlay do it) so the error phase renders `EditorialWorkflowErrorText` plus the anchor, same as the confirming phase.

packages/tinacms/src/toolkit/form-builder/branch-deleted-modal.tsx
```diff
@@ -97,11 +98,12 @@ export const BranchDeletedModal = ({
...
               </p>
 
-              {errorMessage && (
+              {errorMessageParts && (
                 <div className='flex items-center gap-1 text-red-700 py-2 px-3 mb-4 bg-red-50 border border-red-200 rounded'>
                   <CircleAlert className='w-5 h-auto text-red-400 flex-shrink-0' />
                   <span className='text-sm'>
-                    <b>Error:</b> {errorMessage}
+                    <b>Error:</b>{' '}
+                    <EditorialWorkflowErrorText parts={errorMessageParts} />
```
joshbermanssw: This modal gets the same `INDEXING_FAILED` copy but drops `errorLink` and has no `whitespace-pre-line`, so the `\n\n` in `indexingFailureCopy` collapses and the docs link never appears. Needs `errorLink` destructured from `useEditorialWorkflow()` on line 46 as well.

```suggestion
                  <span className='text-sm whitespace-pre-line'>
                    <b>Error:</b>{' '}
                    <EditorialWorkflowErrorText parts={errorMessageParts} />
                    {errorLink && (
                      <>
                        {' '}
                        <a
                          className='underline text-tina-orange-dark font-medium'
                          href={errorLink.url}
                          target='_blank'
                          rel='noreferrer'
                        >
                          {errorLink.label}
                        </a>
                      </>
                    )}
```

That's now the third copy of this box. Lifting the whole thing (icon, span, anchor) into `EditorialWorkflowErrorText` and storing `error?: EditorialWorkflowErrorCopy` as one state value in the hook instead of two would stop the three modals drifting apart.

joshbermanssw (changes requested):
Nice direction; the CMS owning this copy is the right call. Three things block it, all on the fallback surfaces rather than the happy path in `CreateBranchModal`:

- `collectionLabelResolver` calls a method that throws on no match, inside the hook's `catch`, so the modal can hang.
- Media indexing failures never reach the new copy; they still print TinaCloud's raw string with the literal `[text](url)`.
- `BranchDeletedModal` and the media overlay drop `errorLink`, so the anchor only exists in one of the three modals.

Smaller notes that can go to a follow-up: the dated `NOTE: [17 Sep 2026] EK` comment has no precedent in the repo (comments are a last resort per `AGENTS.md`), the JSDoc on lines 120/126/137 restates the names, and the test "keeps message in step with parts" refers to a `message` field that was removed in the last commit.

packages/tinacms/src/toolkit/form-builder/editorial-workflow-utils.ts
kulesy: Fixed in 43b3b49. Typed as Pick<TinaSchema, 'getCollectionByFullPath'> and guarded the way sidebar-body.tsx does, and the test now throws the way the real schema does instead of returning undefined.

packages/tinacms/src/toolkit/components/media/media-workflow-overlay.tsx
kulesy: Ran this against a real failing index: the overlay shows TinaCloud's own "Failed to complete media operation. Please try again.", because media-editorial-workflow.ts sets no errorCode and no INDEXING_FAILED anywhere, unlike the content path 4069 changed.

The overlay now carries the raw error through getEditorialWorkflowError, so media picks the copy up for free once TinaCloud sends the code, but the copy itself needs a TinaCloud change. Happy to raise that.

packages/tinacms/src/toolkit/form-builder/branch-deleted-modal.tsx
kulesy: Took the bigger version you suggested. The whole box (icon, whitespace-pre-line span, parts, anchor) is now EditorialWorkflowErrorBox and all three modals use it, with the hook holding one error value instead of two.

kulesy:
All three addressed in 43b3b49, with replies on the threads.

One thing that has no line to sit on: media isn't the headline case in #7540. The Approach and the acceptance criteria both name the create-branch modal, and media isn't mentioned. Worth fixing either way, but it's scope the issue didn't ask for.

joshbermanssw (approved):
LGTM

Issue #3089, opened 2022-08-10
Tina accepts `templates: []`, but builds an invalid schema out of it


L-A:
## Description

When a field specifies `templates: []`, Tina will silently generate an invalid GraphQL schema.

### Steps to reproduce

Create a minimal collection where a field uses an empty `templates` list, for example:

```javascript
...
collections: [
    {
      label: "Pages",
      name: "page",
      path: "content/pages",
      format: "mdx",
      fields: [
        {
          name: "body",
          label: "Main content",
          type: "rich-text",
          isBody: true,
          templates: []      // ← This is the important line
        }
      ]
    }
] 
...
```

### Expected result

Tina should either:
- Error out because it can't use an empty array as a list of possible templates,
- or silently ignore it, and build a working schema that doesn't expect templates

### Actual result

Tina declares an object type, for which it defines zero fields. Because this is an invalid GraphQL schema, all requests return this:

```javascript
{
  "errors": [
    {
      "message": "Input Object type PageBodyFilter must define one or more fields."
    }
  ]
}
```

#### Tinacms Package Versions

```text
d
├── @tinacms/cli@0.61.1
├── tinacms@0.68.15

```

#### Tools

| Tool    | Version |
| ------- | ------- |
| browser |    Firefox 103.0.1     |
| os      |     macOS 12.4 (21F79)   |
| node    |    16.15.0    |
| npm     |    8.5.5    |

PR #7547, opened 2026-09-10
fix: reject templates with empty fields arrays (#7490)

### LINKED ISSUE #7490: 🐛 Schema validation allows a template with empty `fields`, producing an invalid GraphQL schema

wakqasahmed:
Fixes #7490.

A template with `fields: []` passes schema validation today but generates a zero-field GraphQL input filter type, which is invalid per the GraphQL spec ("Input Object type ...Filter must define one or more fields"). On TinaCloud that breaks every content API request for the project.

The sibling case - `templates: []` on a collection - already got a `.min(1, 'Property \`templates\` cannot be empty.')` guard for the same reason (#3089). Added the equivalent guard to `Template.fields` in `schema.ts`, and found it needed a second copy in `fields.ts`'s `TemplateTemp` too - rich-text/object-field templates validate through that separate lazy schema, not through `schema.ts`'s `Template`, so `schema.ts`'s guard alone didn't cover them.

One related thing the issue flags but this doesn't close: a template whose fields are *all* `displayOnly` or `password` type hits the same zero-field-filter wall, since `_buildFieldFilter` excludes those types when building the filter. A plain `.min(1)` on the array length doesn't catch that (array length 1, still zero-field filter). Duplicating `_buildFieldFilter`'s exact exclusion list into this validation layer felt like the wrong tradeoff - it's private policy that lives in the GraphQL builder, and a copy here would just be another thing to keep in sync. Left as a known gap rather than reaching for a fix that's more architecture change than the issue's own acceptance criteria ask for.

Added test cases for both the collection-template and rich-text-template empty-`fields` rejection, plus confirmed the existing "at least one real field" cases still pass.

### DIFF
--- a/packages/@tinacms/schema-tools/src/validate/fields.ts
+++ b/packages/@tinacms/schema-tools/src/validate/fields.ts
@@ -149,7 +149,9 @@ export const TinaFieldZod: z.ZodType<TinaFieldType> = z.lazy(() => {
     .object({
       label: z.string().optional(),
       name,
-      fields: z.array(TinaFieldZod),
+      fields: z
+        .array(TinaFieldZod)
+        .min(1, 'Property `fields` cannot be empty.'),
       match: z
         .object({
           start: z.string(),
--- a/packages/@tinacms/schema-tools/src/validate/index.spec.ts
+++ b/packages/@tinacms/schema-tools/src/validate/index.spec.ts
@@ -345,6 +345,42 @@ const schemaWithEmptyTemplates: Schema = {
     },
   ],
 };
+const schemaWithEmptyTemplateFields: Schema = {
+  collections: [
+    {
+      name: 'foo',
+      path: 'foo/bar',
+      templates: [
+        {
+          name: 'bar',
+          label: 'Bar',
+          fields: [],
+        },
+      ],
+    },
+  ],
+};
+const schemaWithEmptyRichTextTemplateFields: Schema = {
+  collections: [
+    {
+      name: 'foo',
+      path: 'foo/bar',
+      fields: [
+        {
+          name: 'body',
+          type: 'rich-text',
+          templates: [
+            {
+              name: 'bar',
+              label: 'Bar',
+              fields: [],
+            },
+          ],
+        },
+      ],
+    },
+  ],
+};
 const schemaWithInvalidFiledNesterUnderRichText = {
   collections: [
     {
@@ -466,6 +502,14 @@ describe('validateSchema', () => {
     expect(() => {
       validateSchema({ schema: schemaWithEmptyTemplates });
     }).toThrow();
+  });
+  it.each([
+    ['collection', schemaWithEmptyTemplateFields],
+    ['rich-text', schemaWithEmptyRichTextTemplateFields],
+  ])('fails when %s template fields is empty', (_templateType, schema) => {
+    expect(() => {
+      validateSchema({ schema });
+    }).toThrow('Property `fields` cannot be empty.');
   });
   it('fails when a deeply nested field under a template is invalid', () => {
     expect(() => {
--- a/packages/@tinacms/schema-tools/src/validate/schema.ts
+++ b/packages/@tinacms/schema-tools/src/validate/schema.ts
@@ -17,7 +17,7 @@ const Template = z
       required_error: 'Missing `label` property. Property `label` is required.',
     }),
     name: name,
-    fields: z.array(TinaFieldZod),
+    fields: z.array(TinaFieldZod).min(1, 'Property `fields` cannot be empty.'),
   })
   .superRefine((val, ctx) => {
     const dups = findDuplicates(val.fields?.map((x) => x.name));
--- a/pnpm-workspace.yaml
+++ b/pnpm-workspace.yaml
@@ -57,7 +57,7 @@ catalog:
     "@graphql-tools/graphql-file-loader": ^7.5.17
     "@graphql-tools/load": ^7.8.14
     "@graphql-tools/utils": ^10.8.1
-    "@headlessui/react": 2.1.8
+    "@headlessui/react": 2.2.10
     "@iarna/toml": ^2.2.5
     "@monaco-editor/loader": ^1.7.0
     "@octokit/rest": ^22.0.1
--- a/.changeset/bump-headlessui-react-19.md
+++ b/.changeset/bump-headlessui-react-19.md
@@ -1,0 +1,5 @@
+---
+'@tinacms/rich-text': patch
+---
+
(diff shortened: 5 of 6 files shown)


packages/@tinacms/schema-tools/src/validate/schema.ts
```diff
@@ -17,7 +17,7 @@ const Template = z
       required_error: 'Missing `label` property. Property `label` is required.',
     }),
     name: name,
-    fields: z.array(TinaFieldZod),
+    fields: z.array(TinaFieldZod).min(1, 'Property `fields` cannot be empty.'),
```
wakqasahmed, line 20: Confirmed this only guards the top-level `TinaCloudCollection.templates` array (collection block/union templates). Anything nested under a `rich-text` or `object` field - even recursively - validates through `TemplateTemp` in fields.ts instead, which is a separate schema object, not a shared reference to this one. So the fields.ts half of this PR is load-bearing, not a defensive copy for symmetry.

packages/@tinacms/schema-tools/src/validate/fields.ts
```diff
@@ -149,7 +149,9 @@ export const TinaFieldZod: z.ZodType<TinaFieldType> = z.lazy(() => {
     .object({
       label: z.string().optional(),
       name,
-      fields: z.array(TinaFieldZod),
+      fields: z
+        .array(TinaFieldZod)
+        .min(1, 'Property `fields` cannot be empty.'),
```
wakqasahmed, line 154: This will also reject a legitimate-looking pattern: a rich-text template representing a self-closing/void embed with no props. `@tinacms/mdx`'s own test suite models exactly that - `packages/@tinacms/mdx/src/next/tests/mdx-basic-nested-objects-3/field.ts` has `templates: [{ name: 'World', fields: [] }]` for `<World />` embedded in markdown. That fixture bypasses `validateSchema` (calls `parseMDX`/`serializeMDX` directly) so it won't fail CI, but it shows `fields: []` is a deliberately-supported shape at the mdx-parsing layer, not just leftover misconfiguration. A real schema author using the same pattern now gets a hard build failure. That may be the correct tradeoff per the issue's acceptance criteria (fail-fast unconditionally), but it's a behavior change beyond "reject broken schemas" that the PR doesn't mention - worth being explicit about it, and maybe pointing at the `_value`-placeholder-field workaround the shortcode fixtures (`wordpress-style`, `unrecognized-shortcodes`) already use for the same situation.

packages/@tinacms/schema-tools/src/validate/index.spec.ts
```diff
@@ -467,6 +503,14 @@ describe('validateSchema', () => {
       validateSchema({ schema: schemaWithEmptyTemplates });
     }).toThrow();
   });
+  it.each([
```
wakqasahmed, line 506: Minor: this is the only `it.each` in this file - every other case is a standalone `it(...)`. Not wrong (arguably nicer than two near-duplicate blocks), just flagging since it's a new pattern for this test file rather than matching what's around it.

pnpm-workspace.yaml
```diff
@@ -57,7 +57,7 @@ catalog:
     "@graphql-tools/graphql-file-loader": ^7.5.17
     "@graphql-tools/load": ^7.8.14
     "@graphql-tools/utils": ^10.8.1
-    "@headlessui/react": 2.1.8
+    "@headlessui/react": 2.2.10
```
wakqasahmed, line 60: This pin bump (and the matching changeset + lockfile changes) is unrelated to #7490 and is already its own open PR (#7518). Looks like this branch was cut from that unmerged branch rather than `main`. Rebasing onto `main` would keep this PR scoped to just the fields-validation fix.

wakqasahmed (commented):
Reviewed cold against #7490, no prior context. The core fix is right: I traced both schemas by hand and `Template.fields` in schema.ts and `TemplateTemp.fields` in fields.ts really are two independent zod objects, not one re-exported into the other. `TemplateTemp` is defined inside the `z.lazy()` closure in fields.ts and never imported by schema.ts, so a template nested under any `rich-text` or `object` field (at any depth, including inside a collection-level `Template`'s own fields) only ever validates through `TemplateTemp` - the schema.ts guard alone would not have caught it. So both edits are needed, not a defensive duplicate.

Two things I'd want addressed before merge:

1. This diff also carries an unrelated `@headlessui/react` 2.1.8 -> 2.2.10 bump (changeset + pnpm-lock.yaml + pnpm-workspace.yaml), which is already its own open PR (#7518). Looks like this branch was cut on top of that unmerged branch instead of `main`. Worth rebasing so this PR's diff is only the validation fix - right now it's asking reviewers/CI to also sign off on a dependency bump that has nothing to do with #7490.

2. There's a real gap in the false-positive analysis: `@tinacms/mdx`'s own test suite already models a rich-text template with `fields: []` on purpose, for a self-closing embed - see `packages/@tinacms/mdx/src/next/tests/mdx-basic-nested-objects-3/field.ts`, `templates: [{ name: 'World', fields: [] }]`, parsing `# Hello <World />`. That test calls `parseMDX`/`serializeMDX` directly and never goes through `validateSchema`, so it won't fail CI - but it shows the mdx layer intentionally supports a template with zero fields for a prop-less custom element. A real author writing the same pattern in `tina/config` will now get a hard build-time rejection (`Property \`fields\` cannot be empty.`) for something the rich-text parser handles fine on its own. Given the issue's acceptance criteria literally says "fails fast ... when any template has an empty fields array" with no carve-out, this might be the intended, accepted tradeoff - but the PR description doesn't mention it, and it's a real behavior change for a pattern that isn't actually broken at the mdx layer, only downstream in the GraphQL filter type. Worth calling out explicitly (and maybe pointing authors at the same `_value`-placeholder-field workaround the shortcode fixtures already use).

Agree with leaving the displayOnly/password-only case out of scope for this PR - I checked `_buildFieldFilter` myself (`packages/@tinacms/graphql/src/builder/index.ts:1101-1263`): `displayOnly` explicitly returns `undefined` and there's no `case 'password'` at all, so both fall through to an implicit `undefined`, confirming the issue's note. Copying that exclusion list into schema-tools purely to catch an all-displayOnly/password template would be real drift risk for a fix this narrow, and it's disclosed candidly in the PR description rather than swept under the rug.

wakqasahmed:
Closing and reopening as a fresh PR — this branch accidentally included an unrelated commit from another local branch (the `@headlessui/react` bump already tracked in #7518), and separately your review caught a real false-positive: the fix as written also rejected a rich-text embed template with no fields, which `@tinacms/mdx`'s own test fixtures rely on for prop-less self-closing elements. Both fixed in the replacement PR, which scopes the empty-fields guard to only where it actually matters (an object field's templates, which do generate a broken GraphQL filter type when empty) and leaves rich-text embed templates alone (their filter type doesn't depend on embed field counts). Thanks for catching both.

PR #7548, opened 2026-09-10
fix: reject templates with empty fields arrays (#7490)

### LINKED ISSUE #7490: 🐛 Schema validation allows a template with empty `fields`, producing an invalid GraphQL schema

wakqasahmed:
Fixes #7490. Replaces #7547 (closed - that branch accidentally carried an unrelated commit from another local branch).

A collection template with `fields: []` passes schema validation today but generates a zero-field GraphQL input filter type via `_buildTemplateFilter`, which is invalid per the GraphQL spec and breaks every content API request for the project on TinaCloud.

The sibling case - `templates: []` on a collection - already got a `.min(1, 'Property \`templates\` cannot be empty.')` guard for the same reason (#3089). Added the equivalent guard to `Template.fields` in `schema.ts`, and to the shared `TemplateTemp` schema in `fields.ts` (used by both `ObjectField.templates` and `RichTextField.templates`) - both go through the same `_buildTemplateFilter` path once a field actually has templates, traced via `_filterCollectionDocumentType`/`TinaSchema.getTemplatesForCollectable` in `@tinacms/graphql`'s builder.

(Earlier revisions of this PR tried to scope the fix to only `ObjectField`, reasoning a rich-text field's own filter type is fixed regardless of its embed templates - that's only true when the field has no templates at all; once it has any, it falls through to the identical code path `object` fields use. A reviewer caught this. `PageBodySponsorshipTiersFilter` from the issue's own repro decodes as collection `page` + field `body` + template `SponsorshipTiers`, confirming the originally reported bug is exactly this rich-text-embed case.)

Also fixed `@tinacms/mdx`'s `mdx-basic-nested-objects-3` test fixture, which modeled a rich-text embed template with `fields: []` (a prop-less self-closing element) - an actual instance of this same bug that its own test never caught, since that test only round-trips MDX parsing and never calls `validateSchema`.

A separate related gap the issue also flags - a template whose fields are all `displayOnly`/`password` type also produces a zero-field filter, since `_buildFieldFilter` excludes those types - isn't closed by this fix. Duplicating that exclusion list into schema-tools to catch it would risk drifting out of sync with the actual GraphQL filter builder, so it's left as a known limitation rather than folded into this narrower, spec-matching fix.

Tests cover: a collection template with empty fields, an object field's template with empty fields, and a rich-text embed template with empty fields - all rejected.

### DIFF
--- a/packages/@tinacms/mdx/src/next/tests/mdx-basic-nested-objects-3/field.ts
+++ b/packages/@tinacms/mdx/src/next/tests/mdx-basic-nested-objects-3/field.ts
@@ -21,7 +21,15 @@ export const field: RichTextField = {
                 {
                   name: 'content',
                   type: 'rich-text',
-                  templates: [{ name: 'World', fields: [] }],
+                  templates: [
+                    {
+                      name: 'World',
+                      // A schema-valid template needs at least one field even
+                      // though this element is used with none in this fixture's
+                      // markdown (a bare self-closing `<World />`).
+                      fields: [{ name: 'label', type: 'string' }],
+                    },
+                  ],
                 },
               ],
             },
--- a/packages/@tinacms/schema-tools/src/validate/fields.ts
+++ b/packages/@tinacms/schema-tools/src/validate/fields.ts
@@ -149,7 +149,9 @@ export const TinaFieldZod: z.ZodType<TinaFieldType> = z.lazy(() => {
     .object({
       label: z.string().optional(),
       name,
-      fields: z.array(TinaFieldZod),
+      fields: z
+        .array(TinaFieldZod)
+        .min(1, 'Property `fields` cannot be empty.'),
       match: z
         .object({
           start: z.string(),
@@ -199,20 +201,6 @@ export const TinaFieldZod: z.ZodType<TinaFieldType> = z.lazy(() => {
             message: duplicateTemplateErrorMessage(dups),
           });
         }
-        // An object field's templates each get their own generated GraphQL
-        // filter type (_buildTemplateFilter), which is invalid GraphQL when
-        // it has no fields. Rich-text embed templates don't have this
-        // problem (their containing rich-text field always gets one fixed
-        // generic filter regardless of embed shape), so this check is
-        // scoped to ObjectField's templates only, not RichTextField's.
-        val?.forEach((template) => {
-          if (template.fields.length === 0) {
-            ctx.addIssue({
-              code: z.ZodIssueCode.custom,
-              message: 'Property `fields` cannot be empty.',
-            });
-          }
-        });
       }),
   });
 
--- a/packages/@tinacms/schema-tools/src/validate/fields.ts
+++ b/packages/@tinacms/schema-tools/src/validate/fields.ts
@@ -199,6 +199,20 @@ export const TinaFieldZod: z.ZodType<TinaFieldType> = z.lazy(() => {
             message: duplicateTemplateErrorMessage(dups),
           });
         }
+        // An object field's templates each get their own generated GraphQL
+        // filter type (_buildTemplateFilter), which is invalid GraphQL when
+        // it has no fields. Rich-text embed templates don't have this
+        // problem (their containing rich-text field always gets one fixed
+        // generic filter regardless of embed shape), so this check is
+        // scoped to ObjectField's templates only, not RichTextField's.
+        val?.forEach((template) => {
+          if (template.fields.length === 0) {
+            ctx.addIssue({
+              code: z.ZodIssueCode.custom,
+              message: 'Property `fields` cannot be empty.',
+            });
+          }
+        });
       }),
   });
 
--- a/packages/@tinacms/schema-tools/src/validate/index.spec.ts
+++ b/packages/@tinacms/schema-tools/src/validate/index.spec.ts
@@ -345,6 +345,68 @@ const schemaWithEmptyTemplates: Schema = {
     },
   ],
 };
+const schemaWithEmptyTemplateFields: Schema = {
+  collections: [
+    {
+      name: 'foo',
+      path: 'foo/bar',
+      templates: [
+        {
+          name: 'bar',
+          label: 'Bar',
+          fields: [],
+        },
+      ],
+    },
+  ],
+};
+const schemaWithEmptyObjectFieldTemplateFields: Schema = {
+  collections: [
+    {
+      name: 'foo',
+      path: 'foo/bar',
+      fields: [
+        {
+          name: 'items',
+          type: 'object',
+          templates: [
+            {
+              name: 'bar',
+              label: 'Bar',
+              fields: [],
+            },
+          ],
+        },
+      ],
+    },
+  ],
+};
+// A rich-text embed template with no fields is a legitimate, intentionally
+// supported shape (a prop-less self-closing element, e.g. `<Divider />`) -
+// unlike an object field's templates, a rich-text field's own GraphQL filter
+// type doesn't depend on its embed templates' field counts, so this must
+// keep validating successfully.
+const schemaWithEmptyRichTextTemplateFields: Schema = {
+  collections: [
+    {
+      name: 'foo',
+      path: 'foo/bar',
+      fields: [
+        {
+          name: 'body',
+          type: 'rich-text',
+          templates: [
+            {
+              name: 'bar',
+              label: 'Bar',
+              fields: [],
+            },
+          ],
+        },
+      ],
+    },
+  ],
+};
 const schemaWithInvalidFiledNesterUnderRichText = {
   collections: [
     {
@@ -466,6 +528,21 @@ describe('validateSchema', () => {
     expect(() => {
       validateSchema({ schema: schemaWithEmptyTemplates });
     }).toThrow();
+  });
+  it('fails when a collection template fields is empty', () => {
+    expect(() => {
+      validateSchema({ schema: schemaWithEmptyTemplateFields });
+    }).toThrow('Property `fields` cannot be empty.');
+  });
+  it('fails when an object field template fields is empty', () => {
+    expect(() => {
+      validateSchema({ schema: schemaWithEmptyObjectFieldTemplateFields });
+    }).toThrow('Property `fields` cannot be empty.');
+  });
+  it('passes when a rich-text embed template fields is empty', () => {
+    expect(() => {
+      validateSchema({ schema: schemaWithEmptyRichTextTemplateFields });
+    }).not.toThrow();
   });
   it('fails when a deeply nested field under a template is invalid', () => {
     expect(() => {
--- a/packages/@tinacms/schema-tools/src/validate/index.spec.ts
+++ b/packages/@tinacms/schema-tools/src/validate/index.spec.ts
@@ -2,8 +2,8 @@
 
 */
 
-import type { Schema } from '../types';
 import { validateSchema } from '.';
+import type { Schema } from '../types';
 
 let consoleErrMock: any;
 beforeEach(() => {
@@ -381,11 +381,11 @@ const schemaWithEmptyObjectFieldTemplateFields: Schema = {
     },
   ],
 };
-// A rich-text embed template with no fields is a legitimate, intentionally
-// supported shape (a prop-less self-closing element, e.g. `<Divider />`) -
-// unlike an object field's templates, a rich-text field's own GraphQL filter
-// type doesn't depend on its embed templates' field counts, so this must
-// keep validating successfully.
+// A rich-text field WITH templates goes through the same
+// _filterCollectionDocumentType/_buildTemplateFilter path an object field's
+// templates do (confirmed by tracing packages/@tinacms/graphql's builder),
+// so an empty-fields embed template is just as invalid here as it is for an
+// object field.
 const schemaWithEmptyRichTextTemplateFields: Schema = {
   collections: [
     {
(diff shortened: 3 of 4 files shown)


wakqasahmed (commented):
Traced the code paths this PR's reasoning rests on (`_buildFieldFilter` and `getTemplatesForCollectable` in `@tinacms/graphql`), and I don't think the central claim holds - details inline. Short version: RichTextField's filter type is only the fixed generic shape when the field has *no* templates at all. Once `field.templates.length > 0` (any real rich-text embed/block usage), `_buildFieldFilter`'s `rich-text` case falls through to `_filterCollectionDocumentType` -> `_buildTemplateFilter`, same as `object`. An embed template with `fields: []` produces the same zero-field `InputObjectTypeDefinition` there.

That matters because the original issue's "Pain" section says "A **rich-text block template** with `fields: []`...", and the repro's error name (`PageBodySponsorshipTiersFilter`) decodes as collection `page` + field `body` (rich-text, by Tina naming convention) + template `SponsorshipTiers`. That reads like the reported bug is a rich-text embed template, which this version explicitly leaves unguarded - so I don't think this closes #7490's actual repro, only the related object-field case.

The mdx false-positive concern from #7547's review is real (checked `mdx-basic-nested-objects-3`'s fixture), but the fix for that seems like it should be scoped to the mdx fixture/workaround, not exempting all of RichTextField.templates from validation - #7547 guarded both ObjectField and RichTextField.

packages/@tinacms/schema-tools/src/validate/fields.ts
```diff
@@ -199,6 +199,20 @@ export const TinaFieldZod: z.ZodType<TinaFieldType> = z.lazy(() => {
             message: duplicateTemplateErrorMessage(dups),
           });
         }
+        // An object field's templates each get their own generated GraphQL
+        // filter type (_buildTemplateFilter), which is invalid GraphQL when
+        // it has no fields. Rich-text embed templates don't have this
+        // problem (their containing rich-text field always gets one fixed
+        // generic filter regardless of embed shape), so this check is
+        // scoped to ObjectField's templates only, not RichTextField's.
+        val?.forEach((template) => {
```
wakqasahmed: This premise ("rich-text field always gets one fixed generic filter regardless of embed shape") only holds when `field.templates` is empty/undefined. Checked `_buildFieldFilter` in `packages/@tinacms/graphql/src/builder/index.ts` (the `rich-text` case, ~line 1222): the fixed `startsWith`/`eq`/`exists` shape is only returned `if (!field.templates || field.templates.length === 0)`. Once there's at least one embed template, it falls through to `_filterCollectionDocumentType(field)` - the identical call the `object` case makes one branch up - which resolves via `TinaSchema.getTemplatesForCollectable` to a `union` (since a rich-text field has `templates`, not `fields`, at its own level) and calls `_buildTemplateFilter(template)` per embed template, same as ObjectField. So an empty-fields rich-text embed template hits the exact same zero-field `InputObjectTypeDefinition` bug this PR is fixing for `object`. The RichTextField.templates guard below (not added here) would need the same check.

packages/@tinacms/schema-tools/src/validate/fields.ts
```diff
@@ -199,6 +199,20 @@ export const TinaFieldZod: z.ZodType<TinaFieldType> = z.lazy(() => {
...
+          if (template.fields.length === 0) {
+            ctx.addIssue({
+              code: z.ZodIssueCode.custom,
+              message: 'Property `fields` cannot be empty.',
+            });
+          }
+        });
       }),
```
wakqasahmed, line 204: The guard added just above (ObjectField.templates) doesn't extend to `RichTextField.templates` a few lines down (not touched by this diff) - per the code trace on the comment above, `RichTextField.templates` needs the same check once `field.templates.length > 0`. Also worth noting the original issue's "Pain" section literally says "A rich-text block template with `fields: []`...", and the repro error `PageBodySponsorshipTiersFilter` decodes (via `NAMER.dataFilterTypeName`) as collection `page` + field `body` (the standard Tina name for a rich-text field) + template `SponsorshipTiers`. That reads like the originally reported bug is specifically this unguarded case.

packages/@tinacms/schema-tools/src/validate/index.spec.ts
```diff
@@ -467,6 +529,21 @@ describe('validateSchema', () => {
...
+    expect(() => {
+      validateSchema({ schema: schemaWithEmptyTemplateFields });
+    }).toThrow('Property `fields` cannot be empty.');
+  });
+  it('fails when an object field template fields is empty', () => {
+    expect(() => {
+      validateSchema({ schema: schemaWithEmptyObjectFieldTemplateFields });
+    }).toThrow('Property `fields` cannot be empty.');
+  });
+  it('passes when a rich-text embed template fields is empty', () => {
```
wakqasahmed: This test only proves schema-tools' `validateSchema` accepts this shape - it doesn't prove the shape is safe downstream. Per the trace on the `RichTextField.templates` schema in fields.ts, building this exact schema into GraphQL (`@tinacms/graphql`'s builder) would hit the same "Input Object type ... must define one or more fields" error the PR is meant to prevent, since `field.templates.length > 0` here routes through `_buildTemplateFilter` rather than the fixed richText filter shape. A test at this layer alone can't validate the PR's core claim about the GraphQL builder's behavior.

packages/@tinacms/schema-tools/src/validate/fields.ts
wakqasahmed: Confirmed you're right - I stopped reading that switch case one branch too early and missed the fallthrough to _filterCollectionDocumentType when field.templates.length > 0. Traced the full call chain myself (including through TinaSchema.getTemplatesForCollectable) and it matches your finding exactly. Reverted to guarding the shared TemplateTemp schema directly so both ObjectField and RichTextField are covered, and confirmed via the PageBodySponsorshipTiersFilter naming decode that this is indeed what the original issue reported. Also found and fixed @tinacms/mdx's mdx-basic-nested-objects-3 fixture, which was an actual instance of this same bug (its own test just never exercises validateSchema so CI never caught it). Pushed in 4582951.

packages/@tinacms/schema-tools/src/validate/fields.ts
wakqasahmed: Fixed - RichTextField.templates now shares the same guarded TemplateTemp schema.

packages/@tinacms/schema-tools/src/validate/index.spec.ts
wakqasahmed: Updated - this test now asserts the rich-text case fails too, matching the corrected fields.ts. Also fixed the mdx-basic-nested-objects-3 fixture separately, which was a real (if uncaught) instance of the same bug.

packages/@tinacms/schema-tools/src/validate/fields.ts
```diff
@@ -149,7 +149,9 @@ export const TinaFieldZod: z.ZodType<TinaFieldType> = z.lazy(() => {
     .object({
       label: z.string().optional(),
       name,
-      fields: z.array(TinaFieldZod),
+      fields: z
+        .array(TinaFieldZod)
+        .min(1, 'Property `fields` cannot be empty.'),
```
wakqasahmed, line 154: Verified: `TemplateTemp` (the schema shared by both `ObjectField.templates` at line 192 and `RichTextField.templates` at line 212, further down this file) carries this `.min(1, ...)` guard once, here, uniformly - no special-casing between object and rich-text embed templates. This is the correct scope: re-traced `_buildFieldFilter`'s `rich-text` case in `packages/@tinacms/graphql/src/builder/index.ts` myself (lines 1222-1248) and confirmed a rich-text field with `field.templates.length > 0` falls through to `_filterCollectionDocumentType`/`_buildTemplateFilter`, the identical path `object` fields use - so an empty-fields rich-text embed template is exactly as invalid as an empty-fields object-field template. The earlier narrower version of this PR (guarding only `ObjectField`) would have missed that.

packages/@tinacms/mdx/src/next/tests/mdx-basic-nested-objects-3/field.ts
```diff
@@ -21,7 +21,15 @@ export const field: RichTextField = {
...
                   name: 'content',
                   type: 'rich-text',
-                  templates: [{ name: 'World', fields: [] }],
+                  templates: [
+                    {
+                      name: 'World',
+                      // A schema-valid template needs at least one field even
+                      // though this element is used with none in this fixture's
+                      // markdown (a bare self-closing `<World />`).
+                      fields: [{ name: 'label', type: 'string' }],
```
wakqasahmed, line 30: Ran this fixture's test directly (`npx vitest run src/next/tests/mdx-basic-nested-objects-3/index.test.ts`) - passes. Confirmed the concern raised in the commit message is real but harmless here: `in.md`'s `<World />` is used with zero props (`node.json` shows `"props": {}`), and this fixture's `index.test.ts` only round-trips `parseMDX`/`serializeMDX` (never calls `validateSchema`), and neither of those requires a declared field to actually appear in the parsed props. So the added `label` field is inert for this test's purposes while making the template schema-valid. Good catch that this fixture was itself an instance of the bug the issue reports - it just wasn't caught by CI because this test path never validates the schema.

packages/@tinacms/schema-tools/src/validate/schema.ts
```diff
@@ -17,7 +17,7 @@ const Template = z
       required_error: 'Missing `label` property. Property `label` is required.',
     }),
     name: name,
-    fields: z.array(TinaFieldZod),
+    fields: z.array(TinaFieldZod).min(1, 'Property `fields` cannot be empty.'),
```
wakqasahmed, line 20: Confirmed this is the collection-level `Template.fields` (used by a collection's own `templates:` array, e.g. union collections), a separate zod object from `TemplateTemp` in `fields.ts`. Ran the full `@tinacms/schema-tools` suite (`npx jest --config jest.config.js`, 11 suites/118 tests) and `npx tsc --noEmit -p .` on this branch - both green, including the 3 new empty-fields regression tests in `index.spec.ts`.

wakqasahmed (commented):
Independent re-verification of the RichTextField re-scoping (commit 4582951). Re-traced everything from scratch rather than trusting the prior review chain, per request. Summary:

1. `fields.ts` - confirmed `TemplateTemp` (shared by `ObjectField.templates` and `RichTextField.templates`) carries `.min(1, 'Property \`fields\` cannot be empty.')` uniformly, no special-casing between the two field types. See inline comment.
2. `builder/index.ts` `_buildFieldFilter`'s `rich-text` case - independently re-traced (not just re-read the prior reviewer's claim): the early-return fixed-shape filter (`startsWith`/`eq`/`exists`) only fires when `!field.templates || field.templates.length === 0`. Once there's at least one template it falls through to `this._filterCollectionDocumentType(field)` at line 1247, the exact same call the `object` case makes at line 1220. That in turn calls `_buildTemplateFilter`, which builds an `InputObjectTypeDefinition` from `template.fields` with no non-empty check - `fields: []` produces a zero-field GraphQL input type, invalid per the GraphQL spec (an Input Object Type must define one or more fields). The prior reviewer's trace holds up under independent re-derivation. This file isn't touched by the PR (correctly so - the fix belongs in schema-tools), so no inline comment there, noting it here instead.
3. Ran `mdx-basic-nested-objects-3/index.test.ts` directly - passes. Confirmed `World` is used prop-less (`<World />`, `"props": {}` in `node.json`) and the fixture's test only round-trips `parseMDX`/`serializeMDX`, which doesn't require declared fields to be present in usage, so adding an unused optional `label` field to satisfy schema validity doesn't break the round-trip. Also ran the full `@tinacms/mdx` suite: 17 pre-existing failures, all in unrelated `markdown-shortcodes-*`/`wordpress-style*` fixtures (micromark tokenizer errors, nothing to do with templates or this PR's diff) - not a regression from this change.
4. Fresh `grep -rn "fields: \[\]"` sweep across `packages/` (this repo has no `examples/` hits at all). None of the hits are a `Template.fields` reachable via `validateSchema` for the two now-guarded paths: forestry-migrate's is an error-recovery map entry never passed through the zod schema; resolver/index.test.ts and TinaSchema.test.ts construct `Template`/`Collection` objects directly and bypass `validateSchema` entirely; datalayer.test.ts/database/index.ts's `{ fields: [] }` are `IndexDefinition` (sort-key) objects, an unrelated `fields` concept; the rest are Form/props mocks. Also confirmed `Collection.fields` in `schema.ts` already had `.min(1)` pre-existing (untouched by this diff) - the only genuinely new coverage here is `Template.fields` (collection-level template list) and the `TemplateTemp` shared by `ObjectField`/`RichTextField`. No missed breakage found.
5. Ran `@tinacms/schema-tools` full suite (`npx jest --config jest.config.js`): 11 suites / 118 tests, all green, including all 3 new empty-fields tests (collection template, object field template, rich-text template). `npx tsc --noEmit -p .`: clean, exit 0.

Conclusion: the correction (commit 4582951) holds up under independent re-derivation - I did not just trust the prior reviewer's or author's claims, I re-read the switch/case and re-ran the tests myself. No additional bugs or missed spots found beyond what's already fixed.

wakqasahmed:
Hi @wicksipedia @kulesy — not sure who owns review for this one — it's been a little while, green and mergeable. Could you take a look or redirect me?

kulesy (approved):

joshbermanssw (approved):

joshbermanssw:
Thanks for the collaboration @wakqasahmed !

kulesy:
Looks good. I traced the rich-text case and you're right that once a field has any templates it goes down the same filter path an object field does, so guarding the shared template schema was the right call rather than just ObjectField.

One thing worth a follow-up: this closes https://github.com/tinacms/tinacms/issues/7490, but the other case that issue mentions, a template whose fields are all displayOnly or just a password, still breaks the content API the same way. Raised that as https://github.com/tinacms/tinacms/issues/7628.

PR #7553, opened 2026-09-11
fix: browser bundling for the Tina client

### LINKED ISSUE #7515: 🐛 Bug - TinaCMS - Tina client pulls node builtins into browser bundles and breaks the build
### Pain

Since 3.1.0, shipped December 2025, the published `dist/client.js` imports node
builtins directly. Webpack cannot resolve them for the browser, so importing the
Tina client into a `use client` component fails the build with four
`UnhandledSchemeError` traces that never once mention Tina. Turbopack stubs them
and builds fine, which is why this looks unreproducible on a Turbopack project.

That split is the trap. Rules runs Turbopack in dev and webpack in `next build`,
so this would pass locally and fail only in CI. Rules and the Website are clear
today because they import the client from server pages and route handlers only.

Introduced by https://github.com/tinacms/tinacms/pull/6240 while fixing
https://github.com/tinacms/tinacms/issues/6241
Supersedes https://github.com/tinacms/tinacms/issues/6260

### Acceptance Criteria

* [x] The node cache stays out of the browser build of the client, either split
  into its own chunk the way 3.0.0 did, or behind a browser export condition
* [x] A client component importing the Tina client builds clean on webpack
* [x] Server builds still cache as they do now
* [x] A test fails if a node builtin survives a browser bundle of the client

### Reproduce Steps

1. Create a Next app with `tinacms@3.12.1`
2. Import the Tina client into a component marked `use client`
3. Build with webpack (`next build --webpack` on Next 16)

Fails on Next 15 and 16 under webpack, passes under Turbopack, and passes from a
server component on every version. Builds clean on tinacms 3.0.0.

harley-j:
## Summary

- publish a browser-specific Tina client entry
- retain the filesystem response cache in the Node entry
- add unit and built-artifact regression coverage

Fixes #7515

## Testing

- TinaCMS package build
- focused Tina client tests (6 passed)
- browser bundle artifact test
- focused TypeScript and Biome checks
- Node filesystem-cache smoke test
- Next.js 15 webpack compilation of the client-component repro

Assisted by Codex:gpt-5.6-sol


joshbermanssw:
Nice fix, the split is minimal and the export map looks right.

One thing on AC4 (\"a test fails if a node builtin survives a browser bundle\"): \`client.browser.ts\` is a plain-string entry so it builds through the Vite lib path, and Vite rewrites Node builtins to a \`__vite-browser-external\` stub instead of leaving \`node:fs\` etc in the output. So the regex in \`tests/build-verification.test.ts\` returns false even if \`node-cache\` leaks back into \`unifiedClient\`, which is the regression shape this test exists to catch.

Suggest also asserting the bundle has no \`__vite-browser-external\` reference, or bundling \`dist/client.browser.js\` with esbuild \`platform: 'browser'\` and asserting it succeeds.

harley-j:
Addressed in 7ee34e3aa. The artifact test now rejects both Vite's __vite-browser-external marker and any quoted node: specifier. I also verified the negative case by temporarily leaking NodeCache into the browser entry: Vite emitted the external stub and the test failed; after restoring the clean entry, it passes.

joshbermanssw (approved):
LGTM

PR #7556, opened 2026-09-13
fix(@tinacms/cli): move to Vite 7 so installs have one esbuild binary

wicksipedia:
## Why

`@tinacms/cli` depends on `esbuild ^0.28.1` and on `vite ^6.4.3`. Vite 6 requires `esbuild ^0.25.0`, so every install of the CLI gets two esbuild binaries of about 10 MB each. #7297 lined the two up when it moved the CLI to Vite 6. Dependabot (#7076) then moved the catalog to esbuild 0.28 and split them again.

Vite 7.3.6 accepts `esbuild ^0.27.0 || ^0.28.0`, so the CLI and Vite now share one binary. Vite 6.4 gets security patches only. Vite 7.3 still gets bug fixes.

## Changes

- `pnpm-workspace.yaml`: add `vite: ^7.3.6` to the catalog. `@tinacms/cli` now uses `catalog:` for vite.
- `bin/tinacms`: raise the Node.js floor from 18 to 22 to match the LTS versions in the TinaCMS docs (22 and 24). On an older version the binary now exits with code 1. Before, it printed an error and exited with 0, so `tinacms build && next build` in CI skipped the Tina build and deployed anyway.
- `plugins.test.ts`: mock `vite`. Vite 7 ships as ESM only, and the Jest module registry cannot load it. The test covers the dev-server middleware and does not use Vite.

## Effects on users

- `tinacms` needs Node.js 22 or later. Sites that build on Node.js 20 must upgrade their CI image.
- The admin build uses the Vite 7 default browser targets: Chrome 107, Edge 107, Firefox 104, and Safari 16. The Vite 6 targets were Chrome 87, Edge 88, Firefox 78, and Safari 14.
- The built kitchen-sink admin is 16 KB smaller after gzip (3.286 MB to 3.270 MB). Build time stays at 11 to 13 seconds.
- Self-hosted backends do not change. The CLI is the only package that moves to Vite 7.

## Testing

- `pnpm --filter @tinacms/cli test`: 28 suites and 359 tests pass.
- In `examples/next/kitchen-sink`, `tinacms dev` and `tinacms build --local --skip-cloud-checks` run on Vite 7. The dev admin loads in a browser, lists the Authors collection, and logs no console errors. The built admin loads and renders.
- `pnpm why esbuild --filter @tinacms/cli --prod` shows one version, 0.28.1.
- `bin/tinacms` exits with 1 when `process.version` is v20.19.0 or v21.7.3, and runs on v22.20.0.
- The CLI code has no use of the APIs that the Vite 7 migration guide removes.
- `pnpm types` in the CLI fails with the same errors on Vite 6 and Vite 7. The errors come from `../mdx/src` and also occur on `main`.

https://claude.ai/code/session_01TrZVtJwM81RkiiWHBGFt1J


.changeset/bump-cli-vite-7.md
```diff
@@ -0,0 +1,11 @@
+---
+"@tinacms/cli": patch
```
joshbermanssw: This drops Node 20, which works on main today, so it is not a patch.

```suggestion
"@tinacms/cli": minor
```

While here: add `"engines": { "node": "22.x || 24.x" }` to the CLI `package.json` so install warns too, and update the three docs that still say Node 18+ (`examples/astro/kitchen-sink/README.md`, `examples/astro/visual-editing/README.md`, `packages/@tinacms/astro/GETTING_STARTED.md`).

joshbermanssw (changes requested):
Code looks good. One label fix and a couple of small follow-ons before merge.

.changeset/bump-cli-vite-7.md
wicksipedia: all done in [26a14c2](https://github.com/tinacms/tinacms/pull/7556/commits/26a14c27323b8a02155269de1e36b0c558a6ba5d)

joshbermanssw (approved):
LGTM

Issue #7602, opened 2026-09-14
TinaCMS doesn't run in Devcontainers


openscript:
### The exact error message

```shell
$ pnpm dev
$ tinacms dev -c "astro dev"
🦙 TinaCMS Dev Server is initializing...
Indexing local files ⠼
```

### Steps to reproduce

1. Create a Devcontainer
```json
{
  "name": "tinacms",
  "image": "mcr.microsoft.com/devcontainers/typescript-node:5-24-trixie"
}
```
2. Scaffold TinaCMS with Astro (default template) like the «Getting started guide with Astro» describes: https://tina.io/docs/frameworks/astro#option-1-scaffold-from-the-starter-template
3. Start with `pnpm dev` and it will hang forever in `Indexing local files ⠼`

### What you expected vs. what actually happened

I would expect it starts up successfully.

### Your environment

- TinaCMS v3.10.0
- Node v24.19.0

Running in a Devcontainer: `mcr.microsoft.com/devcontainers/typescript-node:5-24-trixie`

### Workaround

There are two errors causing this:

https://github.com/tinacms/tinacms/issues/7294 and the hard-coded `localhost` when connecting to the db which resolves to IPv6 ::1 in my environment.

When I patch both problems it works.

#### `@tinacms__graphql@2.4.11.patch`

```diff
diff --git a/dist/index.js b/dist/index.js
index bef0a4a00bca7f98f6aeeddccb9fdbc8c97c882e..5f622eff648ff1261b3257d011f97e9f7cd1de77 100644
--- a/dist/index.js
+++ b/dist/index.js
@@ -6638,6 +6638,8 @@ var handleCollectionField = (info, args, resolver) => {
 import { ManyLevelGuest } from "many-level";
 import { pipeline } from "readable-stream";
 import { connect } from "net";
+var CONNECT_RETRIES = 50;
+var CONNECT_RETRY_DELAY_MS = 100;
 var TinaLevelClient = class extends ManyLevelGuest {
   port;
   _connected = false;
@@ -6647,10 +6649,32 @@ var TinaLevelClient = class extends ManyLevelGuest {
   }
   openConnection() {
     if (this._connected) return;
-    const socket = connect(this.port);
-    pipeline(socket, this.createRpcStream(), socket, () => {
-      this._connected = false;
-    });
+    // The datalayer server's listen() may not have completed yet; retry
+    // instead of leaving operations queued forever (tinacms/tinacms#7294).
[...]

Issue #7557, opened 2026-09-13
🐛 Image field - an unknown accept value silently disables the filter


kulesy:
### 🩸 Pain

A typo in an image field's `accept` leaves the field taking any file, with nothing logged and nothing shown to the developer. `resolveMediaAccept` drops values it does not recognise, but nothing downstream treats the empty result as a failure. The field's own check is skipped when the list is empty, and the dropzone falls back to the global media accept. So `accept: 'docx'` reads as no restriction at all.

A TypeScript config catches this at compile time because the option is typed, so it only bites JavaScript configs.

### 🛠️ Approach

- [ ] Fail the build, or the schema validation, when `accept` carries a value outside the supported vocabulary
- [ ] Name the unknown value in the error and list what is accepted
- [ ] Cover the single-value and array forms, and list image fields

### ✅ Acceptance Criteria

- [ ] An unknown `accept` value stops the build with an error naming the value
- [ ] A JavaScript config gets the same signal TypeScript already gives
- [ ] A valid `accept` value still restricts the field the way it does today

### 🔗 Related

- https://github.com/tinacms/tinacms/pull/7477
- https://github.com/tinacms/tinacms/issues/7481
- https://github.com/tinacms/tinacms/issues/7474
- https://github.com/tinacms/tina.io/pull/4871

joshbermanssw:
✅ Done by [Tyagiquamar](https://github.com/Tyagiquamar)

Issue #7513, opened 2026-09-03
TinaCloud content API returns corrupted body when the client accepts zstd (ERR_CONTENT_DECODING_FAILED in Chrome/Brave)


Primata:
## Summary

When a browser sends `Accept-Encoding: gzip, deflate, br, zstd`, large responses from
`https://content.tinajs.io/2.4/content/<clientId>/github/<branch>` come back with
`Content-Encoding: zstd`, but the body is not valid zstd. The bytes look like the compressed
stream was passed through a UTF-8 conversion: every byte that is not valid UTF-8 has been
replaced with `EF BF BD` (U+FFFD). The zstd magic number `28 B5 2F FD` arrives as
`28 EF BF BD 2F EF BF BD`.

Chrome and Brave advertise zstd, so the admin UI fails with `net::ERR_CONTENT_DECODING_FAILED`
on the schema check and shows the "An unexpected error occurred while validating your Tina
schema" dialog. Login and small queries work because small responses are not compressed.
Safari, which does not advertise zstd, works.

The same request with `Accept-Encoding: gzip` or `Accept-Encoding: br` returns a valid body.

## Reproduction

```bash
URL=https://content.tinajs.io/2.4/content/<clientId>/github/main
Q='{"query":"{ __schema { types { name } } }"}'

# zstd: corrupted
curl -s -D - -o zstd.bin -H "X-API-KEY: <read-only token>" -H 'content-type: application/json' \
  -H 'accept-encoding: zstd' -d "$Q" "$URL" | grep -i content-encoding
xxd -l 8 zstd.bin          # 28ef bfbd 2fef bfbd  (expected 28b5 2ffd ...)
zstd -dc zstd.bin          # zstd: unsupported format

# gzip: fine
curl -s -o gzip.bin -H "X-API-KEY: <read-only token>" -H 'content-type: application/json' \
  -H 'accept-encoding: gzip' -d "$Q" "$URL"
gunzip -c gzip.bin | head -c 100   # valid JSON
```

Any query whose response is large enough to be compressed reproduces it. A tiny query such as
`{ __typename }` is returned uncompressed and works.

## Browser symptoms

- Brave and Chrome 152 on macOS, extensions disabled, private window: the admin at `/admin`
  logs in, then fails on `checkGraphqlSchema` with `TypeError: Failed to fetch`. The network
  panel shows the POST to `content.tinajs.io/.../github/main` as
  `net::ERR_CONTENT_DECODING_FAILED`.
[...]

Issue #7512, opened 2026-09-03
TinaCloud content endpoint returns zstd response that fails browser decoding


rbedellasmgroup:

rbedellasmgroup:
### The exact error message

In Microsoft Edge, Tina admin fails while loading a collection:

```text
[TypeError] GetCollection failed: Failed to fetch
```

The browser console shows the underlying network failure:

```text
POST https://content.tinajs.io/2.4/content/7540c807-db53-45eb-ab1a-8c23a77f76d4/github/master
net::ERR_CONTENT_DECODING_FAILED 200 (OK)
```

Related console output:

```text
Unexpected error checking schema: TypeError: Failed to fetch
Uncaught (in promise) TypeError: Failed to fetch
Uncaught TypeError: Cannot read properties of undefined (reading 'name')
```

### Steps to reproduce

1. Open the Tina admin at `https://docs.asmgroupinc.com/admin/#/collections/doc/~/stm`.
2. Sign in if prompted.
3. Attempt to load the `doc` collection path `~/stm`.
4. Observe the Tina error modal and the console/network errors above.
5. Open Edge DevTools > Network conditions > Accepted Content-Encodings.
6. Deselect `zstd`, leaving `deflate`, `gzip`, and `br` enabled.
7. Refresh the page.
8. The page loads successfully when `zstd` is not advertised.

### What you expected vs. what actually happened

Expected: Tina admin should load the collection normally.

Actual: TinaCloud responds with HTTP 200 but `Content-Encoding: zstd`, and the browser fails to decode the response before TinaCMS JavaScript can read it. This causes `fetch` to reject and the admin displays `[TypeError] GetCollection failed: Failed to fetch`.

Response headers from a failing request:

```text
Status Code: 200 OK
Access-Control-Allow-Credentials: true
Access-Control-Allow-Origin: https://docs.asmgroupinc.com
Apigw-Requestid: DIqqSgpkIAMEPXw=
Content-Encoding: zstd
Content-Length: 22290
Content-Type: text/plain; charset=utf-8
Date: Thu, 03 Sep 2026 18:36:49 GMT
Tinacms-Graphql-Project-Version: 1.6.1
Tinacms-Graphql-Version: 1.6.1
Vary: Accept-Encoding, Origin
```

The fact that the same page loads after removing `zstd` from accepted encodings makes this look like a TinaCloud/API gateway zstd response encoding issue rather than an application schema issue.

### Your environment

[...]

PR #7604, opened 2026-09-16
ci: flag likely duplicate issues with an agentic workflow

joshbermanssw:
cc: —
Related PBI: none

## TL;DR

Re-reports of open bugs arrive in the ~44 human-filed issues a month (three open issues currently describe the same zstd content-encoding bug: #7509, #7512, #7513) and sit until someone spots them. This adds the repo's first `gh aw` workflow: on `issues.opened`, a Copilot agent searches open issues for the same root cause and, at high confidence, leaves one comment naming the canonical issue and applies the new `🤖 Duplicate` label.

## Reviewer notes

- **Blast radius is one comment + one label.** The agent token is `issues: read`; writes go through `safe-outputs` capped at `max: 1` each, label allowlist of exactly `🤖 Duplicate`, target pinned to the triggering issue. Never closes.
- **`roles: all` is deliberate.** gh-aw defaults to maintainers only, which would skip every community reporter.
- **`skip-bots: [github-actions]`** keeps the starter-template failure issues out; that workflow already dedupes by title.
- **Needs org-level Copilot.** `copilot-requests: write` worked in the trial repo; confirm it is allowed for the `tinacms` org (or add `COPILOT_GITHUB_TOKEN`) before the first real run.

## Tests

Three `gh aw trial` runs against the live issue set, staged so nothing was written to this repo:

- #7513 → high, matched #7509, comment + label
- #7602 → medium, matched #7294, comment only (reporter already links #7294 and adds a second cause)
- #7557 → `noop`, no candidate

## Links

- [gh-aw](https://github.com/github/gh-aw)
- [Trial runs](https://github.com/joshbermanssw/gh-aw-trial/actions)

---
Changes made by Claude Opus 5 via Claude Code.


.github/scripts/dedupe-issue.mts
```diff
@@ -0,0 +1,302 @@
...
+      title: c.title,
+      probability: answers[`same::${c.number}`].noul,
+    })),
+  };
+};
+
+const errorLine = (issue: Issue) => {
+  const section = (issue.body ?? '')
+    .split(/^### /m)
+    .find((s) => s.startsWith('The exact error message'));
```
kulesy: Only 8 of our 315 open issues were filed through the bug report form, so this line is missing from almost every duplicate comment, and on the zstd pair it picks up a section marker inside the code fence rather than the error.

.github/workflows/dedupe-issue.yml
```diff
@@ -0,0 +1,46 @@
...
+        with:
+          node-version-file: .nvmrc
+
+      - name: Compare against open issues
+        run: node .github/scripts/dedupe-issue.mts
+        env:
+          GH_TOKEN: ${{ github.token }}
+          TYPESAFE_API_KEY: ${{ secrets.TYPESAFE_API_KEY }}
+          ISSUE_NUMBER: ${{ github.event.issue.number || inputs.issue_number }}
+          DRY_RUN: ${{ inputs.dry_run == true }}
```
kulesy: This is always false on the issues.opened trigger, so the only way to see how 0.85 and 0.6 behave on real issues is to let the bot comment on them.

.github/scripts/dedupe-issue.mts
joshbermanssw: Verified: 9 of 313 open issues use the template heading, and #7513 does not use the template at all. Removed the quote entirely in 86effe957 rather than patching the extraction; the comment names the winner's title, which carries the same signal.

.github/workflows/dedupe-issue.yml
joshbermanssw: Correct. 86effe957 adds a repository variable, DEDUPE_DRY_RUN; set it to "true" and the issues.opened path writes the shortlist table to the job summary without commenting. Default stays live.

kulesy (approved):

PR #7613, opened 2026-09-17
fix(tinacms): send the same TinaCloud token from editorial workflow and content requests

### LINKED ISSUE #7612: "Save to new branch" fails with "Failed to complete workflow" on tinacms 3.12.1 and later
### 🩸 Pain

Since tinacms 3.12.1, "Save to new branch" on a TinaCloud site fails with "Failed to complete workflow. Please try again." and leaves an empty branch. Normal saves work. Reported on Discord on 17 Sep 2026 (tinacms 3.13.0, Astro).

One session sends two different tokens. Content requests prefer the ID token, but editorial workflow requests go through the auth provider's fetch helper, which has preferred the access token since https://github.com/tinacms/tinacms/pull/7445. TinaCloud reads the editor's email from the token and a Cognito access token has no email claim, so the workflow's commit step fails.

A session with no ID token at all (the WorkOS redirect login stores only an access token) still sends the access token after this fix, so the TinaCloud-side hardening is what closes it fully.

### 🛠️ Approach

- [ ] One shared helper picks the token for both request paths: ID token first, access token as fallback, as 3.11.0 sent everywhere
- [ ] Tests that both paths pick the same token

### ✅ Acceptance Criteria

- [ ] "Save to new branch" on a TinaCloud site with a Cognito session completes and opens the PR

### Related

- https://github.com/tinacms/tinacms/pull/7445
- https://github.com/tinacms/tinacms/pull/7304

kulesy:
Closes #7612

**TL;DR** Content requests and editorial workflow requests now pick the TinaCloud token through one shared helper, ID token first. "Save to new branch" works again on a Cognito session, and ordinary saves send exactly what they sent before.

**Pain:** since tinacms 3.12.1 every "Save to new branch" on a TinaCloud site fails with "Failed to complete workflow. Please try again." and leaves an empty branch behind.

**Solution:** one helper chooses the bearer token for both request paths, preferring the ID token and falling back to the access token, which is what versions up to 3.11.0 sent everywhere.


harley-j (approved):
LGTM

PR #7614, opened 2026-09-17
fix(schema-tools): reject unknown accept values on image fields

### LINKED ISSUE #7557: 🐛 Image field - an unknown accept value silently disables the filter

Tyagiquamar:
Fixes #7557.M An unknown \ccept\ value on an image field (for example \ccept: 'docx'\) resolved to an empty allow-list, and the field check is skipped when the list is empty, so the typo silently disabled the file filter with nothing logged. TypeScript configs catch this at compile time; JavaScript configs got no signal.M This change validates \ccept\ in the schema (zod) against the same vocabulary \
esolveMediaAccept\ understands (categories plus extensions), for both the single-value and array forms and for list fields. An unknown value now fails validation with an error naming the value and listing what is accepted.M Verified in Docker (node:22-bookworm): full @tinacms/schema-tools jest suite 11 suites / 119 tests pass (4 new accept cases included, confirmed failing before the fix), tsc exit 0, biome lint and format exit 0.


Tyagiquamar:
Hi, just following up on this when you get a chance. The branch is up to date and checks are green. If it looks good from your side, it should be ready to merge. Happy to make any changes if needed. Thanks!

packages/@tinacms/schema-tools/src/validate/fields.ts
```diff
@@ -48,6 +53,29 @@ const TinaField = z.object({
...
+// Media accept values
+// ==========
+const mediaAcceptValues = [
+  ...(Object.keys(MEDIA_CATEGORIES) as (keyof typeof MEDIA_CATEGORIES)[]),
+  ...MEDIA_EXTENSIONS,
+] as unknown as [MediaAccept, ...MediaAccept[]];
+const validMediaAcceptList = mediaAcceptValues
+  .map((value) => `\`${value}\``)
+  .join(', ');
+const MediaAcceptValue = z.enum(mediaAcceptValues, {
```
joshbermanssw: Suggestion: build the enum from an `as const` tuple so no assertion is needed. `z.enum(CONTENT_FORMATS)` in `validate/schema.ts:55` already does exactly this cast-free, fed by an `as const` array in `types/index.ts`.

Two separate things force the casts here: `Object.keys()` returns `string[]` and throws away the literal types (the inner `as`), and a spread-built array cannot be proven non-empty, which is what `z.enum` requires (the outer one). The `as unknown as` hop is also stronger than needed; a single `as [MediaAccept, ...MediaAccept[]]` compiles, and going through `unknown` erases the relationship so a future non-string entry in `MEDIA_EXTENSIONS` would pass silently.

Add beside `MEDIA_CATEGORIES` in `packages/@tinacms/schema-tools/src/types/index.ts`:

```ts
export const MEDIA_CATEGORY_NAMES = [
  'image', 'video', 'audio', 'document',
] as const satisfies readonly MediaCategory[];

export const MEDIA_ACCEPT_VALUES = [
  ...MEDIA_CATEGORY_NAMES,
  ...MEDIA_EXTENSIONS,
] as const;
```

The import here then becomes `import { MEDIA_ACCEPT_VALUES } from '../types/index';`, worth folding into the existing `'../types/index'` import on line 2, and this block collapses to:

```suggestion
const validMediaAcceptList = MEDIA_ACCEPT_VALUES.map(
  (value) => `\`${value}\``
).join(', ');
const MediaAcceptValue = z.enum(MEDIA_ACCEPT_VALUES, {
```

One caveat worth knowing: `satisfies readonly MediaCategory[]` catches a wrong name but not an omitted one, so a fifth category added to `MEDIA_CATEGORIES` would quietly go unvalidated. If that is worth guarding:

```ts
type MissingCategory = Exclude<MediaCategory, (typeof MEDIA_CATEGORY_NAMES)[number]>;
const _allCategoriesListed: MissingCategory extends never ? true : MissingCategory = true;
```

Dropping a category from the list then fails with `Type 'true' is not assignable to type '"document"'`, naming the one you missed.

packages/@tinacms/schema-tools/src/validate/fields.ts
```diff
@@ -86,6 +114,13 @@ const ImageField = TinaScalerBase.extend({
     required_error: typeRequiredError,
   }),
   uploadDir: z.function().args(z.any()).returns(z.string()).optional(),
+  accept: z
+    .union([MediaAcceptValue, z.array(MediaAcceptValue)], {
+      errorMap: () => ({
+        message: `Invalid \`accept\` property. \`accept\` must be a ${validMediaAcceptList} or an array of them.`,
+      }),
+    })
+    .optional(),
```
joshbermanssw: Suggestion: drop this outer `errorMap` and keep the union as-is.

`z.union([A, z.array(A)])` is the right definition for "one or many", the issue is only this error map layered on top. It cannot suppress the branch errors underneath it, because `util/parseZodErrors.ts` recursively expands every entry in `issue.unionErrors` and appends them all. So it adds a third message rather than replacing the other two, and it reprints the full 22-value vocabulary that the enum's own message already lists.

Removing it also removes the ungrammatical "must be a `image`, `video`, ..." phrasing, and the value-naming message from `MediaAcceptValue` (the one the acceptance criterion asks for) still comes through.

```suggestion
  accept: z.union([MediaAcceptValue, z.array(MediaAcceptValue)]).optional(),
```

That leaves `Expected array, received string` as the one remaining bit of noise for a single-value typo. That one is `parseZodError`'s doing rather than this PR's, and worth a follow-up to collapse `unionErrors` instead of concatenating them, which would tidy every union error in the schema.

packages/@tinacms/schema-tools/src/validate/fields.ts
```diff
@@ -48,6 +53,29 @@ const TinaField = z.object({
...
+const MediaAcceptValue = z.enum(mediaAcceptValues, {
+  errorMap: (issue, ctx) => {
+    if (issue.code === 'invalid_enum_value') {
+      return {
+        message: `Invalid \`accept\` value \`${
+          typeof ctx.data === 'string' ? ctx.data : JSON.stringify(ctx.data)
+        }\`. Must be one of: ${validMediaAcceptList}.`,
+      };
+    }
+    return { message: issue.message ?? '' };
```
joshbermanssw: `issue.message` is not set yet for every issue code reaching this branch (`invalid_type` is the one that bites here), so this falls through to `''` and emits a blank error line in the array case. `ctx.defaultError` is the contract's fallback and carries zod's own wording.

```suggestion
    return { message: ctx.defaultError };
```

joshbermanssw (commented):
Two small changes that cut the duplicated `accept` error output, keeping the union definition intact.

Tyagiquamar:
All three requested changes are addressed in 00e2091 (validated locally, matching repo CI for schema-tools):

- Enum casts removed: added MEDIA_CATEGORY_NAMES / MEDIA_ACCEPT_VALUES as const tuples in types/index.ts (same pattern as z.enum(CONTENT_FORMATS)); fields.ts now uses z.enum(MEDIA_ACCEPT_VALUES) with no assertions.
- Outer accept errorMap removed: accept is now z.union([MediaAcceptValue, z.array(MediaAcceptValue)]).optional(); the inner enum message is preserved.
- Fallback changed to ctx.defaultError so invalid_type no longer yields a blank message.

Validation: jest @tinacms/schema-tools 11 suites / 119 tests pass (including invalid single, array, and list accept cases plus the valid-values case), pnpm tsc clean, biome check clean on both touched files.

joshbermanssw:
Thanks for applying the suggestions [Tyagiquamar](https://github.com/Tyagiquamar) - could you please sign the `license/cla`

Tyagiquamar:
done

joshbermanssw (approved):
Thanks, all three suggestions are in and look good. The enum mirrors exactly what `resolveMediaAccept` matches, so this only turns values that were already silently ignored into errors; no working config breaks.

Optional nit, not blocking: the new `MEDIA_ACCEPT_VALUES` import can fold into the existing `'../types/index'` import on line 2 (`import { MEDIA_ACCEPT_VALUES, type TinaField as TinaFieldType } from '../types/index';`).

